PR #130, opened 2026-02-28
fix(deps): replace pre-release agentscope pin with stable 1.0.16

### LINKED ISSUE #47: [Bug]: Avoid pre-release dependencies (agentscope, reme-ai) to ensure pip install works on common PyPI mirrors/proxies
## Description

`pyproject.toml` includes pre-release dependencies (e.g. `agentscope==1.0.16.dev0`, `reme-ai==0.3.0.0b5`). Some PyPI mirrors/proxies don’t host or allow pre-releases, causing install failures.

## Component(s) Affected

## Environment

- **CoPaw version:** main (current `pyproject.toml`)
- **OS:** Any
- **Install method:** pip
- **Python version:** 3.10+

## Steps to Reproduce

1. Use a PyPI mirror/proxy that doesn’t include pre-releases.
2. Run `pip install copaw`.

## Actual vs Expected

- **Actual:** Install fails because pre-release versions can’t be resolved from the mirror/proxy.
- **Expected:** Default dependencies use stable releases; pre-releases (if needed) should be optional (e.g. `copaw[dev]`) or guarded by a relaxed stable constraint.

## Logs

```text
ERROR: Could not find a version that satisfies the requirement agentscope==1.0.16.dev0 (from copaw)
ERROR: No matching distribution found for agentscope==1.0.16.dev0
```

fancyboi999:
## Summary

Fix dependency pin to avoid pre-release-only install failures on common PyPI mirrors/proxies.

Closes #47

## Root Cause

`pyproject.toml` pinned:
- `agentscope==1.0.16.dev0` (pre-release)

Many mirrors/proxies do not include or allow pre-release packages, causing `pip install copaw` to fail.

## Change

- Replace `agentscope==1.0.16.dev0` with stable `agentscope==1.0.16`.
- Keep `reme-ai==0.3.0.0` unchanged (already stable on PyPI).

## Verification

- Dry-run resolution with stable constraints:

```bash
python -m pip install --dry-run --no-deps --no-cache-dir \
  "agentscope==1.0.16" "reme-ai==0.3.0.0"
```

Observed result:
- Would install `agentscope-1.0.16`
- Would install `reme_ai-0.3.0.0`

- Pre-commit check on changed file passed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ description = "CoPaw is a **personal assistant** that runs in your own environme
 readme = "README.md"
 requires-python = ">=3.10,<3.14"
 dependencies = [
-    "agentscope==1.0.16.dev0",
+    "agentscope==1.0.16",
     "agentscope-runtime==1.1.0",
     "discord-py>=2.3",
     "dingtalk-stream>=0.24.3",


xieyxclack:
Thank you for your suggestion. Although we would also like to upgrade the dependency versions to the official release, we have found that there are some behavioral differences after upgrading. We are currently fixing these issues, and will continue to maintain the current dependencies for now.

xieyxclack:
We will update the dependency of agentscope to 1.0.17 soon, so we are closing this PR.

PR #156, opened 2026-02-28
fix(providers): add_model/remove_model cannot find custom providers

### LINKED ISSUE #95: [Bug]: copaw models add-model提示找不到provider
root@app-75113-973d4790147911f1b38b1d776d1a7c7e-85446cf894-cx9h7:/code# copaw models add-provider mm -n m -u https://api.siliconflow.cn/v1
✓ Custom provider 'm' (mm) created.
  base_url: https://api.siliconflow.cn/v1
  Run 'copaw models add-model' to add models, then 'copaw models config-key' to set the API key.

root@app-75113-973d4790147911f1b38b1d776d1a7c7e-85446cf894-cx9h7:/code# copaw models add-model mm -m x -n x
Error: Provider 'mm' not found.

root@app-75113-973d4790147911f1b38b1d776d1a7c7e-85446cf894-cx9h7:/code# copaw models add-model m -m x -n x
Error: Provider 'm' not found.

root@app-75113-973d4790147911f1b38b1d776d1a7c7e-85446cf894-cx9h7:/code# copaw models add-model modelscope -m x -n x
✓ Model 'x' (x) added to 'modelscope'.

lailoo:
## Summary

- **Bug**: `add_model()` and `remove_model()` cannot find custom providers because they query the in-memory `PROVIDERS` registry before loading `providers.json`
- **Root cause**: In `store.py`, `PROVIDERS.get(provider_id)` is called before `load_providers_json()`, which is the function that populates custom providers into the registry via `sync_custom_providers()`
- **Fix**: Move `load_providers_json()` call before the `PROVIDERS.get()` lookup in both functions

Fixes #95

## Problem

When a user creates a custom provider and then tries to add a model to it via CLI, the operation fails with `Provider 'xxx' not found.`

The root cause is in `src/copaw/providers/store.py`: both `add_model()` and `remove_model()` call `PROVIDERS.get(provider_id)` **before** calling `load_providers_json()`. Since custom providers are only registered into the in-memory `PROVIDERS` dict when `load_providers_json()` runs `sync_custom_providers()`, the lookup always fails for custom providers in a fresh CLI process.

The Web UI (`copaw app`) calls the same `add_model()` function via its API endpoint, so it is also affected in theory. However, since `copaw app` is a long-running process that calls `load_providers_json()` at startup, the bug is much harder to trigger there.

## Reproduction (on `main` branch)

```
$ copaw models add-provider testbug -n "TestBug" -u "https://api.example.com/v1"
✓ Custom provider 'TestBug' (testbug) created.

$ copaw models add-model testbug -m test-model -n "Test Model"
Error: Provider 'testbug' not found.
```

## Changes

- `src/copaw/providers/store.py` — In `add_model()` and `remove_model()`, moved `data = load_providers_json()` before `defn = PROVIDERS.get(provider_id)` so that custom providers are loaded into the registry before the lookup.

## Verification (on fix branch)

```
$ copaw models add-provider testbug -n "TestBug" -u "https://api.example.com/v1"
✓ Custom provider 'TestBug' (testbug) created.

$ copaw models add-model testbug -m test-model -n "Test Model"
✓ Model 'Test Model' (test-model) added to 'testbug'.

$ copaw models remove-model testbug -m test-model
✓ Model 'test-model' removed from 'testbug'.
```

## Test plan

- [x] Reproduced bug on `main` branch: `add-model` fails with "Provider not found"
- [x] Verified fix on fix branch: `add-model` and `remove-model` both work correctly
- [x] 4 regression tests added in `tests/test_provider_store.py`, all pass
- [x] Built-in provider `add-model` still works (covered by test)
- [x] Nonexistent provider still raises `ValueError` (covered by test)

## Effect on User Experience

**Before:** Creating a custom provider succeeds, but adding a model to it immediately fails with "Provider 'xxx' not found."
**After:** Adding and removing models on custom providers works as expected.

### DIFF
--- a/src/copaw/providers/store.py
+++ b/src/copaw/providers/store.py
@@ -162,8 +162,13 @@ def _ensure_all_providers(providers: dict[str, ProviderSettings]) -> None:
 # -- Load / Save --
 
 
-def load_providers_json(path: Optional[Path] = None) -> ProvidersData:
-    """Load providers.json, creating/repairing as needed."""
+def _load_and_sync(path: Optional[Path] = None) -> ProvidersData:
+    """Load providers.json and sync registries, but do NOT save back.
+
+    This is the internal workhorse used by both ``load_providers_json``
+    (which saves after loading) and mutator functions (which save after
+    their own modifications, avoiding a redundant double-write).
+    """
     if path is None:
         path = get_providers_json_path()
 
@@ -197,6 +202,12 @@ def load_providers_json(path: Optional[Path] = None) -> ProvidersData:
         active_llm=active_llm,
     )
     _validate_active_llm(data)
+    return data
+
+
+def load_providers_json(path: Optional[Path] = None) -> ProvidersData:
+    """Load providers.json, creating/repairing as needed."""
+    data = _load_and_sync(path)
     save_providers_json(data, path)
     return data
 
@@ -234,7 +245,7 @@ def update_provider_settings(
     base_url: Optional[str] = None,
 ) -> ProvidersData:
     """Partially update a provider's settings. Returns updated state."""
-    data = load_providers_json()
+    data = _load_and_sync()
     cpd = data.custom_providers.get(provider_id)
 
     if cpd is not None:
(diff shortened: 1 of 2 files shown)


zhijianma (commented):
Please resolve the Copilot comments, and I would suggest calling `load_providers_json` before using PROVIDERS.

zhijianma:
@lailoo 
Plz see the comments

zhijianma:
Thank you for your contribution to CoPaw
Fixed in #442 and #457

PR #157, opened 2026-02-28
feat(config): add global timezone setting for tools and cron

### LINKED ISSUE #154: [Feature]: 内置的当前时间获取工具只能获取UTC时间，希望可以手动更改时区
很多时候模型获取到UTC时间，然后创建定时任务，但用户描述的是基于当前时区的相对时间，比如10分钟后
模型就要把10分钟后换算成UTC时间再定时，这样换算很容易出错。稳定性极差
并且模型的回复中经常提到换算过程，导致回复内容冗长。

lailoo:
## Summary

- **Feature**: Add a global `timezone` field to `config.json` so tools and cron scheduler use the user's timezone instead of hardcoded UTC
- **Root cause**: `CronManager` initialized with `timezone="UTC"` in `_app.py`; `ScheduleSpec` and CLI `--timezone` both default to `"UTC"`; no way for users to configure their timezone globally
- **Fix**: Add `timezone` field to `Config`, wire it through `get_current_time` tool, `CronManager`, `ScheduleSpec`, and CLI

Fixes #154

## Problem

When a user says "10分钟后提醒我" (remind me in 10 minutes), the agent:
1. Calls `get_current_time` → gets local time (e.g. 09:00 CST)
2. Creates a cron job for 09:10 → but `ScheduleSpec.timezone` defaults to `"UTC"`
3. Cron fires at 09:10 UTC = 17:10 CST → **8 hours late**

Additionally, the model often tries to manually convert between local time and UTC, which is error-prone and produces verbose responses explaining the conversion process.

Root causes:
- `_app.py`: `CronManager` hardcoded `timezone="UTC"`
- `models.py`: `ScheduleSpec.timezone` defaults to `"UTC"`
- `cron_cmd.py`: CLI `--timezone` defaults to `"UTC"`
- No global timezone configuration in `config.json`

## Changes

- `src/copaw/config/config.py` — Add `timezone: str = ""` field to `Config`
- `src/copaw/config/utils.py` — Add `get_timezone()` that reads config, falls back to system local timezone detection (via `/etc/localtime` symlink, `TZ` env, or UTC offset)
- `src/copaw/config/__init__.py` — Export `get_timezone`
- `src/copaw/app/_app.py` — Pass `get_timezone()` to `CronManager` instead of hardcoded `"UTC"`
- `src/copaw/agents/tools/get_current_time.py` — Use `get_timezone()` + `ZoneInfo` instead of `datetime.now().astimezone()`
- `src/copaw/app/crons/models.py` — Add validator on `ScheduleSpec.timezone` to resolve empty string from config
- `src/copaw/cli/cron_cmd.py` — Change `--timezone` default from `"UTC"` to `None` (resolves from config)
- `src/copaw/app/crons/manager.py` — Replace deprecated `datetime.utcnow()` with `datetime.now(timezone.utc)`

## After fix

- `get_timezone()` auto-detects system timezone (e.g. `Asia/Shanghai`) or reads from config
- `CronManager` and `ScheduleSpec` use the detected timezone instead of UTC
- Agent no longer needs to do timezone conversion when creating reminders

Users can also set timezone explicitly in `~/.copaw/config.json`:
```json
{
  "timezone": "Asia/Shanghai"
}
```

## Test plan

- [x] Verified: `get_timezone()` returns `Asia/Shanghai` (auto-detected from system)
- [x] Verified: config override (`America/New_York`) works correctly
- [x] 7 regression tests in `tests/test_timezone_config.py`, all pass
- [x] All existing tests pass

## Effect on User Experience

**Before:** Agent gets local time but creates cron jobs in UTC. User says "10分钟后提醒我" and gets reminded 8 hours late. Model outputs verbose timezone conversion explanations.
**After:** Tools and cron scheduler use the same timezone (auto-detected or user-configured). Reminders fire at the correct local time. No timezone conversion needed in model responses.


lailoo:
Closing this PR — the issue was not reproduced in a real running environment.

On macOS, `get_current_time` on the main branch already returns correct local time (CST) via `datetime.now().astimezone()`, with no visible difference from the fix branch output. The CronManager hardcoded-UTC scheduling issue requires creating actual cron jobs to reproduce, which was not completed.

Without sufficient reproduction evidence, closing this PR for now.

PR #158, opened 2026-02-28
fix(formatter): preserve reasoning_content for thinking models (#155)

### LINKED ISSUE #155: [Bug]:用kimi2.5模型，使用过程中一直报错
报错内容如下，不知道该如何传递参数或者关闭思考模式
Unknown agent error: BadRequestError: Error code: 400 - {'error': {'message': 'thinking is enabled but reasoning_content is missing in assistant tool call message at index 2', 'type': 'invalid_request_error'}} (Details: /var/folders/bj/w3yg9cxx7ln3pbqppl3_gv700000gn/T/copaw_query_error_o81tkgdr.json)

lailoo:
## Summary

- **Bug**: Kimi K2.5 (and other thinking models) fail with "reasoning_content is missing in assistant tool call message" during multi-turn tool calling
- **Root cause**: `FileBlockSupportFormatter._format()` in `model_factory.py` delegates to `OpenAIChatFormatter` which silently drops thinking blocks, losing `reasoning_content` from assistant messages
- **Fix**: After base formatting, inject `reasoning_content` from thinking blocks into the corresponding formatted assistant messages

Fixes #155

## Problem

When using Kimi K2.5 (which has thinking mode enabled by default):
1. Model returns assistant message with `reasoning_content` + `tool_calls`
2. CoPaw saves the response as `ThinkingBlock` + `ToolUseBlock` in memory
3. On the next turn, the formatter drops the `ThinkingBlock` (logged as "Unsupported block type thinking")
4. The API receives an assistant message without `reasoning_content` and rejects it

**Before fix (Web UI on main branch):**

image

Error log:
```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'thinking is enabled but
reasoning_content is missing in assistant tool call message at index 6',
'type': 'invalid_request_error'}}
```

Full traceback:
```
react_agent.py:298 → reply() → super().reply()
_react_agent.py:437 → reply() → _reasoning()
_react_agent.py:563 → _reasoning() → self.model()
_openai_model.py:289 → __call__() → client.chat.completions.create()
→ openai.BadRequestError: 400 - thinking is enabled but reasoning_content
  is missing in assistant tool call message at index 6
```

**Before fix (real API call reproduction):**
```
Step 3: Send next turn WITHOUT reasoning_content → expect error
  ❌ FAIL — BUG CONFIRMED!
  Error: Error code: 400 - {'error': {'message': 'thinking is enabled but
  reasoning_content is missing in assistant tool call message at index 2',
  'type': 'invalid_request_error'}}
```

## Changes

- `src/copaw/agents/model_factory.py` — In `FileBlockSupportFormatter._format()`, collect thinking content from assistant messages before base formatting, then inject as `reasoning_content` into the formatted output. Pattern follows agentscope's existing `DeepSeekChatFormatter`.

**After fix (Web UI on fix branch):**

image

**After fix (real API call verification):**
```
Step 3: Format through CoPaw's formatter
  reasoning_content in formatted msg: YES

Step 4: Send formatted messages to Kimi K2.5 API
  ✅ PASS — Response: The current time is 2:30 PM (14:30) on March 1, 2026 (CST, UTC+0800).
```

## Test plan

- [x] Reproduced on main with real Kimi K2.5 API call (400 error confirmed)
- [x] Reproduced on main via Web UI (same 400 error)
- [x] Verified fix with real Kimi K2.5 API call (multi-turn tool calling succeeds)
- [x] New test: `test_reasoning_content_preserved` — thinking blocks → reasoning_content in formatted output
- [x] New test: `test_no_reasoning_when_no_thinking` — no false positives
- [x] New test: `test_multiple_assistant_messages` — each assistant msg gets correct reasoning_content
- [x] All 3 regression tests pass

## Effect on User Experience

**Before:** Using Kimi K2.5 (or any thinking model) with tools causes immediate failure on the second turn. Users cannot use thinking models with CoPaw at all.
**After:** Thinking models work correctly. `reasoning_content` is preserved across turns, enabling multi-step tool calling with models like Kimi K2.5, DeepSeek-R1, etc.

## Release Notes

* **Bug Fixes**
  * Improved handling of thinking blocks to ensure they are properly preserved and injected as reasoning content in assistant messages during message formatting.

* **Tests**
  * Added test coverage to verify thinking blocks are correctly preserved as reasoning content in formatted output, including edge cases with multiple assistant messages and scenarios without thinking blocks.


src/copaw/agents/model_factory.py
```diff
@@ -73,13 +73,45 @@ class FileBlockSupportFormatter(base_formatter_class):
...
-            tool messages.
+            - Sanitizes improperly paired tool messages.
+            - After base formatting, injects ``reasoning_content`` from
+              thinking blocks into assistant messages so that APIs like
+              Kimi K2.5 (which require it) don't reject the request.
             """
             msgs = _sanitize_tool_messages(msgs)
-            return await super()._format(msgs)
+
+            # Collect thinking content per assistant msg (in order).
```
ekzhu, line 85: Should we apply this kind of fix to AgentScope library's formatter instead of CoPaw? @rayrayraykk @DavdGao

rayrayraykk: Agree

ekzhu: @lailoo can you make this fix to AgentScope library instead? It will help to keep this code base focusing on the application not the model interaction layer.

lailoo: Good point!  I'll move this fix to the AgentScope library instead. Will create a PR there and link it back here.
@ekzhu

ekzhu: Thank you! Closing this PR.

ekzhu:
Closing this PR in favor of a fix directly to agentscope library.

PR #192, opened 2026-03-01
Prevent deadlock on large stdout and add regression tests

dipeshbabu:
- This fixes `execute_shell_command()` to use `proc.communicate()` under `asyncio.wait_for(...)` instead of awaiting `proc.wait()` before draining pipes.
- This prevents potential deadlocks/timeouts when commands write large stdout/stderr (pipe buffer fills before exit).
- Add regression tests for large stdout and empty command handling.
- Fix a Windows-only failure in `execute_shell_command` where an invalid default `WORKING_DIR` could raise `WinError 267` and prevent commands from running. It adds safe cwd selection with a fallback to `Path.cwd()` and retries subprocess creation if the provided cwd is rejected.
- Minor docstring correction for timeout default.

### DIFF
--- a/src/copaw/agents/tools/shell.py
+++ b/src/copaw/agents/tools/shell.py
@@ -8,10 +8,36 @@ import locale
 from pathlib import Path
 from typing import Optional
 
-from agentscope.tool import ToolResponse
 from agentscope.message import TextBlock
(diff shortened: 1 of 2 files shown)


zhijianma:
fixed in #672

PR #208, opened 2026-03-01
feat(providers): add Responses adapter with UI chat-model selection

Asuta:
## Summary
- add `OpenAIResponsesChatModel` in CoPaw (internal adapter, no external proxy), sending requests to Responses API and converting responses back to existing chat blocks
- expose provider `chat_model` end-to-end (store/router/CLI + Models UI select + card display + i18n), so users can switch between `OpenAIChatModel` and `OpenAIResponsesChatModel` visually
- add chat-model normalization + fallback warning logs for invalid values and regression tests covering three real failures:
  - unsupported `input[].name`
  - missing `tools[].name` for Responses format
  - wrong assistant history block type (`input_text` vs `output_text`)

## Validation
- `npm run build` (console)
- `python -m compileall src/copaw`
- `PYTHONPATH=src pytest -q tests/providers`
- manual browser verification (multi-turn chat works after switching to Responses)

### DIFF
--- a/src/copaw/providers/registry.py
+++ b/src/copaw/providers/registry.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import re
+import logging
 from typing import TYPE_CHECKING, List, Optional, Type
 
 from agentscope.model import ChatModelBase, OpenAIChatModel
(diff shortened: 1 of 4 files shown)


Asuta:
Closing this PR in favor of a refreshed branch that includes the follow-up Responses streaming fix.

PR #215, opened 2026-03-01
feat(providers): add Responses adapter with UI chat-model selection

Asuta:
## Summary
- add internal `OpenAIResponsesChatModel` support in CoPaw (no external proxy), with request/response conversion between Responses API and existing chat block format
- expose provider `chat_model` through backend + Models UI (select/toggle), so users can switch between `OpenAIChatModel` and `OpenAIResponsesChatModel` visually
- add backend normalization + warning fallback for invalid `chat_model` values
- add regression tests for prior compatibility issues (`input[].name`, `tools[].name`, assistant-history block type)
- add true Responses streaming support (`stream=True` and event parsing for text deltas, tool-call argument deltas, and usage at completion)

## Validation
- `npm run build` (console)
- `python -m compileall src/copaw`
- `PYTHONPATH=src pytest -q tests/providers`
- manual browser checks: switched to Responses mode and verified multi-turn streaming output

### DIFF
--- a/src/copaw/agents/tools/shell.py
+++ b/src/copaw/agents/tools/shell.py
@@ -5,6 +5,10 @@
 
 import asyncio
 import locale
+import os
+import shlex
+import shutil
+import subprocess
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


zhijianma:
@Asuta 

We have refactored the provider module to improve performance/readability.

Please confirm if this PR is still needed. If so, kindly update the code and resolve the conflicts.

Thank you again for your contribution!

zhijianma:
Outdated

PR #357, opened 2026-03-02
fix(install): avoid uv parallel install lock failures on Windows

### LINKED ISSUE #301: [Bug]: install.ps1 fails on Windows with "Access Denied" due to uv parallel install
## Description

  Running the installer on Windows fails with "拒绝访问 (os error -2147024891)" (Access Denied) during `uv pip install`.
   The error occurs on a random package each time (e.g., `charset-normalizer`, `httpx`, `distro`).

  ## Steps to Reproduce

  1. Windows 11 Pro (10.0.26200)
  2. Run:
     ```powershell
     irm https://copaw.agentscope.io/install.ps1 | iex
     ```
  3. Installation fails at `uv pip install copaw` step.

  ## Error Output

  ```
  [copaw] Installing copaw from PyPI...
  error: Failed to install: charset_normalizer-3.4.4-cp312-cp312-win_amd64.whl (charset-normalizer==3.4.4)
    Caused by: 拒绝访问。 (os error -2147024891)
  [copaw] Installation failed
  ```

  The failing package is different each run (`httpx`, `distro`, etc.), indicating it is not package-specific.

  ## Root Cause

  `uv` (v0.10.7) installs packages in parallel by default. On Windows, concurrent file writes cause file-locking
  conflicts resulting in Access Denied errors. copaw has 214 dependencies, making the collision probability high.

  ## Workaround

  Setting `UV_CONCURRENT_INSTALLS=1` before `uv pip install` resolves the issue:

  ```powershell
  $env:UV_CONCURRENT_INSTALLS = "1"
  uv pip install copaw --python "$VenvPython" --prerelease=allow
  ```

  Installation completes successfully with this change (214 packages installed in ~31s).

  ## Suggested Fix

  In `install.ps1`, add before the `uv pip install` calls:

  ```powershell
  # Workaround for uv parallel install failures on Windows
  if ($IsWindows -or $env:OS -eq "Windows_NT") {
      $env:UV_CONCURRENT_INSTALLS = "1"
  }
  ```

  ## Environment

  - **OS:** Windows 11 Pro 10.0.26200
  - **uv:** 0.10.7
  - **Python:** 3.12.12 (managed by uv)
  - **Shell:** PowerShell

liuxiaopai-ai:
## Summary
- set `UV_CONCURRENT_INSTALLS=1` in `scripts/install.ps1` before any `uv pip install` call
- keep install flow unchanged, only enforce serial wheel installation on Windows

## Why
Issue reports show intermittent `Access Denied (os error -2147024891)` while `uv pip install` runs in parallel on Windows. Serial install avoids Windows file-lock races and makes installer behavior stable.

## Issue Mapping
Fixes #301

## Validation
- code review: only installer env var setup added; no behavior changes outside install concurrency
- local shell verification of diff and script flow
- note: `pwsh` binary is not available in this execution environment, so I could not run PowerShell parser/runtime checks here

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -123,6 +123,11 @@ function Ensure-Uv {
 
 Ensure-Uv
 
+# On Windows, uv parallel wheel installs can intermittently fail with
+# file-lock "Access Denied" errors. Force serial install for reliability.
+$env:UV_CONCURRENT_INSTALLS = "1"
+Write-Info "Set UV_CONCURRENT_INSTALLS=1 for stable package installation on Windows"
+
 # ── Step 2: Create / update virtual environment ──────────────────────────────
 if (Test-Path $CopawVenv) {
     Write-Info "Existing environment found, upgrading..."


xieyxclack:
We provide a robust fix in #484 

Thank you for your contribution

PR #361, opened 2026-03-02
fix(feishu): resolve cron send target via receive_id fallbacks

### LINKED ISSUE #350: [Bug]: 飞书 receive_id 问题
我设置了飞书 channels，可以正常在飞书中交互，但是cron 主动发送的消息不行，分析下看起来是receive_id 没有设置。需要在cron发送给飞书消息时候加入open_id

liuxiaopai-ai:
## Summary
- fix Feishu session-key suffix fallback to match against stored `receive_id` (not `receive_id_type`)
- add fallback to `meta.user_id` when it carries an `open_id` (`ou_*`) and store lookup misses
- add async regression tests for both fallback paths

## Why
Cron proactive messages can fail when the short session key cannot be resolved from store. The existing suffix fallback compared the wrong tuple field and could never match real open_id/chat_id values.

## Issue Mapping
Fixes #350

## Validation
- `PYTHONPATH=src pytest -q tests/channels/test_feishu_receive_id_resolution.py`
- `python -m compileall src/copaw/app/channels/feishu/channel.py tests/channels/test_feishu_receive_id_resolution.py`

### DIFF
--- a/src/copaw/app/channels/feishu/channel.py
+++ b/src/copaw/app/channels/feishu/channel.py
@@ -1347,13 +1347,23 @@ class FeishuChannel(BaseChannel):
                 if len(suffix) >= 4:
                     async with self._receive_id_lock:
                         for _, v in self._receive_id_store.items():
-                            if v[0] and str(v[0]).endswith(suffix):
+                            # v = (receive_id_type, receive_id)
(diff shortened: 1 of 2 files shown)


xieyxclack:
fixed in #464

Issue #557, opened 2026-03-04
[Bug]: Image messages cause DashScope 400 error and poison session history


beiyu:
## Description

When a user sends an image via Telegram, the agent crashes with a DashScope API 400 error:

```
<400> InternalError.Algo.InvalidParameter: The provided URL does not appear to be valid.
```

The Telegram channel downloads the image to a local path (e.g. `~/.copaw/media/telegram/6213babe9efc.jpg`). During message processing, `_update_block_with_local_path()` converts this to a `file:///` URI via `Path.as_uri()`. When the DashScope formatter later checks `_is_accessible_local_file(url)`, it calls `os.path.isfile("file:///...")` which returns `False`, so the `file:///` URI is sent to the DashScope API as-is — which rejects it as an invalid URL.

**Root cause:** `_update_block_with_local_path()` in `message_processing.py` uses `Path(local_path).as_uri()` for image/video blocks, producing a `file:///` URI. The downstream DashScope formatter expects a bare absolute path (e.g. `/Users/.../xxx.jpg`) to correctly detect it as a local file and prepend `file://` itself.

**Data flow:**
```
Telegram photo
  → _download_telegram_file() → /Users/.../.copaw/media/telegram/xxx.jpg
  → ImageContent(image_url="/Users/.../.copaw/media/telegram/xxx.jpg")
  → agentscope_runtime adapter → ImageBlock(source=URLSource(url="/Users/.../xxx.jpg"))
  → process_file_and_media_blocks_in_message()
    → _update_block_with_local_path()
    → source.url = Path(local_path).as_uri()  →  "file:///Users/.../xxx.jpg"  ← BUG
  → DashScope formatter _format_dashscope_media_block():
    → _is_accessible_local_file("file:///Users/.../xxx.jpg") → False
    → sends "file:///Users/.../xxx.jpg" as web URL to DashScope API → 400
```

Note: the `audio` block branch in the same function also uses `as_uri()`, but includes a `media_type` field and may follow a different formatter path. It should be audited as well.

**Suggested fix:** Replace `Path(local_path).as_uri()` with `str(Path(local_path).resolve())` for image/video blocks in `_update_block_with_local_path()`, so the formatter receives a bare path and can apply its own `file://` prefix correctly.

**Security considerations:** None. The media path is already validated against `_ALLOWED_MEDIA_ROOT`.

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
[...]

PR #559, opened 2026-03-04
fix: remove failed user messages from memory to prevent session poisoning

beiyu:
## Summary

Related: #557

When `query_handler` raises an exception (e.g. a DashScope 400 from a malformed image URI), the user's input message has already been added to agent memory by `ReActAgent.reply()`. The `finally` block then calls `save_session_state()`, persisting the poisoned memory to disk. Every subsequent request in the same session replays the bad message and fails again — the session is permanently broken until the user runs `/start`.

This PR removes the failed user messages from memory in the `except` block before the session state is saved, so a single error does not permanently break the session.

## Changes

- In `runner.py` `query_handler`: on exception, delete the current request's `msgs` from `agent.memory` by their IDs before re-raising
- Wrapped in try/except so a failure to clean up memory does not mask the original error

## How it works

```
ReActAgent.reply(msg)
  → self.memory.add(msg)     # msg is in memory now
  → LLM call fails           # exception
  → propagates to runner

runner.query_handler except block:
  → agent.memory.delete([msg.id])  # ← NEW: remove the poisoned message
  → raise

runner.query_handler finally block:
  → save_session_state()     # saves clean memory (without the bad msg)
```

## Testing

- Send image via Telegram → 400 error → send text in same session → works (was broken before)
- Normal text conversation → no change in behavior

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error recovery: when processing fails, the app now captures memory state before cleanup and removes the affected user messages from session memory to prevent malformed content from persisting. Deletion issues are logged but do not mask the original error, improving stability and observability during failures.

### DIFF
--- a/src/copaw/app/runner/runner.py
+++ b/src/copaw/app/runner/runner.py
@@ -150,11 +150,38 @@ class AgentRunner(Runner):
                 await agent.interrupt()
             raise
         except Exception as e:
+            # Dump before cleanup so the debug file captures the full
+            # memory state including the message that caused the error.
             debug_dump_path = write_query_error_dump(
                 request=request,
                 exc=e,
                 locals_=locals(),
             )
+
+            # Remove the user messages that were added to memory before the
+            # error.  Without this, the malformed content (e.g. a bad image
+            # URI) is persisted by save_session_state in the finally block
+            # and poisons every subsequent request in this session.
+            if agent is not None and msgs:
+                try:
+                    msg_ids = [
+                        m.id for m in (
+                            msgs if isinstance(msgs, list) else [msgs]
+                        )
+                        if hasattr(m, "id")
+                    ]
+                    if msg_ids:
+                        await agent.memory.delete(msg_ids)
+                        logger.info(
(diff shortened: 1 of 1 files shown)


jinliyl:
@beiyu 
Thank you very much for your PR! Clearing the context by overwriting the code on failure might seem a bit heavy-handed. Instead, you can simply use the /clear command to clear the context directly.

beiyu:
> @beiyu Thank you very much for your PR! Clearing the context by overwriting the code on failure might seem a bit heavy-handed. Instead, you can simply use the /clear command to clear the context directly.

@jinliyl Thanks for the feedback! The /clear command does work as a manual workaround, but the issue is that users don't know why the bot stopped responding — they just see silence after sending an image. From their perspective, the bot is broken.

This PR targets the automatic recovery case: when a single bad message causes an LLM error, the session should not be permanently poisoned. The user's next message should just work, without requiring them to diagnose the problem and run /clear.

The cleanup only removes the specific user message(s) from the current failed turn — not the entire conversation history. So it's more surgical than /clear, which wipes everything.

PR #633, opened 2026-03-04
[codex] fix responses-api fallback and gateway compatibility

wypgb:
﻿## Summary
This PR adds an OpenAI-compatible resilience layer to OpenAIChatModelCompat for gateways that do not fully support chat.completions but do support the Responses API.

The change is designed to prevent hard failures seen in real CoPaw sessions with third-party gateways. Specifically, it introduces an automatic fallback from chat.completions to /v1/responses, plus robust request/response conversion logic so ReAct-style tool loops continue working.

## User Impact
Before this change, users configured with some OpenAI-compatible gateways could hit repeated runtime failures such as:
- PermissionDeniedError: Your request was blocked.
- BadRequestError: Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.
- AttributeError: 'str' object has no attribute 'model_dump'

These failures prevented the agent from completing normal turns and tool-augmented turns.

After this change, the same scenarios proceed successfully by falling back to Responses API with compatible payload mapping.

## Root Cause
1. CoPaw runtime primarily uses chat.completions semantics.
2. Some gateways reject this path and require /v1/responses.
3. Responses API payload requirements can vary by role and content block type (for example assistant history blocks requiring output_text rather than input_text).
4. Some gateways return non-standard response shapes (for example plain string), while previous code assumed an object with model_dump().

### 1) Automatic fallback to Responses API
- In OpenAIChatModelCompat.__call__, when known compatibility errors are detected, the model call retries via Responses API (/v1/responses).

### 2) Request conversion for fallback path
- Added conversion from chat-completions message format to responses input format.
- Added conversion for tool schemas and tool choice.
- Added role-aware text block mapping:
  - ssistant history text -> output_text
  - user/system/developer text -> input_text

### 3) Response conversion back to AgentScope format
- Added conversion from responses output items to ChatResponse blocks, including tool calls (	ool_use) and plain text.
- Preserved usage information where available.

### 4) Robust handling of heterogeneous response shapes
- _response_to_chat_response now supports response objects, dicts, and plain strings, avoiding crashes when gateways return unexpected shapes.

## Validation Performed
I ran live compatibility checks against an OpenAI-compatible gateway that previously failed:
- Normal chat turn via fallback path.
- Tool-calling turn with function call parsing.
- Assistant-history turn (role/content block compatibility regression).
- Defensive conversion path where response can be plain string.

Additionally, the updated module compiles successfully (py_compile).

## Files Changed
- src/copaw/providers/openai_chat_model_compat.py

## Notes
This is intentionally scoped to compatibility behavior inside OpenAIChatModelCompat and does not change provider configuration UX or defaults.

### DIFF
--- a/src/copaw/providers/openai_chat_model_compat.py
+++ b/src/copaw/providers/openai_chat_model_compat.py
@@ -4,12 +4,15 @@
 from __future__ import annotations
 
 import json
+import time
 from datetime import datetime
 from types import SimpleNamespace
 from typing import Any, AsyncGenerator, Type
+from urllib.parse import urlsplit, urlunsplit
 
 from agentscope.model import OpenAIChatModel
 from agentscope.model._model_response import ChatResponse
+from agentscope.model._model_usage import ChatUsage
 from pydantic import BaseModel
 
 
@@ -55,10 +58,7 @@ def _sanitize_tool_call(tool_call: Any) -> Any | None:
         has_name
         and has_arguments
         and isinstance(raw_name, str)
-        and isinstance(
-            raw_arguments,
-            str,
-        )
+        and isinstance(raw_arguments, str)
     ):
         return tool_call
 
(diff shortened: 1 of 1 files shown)


src/copaw/providers/openai_chat_model_compat.py
```diff
@@ -130,6 +130,366 @@ def _sanitize_stream_item(item: Any) -> Any:
...
+    parts = urlsplit(base_url)
+    path = (parts.path or "").rstrip("/")
+    if not path:
+        path = "/v1"
+    elif not path.endswith("/v1"):
+        path = f"{path}/v1"
+    return urlunsplit(
+        (parts.scheme, parts.netloc, path, parts.query, parts.fragment),
+    )
+
```
zhijianma, line 144: The v1 suffix in base_url should be optional since some self-hosted model services may not support it

PR #657, opened 2026-03-04
Feature: add session overflow auto-protection in query_handler

Atletico1999:
## Description

Adds automatic session overflow protection to prevent the agent from becoming permanently unresponsive when conversation history exceeds the model's context window limit.

**Problem:** When session history accumulates beyond the model's token limit (e.g. 200K for Claude), every query fails with `prompt is too long`. The existing [MemoryCompactionHook](cci:2://file:///C:/ProgramData/miniconda3/envs/XXX/Lib/site-packages/copaw/agents/hooks/memory_compaction.py:64:0-203:19) cannot recover because compaction itself requires an LLM call — which also fails when the prompt is too long. This creates a deadlock where no messages can be processed.

**Solution:** Two LLM-free safety nets added to `AgentRunner.query_handler`:

- **Layer 1 (proactive):** After [load_session_state](cci:1://file:///C:/ProgramData/miniconda3/envs/XXX/Lib/site-packages/agentscope/session/_json_session.py:73:4-121:13), check the session file size. If it exceeds 2 MB, truncate memory to the last 5 messages without any LLM call.
- **Layer 2 (reactive):** Catch `prompt is too long` errors, clear all memory, rebuild system prompt, and retry the query once.

Both layers operate without LLM calls, ensuring recovery even when the API rejects requests due to prompt length.

## Type of Change

- [x] Bug fix
- [x] New feature

## Component(s) Affected

## Checklist

- [x] Ready for review

## Testing

1. Simulate a session file larger than 2 MB (e.g. through frequent cron tasks accumulating history)
2. Send a message — Layer 1 should truncate and log: `Session overflow protection: truncated N messages`
3. If a `prompt is too long` error still occurs, Layer 2 catches it, clears memory, and retries

## Additional Notes

- The 2 MB threshold and 5-message keep count are class constants (`_SESSION_FILE_SIZE_LIMIT`, `_SESSION_KEEP_RECENT`) — easy to adjust or make env-configurable later.
- Fully backward-compatible: no behavior change for sessions under the size limit.

## Release Notes

* **Bug Fixes**
  * Improved handling of extended conversations with automatic session memory management
  * Added protection to prevent overflow errors when conversations exceed size limits
  * Enhanced error detection and automatic retry mechanism for prompt-related failures

### DIFF
--- a/src/copaw/agents/tools/browser_control.py
+++ b/src/copaw/agents/tools/browser_control.py
@@ -17,7 +17,9 @@ import os
 import subprocess
 import sys
 import time
-from concurrent.futures import ThreadPoolExecutor  # pylint: disable=no-name-in-module
+from concurrent.futures import (
+    ThreadPoolExecutor,
+)  # pylint: disable=no-name-in-module
 from typing import Any, Optional
 
 from agentscope.message import TextBlock
@@ -1135,16 +1137,16 @@ async def _action_screenshot(
                 await _run_sync(
                     locator.screenshot,
                     path=path,
-                    type=screenshot_type
-                    if screenshot_type == "jpeg"
-                    else "png",
+                    type=(
+                        screenshot_type if screenshot_type == "jpeg" else "png"
+                    ),
                 )
             else:
                 await locator.screenshot(
                     path=path,
-                    type=screenshot_type
-                    if screenshot_type == "jpeg"
-                    else "png",
+                    type=(
+                        screenshot_type if screenshot_type == "jpeg" else "png"
+                    ),
                 )
         else:
             if frame_selector and frame_selector.strip():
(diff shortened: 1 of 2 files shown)


erknie:
This is a serious problem which has hit me multiple times forcing reboots of the server running CoPaw and the backend llama-server PC.

Atletico1999:
Thanks @erknie  for confirming this — good to know I'm not the only one hitting it.

Took another look at the CodeRabbit feedback on `_is_prompt_too_long()`. Fair point about the substring matching being fragile. Two things I can do:

1. Check structured error fields first (exception class, status code, error code) before falling back to substring matching
2. Make the memory reset less aggressive — trim oldest messages instead of wiping everything, unless we're sure it's a real context overflow

Let me know if that direction works and I'll push an update.

xieyxclack:
It seems that this PR includes many unrelated files.

Atletico1999:
I've cleaned up the PR.

Specifically:
1. **Clean Rebase**: I rebased the branch on the latest `main`, so the unrelated file changes are gone now. The PR is now strictly focused on the session overflow protection in `runner.py`.
2. **System Prompt Preservation**: I added a helper to ensure the `system` message is always preserved during truncation or retry-reset. This keeps the agent's core instructions intact even when the context window is near its limit.
3. **Hardened Detection**: The `_is_prompt_too_long` check now prioritizes looking at structured error fields and status codes from model providers, keeping substring matching only as a fallback.
4. **Conservative Truncation**: I've softened the retry strategy to trim oldest messages instead of a full memory wipe, and fixed the boundary condition for the overflow check as suggested.

PR #749, opened 2026-03-05
fix(console): handle event stream interruptions gracefully

sidonsoft:
## Problem

Console channel's consume_one() silently loses messages when the event stream is interrupted during processing.

## Solution

Added explicit exception handling and user notification for stream interruptions.

## Changes

- Added stream_error tracking variable
- Separated exception handling for stream vs processing callbacks
- Added user notification when stream is interrupted
- Improved docstring for consume_one()
- Created EVENT_STREAM_FIX.md documentation

## CI Status

- ✅ Console format check: pass
- ✅ Website format check: pass  
- ✅ CodeRabbit: pass (all feedback addressed)
- ⏳ Run job: pending (expected to pass)

### DIFF
--- a/EVENT_STREAM_FIX.md
+++ b/EVENT_STREAM_FIX.md
@@ -1,0 +1,56 @@
+# Event Stream Resilience Fix for Console Channel
+
+## Problem
+
+Console channel's `consume_one()` silently loses messages when the event stream is interrupted during processing. This occurs when:
+
+- Memory compaction runs during response generation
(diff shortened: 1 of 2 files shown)


sidonsoft:
Updated with addressing all CodeRabbit and Gemini Code Assist review comments:

- Narrowed exception scope so processing bugs are not mislabeled as stream interruptions
- Gated on_reply_sent to only fire when a reply was actually emitted
- Sanitized callback error messages (don't expose internal details)

PR #760, opened 2026-03-05
feat(runner): detect context length errors and suggest /compact

sidonsoft:
## Problem
When context exceeds model limits, users see:
AGENT_UNKNOWN_ERROR: Unknown agent error: BadRequestError: 400 Bad Request: prompt too long; exceeded max context length

## Solution
Detect BadRequestError with context limit patterns and convert to ModelContextLengthExceededException with helpful /compact suggestion.

## Changes
- _is_context_length_error(exc) - helper to detect context limit errors
- Intercept in query_handler exception block before generic handling
- Raise ModelContextLengthExceededException with suggestion details

## Testing
- AST validity: ✅
- Line length (≤79): ✅
- Trailing whitespace: ✅
- Unit tests: 3/3 passed

## Result
**Before:** AGENT_UNKNOWN_ERROR: Unknown agent error: BadRequestError...
**After:** Model context length exceeded: gpt-4 (with suggestion to run /compact)

Ready for review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for context capacity limitations. Users now receive clearer, user-friendly error messages with actionable guidance when processing capacity is exceeded, helping them quickly understand and resolve the issue.

### DIFF
--- a/src/copaw/app/runner/runner.py
+++ b/src/copaw/app/runner/runner.py
@@ -34,6 +34,35 @@ from ...constant import (
 logger = logging.getLogger(__name__)
 
 
+def _is_context_length_error(exc: Exception) -> bool:
+    """Check if exception is a model context length exceeded error.
+
+    Args:
+        exc: The exception to check.
+
+    Returns:
+        True if this is a context length error from the API.
+    """
+    exc_str = str(exc).lower()
+    exc_type = type(exc).__name__.lower()
+
+    # Check for BadRequest from LiteLLM/OpenAI
+    if "badrequest" not in exc_type:
+        return False
+
+    # Common patterns in context length error messages
+    patterns = [
+        "prompt too long",
+        "context length",
+        "exceeded max context",
+        "token limit",
+        "maximum context length",
+        "too many tokens",
+    ]
+
+    return any(pattern in exc_str for pattern in patterns)
+
+
 class AgentRunner(Runner):
     def __init__(self) -> None:
         super().__init__()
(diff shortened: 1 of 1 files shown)


jinliyl:
@sidonsoft 
Thank you for your feedback. Manually prompting users to perform compaction isn't an ideal solution—we’re much more excited about enabling automatic compaction.

Please feel free to report your issue by opening an issue on GitHub.  
You're also welcome to contribute by helping improve the automatic compaction implementation in the following file:  
https://github.com/agentscope-ai/CoPaw/blob/main/src/copaw/agents/hooks/memory_compaction.py

jinliyl (commented):
Thanks for the PR — the direction is solid and this addresses a real UX pain point. A few suggestions before merging:

1. **Exception chain traversal**: `_is_context_length_error` only inspects the top-level exception. If LiteLLM or another middleware wraps the original `BadRequestError`, the check will miss it. Consider walking `__cause__` / `__context__` to inspect the full chain.

2. **Broader exception type matching**: The `"badrequest" not in exc_type` guard is too narrow — different providers use different class names (`InvalidRequestError`, `APIError`, etc.). Either widen the type check or remove it and rely solely on message pattern matching (the patterns list is already specific enough to avoid false positives).

3. **File path concern**: The diff targets `src/copaw/app/runner/runner.py` — could you confirm this path is correct on the target branch? On `main` the equivalent file appears to be under `src/qwenpaw/`.

Minor nits:
- The lazy import inside the except block is fine for avoiding circular deps, but a comment noting *why* it's there would help future readers.
- Consider logging the original exception at DEBUG level before re-raising, so it's still traceable in logs.

Overall the change is well-scoped and the test coverage looks good. Happy to approve once the exception chain handling is addressed. 👍

PR #851, opened 2026-03-06
fix(dingtalk): correctly skip messages when sender info missing

dipeshbabu:
This fixes a bug in DingTalk sender parsing: `sender_from_chatbot_message()` never returned `should_skip=True` because `suffix` was always non-empty (`"????"`). Now we compute skip based on presence of `sender_id` and `nickname`. Added unit tests for missing sender info, nickname-only, and id suffix behavior.


xieyxclack:
Thank you very much for your contribution!

xieyxclack:
@dipeshbabu Based on your testing, what kinds of messages are skipped? Can you give me an example?

dipeshbabu:
Hi @xieyxclack, thanks for the question.

In this PR, the only messages that become “skipped” are the ones where **both** of the sender fields are missing, meaning:

* `sender_id` is empty or not present, and
* `sender_nick` (nickname) is empty or not present.

These are typically malformed system events or partial payloads rather than real user chat messages.

Concrete examples:

1. **Missing both fields (will be skipped now)**

```json
{
  "sender_id": "",
  "sender_nick": "",
  "text": "hello"
}
```

Previously this would still be processed because the code forced `suffix = "????"` and `skip = not suffix and not nickname` could never become true. After the fix, this case returns `skip=True`.

2. **Nickname present, sender_id missing (not skipped)**

```json
{
  "sender_id": "",
  "sender_nick": "Alice",
  "text": "hello"
}
```

This remains a valid message and is not skipped.

3. **sender_id present, nickname missing (not skipped)**

```json
{
  "sender_id": "abcdef123456",
  "sender_nick": "",
  "text": "hello"
}
```

Also not skipped. The sender will be `unknown#3456`.

So the change is intentionally narrow: we only skip payloads that are missing *all* sender identity information, which previously were incorrectly treated as valid senders like `unknown#????`. I added a unit test to lock this in.

---

If you have a real DingTalk event type that legitimately lacks both fields, I can adjust the condition to also check message type, but I haven’t seen a normal user message come through with both missing.

dipeshbabu:
@xieyxclack could you review?

xieyxclack:
Thank you. Plz remove the test file here

dipeshbabu:
> Thank you. Plz remove the test file here

@xieyxclack just removed the test file.

PR #868, opened 2026-03-07
支持接入AnthropicChatModel；

lizeruicq:
支持AnthropicChatModel下图片的接收和识别，原先接入AnthropicChatModel模型时，发送图片后会报错

### DIFF
--- a/src/copaw/agents/tools/send_file.py
+++ b/src/copaw/agents/tools/send_file.py
@@ -13,6 +13,10 @@ from agentscope.message import (
 )
 
 from ..schema import FileBlock
+from ..utils.message_processing import (
+    _file_to_base64_source,
+    _should_use_base64_for_media,
+)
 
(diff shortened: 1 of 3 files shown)


xieyxclack:
I'm not sure if this is a good idea. If you use base64 for storage, it could potentially cause issues when a tool call returns multimodal information, by putting these base64 results into the context? This is also why the repo uses the file://.

lizeruicq:
> I'm not sure if this is a good idea. If you use base64 for storage, it could potentially cause issues when a tool call returns multimodal information, by putting these base64 results into the context? This is also why the repo uses the file://.

不知是不是目前接入已经支持的claudechatmodel在获取图片时必须是base64？，直接发送图片给copaw会报错invalid request，同时当ai向我返回图片后，再次与copaw对话也会报错

PR #1086, opened 2026-03-09
fix: Handle missing local media files during memory compaction

jsirish:
This PR handles a `TypeError` that occurs at `agentscope.formatter._openai_formatter._to_openai_image_url` during memory compaction in CoPaw when a local image file (e.g., from Telegram) is deleted.

Previously, `_to_openai_image_url` would try to process a non-existent local file URL (e.g. `file://...`), leading to an invalid image format error.

This monkey patch checks for the existence of local media files (`audio`, `image`, `video`) and replaces the block with a text placeholder if the file is missing, preventing CoPaw from crashing during memory compaction.

### DIFF
--- a/src/copaw/agents/model_factory.py
+++ b/src/copaw/agents/model_factory.py
@@ -41,6 +41,9 @@ def _file_url_to_path(url: str) -> str:
     return s
 
 
+import urllib.parse
+import os
+
 def _monkey_patch(func):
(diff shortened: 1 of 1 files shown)


xieyxclack:
Please run `pre-commit run --all-files` to format the code

zhijianma:
Thanks for your contribution. Missing local media files are now replaced with safe text placeholders before formatting in #3635. Please update to the latest version and try it. Closing as completed.

PR #1182, opened 2026-03-10
fix: fuzzy JSON repair and error feedback for malformed tool-call inputs

Atletico1999:
## Description

`_repair_empty_tool_inputs` currently tries only `json.loads(raw_input)` to
recover a `tool_use` block whose `input` is `{}`. When that single attempt
fails (e.g. because the LLM emitted single-quoted keys, unquoted keys, or
trailing commas), it silently logs a warning and leaves `input={}` unchanged.
The agent loop then stalls on the broken tool call with no error surfaced to
the model, which re-generates the same malformed call on every iteration until
`max_iters` is exhausted.

This PR upgrades `_repair_empty_tool_inputs` to a **three-stage pipeline**:

1. **Strict parse** (existing) — `json.loads(raw_input)`.
2. **Fuzzy repair** (new) — `_fuzzy_repair_json()` applies five lightweight
   string fixes in order: BOM/whitespace strip, single-quote→double-quote,
   unquoted-key quoting, trailing-comma removal, and brace-block extraction.
3. **Error-feedback injection** (new) — if both stages above fail, sets
   `input = {"_parse_error": True, "_error_message": "...retry with valid JSON..."}`.
   The tool executor then returns a human-readable error that the model reads
   in the next ReAct step and uses to self-correct.

**Related Issue:** N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

Run the new test file:

```bash
pytest tests/test_tool_message_utils.py -v

### DIFF
--- a/src/copaw/agents/tools/browser_control.py
+++ b/src/copaw/agents/tools/browser_control.py
@@ -17,7 +17,9 @@ import os
 import subprocess
 import sys
 import time
-from concurrent.futures import ThreadPoolExecutor  # pylint: disable=no-name-in-module
+from concurrent.futures import (
+    ThreadPoolExecutor,
+)  # pylint: disable=no-name-in-module
 from typing import Any, Optional
 
 from agentscope.message import TextBlock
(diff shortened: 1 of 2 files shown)


Atletico1999:
"Thanks for the review! 

I've updated the PR to address all the feedback: 

1. Removed raw_input from logging to prevent PII exposure. 
2. Implemented a robust _extract_json_block(text) parser to handle nested braces instead of regex. 
3. Reordered the repair pipeline so that brace extraction happens after the other string formatting fixes. 

The unit tests have also been updated to verify these changes."

PR #1451, opened 2026-03-13
fix(cron): default schedules to local timezone

### LINKED ISSUE #1387: [Question]: 将创建定时任务默认使用本地时区记录到长记忆后，创建任务依然调用UTC时区
看到新版代码中已经支持使用 COPAW_TIMEZONE 环境变量或使用 tzlocal 检测系统本地时区，但是创建定时任务时依然默认为UTC，告诉CoPaw将创建定时任务默认使用本地时区记录到长记忆后依然不能改变，除非每次创建时强制要求本地时间

Image

Image

Image

Alexxigang:
## Summary
- default cron schedules to `COPAW_TIMEZONE` or the detected local timezone instead of hardcoding `UTC`
- apply the same timezone resolution to app startup, service restart, API-backed schedule defaults, and CLI cron creation
- add regression tests for env override, tzlocal detection, datetime fallback, schedule defaults, and CLI payload generation

## Why this fix
`CronManager` was initialized with `timezone="UTC"`, `ScheduleSpec.timezone` defaulted to `UTC`, and `copaw cron create` also defaulted to `UTC`. That meant cron jobs created without an explicit timezone were persisted and executed in UTC even when the machine or `COPAW_TIMEZONE` was set to a local timezone, which matches the bug report in #1387.

This patch centralizes timezone resolution in one helper, prefers `COPAW_TIMEZONE`, falls back to detected local timezone, and only uses `UTC` as the final fallback.

## Validation
- `PYTHONPATH=src python -m pytest tests/unit/app/test_cron_timezone.py -q` -> `9 passed`
- `PYTHONPATH=src python -c "from pathlib import Path; files = [Path('src/copaw/app/crons/timezone_utils.py'), Path('src/copaw/app/crons/models.py'), Path('src/copaw/app/crons/manager.py'), Path('src/copaw/app/_app.py'), Path('src/copaw/cli/cron_cmd.py'), Path('tests/unit/app/test_cron_timezone.py')]; [compile(p.read_text(encoding='utf-8'), str(p), 'exec') for p in files]; print('ok')"` -> `ok`

Closes #1387

### DIFF
--- a/src/copaw/app/_app.py
+++ b/src/copaw/app/_app.py
@@ -29,6 +29,7 @@ from .mcp import MCPClientManager, MCPConfigWatcher  # MCP hot-reload support
 from .runner.repo.json_repo import JsonChatRepository
 from .crons.repo.json_repo import JsonJobRepository
 from .crons.manager import CronManager
+from .crons.timezone_utils import get_default_timezone
 from .runner.manager import ChatManager
 from .routers import router as api_router
 from .routers.voice import voice_router
@@ -94,7 +95,7 @@ async def lifespan(
         repo=repo,
         runner=runner,
         channel_manager=channel_manager,
-        timezone="UTC",
+        timezone=get_default_timezone(),
     )
     await cron_manager.start()
 
@@ -336,7 +337,7 @@ async def lifespan(
             repo=job_repo,
             runner=runner,
             channel_manager=new_channel_manager,
-            timezone="UTC",
+            timezone=get_default_timezone(),
         )
         try:
             await new_cron_manager.start()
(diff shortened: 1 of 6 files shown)


howyoungchen:
Hello, I also noticed this issue and tried to address it in [https://github.com/agentscope-ai/CoPaw/pull/1432](https://github.com/agentscope-ai/CoPaw/pull/1432).

My approach was:

1. Modify the cron skill so that the LLM is explicitly told that the cron tool only accepts UTC time.
2. Insert the current UTC time into the system prompt.
3. Change the time retrieval tool so that it returns UTC time.

Combined with the user’s time zone in the existing `profile.md`, all time-related handling is left to the LLM, which resolves the time inconsistency issue.

My understanding is that your approach is to use the `COPAW_TIMEZONE` environment variable as the time zone for cron input time.

I think this is a very good solution.

However, would users need to add this environment variable manually? If we do decide to use this environment variable, it seems the documentation should also explain this, and all time-related code should be updated consistently, for example:

1. The time retrieval tool
2. The system prompt insertion for time

xieyxclack:
There is a follow-up PR: https://github.com/agentscope-ai/CoPaw/pull/1535

@Alexxigang Any feedback or suggestions are welcome!

lalaliat:

PR #1467, opened 2026-03-14
fix(cron): reject invalid cron expressions on write path, closes #1443

### LINKED ISSUE #1443: Bug: Invalid persisted cron schedule can fail app startup; should degrade to per-job error
## CoPaw Version

from source (branch: feat/upstream/bootstrap-dev, based on latest main around 2026-03-13)

## Description

When persisted cron jobs are loaded at startup, some cron expressions pass basic 5-field checks but fail APScheduler semantic validation. In this case, app startup fails in lifespan instead of isolating the faulty job.

Example problematic expression: 0 */30 8-18 * 1-5

What happened:
- app startup failed
- service was unavailable
- additional async shutdown noise appeared in logs

What I expected:
- app startup should still succeed
- only invalid job should be skipped and marked as error
- create or replace cron paths should reject invalid schedules with clear validation errors

Related PR(s): none yet

Security considerations: none (stability and validation bug)

## Component(s) Affected

- [ ] Skills
- [x] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- CoPaw version: from source (feature branch)
- OS: macOS
- Install method: from source
- Python version: 3.13.1

## Steps to Reproduce

1. Persist a cron job with expression 0 */30 8-18 * 1-5 (for example via existing cron config data).
2. Start app with copaw app.
3. Observe startup failure in cron registration during lifespan.

## Actual vs Expected

- Actual:
  - startup fails due to cron validation error
  - app exits
  - extra async generator shutdown noise appears after the primary failure
- Expected:
  - startup succeeds
  - invalid cron job is skipped and marked error
  - valid jobs continue to run
  - invalid create or replace requests are rejected with actionable 4xx message

## Logs / Screenshots

Key error lines:

ValueError: Error validating expression '*/30': the step value (30) is higher than the total range of the expression (23)
ERROR: Application startup failed. Exiting.

Then secondary noise during shutdown (for example RuntimeError related to cancel scope and async generator closing).

## Additional Notes

Proposed fix:
1. Startup fault isolation at per-job level
- register jobs one by one
- skip invalid job and keep startup successful
- set state as error with reason

2. Add semantic validation on write paths
- API create or replace cron job
- CLI create or replace cron job
- validate by building APScheduler trigger server-side

3. Surface recovery information
- invalid job state should be visible in list or state APIs and console
- no backend auto-rewrite of cron expressions to avoid unintended behavior

futuremeng:
Phase 1 (startup resilience):
- CronManager.start() wraps per-job _register_or_update() in try/except; invalid jobs are marked last_status=error and startup continues.

Phase 2 (write-path validation):
- Add module-level validate_cron_trigger(cron, timezone) that delegates to APScheduler CronTrigger as the authoritative validator.
- Call it in create_or_replace_job() before upsert_job(), so invalid cron expressions are never persisted to jobs.json.
- api.py: POST /cron/jobs and PUT /cron/jobs/{id} now catch ValueError and return HTTP 422 with a descriptive detail message.
- cron_cmd.py: CLI create command performs early local validation before sending, giving immediate feedback without a server round-trip.

Tests:
- test_start_skips_invalid_cron_and_keeps_valid_job
- test_create_or_replace_raises_on_invalid_cron
- test_create_or_replace_raises_without_started

## Description

Invalid cron expressions (e.g. `*/30` in the hour field) were silently accepted by the Pydantic model validator (which only normalises field count) but then rejected by APScheduler's `CronTrigger` at startup, causing `Application startup failed. Exiting.`.

This PR fixes the crash (Phase 1) and closes the write-path gap that allowed the bad expression to be persisted in the first place (Phase 2).

**Related Issue:** Fixes #1443

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [x] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

Run the unit tests:

```bash
pytest tests/unit/app/crons/ -v
```

To reproduce the original crash: persist a job with `cron: "0 */30 * * *"` to `~/.copaw/jobs.json`, then start the app. After this fix the app starts normally and the invalid job shows as `last_status=error`.

```bash
pre-commit run --all-files

pytest tests/unit/app/crons/ -v
# tests/unit/app/crons/test_manager.py::test_start_skips_invalid_cron_and_keeps_valid_job PASSED
# tests/unit/app/crons/test_manager.py::test_create_or_replace_raises_on_invalid_cron PASSED
# tests/unit/app/crons/test_manager.py::test_create_or_replace_raises_without_started PASSED
# 3 passed in 1.06s
```

## Additional Notes

The root cause: `ScheduleSpec.normalize_cron_5_fields` (Pydantic validator) only checks field count; it does not validate APScheduler semantics. A step like `*/30` on the hour field (range 0–23) is structurally valid but semantically rejected by APScheduler. The new `validate_cron_trigger()` helper uses APScheduler itself as the authoritative gate on every write path.

### DIFF
--- a/docs/cron-invalid-schedule-proposal.md
+++ b/docs/cron-invalid-schedule-proposal.md
@@ -1,0 +1,121 @@
+# Cron Invalid Schedule Handling Proposal
+
+## Scope
+
+- Upstream target: agentscope-ai/CoPaw main
+- Local fork impact: keep in sync after upstream discussion/merge
+
+This proposal addresses a startup stability gap: invalid cron expressions in persisted jobs should not fail app startup.
+
+## Problem Statement
+
+When a persisted cron job contains an expression that passes basic shape checks (5 fields) but fails APScheduler semantic validation, app startup can fail during cron registration.
+
+Example:
+- cron: 0 */30 8-18 * 1-5
+- failure: step value 30 exceeds hour range 0-23
+
+Observed effect:
+- app fails in lifespan startup
+- noisy secondary shutdown errors may appear from other async subsystems
+
+## Why This Matters
+
+- Reliability: one bad job should not block all services.
+- Operability: users need clear visibility of invalid jobs and easy recovery.
+- Roadmap alignment: fits self-healing/daemon direction.
+
+## Proposed Design
+
+### 1) Startup resilience (task-level)
+
+At startup, register cron jobs one by one:
+- if valid: schedule normally
+- if invalid: skip scheduling that job, set runtime state to error, keep app startup successful
+
+Status shape (existing model can be reused):
+- last_status = error
+- last_error = invalid schedule: <validation message>
+- next_run_at = null
+
+### 2) Validation on create/update
+
+Add semantic validation before persistence for all write paths:
+- API create/replace cron job
+- CLI cron create (and any edit/replace command path)
+- Console form submit path (server-side is authoritative)
+
+Validation rule:
+- build APScheduler trigger with provided cron/timezone
+- if trigger build fails, reject request with clear 4xx error and actionable message
+
+### 3) Recovery UX
+
+Expose invalid jobs clearly in list/state views:
+- display error status and message
+- provide edit/resave flow to fix and reactivate
+
+No auto-rewrite of expressions in backend.
+
+## Non-Goals
+
+- Automatic transformation of invalid cron expressions
+- Backfilling complex migration for historical job data beyond state marking
+
+## Compatibility
+
+- Backward compatible for valid jobs
+- Invalid historical jobs no longer crash startup
+- Existing storage format remains usable
+
+## Acceptance Criteria
+
+1. App starts successfully even when persisted jobs include invalid cron entries.
+2. Invalid jobs are marked error and omitted from scheduler registration.
+3. Creating/replacing a job with invalid cron returns a clear validation error.
+4. Unit tests cover:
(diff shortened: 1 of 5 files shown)


lalaliat:
Thank you for your contribution!

The reason cron jobs with problematic expressions could previously be created is that, in older versions, the `CronManager` was only started when heartbeat monitoring was enabled. Consequently, in some cases these faulty cron jobs were not being validated or monitored.

In the latest version introducing multi-agent support, the `CronManager` is now enabled by default. As a result, it is no longer possible to create cron jobs with problematic expressions (#1670).

Furthermore, to prevent any legacy invalid cron jobs from blocking the CoPaw startup, we have implemented a mechanism to skip such jobs and set their status to `disabled`. This ensures that CoPaw can now start normally without interruption (#1734).

We truly appreciate your effort and contribution. If you discover any other issues or have further suggestions for improvement, please feel free to share them!

futuremeng:
Closing this PR because the related direction is no longer aligned with the current roadmap, and the branch has drifted significantly from main. We may revisit with a smaller, refreshed PR later if needed.

PR #1468, opened 2026-03-14
fix(chat): copy text content instead of raw response JSON

lllcy:
## Summary
- override the console reply copy action so it copies assistant-readable text instead of the full response JSON payload
- extract assistant `text` and `refusal` content first, and fall back to JSON only when no copyable text exists
- keep the regenerate action and reuse the existing copy success and failure toasts

## Testing
- `uv tool run --from pre-commit pre-commit run --all-files`
- `uv tool run --with pytest --with pytest-asyncio --with pytest-cov pytest` *(fails during collection in the current environment because required project dependencies like `copaw`, `httpx`, and `click` are not installed)*

### DIFF
--- a/console/src/pages/Chat/index.tsx
+++ b/console/src/pages/Chat/index.tsx
@@ -1,21 +1,28 @@
 import {
   AgentScopeRuntimeWebUI,
-  type IAgentScopeRuntimeWebUIOptions,
+  IAgentScopeRuntimeWebUIOptions,
 } from "@agentscope-ai/chat";
-import { useCallback, useMemo, useState } from "react";
-import { Modal, Button, Result, message } from "antd";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


lllcy:
Addressed the actionable review feedback in follow-up commit `eae126c`.

- wrapped `copyResponse` in `useCallback` and added it to the `options` memo dependencies to avoid stale translations and handlers
- hardened the `execCommand("copy")` fallback with `focus()`, boolean result checking, and `try/finally` cleanup
- removed the deep internal type import from `@agentscope-ai/chat` and replaced it with minimal local types for the fields this page actually reads

I also double-checked the regenerate concern against the runtime UI implementation: `actions.list` is combined with the built-in regenerate action when `replace: true`, so this PR still keeps regenerate enabled.

lllcy:
Closing this PR in favor of a fresh branch rebased from the current upstream main. A clean replacement PR will follow.

PR #1471, opened 2026-03-14
fix(chat): copy text content instead of raw response JSON

### LINKED ISSUE #1457: [Bug]:点击复制回复内容应为Markdown格式，但现在点击复制回复的内容是Json格式
## CoPaw Version

[Provide the version of CoPaw you are using, e.g. 0.x.x or git commit hash.]
0.0.7(包括之前所有版本)

## Description
点击复制回复内容应为Markdown格式，但现在点击复制回复的内容是Json格式。

## Logs / Screenshots

Image

lllcy:
## Summary
- override the console reply copy action so it copies assistant-readable text instead of the full response JSON payload
- extract assistant `text` and `refusal` content first, and fall back to JSON only when no copyable text exists
- keep the built-in regenerate action while hardening the clipboard fallback and avoiding stale translation handlers

Fixes #1457

## Testing
- `uv tool run --from pre-commit pre-commit run --files console/src/pages/Chat/index.tsx`

### DIFF
--- a/console/src/pages/Chat/index.tsx
+++ b/console/src/pages/Chat/index.tsx
@@ -23,7 +23,7 @@ type CopyableContent = {
 
 type CopyableMessage = {
   role?: string;
-  content?: CopyableContent[];
+  content?: string | CopyableContent[];
 };
 
 type CopyableResponse = {
@@ -67,7 +67,7 @@ function extractCopyableText(response: CopyableResponse): string {
     return chunks.filter(Boolean).join("\n\n").trim();
   };
 
-  return collectText(true) || collectText(false) || JSON.stringify(response);
+  return collectText(true) || JSON.stringify(response);
 }
 
 async function copyText(text: string) {


lllcy:
Follow-up changes pushed in `e913a5e`.

- removed the unused `index.module.less` side-effect import
- simplified `CopyableContent`
- hardened `extractCopyableText` so it also handles string content and safely skips non-array/non-string content

I did not change `replace: true` because the current `@agentscope-ai/chat` implementation still appends the built-in regenerate action when `replace` is enabled.

xieyxclack:
@lllcy Thank you for your contribution! 

One suggestion: `t("common.copy")` — this key doesn't exist in the locale files (en/zh/ja/ru). The copy button tooltip will show the raw key string instead of localized text. Please add a `"copy"` entry to the `common` section in each locale file.

lllcy:
Thanks for catching that — fixed in `cfa41fd`.

Added `common.copy` to the console locale files:
- `console/src/locales/en.json`
- `console/src/locales/zh.json`
- `console/src/locales/ja.json`
- `console/src/locales/ru.json`

lllcy:
Follow-up changes pushed in `65b26db`.

- updated `CopyableMessage.content` to match the handled runtime shape (`string | CopyableContent[]`)
- restricted the copy fallback so it uses assistant-readable content first and only falls back to raw JSON when no assistant text is available

I still did not change `replace: true` because the current `@agentscope-ai/chat` implementation appends the built-in regenerate action when `replace` is enabled.

PR #1525, opened 2026-03-15
fix(cron): isolate invalid persisted schedules

### LINKED ISSUE #1443: Bug: Invalid persisted cron schedule can fail app startup; should degrade to per-job error

Alexxigang:
## Summary
- skip invalid persisted cron jobs during startup instead of aborting the whole app
- mark skipped jobs with an `error` runtime state so they remain visible for recovery
- validate cron schedules before persisting create/replace requests and return clear 400 errors from the API
- expand regression coverage for startup isolation, write-path validation, and API error translation

## Why this fix
CoPaw currently accepts cron expressions that pass the basic 5-field parser but fail APScheduler's semantic validation. If such a job is already persisted, startup fails while replaying jobs and the whole service becomes unavailable.

This patch isolates startup failures to the offending job and adds write-path validation so invalid schedules are rejected before they can be stored.

Closes #1443.

## Validation
- `PYTHONPATH=src python -m pytest tests/unit/app/test_cron_startup_validation.py tests/unit/app/test_cron_manager_validation.py -q`
- Result: `9 passed`
- `PYTHONPATH=src python -c "from pathlib import Path; files = [Path('src/copaw/app/crons/manager.py'), Path('src/copaw/app/crons/api.py'), Path('tests/unit/app/test_cron_manager_validation.py')]; [compile(p.read_text(encoding='utf-8'), str(p), 'exec') for p in files]; print('ok')"`
- Result: `ok`
- Note: the test files use lightweight stubs for `apscheduler` and `fastapi` because those runtime dependencies are not installed in this local environment; the assertions are focused on CoPaw's startup isolation and API error-mapping logic.

### DIFF
--- a/src/copaw/app/crons/api.py
+++ b/src/copaw/app/crons/api.py
@@ -41,7 +41,10 @@ async def create_job(
     # server generates id; ignore client-provided spec.id
     job_id = str(uuid.uuid4())
     created = spec.model_copy(update={"id": job_id})
-    await mgr.create_or_replace_job(created)
+    try:
+        await mgr.create_or_replace_job(created)
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e)) from e
     return created
 
 
@@ -53,7 +56,10 @@ async def replace_job(
 ):
     if spec.id != job_id:
         raise HTTPException(status_code=400, detail="job_id mismatch")
-    await mgr.create_or_replace_job(spec)
+    try:
+        await mgr.create_or_replace_job(spec)
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e)) from e
     return spec
 
 
--- a/src/copaw/app/crons/manager.py
+++ b/src/copaw/app/crons/manager.py
@@ -60,7 +60,17 @@ class CronManager:
 
             self._scheduler.start()
             for job in jobs_file.jobs:
-                await self._register_or_update(job)
+                try:
+                    await self._register_or_update(job)
+                except Exception as e:  # pylint: disable=broad-except
+                    self._mark_job_error(job.id, e)
+                    logger.warning(
+                        "cron startup: skipping invalid persisted job "
+                        "job_id=%s name=%s error=%s",
+                        job.id,
+                        job.name,
+                        repr(e),
+                    )
 
             # Heartbeat: one interval job when enabled in config
             hb = get_heartbeat_config()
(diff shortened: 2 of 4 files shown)


lalaliat:
Thank you for your contribution! However, after testing this PR, we found that it does not resolve the startup failure issue. Additionally, this problem has already been fixed in the latest version:

1. Cron jobs with problematic expressions can no longer be created (#1670).
2. If any historical invalid cron jobs exist, they are now skipped during startup and their status is set to disabled, ensuring that the CoPaw application starts successfully without interruption (#1734).

Nevertheless, we truly appreciate your effort and contribution. If you discover any other issues or have further suggestions for improvement, please feel free to submit them!

lalaliat:

PR #1538, opened 2026-03-15
fix(formatter): preserve reasoning content on surviving assistant turns

### LINKED ISSUE #1532: reasoning_content injection fails with "Assistant message count mismatch" warning in FileBlockSupportFormatter
## Description

The `FileBlockSupportFormatter._format` method in `copaw/agents/model_factory.py` fails to inject `reasoning_content` into API requests because the assistant message count before and after formatting never matches.

This causes:
1. Every request to log a warning: "Assistant message count mismatch after formatting (X before, Y after). Skipping reasoning_content injection."
2. `reasoning_content` from thinking blocks is never passed to models that support it
3. Excessive log growth

## Root Cause

The code at lines 162-182 in `model_factory.py` assumes the number of assistant messages remains constant after parent class formatting:

```python
if reasoning_contents:
    in_assistant = [m for m in msgs if m.role == "assistant"]
    out_assistant = [m for m in messages if m.get("role") == "assistant"]
    if len(in_assistant) != len(out_assistant):
        logger.warning(
            "Assistant message count mismatch after formatting "
            "(%d before, %d after). "
            "Skipping reasoning_content injection.",
            len(in_assistant),
            len(out_assistant),
        )
```

However, the parent `OpenAIChatFormatter._format` skips "empty" messages (those with only `thinking` blocks but no `content` or `tool_calls`) at line 354-356:

```python
# When both content and tool_calls are None, skipped
if msg_openai["content"] or msg_openai.get("tool_calls"):
    messages.append(msg_openai)
```

This causes `in_assistant` to always be larger than `out_assistant`, preventing `reasoning_content` injection.

## Suggested Fix

Inject `reasoning_content` sequentially to assistant messages based on thinking block order, rather than matching by exact message count:

```python
# Option 1: Sequential injection based on thinking block order
if reasoning_contents:
    out_assistant = [
        m for m in messages if m.get("role") == "assistant"
    ]
    # Inject reasoning_content to assistant messages in order
    reasoning_values = list(reasoning_contents.values())
    for i, out_msg in enumerate(out_assistant):
        if i < len(reasoning_values):
            out_msg["reasoning_content"] = reasoning_values[i]
```

## Environment

- copaw version: 0.0.7
- Python version: 3.12
- agentscope version: 1.0.16dev/1.0.16

## Log Sample

\`\`\`
2026-03-15 21:44:48 | Assistant message count mismatch after formatting (140 before, 135 after). Skipping reasoning_content injection.
2026-03-15 21:45:17 | Assistant message count mismatch after formatting (139 before, 134 after). Skipping reasoning_content injection.
\`\`\`

The count difference is consistently ~3-5 messages per request.

Alexxigang:
## Summary
- preserve `reasoning_content` for assistant messages that are still present after the base formatter runs
- stop treating dropped thinking-only assistant turns as a hard mismatch
- add unit coverage for dropped thinking-only turns and assistant-turn alignment

## Why this fix
`FileBlockSupportFormatter` collected reasoning blocks before calling the base formatter, but then required the input and output assistant counts to match exactly. AgentScope drops assistant turns that only contain thinking blocks, so the counts diverged and CoPaw skipped every `reasoning_content` injection while logging a warning on every request.

This change aligns reasoning injection only against assistant messages that actually survive formatting, which preserves valid reasoning payloads without spamming mismatch warnings.

Fixes #1532.

## Validation
```powershell
$env:PYTHONPATH='D:/project/me/githubAutoPR/copaw-git/src'; pytest tests/unit/agents/test_model_factory_reasoning_content.py -q
```

Result: `2 passed`

### DIFF
--- a/src/copaw/agents/model_factory.py
+++ b/src/copaw/agents/model_factory.py
@@ -101,6 +101,27 @@ def _get_formatter_for_chat_model(
     )
 
 
+def _assistant_message_survives_format(msg: Msg) -> bool:
+    """Best-effort check for assistant messages kept by AgentScope formatters."""
+    if msg.role != "assistant":
+        return False
+
+    if isinstance(msg.content, str):
+        return bool(msg.content)
+
+    for block in msg.get_content_blocks():
+        block_type = block.get("type")
+        if block_type == "thinking":
+            continue
+        if block_type == "text":
+            if block.get("text"):
+                return True
+            continue
+        return True
+
+    return False
+
+
 def _create_file_block_support_formatter(
     base_formatter_class: Type[FormatterBase],
 ) -> Type[FormatterBase]:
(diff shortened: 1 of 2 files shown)


zhijianma:
Thanks for your contribution. Reasoning-content preservation is now supported by the current model pipeline, including #5687. Please update to the latest version and try it. Closing as completed.

PR #1557, opened 2026-03-16
fix(agents): inject reasoning_content sequentially to avoid count mismatch

### LINKED ISSUE #1532: reasoning_content injection fails with "Assistant message count mismatch" warning in FileBlockSupportFormatter

mvanhorn:
## Summary

`FileBlockSupportFormatter._format` in `model_factory.py` never injects `reasoning_content` into API requests because the assistant message count before and after formatting never matches. Every request logs:

```
Assistant message count mismatch after formatting (140 before, 135 after). Skipping reasoning_content injection.
```

The root cause: the parent `OpenAIChatFormatter._format` drops assistant messages that contain only thinking blocks (no `content` or `tool_calls`), so `out_assistant` is always smaller than `in_assistant`.

## Changes

- **`model_factory.py`**: Replace the count-based guard with sequential injection. Reasoning values are collected from input assistant messages in order, then assigned to output assistant messages by index. This handles the count difference from dropped thinking-only messages without skipping injection.

## Before/After

**Before**: `reasoning_content` is never injected. Warning logged on every request. Models that support extended thinking never receive the reasoning chain.

**After**: `reasoning_content` is injected to output assistant messages in order. No warning logged. Models receive the reasoning chain as expected.

Fixes #1532

This contribution was developed with AI assistance (Claude Code).


mvanhorn:
Addressed in ba7119d. The sequential injection was indeed misaligning reasoning with the wrong messages. Now predicts which assistant messages survive the parent formatter (drops thinking-only messages without tool_calls) and injects reasoning only into the aligned output. Re-added the count mismatch safeguard as a fallback.

src/copaw/agents/model_factory.py
```diff
@@ -160,26 +160,40 @@ async def _format(self, msgs):
...
+                # predict survivors and collect reasoning only for those.
+                aligned_reasoning = []
+                for m in (msg for msg in msgs if msg.role == "assistant"):
+                    is_thinking_only = (
+                        isinstance(m.content, list)
+                        and m.content
+                        and all(b.get("type") == "thinking" for b in m.content)
+                    )
+                    if not (
+                        is_thinking_only and not getattr(m, "tool_calls", None)
```
qbc2016: `not getattr(m, "tool_calls", None)` is redundant — `Msg` has no `tool_calls` attribute (it always evaluates to `None`). Please simplify to `if not is_thinking_only`.

mvanhorn:
Ran `pre-commit run --all-files` and fixed two issues in fe8d8f5:
- Replaced `str | list[dict]` union syntax with `Union[str, List[dict]]` (mypy v1.7 defaults to pre-3.10 syntax rules)
- Added pylint disable for too-many-statements on the factory function

All hooks pass now (mypy, black, flake8, pylint, prettier).

src/copaw/agents/model_factory.py
mvanhorn: Simplified in ea44e2a - removed the redundant `getattr(m, "tool_calls", None)` check since `Msg` has no `tool_calls` attribute.

qbc2016 (commented):
Please merge main and fix the conflicts. Others lgtm.

mvanhorn:
Thanks for the review and merge!

PR #1600, opened 2026-03-16
fix: handle None agents.defaults in get_heartbeat_config and always start CronManager

webnoise:
## Problem

Application startup fails with:
```
AttributeError: 'NoneType' object has no attribute 'heartbeat'
```
This happens because `config.agents.defaults` is `None` by default (users who haven't configured the `defaults` section), and `get_heartbeat_config()` accesses `.heartbeat` without a None guard.

Additionally, `CronManager.start()` was only called when `heartbeat.enabled == True`, which means user-defined cron jobs would never be scheduled when heartbeat is disabled.

## Fix

- `src/copaw/config/utils.py`: Add `None` check for `config.agents.defaults` before accessing `.heartbeat`
- `src/copaw/app/workspace.py`: Always call `cron_manager.start()`; heartbeat enablement is already controlled internally inside `CronManager.start()` via `get_heartbeat_config()`

### DIFF
--- a/src/copaw/config/utils.py
+++ b/src/copaw/config/utils.py
@@ -363,10 +363,8 @@ def save_config(config: Config, config_path: Optional[Path] = None) -> None:
 def get_heartbeat_config() -> HeartbeatConfig:
     """Return effective heartbeat config (from file or default 30m/main)."""
     config = load_config()
-    if config.agents.defaults is None:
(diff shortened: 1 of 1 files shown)


zhijianma:
@webnoise 

Thank you for you contribution to CoPaw. 

Plz resovle the conflicts.

PR #1848, opened 2026-03-19
feat(mcp): improve import templates and preserve description

futuremeng:
## Summary
- add click-to-fill templates in MCP create modal for Standard / Direct / Single formats
- add click-to-fill streamable_http remote template
- preserve imported description with robust normalization (supports description/desc/remark)
- accept `streamable-http` transport alias in backend MCP config normalization
- update MCP docs (zh/en) with richer import examples and remote template

## Validation
- npm --prefix console run build


zhijianma:
@futuremeng 

I think we need to support and be compatible with more types of MCP configurations. This template doesn't seem very useful since people just copy and paste anyway. I believe it's more important to focus on implementing robust validation checks.


image

PR #1896, opened 2026-03-20
fix: support audio content using top-level data field

### LINKED ISSUE #1516: [Bug]: AudioContent not supported in Telegram channel - Fix
# Bug Report: AudioContent not supported in Telegram channel

## Summary

When a user sends a voice message in Telegram, CoPaw cannot process it because `AudioContent` is not properly converted to a format that the LLM can understand.

## Problem

The issue occurs in the message formatting pipeline:

1. **Telegram channel** receives voice message → downloads as `.oga` file
2. **Channel** creates `AudioContent(type='audio', data='/path/to/file.oga', format='ogg')`
3. **agentscope_runtime** converts `AudioContent.model_dump()` → `{'type': 'audio', 'data': '/path/to/file.oga', 'format': 'ogg'}`
4. **Formatter** receives block with `{'type': 'audio', 'source': {'type': 'base64', 'data': '/path/to/file.oga', 'media_type': 'audio/ogg'}}`
5. **Somewhere** the code tries `base64.b64decode('/path/to/file.oga')` → `Incorrect padding` error

## Error

```
Failed to download file from base64: Incorrect padding
Failed to process audio block: Incorrect padding
```

## Root Cause

The `agentscope_runtime` or `agentscope` creates a `source` dict with `type='base64'` but puts a **file path** in the `data` field instead of actual base64-encoded content.

Expected:
```python
{'type': 'audio', 'source': {'type': 'base64', 'data': 'base64_encoded_string...', 'media_type': 'audio/ogg'}}
```

Actual:
```python
{'type': 'audio', 'source': {'type': 'base64', 'data': '/root/.copaw/media/telegram/xxx.oga', 'media_type': 'audio/ogg'}}
```

## Workaround

Currently, the workaround is to transcribe audio in the Telegram channel itself (before passing to the agent):

```python
# In channel.py
if local_path:
    response = requests.post(
        "http://host.docker.internal:8000/v1/audio/transcriptions",
        files={"file": open(local_path, "rb")},
        data={"model": "systran/faster-whisper-base"}
    )
    text = response.json().get("text", "")
    content_parts.append(TextContent(type=TEXT, text=f"[Voice: {text}]"))
```

### Option 1: Fix in agentscope_runtime

In the conversion from `AudioContent` to block dict, check if `data` is a file path and convert to base64:

```python
# In agentscope_runtime/engine/schemas/agent_schemas.py or similar
def audio_content_to_block(content: AudioContent) -> dict:
    if os.path.exists(content.data):
        with open(content.data, "rb") as f:
            base64_data = base64.b64encode(f.read()).decode("utf-8")
        return {
            'type': 'audio',
            'source': {
                'type': 'base64',
                'data': base64_data,
                'media_type': f'audio/{content.format}'
            }
        }
    # ... handle existing base64 data
```

### Option 2: Support file paths in formatter

In `_to_openai_audio_data()` in `agentscope/formatter/_openai_formatter.py`:

```python
def _to_openai_audio_data(source: dict) -> dict:
    if source.get("type") == "base64" and source.get("data"):
        data = source["data"]
        # Check if data is a file path (not actual base64)
        if os.path.exists(data):
            with open(data, "rb") as f:
                data = base64.b64encode(f.read()).decode("utf-8")
        # ... rest of processing
```

## Files to Modify

- `agentscope_runtime/engine/schemas/agent_schemas.py` - AudioContent conversion
- OR `agentscope/formatter/_openai_formatter.py` - `_to_openai_audio_data()` function

## Testing

1. Send voice message to Telegram bot
2. Agent should receive transcribed text or audio content
3. No "Incorrect padding" error

## Impact

This feature is needed by many users who want to:
- Send voice messages in Telegram
- Use voice notes instead of typing
- Support accessibility features

## Environment

- CoPaw v0.0.7
- AgentScope 1.0+
- python-telegram-bot 20+
- Docker on Windows

sidonsoft:
## Summary

Fix audio/voice message processing when an audio block uses a top-level `data` field instead of a `source` dict.

This affects the path used by Telegram voice/audio content in `v0.1.0`, where `AudioContent` is created with `data=...` but `message_processing.py` only looks for `source`.

## What changed

- Added compatibility handling in `src/copaw/agents/utils/message_processing.py`:
  - for `block_type == "audio"`, if `source` is missing but `data` is present, normalize it into a `{"type": "url", "url": ...}` source
- Added targeted unit tests covering:
  - `_extract_source_and_filename()` with Telegram-style audio blocks
  - `_process_single_block()` using an audio block with `data=file://...`

## Why this approach

This is a smaller, safer compatibility fix than changing Telegram channel output. It also makes the processing layer more data-agnostic for any path that emits `AudioContent(data=...)`.

## Validation

Targeted tests:

```bash
.venv/bin/python -m pytest -q tests/unit/agents/utils/test_message_processing.py
```

Result: `2 passed`

Additional sanity slice:

```bash
.venv/bin/python -m pytest -q tests/unit/workspace/test_prompt.py tests/unit/cli/test_cli_version.py
```

Result: `6 passed`

Closes #1516

### DIFF
--- a/src/copaw/agents/utils/message_processing.py
+++ b/src/copaw/agents/utils/message_processing.py
@@ -120,6 +120,7 @@ def _media_type_from_path(path: str) -> str:
         ".mp3": "audio/mp3",
         ".opus": "audio/opus",
         ".ogg": "audio/ogg",
+        ".oga": "audio/ogg",
         ".flac": "audio/flac",
         ".m4a": "audio/mp4",
         ".aac": "audio/aac",
@@ -328,19 +329,30 @@ async def _process_single_block(
     if source is None:
         return None
 
-    # Normalize: when source is "base64" but data is a local path (e.g.
-    # DingTalk voice returns path), treat as url only if under allowed dir.
+    # Normalize: when source is "base64" but data is really a local file
+    # path or file:// URI (for example some voice-message paths), treat it
+    # as a URL source instead of trying to base64-decode it.
     if (
         block_type == "audio"
         and isinstance(source, dict)
         and source.get("type") == "base64"
     ):
         data = source.get("data")
-        if isinstance(data, str) and os.path.isfile(data):
+        local_path = None
+        if isinstance(data, str) and data:
+            if os.path.isfile(data):
+                local_path = data
+            else:
+                parsed = urllib.parse.urlparse(data)
+                if parsed.scheme == "file":
+                    candidate = urllib.parse.unquote(parsed.path)
+                    if os.path.isfile(candidate):
+                        local_path = candidate
+        if local_path:
             block["source"] = {
                 "type": "url",
-                "url": Path(data).as_uri(),
-                "media_type": _media_type_from_path(data),
+                "url": Path(local_path).as_uri(),
+                "media_type": _media_type_from_path(local_path),
             }
             source = block["source"]
 
@@ -412,6 +424,10 @@ async def process_file_and_media_blocks_in_message(msg) -> None:
         downloaded_files = []
 
         for i, block in enumerate(message.content):
+            # Convert Pydantic model objects (e.g., AudioContent, ImageContent
+            # from agentscope_runtime) to dicts for uniform processing.
+            if hasattr(block, "model_dump"):
+                block = block.model_dump()
             if not isinstance(block, dict):
                 continue
 
--- a/src/copaw/agents/utils/message_processing.py
+++ b/src/copaw/agents/utils/message_processing.py
@@ -91,9 +91,7 @@ def _extract_source_and_filename(block: dict, block_type: str):
         if isinstance(data, str) and data:
             parsed = urllib.parse.urlparse(data)
             filename = (
-                os.path.basename(parsed.path)
-                or os.path.basename(data)
-                or None
+                os.path.basename(parsed.path) or os.path.basename(data) or None
             )
             return {"type": "url", "url": data}, filename
 
@@ -109,6 +107,44 @@ def _extract_source_and_filename(block: dict, block_type: str):
             filename = os.path.basename(parsed.path) or None
 
     return source, filename
+
+
+def _normalize_audio_source_for_local_file(
+    block_type: str,
+    block: dict,
+    source: dict,
+):
+    """Normalize audio blocks whose payload is a local path or file URI."""
+    if block_type != "audio" or not isinstance(source, dict):
+        return source
+
+    if source.get("type") != "base64":
+        return source
+
+    data = source.get("data")
+    if not isinstance(data, str) or not data:
+        return source
+
+    local_path = None
+    if os.path.isfile(data):
+        local_path = data
+    else:
+        parsed = urllib.parse.urlparse(data)
+        if parsed.scheme == "file":
+            candidate = urllib.parse.unquote(parsed.path)
+            if os.path.isfile(candidate):
+                local_path = candidate
+
+    if not local_path:
+        return source
+
+    normalized = {
+        "type": "url",
+        "url": Path(local_path).as_uri(),
+        "media_type": _media_type_from_path(local_path),
+    }
+    block["source"] = normalized
+    return normalized
 
 
 def _media_type_from_path(path: str) -> str:
@@ -309,6 +345,7 @@ async def _process_audio_block(
     return False
 
 
+# pylint: disable=too-many-branches
 async def _process_single_block(
     message_content: list,
     index: int,
--- a/src/copaw/agents/utils/message_processing.py
+++ b/src/copaw/agents/utils/message_processing.py
@@ -74,9 +74,28 @@ async def _process_single_file_block(
 
 
 def _extract_source_and_filename(block: dict, block_type: str):
-    """Extract source and filename from a block."""
+    """Extract source and filename from a block.
+
+    Supports both the historical ``source``-dict format and audio blocks
+    that carry their payload in a top-level ``data`` field (for example
+    Telegram `AudioContent(data=...)`).
+    """
     if block_type == "file":
         return block.get("source", {}), block.get("filename")
+
+    # Compatibility: some audio content blocks use a top-level `data`
+    # field instead of a `source` dict. Treat the data value as a URL or
+    # local path and normalize it into the same shape used elsewhere.
+    if block_type == "audio" and not block.get("source"):
+        data = block.get("data")
+        if isinstance(data, str) and data:
+            parsed = urllib.parse.urlparse(data)
+            filename = (
+                os.path.basename(parsed.path)
+                or os.path.basename(data)
+                or None
+            )
+            return {"type": "url", "url": data}, filename
 
     source = block.get("source", {})
     if not isinstance(source, dict):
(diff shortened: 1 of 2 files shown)


sidonsoft:
Follow-up: I found and fixed the actual live Telegram voice-note path too.

In the running app, the failing block was arriving as:

```python
{
  "type": "audio",
  "source": {
    "type": "base64",
    "media_type": "audio/None",
    "data": "file:///.../telegram/voice-....oga"
  }
}
```

So there were really two compatibility issues:

1. audio blocks using top-level `data`
2. audio blocks using `source.type == "base64"` where `source.data` is actually a `file://...` URI, not real base64

This PR now handles both cases by normalizing local paths / file URIs into URL-style sources before the base64 decoder is reached.

Also fixed `.oga` media type normalization so Telegram voice-note files map to `audio/ogg` instead of `audio/octet-stream`.

Updated validation:

```bash
.venv/bin/python -m pytest -q tests/unit/agents/utils/test_message_processing.py
# 3 passed

.venv/bin/python -m pytest -q tests/unit/workspace/test_prompt.py tests/unit/cli/test_cli_version.py
# 6 passed
```

sidonsoft:
Quick reviewer note: the current PR now covers **all of the failure shapes I saw locally**, not just one Telegram variant.

The compatibility handling now covers:

1. **Audio blocks with top-level `data`**
   - e.g. `{"type": "audio", "data": ...}`

2. **Audio blocks with `source.type == "base64"` where `source.data` is actually a local path / `file://...` URI**
   - this was the real live Telegram voice-note failure path in my logs

3. **Pydantic content objects that need `model_dump()` before processing**
   - `process_file_and_media_blocks_in_message()` now converts model-like blocks to dicts before the normal media path runs

4. **`.oga` media type normalization**
   - mapped to `audio/ogg` so Telegram voice-note files don’t fall through as `audio/octet-stream`

So the intent of the PR is:
- keep channel-side behavior unchanged
- make message processing more tolerant / data-agnostic
- normalize multiple equivalent audio-content shapes into one common handling path

Targeted tests were added for:
- top-level `data`
- `base64` + `file://...` source payloads
- audio normalization through `_process_single_block()`

zhijianma:
@sidonsoft 

Plz see my [ test
](https://github.com/agentscope-ai/CoPaw/issues/1516#issuecomment-4094672436)

sidonsoft:
Thanks — yes, this aligns with the intended behavior.

The important point is that the processing layer should normalize audio inputs based on what the payload **actually is**, not just the nominal `source.type`.

So if an audio block comes through as:

- `source.type == "base64"`
- but `source.data` is actually a local file path (or `file://` URI)

then converting it into a URL-style source before downstream processing is the correct compatibility behavior.

That is also why I’ve been framing this as a **message-processing normalization** issue rather than a Telegram-only workaround. Your console-uploaded MP3 example shows the same mismatch can appear in other paths too, not just Telegram voice notes.

So from my side, your test is a good confirmation that the normalization approach is the right shape for the fix.

zhijianma:
@sidonsoft 

You may have overlooked a crucial processing step.

In the runner, there is a conversion from Runtime Message to AgentScope Message that transforms AudioContent into AudioBlock.


```python

type_mapping = {
                "text": (TextBlock, "text"),
                "image": (ImageBlock, "image_url"),
                "audio": (AudioBlock, "data"),
                "data": (TextBlock, "data"),
                "video": (VideoBlock, "video_url"),
                "file": (FileBlock, "file_url"),
            }
....
....
for cnt in message.content:   # AudioContent  in AgenetScope-Runtime
      cnt_type = cnt.type or "text" 
      block_cls, attr_name = type_mapping[cnt_type]  #AudioBlock in AgentScope
      value = getattr(cnt, attr_name)
      elif cnt_type == "audio":
                    if (
                        value
                        and isinstance(value, str)
                        and value.startswith(
                            "data:",
                        )
                    ):
                        mediatype_part = value.split(";")[0].replace(
                            "data:",
                            "",
                        )
                        base64_data = value.split(",")[1]
                        base64_source = Base64Source(
                            type="base64",
                            media_type=mediatype_part,
                            data=base64_data,
                        )
                        msg_content.append(
                            block_cls(type=cnt_type, source=base64_source),
                        )
                    else:
                        parsed_url = urlparse(value)
                        if parsed_url.scheme and parsed_url.netloc:
                            url_source = URLSource(type="url", url=value)
                            msg_content.append(
                                block_cls(type=cnt_type, source=url_source),
                            )
                        else:
                            audio_extension = getattr(cnt, "format")
                            base64_source = Base64Source(
                                type="base64",
                                media_type=f"audio/{audio_extension}",
                                data=value,
                            )
                            msg_content.append(
                                block_cls(type=cnt_type, source=base64_source),
                            )
```

sidonsoft:
## Upstream Fix

I've opened a matching PR in **agentscope-runtime** that fixes the root cause identified in https://github.com/agentscope-ai/CoPaw/pull/1896#issuecomment-4096810319:

**https://github.com/agentscope-ai/agentscope-runtime/pull/466**

### Summary

The runner's `message_to_agentscope_msg()` now detects local file paths before falling back to base64:

```python
if value and isinstance(value, str) and os.path.isfile(value):
    # Local file path → convert to file:// URL
    url_source = URLSource(
        type="url",
        url=Path(value).as_uri(),
        media_type=f"audio/{audio_extension}" if audio_extension else None,
    )
```

### Coordination

| PR | Location | Purpose |
|----|----------|---------|
| **#466** (agentscope-runtime) | Root cause | Produces correct blocks from the start |
| **#1896** (CoPaw) | Downstream | Defensive handling for legacy/edge cases |

Both PRs should be merged. The downstream fix in this PR remains valuable for:
1. Backwards compatibility with older runtime versions
2. Defense-in-depth against similar issues in other code paths

sidonsoft:
## Bug Found: Bare paths returned without file:// scheme

The `_extract_source_and_filename` function returns bare paths like `/tmp/voice.ogg` as `{"type": "url", "url": data}` without normalizing to `file://` URLs. This causes downstream `download_file_from_url()` to fail.

### The Problem

```python
return {"type": "url", "url": data}, filename  # data = "/tmp/voice.ogg"
```

This gets passed to `_process_single_file_block` which sees `parsed.scheme == ""` and tries to fetch it as a remote URL.

### Fix

Normalize bare local paths to `file://` URLs:

```python
if parsed.scheme and parsed.netloc:
    # Full URL (https://, http://, etc.)
    return {"type": "url", "url": data}, filename
elif parsed.scheme == "file":
    # Already a file:// URL
    return {"type": "url", "url": data}, filename
elif os.path.isfile(data):
    # Bare local path → convert to file:// URL
    return {
        "type": "url",
        "url": Path(data).as_uri(),
        "media_type": _media_type_from_path(data),
    }, filename
else:
    # Unknown - pass through as-is (may be base64 or invalid)
    return {"type": "url", "url": data}, filename
```

I've pushed this fix to a branch on my fork. The fix should be incorporated into this PR before merging.

zhijianma:
@sidonsoft 

So, do you think this PR still needs to be merged?

zhijianma:
@sidonsoft 

Invalid

PR #2045, opened 2026-03-22
feat(providers): add multimodal probe validation and capability baseline

leoleils:
## Description

Add multimodal probe validation with capability baseline comparison for all built-in providers. This enables:
- Structured logging (timing + result) for all probe operations
- Documentation-based default annotations for unprobed models
- Automatic discrepancy detection between probe results and expected baselines
- Ollama multimodal probing via OpenAI-compatible endpoint

**Related Issue:** Relates to multimodal model support feature

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

```bash
# Run all unit tests
pytest tests/unit/ -x -q

# Run specific test files for this feature
pytest tests/unit/providers/test_capability_baseline.py \
       tests/unit/providers/test_capability_baseline_props.py \
       tests/unit/providers/test_multimodal_prober.py \
       tests/unit/providers/test_ollama_provider.py \
       tests/unit/providers/test_provider_manager.py -v

### DIFF
--- a/console/src/pages/Chat/index.tsx
+++ b/console/src/pages/Chat/index.tsx
@@ -589,9 +589,7 @@ export default function ChatPage() {
               const isLt10M = file.size / 1024 / 1024 < 10;
               if (!isLt10M) {
                 message.error(t("chat.attachments.fileSizeLimit"));
-                return options.onError?.(
-                  new Error("File size exceeds 10MB"),
-                );
+                return options.onError?.(new Error("File size exceeds 10MB"));
               }
 
               options.onProgress?.({ percent: 0 });
@@ -599,9 +597,7 @@ export default function ChatPage() {
               options.onProgress?.({ percent: 100 });
               options.onSuccess({ url: chatApi.fileUrl(res.url) });
             } catch (e) {
-              options.onError?.(
-                e instanceof Error ? e : new Error(String(e)),
-              );
+              options.onError?.(e instanceof Error ? e : new Error(String(e)));
             }
           },
         },

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -89,3 +89,17 @@ CLAUDE.md
...
 .cache/  # Includes conda_unpack_wheels/ for Windows packaging workaround
+
+# Hypothesis test cache
+.hypothesis/
+src/.hypothesis/
+
+# Kiro IDE
+.kiro/
+
+# Stray image/doc files (not part of source)
```
pan-x-c: Please remove these file-specific annotations. If they are indeed necessary, please restrict the file paths—specifically by utilizing single folders and regular expressions—to minimize the number of entries.

src/copaw/providers/gemini_provider.py
```diff
@@ -128,3 +138,194 @@ def get_chat_model_instance(self, model_id: str) -> ChatModelBase:
...
+                result = True, f"Image supported (answer={answer!r})"
+                elapsed = time.monotonic() - start_time
+                logger.info(
+                    "Image probe done: model=%s result=%s %.2fs",
+                    model_id,
+                    result[0],
+                    elapsed,
+                )
+                return result
+            result = (
```
pan-x-c, line 213: Whether a model can correctly identify images should not serve as the basis for determining whether it has the capability.

src/copaw/providers/multimodal_prober.py
```diff
@@ -0,0 +1,440 @@
...
+    supports_video: bool = False
+    image_message: str = ""
+    video_message: str = ""
+
+    @property
+    def supports_multimodal(self) -> bool:
+        return self.supports_image or self.supports_video
+
+
+async def probe_image_support(
```
pan-x-c: This method is only for `openai`, which should be moved into `OpenAIProvider`

src/copaw/providers/multimodal_prober.py
```diff
@@ -0,0 +1,440 @@
...
+    model_id: str,
+    video_url: str,
+    timeout: float,
+    *,
+    start_time: float,
+) -> tuple[bool, str] | None:
+    """Try a single video URL format. Return None to try next."""
+    is_http = video_url == _PROBE_VIDEO_URL
+    req_timeout = timeout * 3 if is_http else timeout
+    client = AsyncOpenAI(
```
pan-x-c: Same as above

src/copaw/providers/multimodal_prober.py
```diff
@@ -0,0 +1,440 @@
...
+        base_url,
+    )
+    start_time = time.monotonic()
+    client = AsyncOpenAI(base_url=base_url, api_key=api_key, timeout=timeout)
+    try:
+        # Use a generous max_tokens because "thinking" models (e.g. Kimi K2.5)
+        # consume tokens for internal reasoning_content before producing the
+        # visible answer in `content`.  With too few tokens the reasoning
+        # exhausts the budget and `content` comes back empty.
+        res = await client.chat.completions.create(
```
pan-x-c: Perhaps the API-related code could be abstracted into helper functions embedded within the provider class, and you can directly invoke the methods provided by that class here.

.gitignore
leoleils: 🙆🏻‍♂️OK

src/copaw/providers/gemini_provider.py
leoleils: Good point — however, the semantic check (verifying the model actually "sees" the image) is intentional. Some OpenAI-compatible APIs silently accept image content in the request without returning a 400 error, but the underlying text-only model simply ignores the image payload. If we only check whether the API rejects the request, these models would be falsely classified as multimodal-capable.

The two-stage approach addresses this: (1) if the API returns 400 or a media-keyword error, it's clearly not supported; (2) if the API accepts the request, we verify the model can actually perceive the image content by asking it to identify a solid-red test image. This filters out "fake multimodal" endpoints that accept but ignore image input.

That said, I agree the current check is strict — a model that genuinely supports vision but gives an inaccurate color answer would be incorrectly marked as unsupported. I'm open to relaxing the verification (e.g. accepting any non-empty response that isn't a refusal) if you think that's a better tradeoff. What do you suggest?

leoleils: In my tests, Qwen 3max accepts base64-encoded images without returning errors—which is precisely why probing scheme includes actual semantic image verification.

zhijianma:
@leoleils 
Hi, please execute `npm run format` in console directory.

pan-x-c:
Using the following command to fix npm format error

```
cd console
npx prettier --write pages/Chat
```

PR #2260, opened 2026-03-25
feat(channel): add WeChat iLink Bot channel

livehl:
## Summary

Add WeChat iLink Bot channel support, allowing CoPaw to connect to WeChat via the official iLink Bot protocol.

## Features

- **QR code scan login**: Generates a QR code for WeChat login, maintains session via long-polling event loop
- **Receive messages**: Text, image, voice (SILK format), file, and video messages
- **Send messages**: Text replies with proper `client_id` and `base_info` fields
- **Media decryption**: AES-128-ECB CDN media download and decryption, aligned with official SDK `parseAesKey` logic (supports hex key format and base64 key format)
- **Frontend UI**: Added WeChat channel option in console ChannelDrawer with i18n support (zh/en/ja/ru)

## Implementation

- Implemented against official SDK `@tencent-weixin/openclaw-weixin@2.0.1`
- `channel.py`: Main channel class with login, polling, and message dispatch
- `client.py`: HTTP client wrapping iLink Bot API endpoints
- `utils.py`: AES-ECB decryption utility with multi-format key support
- Registered in `channels/registry.py` and config fields added to `config.py`

## Testing

End-to-end tested:
- Text message receive & echo reply ✅
- Image download & decrypt (JPEG) ✅
- File download & decrypt (PDF) ✅
- Voice download & decrypt (SILK, ASR verified) ✅

### DIFF
--- a/src/copaw/app/channels/weixin/channel.py
+++ b/src/copaw/app/channels/weixin/channel.py
@@ -64,7 +64,7 @@ class WeixinChannel(BaseChannel):
         - Group chat:    weixin:group:<group_id>
     """
 
-    channel = "weixin"
+    channel = "wechat"
 
     def __init__(
         self,
@@ -447,14 +447,20 @@ class WeixinChannel(BaseChannel):
                     msgs: List[Dict[str, Any]] = data.get("msgs") or []
                     for msg in msgs:
                         await self._on_message(msg, client)
-                    # Only warn+retry on non-zero ret when no messages
+                    # ret=-1 is normal long-poll timeout (no new messages)
                     if ret != 0 and not msgs:
-                        logger.warning(
-                            "weixin getupdates non-zero ret=%s"
-                            " (no msgs), retry in 3s",
-                            ret,
-                        )
-                        await asyncio.sleep(3)
+                        if ret == -1:
+                            logger.debug(
+                                "weixin getupdates timeout (ret=-1)"
+                                ", continue polling",
+                            )
+                        else:
+                            logger.warning(
+                                "weixin getupdates non-zero ret=%s"
+                                " (no msgs), retry in 3s",
+                                ret,
+                            )
+                            await asyncio.sleep(3)
                 except asyncio.CancelledError:
                     break
                 except Exception:


hongxicheng:
I just tested this PR, and the QR code wasn't printed when I first launched the WeChat channel, and there were errors in the code.

src/copaw/app/channels/weixin/channel.py
```diff
@@ -0,0 +1,785 @@
...
+from ....constant import DEFAULT_MEDIA_DIR
+from ..base import (
+    BaseChannel,
+    ContentType,
+    OnReplySent,
+    OutgoingContentPart,
+    ProcessHandler,
+)
+from ..utils import split_text
+from .client import ILinkClient
```
hongxicheng:
```suggestion
from .client import ILinkClient, _DEFAULT_BASE_URL
```

src/copaw/app/channels/weixin/channel.py
```diff
@@ -0,0 +1,785 @@
...
+            filter_thinking=filter_thinking,
+            dm_policy=dm_policy,
+            group_policy=group_policy,
+            allow_from=allow_from,
+            deny_message=deny_message,
+        )
+        self.enabled = enabled
+        self.bot_token = bot_token
+        self.bot_prefix = bot_prefix
+        self._base_url = base_url or ""
```
hongxicheng:
```suggestion
        self._base_url = base_url or _DEFAULT_BASE_URL
```

hongxicheng:
After I modified the two pieces of code above, I could see the QR code link in the log, but it was not saved to copaw after I copied it to the browser and scanned it.

livehl:
## Update: QR Code Login UI & Bug Fixes

Following the initial review and testing feedback, here is a summary of the recent changes:

### 🐛 Bug Fixes

**QR code not displaying (ORB blocking)**
- WeChat API returns a URL for `qrcode_img_content` instead of a base64 image. Direct `` loading was blocked by the browser's Opaque Response Blocking (ORB) policy.
- Fix: generate the QR code image **server-side** using [`segno`](https://github.com/heuer/segno) (pure Python, no binary deps) from the `qrcode` string, return it as base64 PNG. No cross-origin requests needed.

**Token not saved after scanning**
- After QR login, `self._base_url` was not updated with the `baseurl` returned by the confirmed response. The polling thread created a new `ILinkClient` using the stale `_base_url`, causing subsequent message polling to fail.
- Fix: propagate `base_url` back to `self._base_url` immediately after login confirmation.

**Multiple "login success" toasts**
- The `setInterval` poll callback could fire multiple times before `clearInterval` took effect, showing duplicate success notifications.
- Fix: added a `weixinConfirmedRef` boolean ref as an atomic guard; subsequent callbacks return early if already confirmed.

### ✨ New Features

**Token auto-fill after scan**
- Added `GET /config/channels/weixin/qrcode/status?qrcode=xxx` backend endpoint that proxies `get_qrcode_status` from the iLink API.
- Frontend polls this endpoint every 2 seconds after the QR code is shown. On `status === "confirmed"`, `bot_token` is automatically filled into the form field and the QR code is hidden. User just clicks Save.

### 🔧 Refactoring

- Moved lazy imports (`uuid`, `urllib.parse.quote`) to module top-level in `client.py`.
- Extracted repeated `base_url` resolution logic into `_get_weixin_base_url()` helper in the config router, eliminating duplication between the two QR endpoints.

hongxicheng:
Hello, I just tested it and the function works fine.would it be convenient to improve the document and add a redirect link in the WeChat channel settings interface？

livehl:
Thanks for the review and testing, @hongxicheng!

## Changes in response to your feedback

### 1. Documentation

Added a dedicated **WeChat Personal (iLink)** section to both English and Chinese channel docs:

- **How it works** — QR code login flow, token persistence, long-polling receive, text-only send
- **QR code login via Console** — step-by-step guide for the scan-to-login UI
- **Config file reference** — full field table (bot_token, bot_token_file, base_url, media_dir, dm_policy, group_policy, allow_from) with defaults
- **Environment variable reference** — for headless / Docker deployments
- Updated the **config overview table** and **multimodal support table** in the appendix

### 2. Redirect link in the settings UI

Added doc redirect links for the WeChat Personal channel in `ChannelDrawer.tsx`:
- Users can now click the doc icon in the channel settings panel to jump directly to the relevant docs section

Commit: `0ed9d4f`

xutianfu123123:
Waiting for WeChat Channel

xieyxclack:
@livehl Thank you for your contribution. Please pass the format checks. Refer to https://github.com/agentscope-ai/CoPaw/blob/main/CONTRIBUTING.md#4-code-and-quality

livehl:
Fixed all format check failures:

**Python (pre-commit)**
- Applied `black` (line-length=79) and `add-trailing-comma` to all modified Python files
- Fixed `flake8 E501` (line too long) in `channel.py`, `client.py`, `utils.py`, `routers/config.py` by wrapping long comments and docstrings
- Removed unused variable `hint_ext` (F841) in `channel.py`

**Prettier (console & website)**
- Ran `prettier --write` on `console/src/**/*.{ts,tsx,json}` — fixed formatting in `channel.ts` and `ChannelDrawer.tsx`
- Ran `prettier --write` on `website/public/docs/channels.zh.md` and `channels.en.md`

Commit: `4ff63a2`

src/copaw/app/routers/config.py
```diff
@@ -152,6 +152,115 @@ async def reload_in_background():
...
+    # The scan target is the URL that WeChat app should open when scanning
+    if qrcode_img_url.startswith("http"):
+        scan_url = qrcode_img_url
+    else:
+        scan_url = (
+            f"https://liteapp.weixin.qq.com/q/7GiQu1"
+            f"?qrcode={qrcode}&bot_type=3"
+        )
+    try:
+        import segno
```
rayrayraykk: move to top

hongxicheng:
Hello, I encountered an issue in the latest testing. The heartbeat and cron of WeChat channels are not working properly. This is because the context_token was not passed when sending messages proactively. According to my research, the context_token seems to be non-expiring. I suggest saving the context_token of the user's last message and then using it when sending messages proactively.

livehl:
> Hello, I encountered an issue in the latest testing. The heartbeat and cron of WeChat channels are not working properly. This is because the context_token was not passed when sending messages proactively. According to my research, the context_token seems to be non-expiring. I suggest saving the context_token of the user's last message and then using it when sending messages proactively.

Fixed. The changes:
Added _user_context_tokens: Dict[str, str] to cache the latest context_token per user when receiving messages.
In send() and send_content_parts(), fall back to the cached token when meta does not carry one — so heartbeat/cron proactive sends will work as long as the user has sent at least one message.

hongxicheng:
> > Hello, I encountered an issue in the latest testing. The heartbeat and cron of WeChat channels are not working properly. This is because the context_token was not passed when sending messages proactively. According to my research, the context_token seems to be non-expiring. I suggest saving the context_token of the user's last message and then using it when sending messages proactively.
> 
> Fixed. The changes: Added _user_context_tokens: Dict[str, str] to cache the latest context_token per user when receiving messages. In send() and send_content_parts(), fall back to the cached token when meta does not carry one — so heartbeat/cron proactive sends will work as long as the user has sent at least one message.

It still doesn't seem to work. I think it's because the IDs obtained from `from_user_id` and `to_user_id` might be inconsistent. I suggest you check that. If possible, you could also persist the `contxt_token` so that it can be retrieved on the next startup without sending a message.

livehl:
Updated. Two fixes:

1. **Persist context_tokens to disk** (`~/.copaw/weixin_context_tokens.json`) — saved on every inbound message, loaded on `start()`. This ensures heartbeat/cron proactive sends work even after a restart, without needing the user to send a message first.

2. **Regarding ID consistency** — traced the full call chain: cron calls `to_handle_from_target(session_id="weixin:<from_user_id>")`, which `_parse_user_id_from_handle` resolves back to `from_user_id`. This matches the key used when storing, so the lookup should be correct. If you still see a mismatch, could you share a log snippet showing the actual `to_handle` and `from_user_id` values? That would help pinpoint the issue.

src/copaw/app/channels/weixin/channel.py
```diff
@@ -0,0 +1,890 @@
...
+
+class WeixinChannel(BaseChannel):
+    """WeChat iLink Bot channel: long-poll receive, HTTP send.
+
+    Session IDs:
+        - Private chat:  weixin:<from_user_id>
+        - Group chat:    weixin:group:<group_id>
+    """
+
+    channel = "weixin"
```
hongxicheng:
```suggestion
    channel = "wechat"
```
The channel name needs to match the frontend name; otherwise, entering "wechat" in cron will not find the corresponding channel.

hongxicheng:
There are no functional issues. I have one last optimization suggestion.In _poll_loop_async (weixin/channel.py), a WARNING is logged every time getupdates returns a non-zero ret with no messages. Since ret=-1 is the normal long-poll timeout (no new messages), this floods the log during idle periods and should be treated as a non-error condition
image

livehl:
Fixed both:

1. **channel name**: renamed `channel = "weixin"` → `channel = "wechat"` to match the frontend.

2. **log noise**: `ret=-1` (normal long-poll timeout with no messages) now logs at DEBUG level instead of WARNING. Only unexpected non-zero ret values still trigger a WARNING and a 3-second retry delay.

qkzsky:
sendmessage can be extended to support sending files and images
Obtain the uploadurl through ilink/bot/getuploadurl and upload it to the cdn before sending it to the user
# Message types
MESSAGE ITEM TYPE TEXT = 1
MESSAGE ITEM TYPE IMAGE = 2
MESSAGE ITEM TYPE VOICE = 3
MESSAGE ITEM TYPE FILE = 4
MESSAGE ITEM TYPE VIDEO = 5

For specific details, you can refer to the implementation of the official package

PR #2304, opened 2026-03-25
fix: treat 404 from models.list as successful connection check

### LINKED ISSUE #2303: [Bug]:  MiniMax provider check_connection() fails with 404 because it calls unsupported /models endpoint
## CoPaw Version

CoPaw, version 0.2.0.post1

## Description

The `check_connection()` method in `AnthropicProvider` calls `client.models.list()` to verify connectivity. MiniMax's API (which is Anthropic-compatible) does not support this endpoint and returns `404 Not Found`.

However, the actual model inference via `messages.create()` works perfectly fine.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- CoPaw version: 0.1.0+
- Python: 3.12+
- Provider: MiniMax (International) / MiniMax (China)

## Steps to Reproduce

1. Configure MiniMax provider with API key
2. Try to activate a MiniMax model
3. The connection check fails with "Anthropic API error"

## Expected Behavior

`check_connection()` should succeed (or at least not fail) when the provider can actually handle model requests.

## Root Cause

In `src/copaw/providers/anthropic_provider.py`, the `check_connection()` method at line 70 calls:
```python
await client.models.list()

## Proposed Fix
Replace check_connection() implementation to use a simple messages.create call with max_tokens=1 instead of models.list(). This endpoint is supported by all Anthropic-compatible providers including MiniMax.


async def check_connection(self, timeout: float = 5) -> tuple[bool, str]:
    """Check if Anthropic provider is reachable."""
    try:
        client = self._client(timeout=timeout)
        # Use messages.create instead of models.list for compatibility
        # with Anthropic-compatible providers like MiniMax
        await client.messages.create(
            model="",  # Will fail if no model, but connection works
            max_tokens=1,
            messages=[{"role": "user", "content": "hi"}],
        )
        return True, ""
    except anthropic.APIError as e:
        if e.status_code == 400:
            # Model required error means connection works
            return True, ""
        return False, f"Anthropic API error: {e}"
    except Exception as e:
        return False, f"Connection failed: {e}"

thakoreh:
Fixes #2303

Anthropic-compatible providers like MiniMax don't support the
/models endpoint and return 404. Since a 404 response means the
server is reachable and responding, this treats it as a successful
connection check instead of failing with an API error.

The 404 only means the endpoint doesn't exist, not that the provider
is unreachable. Actual inference via messages.create() works fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,4 @@ src/.hypothesis/
 
 # Kiro IDE
 .kiro/
+.venv2/
--- a/src/copaw/providers/anthropic_provider.py
+++ b/src/copaw/providers/anthropic_provider.py
@@ -70,6 +70,9 @@ class AnthropicProvider(Provider):
             await client.models.list()
             return True, ""
         except anthropic.NotFoundError:
+            # Anthropic-compatible providers (e.g. MiniMax) may not
+            # support /models. A 404 here means the server is reachable
+            # and responding, but API key validity is not confirmed.
             return True, ""
         except anthropic.APIError:
             return False, "Anthropic API error"
--- a/src/copaw/providers/anthropic_provider.py
+++ b/src/copaw/providers/anthropic_provider.py
@@ -69,6 +69,8 @@ class AnthropicProvider(Provider):
             client = self._client(timeout=timeout)
             await client.models.list()
             return True, ""
+        except anthropic.NotFoundError:
+            return True, ""
         except anthropic.APIError:
             return False, "Anthropic API error"
         except Exception:


wangfei010313:
This PR is no longer needed because the underlying connection-check issue has already been addressed in the current `main` branch.

The current implementation handles `404` and `405` responses from `models.list()` by falling back to a lightweight `messages.create()` request. This is more robust than treating every `/models` 404 as an immediate success: it confirms that the provider's actual inference endpoint is reachable while still avoiding false negatives for Anthropic-compatible proxies that do not expose `/models`.

lalaliat:

PR #2315, opened 2026-03-26
fix(mcp): disable hot reload for unsafe stdio clients

guodaxia103:
## Summary

This PR adds a safety guard for MCP clients that cannot be reliably reloaded in-process, especially stdio-based single-session bridges such as `chrome-mcp-stdio`.

## Problem

CoPaw currently supports zero-downtime agent reloads and MCP hot reloads. This works for many MCP transports, but some stdio MCP clients are not safe to reconnect within the same process.

In practice, `chrome-mcp-stdio` can work after a cold restart, but repeated in-process reloads can leave the old stdio/native-host chain in a bad state and cause errors such as:

- `session interrupted while listing tools`
- `Cancelled via cancel scope`

## Changes

- Add `hot_reload_safe: bool = True` to `MCPClientConfig`
- Skip in-process MCP reload for clients marked `hot_reload_safe=false`
- Skip agent in-process reload when the agent contains MCP clients marked `hot_reload_safe=false`
- Improve MCP close cleanup by forcing cleanup after `close()`
- Add unit tests for the reload-safety decision logic

## Validation

- Rebased onto latest `origin/main`
- `python -m compileall src/copaw` passed
- Verified locally that `chrome-mcp-stdio` works after a full CoPaw restart and no longer depends on in-process hot reload

## 中文说明

这个 PR 为一类“不适合在同一进程内热重载”的 MCP 客户端增加了安全保护，重点场景是 `chrome-mcp-stdio` 这类基于 stdio 的单会话桥接客户端。

### 问题背景

CoPaw 目前支持 agent 零停机重载和 MCP 热重载，这对多数 MCP 传输方式是可行的，但对某些 stdio MCP 客户端并不安全。

实际测试中，`chrome-mcp-stdio` 在完整重启后可以正常工作，但在同一进程内反复热重载时，旧的 stdio/native-host 链路可能无法被可靠释放，从而触发以下问题：

- `session interrupted while listing tools`
- `Cancelled via cancel scope`

### 这次修改

- 在 `MCPClientConfig` 中新增 `hot_reload_safe: bool = True`
- 当客户端被标记为 `hot_reload_safe=false` 时，跳过进程内 MCP 热重载
- 当 agent 包含 `hot_reload_safe=false` 的 MCP 客户端时，跳过 agent 的进程内热重载
- 增强 MCP 关闭时的清理逻辑，避免 stdio 资源残留
- 补充对应的单元测试

### 本地验证

- 已 rebase 到最新 `origin/main`
- `python -m compileall src/copaw` 通过
- 本地已验证 `chrome-mcp-stdio` 在完整重启 CoPaw 后可正常工作，不再依赖进程内热重载


src/copaw/app/mcp/watcher.py
```diff
@@ -261,6 +261,16 @@ async def _handle_client_update(
 
     async def _reload_single_client(self, key: str, new_cfg) -> None:
         """Reload a single client with retry tracking."""
+        if not getattr(new_cfg, "hot_reload_safe", True):
+            logger.info(
```
xieyxclack: Maybe use `logger.debug` here?

src/copaw/app/runner/daemon_commands.py
```diff
@@ -140,6 +141,8 @@ async def run_daemon_restart(context: DaemonContext) -> str:
                     "- Agent not currently loaded. "
                     "Will reload on next request."
                 )
+        except AgentReloadRequiresRestartError as e:
```
xieyxclack, line 144: Is it possible to use or inherit an exception class from agentscope-runtime here to ensure consistency in the design?

Refer to: https://github.com/agentscope-ai/agentscope-runtime/blob/main/src/agentscope_runtime/engine/schemas/exception.py

src/copaw/app/multi_agent_manager.py
```diff
@@ -245,6 +258,16 @@ async def reload_agent(self, agent_id: str) -> bool:
             return False
 
         agent_ref = config.agents.profiles[agent_id]
+        requires_serial_reload = self._requires_serial_reload(agent_ref)
```
xieyxclack: Since `agent_ref` is an instance of `AgentProfileRef`, it does not contain the mcp attribute, right?

xieyxclack (commented):
Thank you for your contribution! Please refer to the inline comments.

Besides, you might need to:
1. Remove the test file (after you test it locally)
2. Add docs to clarify the newly added config term

Please let us know if you have any suggestions :)

src/copaw/app/mcp/watcher.py
guodaxia103: Addressed in `f2dcc4d`: switched this watcher message from `logger.info` to `logger.debug` to keep it low-noise during config polling.

src/copaw/app/runner/daemon_commands.py
guodaxia103: Addressed in `f2dcc4d`: `AgentReloadRequiresRestartError` now inherits from the existing `agentscope-runtime` exception hierarchy via `ConfigurationException`, so the reload guard stays consistent with the runtime exception design.

src/copaw/app/multi_agent_manager.py
guodaxia103: Good catch. Addressed in `f2dcc4d`: the reload guard now loads the full workspace `AgentProfileConfig` via `load_agent_config(agent_id)` before checking MCP reload safety, instead of reading from the root-level `AgentProfileRef`. I also added a regression test for this path.

rayrayraykk (commented):
Image When triggered via the API, many MCP processes remain as zombie processes.

And when we exit instead of reload, the MCP remained unkilled.

I suggest to consider refine like reusable service like memory manager to MCP manager:
```python
        sm.register(
            ServiceDescriptor(
                name="memory_manager",
                service_class=MemoryManager,
                init_args=lambda ws: {
                    "working_dir": str(ws.workspace_dir),
                    "agent_id": ws.agent_id,
                },
                post_init=lambda ws, mm: setattr(
                    ws._service_manager.services["runner"],
                    "memory_manager",
                    mm,
                ),
                start_method="start",
                stop_method="close",
                reusable=True,  # enable this!
                priority=20,
                concurrent_init=True,
            ),
        )
```

guodaxia103:
@rayrayraykk I addressed this suggestion in `c317035`.

Changes made:
- marked `mcp_manager` as a reusable workspace service, aligned with the existing `memory_manager` pattern
- updated `create_mcp_service()` to reuse and rebind the existing manager instead of re-initializing clients during workspace reload
- fixed `ServiceManager._stop_service()` so reused services are still stopped on final shutdown; previously they were skipped even with `final=True`, which could leave reused services alive after process exit
- added focused unit coverage for both behaviors

Local validation:
- `python -m pytest tests/unit/workspace/test_workspace.py -q` -> `7 passed`
- `pre-commit run --files src/copaw/app/workspace/service_factories.py src/copaw/app/workspace/service_manager.py src/copaw/app/workspace/workspace.py tests/unit/workspace/test_workspace.py` -> passed
- `python -m compileall src/copaw` -> passed

xieyxclack:
Closed as outdated

Issue #2814, opened 2026-04-02
[Bug]: Multi-agent chat history is empty for running callee agent


YingchaoX:
## CoPaw Version

CoPaw, version `1.0.1b1`

## Description

In multi-agent chat, when agent A is chatting with agent B and agent B is still running/streaming, opening agent B's chat page can show an empty chat history.

What actually happens:
- Agent A's chat page can still show the in-progress conversation.
- Agent B is visibly running, but when switching to B's chat page, the chat history may appear empty until the run finishes.

What I expected:
- Agent B's chat page should show the current in-progress conversation history while B is still running, not only after completion.

From code-path analysis, this appears to be caused by a mismatch between:
- the live in-memory stream state used by the active chat UI, and
- the persisted session state used by `GET /chats/{chat_id}` for history loading.

The backend chat history endpoint reads session state from disk. However, session state is only saved at the end of a run in the runner `finally` block. During generation, the history endpoint may therefore return `messages=[]` even though the task is still running.

The frontend already contains a special fallback for the latest user message during reconnect, but that fallback only seems to cover messages initiated from the current frontend chat UI. It does not appear to cover agent-to-agent internal messages, which makes this issue especially visible in multi-agent conversations.

**Related PR(s):** #1375

**Security considerations:** None known. This appears to be a chat state synchronization / UI consistency issue, not an auth or data exposure issue.

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **CoPaw version:** `1.0.1b1`
- **OS:** macOS Darwin 25.3.0 (arm64)
- **Install method:** from source
- **Python version (if applicable):** `3.10.20`

## Steps to Reproduce

1. Start CoPaw with multi-agent support enabled.
2. Create or use two agents, A and B.
3. Trigger a conversation where A sends work/messages to B and B starts a long-running response.
4. While B is still running, switch the console UI to agent B and open B's chat page.
5. Observe that B's chat page may show no chat history until the run completes.

## Actual vs Expected

- **Actual:** While agent B is still running, B's chat page may show an empty chat history even though the conversation is in progress.
[...]

PR #2896, opened 2026-04-03
Fix/chat session history hardening

YingchaoX:
## Description

This PR adds a temporary backend-side mitigation for in-progress multi-agent chat history visibility and session file safety.

It does three things:
1. Makes running chat history more recoverable by surfacing runtime history for active chats in `GET /chats/{chat_id}` instead of relying only on persisted session state.
2. Reduces intermediate session writes from time-based saves to message-boundary saves, to lower write frequency and log noise.
3. Hardens session persistence with atomic file writes and tolerant JSON read fallback, so transient partial reads do not cause `/api/chats/{id}` to fail with `500`.

This is intentionally a temporary solution. It improves correctness and reduces the immediate user-facing issue, but it does not implement a true live-stream model for the callee chat page. The long-term direction should still be to separate runtime state from persisted history more cleanly.

**Related Issue:** Relates to #2814

**Security Considerations:** No known security impact. This change is limited to chat history persistence/retrieval behavior and file-write safety.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

1. Start a multi-agent workflow where agent A calls agent B.
2. While B is still running, open B's chat page.
3. Verify that refreshing/re-entering B's page shows in-progress history instead of staying empty.
4. Verify that session reads do not fail even if a chat history request races with an intermediate session write.
5. Verify that final persisted history is still available after completion.

```bash
uv run pre-commit run --all-files
# sort simple yaml files...............................(no files to check)Skipped
# check xml................................................................Passed
# check toml...............................................................Passed
# check docstring is first.................................................Passed
# check json...............................................................Passed
# detect private key.......................................................Passed

```
## Additional Notes

- This PR does not make the callee chat page truly live-streaming while the user stays on the page. It only makes the in-progress history recoverable and safer to read.
- A more complete follow-up would move callee runtime display onto a dedicated runtime-state source instead of relying on persisted session history as an intermediate layer.


zhijianma:
@YingchaoX 

To be honest,  this is not an ideal solution, as you cannot see the streaming status of a sub-agent's task when switching to it.
We plan to explore using console/chat instead of agent/process, leveraging reconnect to improve the overall experience. I've already done some preliminary testing on this approach.

YingchaoX:
That makes sense. I’m also looking forward to a more elegant enhancement here in the future.

Issue #2967, opened 2026-04-05
[Bug]: execute_shell_command may bypass File Guard


jerry78424:
## CoPaw Version

1.0.1

## Description

**Security Test Observation:** Tools `read_file`, `write_file`, and `edit_file` cannot access protected directories without approval. When these tools are disabled or being prompted, the agent falls back to `execute_shell_command` and executes successfully without requiring explicit approval.

This behavior was observed during a test where:
- Primary file operation tools (read_file, write_file, edit_file) are disabled or asked not to use them
- The agent automatically falls back to using `execute_shell_command` as an alternative method
- `execute_shell_command` successfully completes the operation without the same approval requirements

**Security considerations:** 
- **Auth/Permission Bypass Risk**: The fallback mechanism may bypass intended access controls if `execute_shell_command` has broader permissions than the dedicated file tools
- **Consistency Concern**: Different tools having different approval requirements could lead to inconsistent security posture
- **Recommendation**: Consider implementing unified permission checks across all tool types, or ensure `execute_shell_command` respects the same directory access policies as the dedicated file tools

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **CoPaw version:** 1.0.1
- **OS:** Windows 11 25H2
- **Install method:** Desktop app
- **Python version (if applicable):** N/A (Desktop app)

## Steps to Reproduce

1. Configure CoPaw agent with restricted file access permissions for certain directories
2. Attempt to read/write files in a protected directory using `read_file`/`write_file`/`edit_file` tools
3. Observe that these tools cannot access the protected directory without approval
4. Disable the tools or prompt the agent to use execute_shell_command directly
5. Agent falls back to `execute_shell_command` (e.g., using shell commands like `type`, `echo > file`)
6. Notice that `execute_shell_command` succeeds without the same approval requirements

## Actual vs Expected

- **Actual:** The agent successfully accesses protected directories through `execute_shell_command`, bypassing the restrictions applied to dedicated file tools
- **Expected:** Either:
  - All tools should respect the same permission/approval requirements consistently
[...]

PR #2978, opened 2026-04-06
[#2967] Fix shell file-guard path detection for fallback commands

Alexxigang:
## Summary
- improves shell file-path extraction so tool guard recognizes Windows-style paths, backslash-relative paths, and redirected output targets
- preserves Unix shell detection while adding support for common shell file-reader commands such as 	ype and Get-Content
- adds focused regression tests for shell fallback access to sensitive files and parity with dedicated file tools

## Why this fix
Issue #2967 reports that execute_shell_command can bypass File Guard when the agent falls back to shell-based file access. The root cause is that shell path extraction missed several common path forms:
- Windows drive paths like C:\...
- backslash-relative paths like .\secret.txt and ..\secret.txt
- redirected output targets and common shell file-read commands

That left sensitive shell-based file access under-detected even though the file guard layer already existed.

## Validation
- python -m py_compile src/copaw/security/tool_guard/guardians/file_guardian.py tests/unit/security/tool_guard/test_file_guardian_shell_access.py
- PYTHONPATH=src python -m pytest tests/unit/security/tool_guard/test_file_guardian_shell_access.py -q
  - result: 6 passed

## AI assistance
This PR was prepared with AI-assisted tooling, with local reproduction, implementation, and validation performed before submission.

### DIFF
--- a/src/copaw/security/tool_guard/guardians/file_guardian.py
+++ b/src/copaw/security/tool_guard/guardians/file_guardian.py
@@ -92,7 +92,7 @@ _MIME_PREFIXES = (
     "model/",
 )
 
-_WINDOWS_DRIVE_RE = re.compile(r"^[a-zA-Z]:[\/]")
+_WINDOWS_DRIVE_RE = re.compile(r"^[a-zA-Z]:")
 _SHELL_FILE_COMMANDS = frozenset(
     {
         "cat",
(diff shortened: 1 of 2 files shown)


Alexxigang:
Addressing the review suggestions now: expanded shell file-command coverage, added flag/multi-target parsing for shell commands, broadened Windows drive-path detection, and extended fallback scanning to string lists/tuples. Revalidated with PYTHONPATH=src python -m pytest tests/unit/security/tool_guard/test_file_guardian_shell_access.py -q (8 passed).

zhijianma:
@Alexxigang 

Thanks for the PR. QwenPaw's file guardian and governance detectors now extract shell file-path candidates — including Windows/backslash paths, $HOME/%USERPROFILE% expansion, and redirected output targets — and apply sensitive-path checks, covering this. The behavior is already in main via #7375 and #7472 (governance). Closing as covered.

PR #2979, opened 2026-04-06
[#2960] Stabilize MCP hot-reload cleanup lifecycle

### LINKED ISSUE #2960: [Bug]: MCP Client Not Cleaned Up on Hot Reload, Causing Persistent CPU Spike
### Description

When any configuration change is saved in the CoPaw web UI, a zero-downtime hot reload is triggered. If an MCP server is enabled, the MCP client is initialized **twice** during the reload but fails to clean up either instance, leaving orphaned async tasks running indefinitely. This causes persistent abnormal CPU usage until the entire `python.exe` process is manually killed.

Disabling all MCP servers completely eliminates the issue — hot reload then completes cleanly with no CPU anomaly.

### Steps to Reproduce

1. Enable at least one MCP server in CoPaw settings
2. Open the CoPaw web UI and make any configuration change (e.g. modify a skill setting)
3. Save — this triggers a hot reload
4. Observe CPU usage immediately after the reload completes

### Observed Behavior

- MCP client connects **twice** during reload:
  ```
  MCP client connected.
  MCP client connected.
  ```
- Both cleanup attempts fail with:
  ```
  Warning: Error during MCP client cleanup: Attempted to exit cancel scope in a different task than it was entered in
  ```
- One CPU core is continuously occupied after the reload, causing approximately **+6% overall CPU usage** and a **+10°C rise in CPU temperature**
- The CPU usage never recovers — it persists until `python.exe` is forcibly terminated
- No errors are raised in normal application flow; the issue is silent from the user's perspective

### Expected Behavior

Hot reload should cleanly shut down and release all MCP client connections. CPU usage should return to idle levels after the reload completes.

### Verification

This was confirmed by toggling MCP servers:
- **MCP enabled** → hot reload triggers the bug every time (100% reproducible)
- **MCP disabled** → hot reload completes cleanly, no CPU anomaly

### MCP Server Configuration at Time of Bug

```json
{
  "key": "huggingface",
  "name": "huggingface",
  "transport": "streamable_http",
  "url": "https://huggingface.co/mcp",
  "headers": {
    "Authorization": "Bearer <token>"
  }
}
```

### Environment

| Field | Value |
|---|---|
| CoPaw version | 1.0.1 |
| OS | Windows 11 25H2 |
| Install method | Command line (`python.exe -m copaw app`) |
| Python version | 3.10.20 |
| Main model | GPT-5.4 |
| ReMe backend vector model | Qwen/Qwen3-Embedding-8B |

### Related Log Snippets

**With MCP enabled (abnormal):**
```
MCP client connected.
MCP client connected.
Warning: Error during MCP client cleanup: Attempted to exit cancel scope in a different task than it was entered in
Warning: Error during MCP client cleanup: Attempted to exit cancel scope in a different task than it was entered in
```

**With MCP disabled (normal):**
```
Old workspace instance stopped: default. Zero-downtime reload completed.
```
No MCP entries, no warnings, clean shutdown.

Alexxigang:
## Summary
- move MCP client connect/replace/remove/close onto a manager-owned lifecycle task
- replace syncio.wait_for() around client connect with syncio.timeout() so connect/close stay on the same task
- stop old config watchers before zero-downtime workspace handoff to avoid redundant reloads on the retiring instance
- add regression tests covering replace/remove/close-all lifecycle behavior

## Why this fix
The cleanup failure came from MCP lifecycle work crossing task boundaries during hot reload. Even with a single manager task, syncio.wait_for() still wrapped connect() in a separate task, so later close() calls could hit the same anyio cancel-scope error reported in the issue. This patch keeps the full lifecycle on one task and pauses old watchers before the new workspace starts, which prevents the retiring instance from reacting to the same config save during handoff.

## Validation
- python -m py_compile src/copaw/app/mcp/manager.py src/copaw/app/workspace/workspace.py src/copaw/app/multi_agent_manager.py tests/unit/app/test_mcp_hot_reload_cleanup.py
- PYTHONPATH=src python -m pytest tests/unit/app/test_mcp_hot_reload_cleanup.py -q

Closes #2960.

### DIFF
--- a/src/copaw/app/mcp/manager.py
+++ b/src/copaw/app/mcp/manager.py
@@ -1,4 +1,4 @@
-﻿# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 """MCP client manager for hot-reloadable client lifecycle management.
 
 This module provides centralized management of MCP clients with support
@@ -82,7 +82,7 @@ class MCPClientManager:
     ) -> None:
         """Replace or add a client with new configuration.
 
-        Flow: connect new (outside lock) 鈫?swap + close old (inside lock).
+        Flow: connect new (outside lock) → swap + close old (inside lock).
         This ensures minimal lock holding time.
 
         Args:
@@ -185,6 +185,8 @@ class MCPClientManager:
             except BaseException as exc:
                 if not request.future.cancelled():
                     request.future.set_exception(exc)
+                if isinstance(exc, (KeyboardInterrupt, SystemExit)):
+                    raise
             else:
                 if not request.future.cancelled():
                     request.future.set_result(result)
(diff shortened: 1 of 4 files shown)


rayrayraykk:
This PR does not resolve the root bug, close since we've fix in #3106

PR #2989, opened 2026-04-06
[#2955] Flatten llama.cpp archives with metadata dirs

### LINKED ISSUE #2955: [Bug]: 无法安装llama.cpp
## CoPaw Version

v1.0.1

## Description

1.点击按钮下载llama
2.下载进度正常走完，toast提示下载完成
3.但这个页面还是没变化，没有列出模型列表，只有那个‘下载llama’的按钮

无论是用pip安装的copaw还是后面我再用exe安装的，  都是这样

Image

## Environment

- OS: Windows 11
- Python version 3.12

Alexxigang:
## Summary
- ignore metadata-only archive entries like __MACOSX and AppleDouble ._* files when flattening downloaded llama.cpp packages
- flatten the real top-level binary directory even when metadata siblings are present
- add regression coverage for Windows zip packages that include __MACOSX sidecar entries

## Why this fix
The current extraction path only flattens archives when there is exactly one top-level directory. Some Windows users receive packages with both the real versioned directory and a sibling __MACOSX metadata directory. In that shape, CoPaw leaves the binaries nested under the versioned folder, so check_llamacpp_installation() never finds llama-server.exe in ~/.copaw/local_models/bin and the UI keeps showing the install button.

## Validation
- python -m py_compile src/copaw/local_models/llamacpp.py tests/unit/local_models/test_llamacpp_backend.py
- inline pytest run with the same google.genai stub pattern already used in repo tests: 	ests/unit/local_models/test_llamacpp_backend.py ? 24 passed

Closes #2955.

### DIFF
--- a/src/copaw/local_models/llamacpp.py
+++ b/src/copaw/local_models/llamacpp.py
@@ -654,13 +654,29 @@ class LlamaCppBackend:
         staging_dir: Path,
         dest_dir: Path,
     ) -> None:
-        extracted_entries = list(staging_dir.iterdir())
+        extracted_entries = [
+            entry
+            for entry in staging_dir.iterdir()
+            if not LlamaCppBackend._is_ignorable_archive_entry(entry)
+        ]
         source_root = staging_dir
         if len(extracted_entries) == 1 and extracted_entries[0].is_dir():
             source_root = extracted_entries[0]
 
         for item in source_root.iterdir():
+            if LlamaCppBackend._is_ignorable_archive_entry(item):
+                continue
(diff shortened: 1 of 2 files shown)


pan-x-c:
Thanks for your contribution. The current implementation ignores meta folders like `__MACOSX`, but a more robust approach would be to only unpack the folder starting with `llama-`, which is the default top-level folder when llama.cpp is correctly packaged. This implementation should be applicable to a wider range of scenarios.

pan-x-c:
Fixed in #3021

Issue #2314, opened 2026-03-26
[Feature]: Support provider-agnostic conversation history so models can switch freely between turns


garyzhang99:
## Summary

Enable CoPaw to switch freely between different model providers during the same conversation session, such as OpenAI-compatible / MiniMax / Anthropic / Gemini / Ollama, without failing on incompatible conversation-history formats.

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Problem / Motivation

CoPaw can already switch the active model/provider for an agent, but existing conversation history is still tightly coupled to the message semantics used by the previously active provider and formatter family.

This becomes a problem because different providers represent chat history differently, especially for:

- reasoning / thinking
- tool calls and tool results
- multimodal content
- role mapping
- provider-specific metadata or extensions

Examples:

- OpenAI-compatible providers expect `messages[]`, `tool_calls`, and `role=tool`
- Anthropic expects inline `thinking/tool_use` blocks and `tool_result` as a new `role=user` message
- Gemini expects `parts[]`, `function_call`, and `function_response`

As a result, when a conversation switches provider family between turns, CoPaw may replay history in a format the new provider cannot accept, causing request errors or broken conversation continuation.

## Goal

The goal is to make CoPaw conversation history provider-agnostic enough that:

- one turn can use one provider family
- the next turn can use another provider family
- the same chat session can continue naturally
- CoPaw does not need fragile provider-pair-specific replay logic everywhere

Example scenarios:

- previous turn: MiniMax
- next turn: Ollama local model
- next turn after that: Anthropic

All of them should be able to continue the same chat session safely.

### Option A: Introduce a CoPaw-owned canonical transcript + provider adapters

CoPaw could persist a provider-agnostic conversation transcript as its own source of truth, and rebuild provider-specific request payloads from that transcript before each model call.

Possible characteristics:

- store normalized semantic conversation history
- isolate provider-specific metadata as optional extensions
- keep replay logic centralized in adapters instead of scattering it across formatters and runtime patches

This is likely the most robust long-term solution.

### Option B: Keep current memory, but add a normalization layer only for provider switching

[...]

PR #3078, opened 2026-04-08
feat(models): better switch(e.g. preserve media memory) within same type of model providers

garyzhang99:
## Description

This PR improves conversation stability when switching between provider models within the same session.

Previously, some request-time repair and multimodal fallback logic could operate directly on the in-memory AgentScope `Msg` objects. 

That meant backend-specific adaptation could mutate stored conversation history, especially when temporarily switching to a stricter or non-multimodal backend. In those cases, switching back later could no longer recover the original full context.

This PR keeps the existing session/history storage format unchanged and makes the OpenAI-compatible path safer in two places:

1. Before formatting, it now builds a normalized copy of the message list instead of repairing the persisted history in place.
2. During media fallback/retry, it avoids stripping media from stored memory and instead applies request-time media stripping only to the formatter path.

As a result, model backends still receive repaired and downgraded messages when needed, while the original persisted history is preserved.

### What Changed

- Added a request-time normalizer for model message formatting
- Reused existing tool-message sanitization on copied messages
- Applied multimodal downgrade on copied messages only
- Wired the normalizer into the model formatter path
- Updated the reasoning/summarizing media fallback path to avoid mutating stored memory
- Kept session persistence, `/dump_history`, `/load_history`, and chat history API behavior unchanged
- Added unit tests for:
  - tool input repair without mutating original messages
  - media downgrade without mutating original messages
  - stable text-only message normalization
  - Model formatter detection / request-time media-strip flag behavior

### Why This Approach

This is intentionally a narrow first step toward provider-family-safe message replay for [#2314](https://github.com/agentscope-ai/CoPaw/issues/2314). It does not introduce a new canonical transcript schema or migrate persisted session files.

The goal is to improve model-switching stability with minimal risk. By keeping persisted history unchanged, a conversation that temporarily downgrades for a weaker OpenAI-compatible backend can later switch back and still reuse the original context instead of a permanently stripped version.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit` locally and it passes for the changed files
- [x] If pre-commit auto-fixed files, I reran checks
- [x] I ran relevant tests locally and they pass
- [ ] Ready for review

## Testing

```bash
pytest -q \
    tests/unit/agents/test_message_request_normalizer.py \
    tests/unit/agents/test_react_agent_formatter_selection.py \
    tests/unit/agents/test_model_factory_message_normalization.py
```

### DIFF
--- a/src/copaw/agents/model_factory.py
+++ b/src/copaw/agents/model_factory.py
@@ -94,6 +94,7 @@ def _supports_multimodal_for_current_model() -> bool:
 def _normalize_messages_for_formatter(
     msgs: list,
     base_formatter_class: Type[FormatterBase],
+    formatter_instance: FormatterBase | None = None,
 ) -> tuple[list, bool, bool]:
     """Return normalized messages and formatter-family flags.
 
@@ -119,6 +120,8 @@ def _normalize_messages_for_formatter(
 
     if is_openai_compatible_formatter:
         supports_multimodal = _supports_multimodal_for_current_model()
+        if getattr(formatter_instance, "_copaw_force_strip_media", False):
+            supports_multimodal = False
         normalized_msgs = normalize_messages_for_openai_compatible(
             msgs,
             supports_multimodal=supports_multimodal,
@@ -540,6 +543,7 @@ def _create_file_block_support_formatter(
             ) = _normalize_messages_for_formatter(
                 msgs,
                 base_formatter_class,
+                self,
             )
 
             reasoning_contents = {}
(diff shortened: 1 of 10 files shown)


garyzhang99:
This PR is outdated. See https://github.com/agentscope-ai/QwenPaw/pull/3359.

PR #3144, opened 2026-04-09
fix(discord): add thread awareness to on_message routing

### LINKED ISSUE #2980: [Bug] Discord: replies go to parent channel instead of thread
## Problem

The Discord channel has no thread awareness. This causes replies to go to the parent channel instead of the thread in two scenarios:

### 1. Thread-starting messages

When a user creates a thread from a message in Discord:
1. `on_message` fires with `message.channel` = the **parent channel** (not the thread)
2. `resolve_session_id` returns `discord:ch:<parent_channel_id>`
3. Bot replies to the parent channel
4. The thread exists (`message.thread` is set) but CoPaw never checks it

**Expected**: reply should go to the newly created thread.

### 2. Cron/proactive sends that create threads

When a cron job dispatches to a channel and the agent creates a Discord thread (e.g., via a script calling the Discord API directly):
1. Cron targets `session_id: discord:ch:<channel_id>`
2. Agent runs a skill that creates a thread and posts content there (works correctly)
3. But the agent's own response still routes to `to_handle = discord:ch:<channel_id>` — the main channel
4. Status messages like "Done." or tool output leak into the main channel

**Expected**: a way to suppress or redirect the agent's own response when the skill already handled output.

## Root cause

`on_message` (line 232-241 of `discord_/channel.py`) collects metadata but never checks:
- `message.channel.type` (thread vs text channel)
- `message.thread` (thread started from this message)
- `isinstance(message.channel, discord.Thread)`

The `meta` dict has no `is_thread`, `thread_id`, or `parent_channel_id` field. Compare with:
- **Telegram** (`telegram/channel.py`): captures `message_thread_id` and passes it to all send calls
- **Mattermost** (`mattermost/channel.py`): uses `mattermost_thread:{root_id}` for thread-scoped sessions

### For problem 1 (thread-starting messages):
In `on_message`, check if `message.thread` exists. If so, set `meta["channel_id"]` to the thread ID so the session resolves to the thread.

### For both problems — `thread_only` config option:
Add `thread_only: bool = False` to `DiscordConfig`. When enabled:
- If a message arrives in a top-level channel, auto-create a thread via `message.create_thread()` and reply there
- If a message arrives in an existing thread, reply normally

### Metadata enrichment:
Add `is_thread`, `thread_id`, and `parent_channel_id` to the Discord meta dict, consistent with how Telegram and Mattermost handle threads.

## Environment

- CoPaw v1.0.1b2
- discord.py (latest)
- macOS

energyd:
## Problem

Closes #2980.

The Discord channel has no thread awareness. This causes two problems:

### 1. Messages inside threads route to parent channel

`on_message` never checks `isinstance(message.channel, discord.Thread)`, so messages sent inside threads get routed to the parent channel session instead of the thread session.

### 2. Thread creation race condition

When a user creates a thread and sends the first message simultaneously, `MESSAGE_CREATE` may arrive before `THREAD_CREATE` completes in discord.py's cache. The starter message gets routed to the parent channel.

## Fix

**Basic thread awareness** (commit 1):
- Detect `isinstance(message.channel, discord.Thread)` → route to thread session
- Detect `message.thread` (thread created from this message) → route to thread session  
- Add `is_thread`, `thread_id`, `parent_channel_id` to meta dict, consistent with Telegram's `message_thread_id` pattern

**Race condition handling** (commit 3):
- `on_thread_create` handler caches `starter_message_id -> thread_id`
- `on_message` checks `message.flags.has_thread` + cache lookup to detect the race
- Unified `_effective_channel_id` routing for all thread scenarios

No changes to the send path — Discord threads have their own channel IDs and `_resolve_target` already resolves them via `get_channel` / `fetch_channel`.

## Testing

- Send a message inside an existing thread → bot replies in the thread (not parent channel)
- Create a thread and send first message simultaneously → bot replies in the thread
- Messages in non-thread channels → no change in behavior

### DIFF
--- a/src/qwenpaw/app/channels/discord_/channel.py
+++ b/src/qwenpaw/app/channels/discord_/channel.py
@@ -127,10 +127,7 @@ class DiscordChannel(BaseChannel):
                         msg_id,
                     )
                     return
-                if (
-                    len(self._processed_message_ids)
-                    >= self._MAX_CACHED_MESSAGE_IDS
-                ):
+                if len(self._processed_message_ids) >= self._MAX_CACHED_MESSAGE_IDS:
                     oldest = self._processed_message_id_queue.popleft()
                     self._processed_message_ids.discard(oldest)
                 self._processed_message_ids.add(msg_id)
@@ -236,25 +233,18 @@ class DiscordChannel(BaseChannel):
                 is_group = message.guild is not None
                 _is_thread = isinstance(message.channel, discord.Thread)
                 _thread_started = (
-                    not _is_thread
-                    and getattr(message, "thread", None) is not None
+                    not _is_thread and getattr(message, "thread", None) is not None
                 )
                 # Race-condition: thread created simultaneously
                 # with the first message. on_thread_create may
                 # have cached the thread_id before on_message
                 # fires for the starter message.
                 _race_thread_id = None
-                if (
-                    not _is_thread
-                    and not _thread_started
-                    and is_group
-                ):
+                if not _is_thread and not _thread_started and is_group:
                     flags = getattr(message, "flags", None)
                     if flags and getattr(flags, "has_thread", False):
-                        _race_thread_id = (
-                            self._recent_thread_starts.get(
-                                str(message.id),
-                            )
+                        _race_thread_id = self._recent_thread_starts.get(
+                            str(message.id),
                         )
                 # Determine effective channel_id for session routing
                 if _is_thread:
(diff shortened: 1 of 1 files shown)


energyd:
Friendly bump — this fixes a fundamental Discord routing issue where thread messages get misrouted to the parent channel. Rebased on latest main, CI is green.

xieyxclack (approved):
LGTM

PR #3252, opened 2026-04-10
fix(runner): truncate oversized tool_result before session persistence

MilongWong:
## Summary

When a tool returns massive output (e.g., 3+ MB of log text from `wait_task` or `execute_shell_command`), the full content is persisted to the session JSON file via `save_session_state()`, which fires in the `finally` block **before** the next turn's `compact_tool_result` pre-reasoning hook can run.

This causes an infinite compression loop that makes the agent unresponsive.

## Root Cause

The `compact_tool_result` hook runs as a `pre_reasoning` hook (start of next turn), but `save_session_state` fires at the **end** of the current turn. Giant tool outputs are saved to disk before truncation can occur.

Confirmed with real incident data (2026-04-10): 3.8MB tool_result from `wait_task` → session bloat → infinite compression → agent unresponsive.

## Fix

Add `_compact_tool_results_in_state()` in `SafeJSONSession.save_session_state()` that truncates oversized tool_result outputs (>50KB) before writing to disk.

## Type of Change

## Component(s) Affected

- [x] Core

## Testing

- Added `tests/test_session_tool_result_guard.py`
- Verified on live CoPaw: 1,800,000 char output → 2,104 chars (855x reduction)
- Normal tool results unaffected

### DIFF
--- a/src/copaw/app/runner/session.py
+++ b/src/copaw/app/runner/session.py
@@ -24,6 +24,115 @@ logger = logging.getLogger(__name__)
 # Characters forbidden in Windows filenames
 _UNSAFE_FILENAME_RE = re.compile(r'[\\/:*?"<>|]')
 
+# Maximum bytes for a single tool_result output before truncation
+_TOOL_RESULT_MAX_BYTES = 50 * 1024  # 50KB
+
+
+def _truncate_tool_result_output(
+    output: Union[str, list],
+    max_bytes: int,
+    tool_name: str = "",
+) -> Union[str, list]:
+    """Truncate a tool_result output if it exceeds max_bytes.
+
+    Args:
+        output: The tool result output (string or list of content blocks)
+        max_bytes: Maximum allowed bytes before truncation
+        tool_name: Name of the tool for the truncation notice
+
+    Returns:
+        Truncated output with a notice if truncated, original otherwise
+    """
+    if isinstance(output, str):
+        if len(output.encode("utf-8", errors="replace")) <= max_bytes:
+            return output
+        return (
+            f"[auto-truncated] Original output exceeded {max_bytes:,} bytes "
+            f"(tool: {tool_name}). "
+            f"First 2000 chars:\n{output[:2000]}\n[truncated]"
+        )
+
+    if isinstance(output, list):
+        for block in output:
+            if not isinstance(block, dict) or block.get("type") != "text":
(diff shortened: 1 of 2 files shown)


jinliyl:
@MilongWong 

Thank you very much for your feedback.

In fact, there is currently a truncation logic applied before each tool call result. For results that are too long, they are offloaded to the file system and truncated in memory.
Please refer to the code: https://github.com/agentscope-ai/QwenPaw/blob/main/src/qwenpaw/agents/hooks/memory_compaction.py#L120
Also check the ReMe code: https://github.com/agentscope-ai/ReMe/blob/main/reme/memory/file_based/components/tool_result_compactor.py
And the actual truncation implementation: https://github.com/agentscope-ai/ReMe/blob/main/reme/memory/file_based/utils/file_utils.py

The default upper limit is 50KB.
Therefore, no tool result will exceed 50KB.

Specifically regarding your issue: Confirmed with real incident data (2026-04-10): a 3.8MB tool_result from wait_task led to session bloat, causing infinite compression and making the agent unresponsive.

Could you please tell me your copaw/qwenpaw version, as well as your ToolResultCompactConfig settings, particularly the max_input_length configuration?
What tool simultaneously reads a 3.8MB tool result, using /history to view the context result?

jinliyl:
I took a look at the execution flow, and the root cause described in the PR doesn't seem accurate.

`compact_tool_result` (now renamed to `_prune_tool_result`) is registered as a **`post_acting`** hook, not a `pre_reasoning` hook. This means it runs **immediately after each tool call**, not at the start of the next turn. The actual execution order is:

```
tool execution → post_acting hook (_prune_tool_result truncates output) → ... → finally block → save_session_state()
```

By the time `save_session_state()` runs in the `finally` block, the oversized tool output should have already been truncated by `_prune_tool_result`. So the described scenario — "giant tool output is saved to disk before truncation can occur" — shouldn't happen with the current architecture.

The fix itself (adding a truncation guard in `save_session_state`) is reasonable as a defense-in-depth measure, but if the 3.8MB incident actually occurred, the real root cause is likely elsewhere — for example, `_prune_tool_result` not covering the specific output format, or the incident predating the `post_acting` hook registration.

Could you double-check whether the incident happened before the pruning hook was in place, or whether there's a code path where `_prune_tool_result` doesn't trigger?

PR #3295, opened 2026-04-12
fix: omit tool_choice=auto for vLLM compatibility to avoid 400 error

jilin6627-spec:
## Summary

When connecting to a self-hosted vLLM OpenAI-compatible API endpoint that does not have `--enable-auto-tool-choice` enabled (which is the default configuration for most vLLM deployments), vLLM will reject any request containing `tool_choice="auto"` with a 400 error:

> "auto" tool choice requires --enable-auto-tool-choice.

This PR fixes the issue by **omitting `tool_choice` entirely when it's set to `auto`**, following the same fix pattern already adopted by OpenClaw/Clawdia for vLLM compatibility:
- Keeps `tools` intact so the model knows what tools are available
- Omits `tool_choice = "auto"` to avoid triggering the vLLM check
- Client-side parsing still works correctly, so full tool calling capability is preserved

## Type of Change
- ✅ Bug fix (non-breaking change that fixes an issue)

## Component(s) Affected
- ✅ Core (token_usage / model_wrapper)

## Checklist
- [x] I have run `pre-commit run --all-files` (or will do so before merging)
- [x] The change is tested and works locally with vLLM endpoint
- [x] No documentation is needed for this small compatibility fix

## Testing
- **Problem**: Connecting QwenPaw to a vLLM OpenAI-compatible endpoint (without `--enable-auto-tool-choice`) with tool calling enabled would result in a 400 error immediately
- **Fix**: After this change, QwenPaw can connect successfully and tool calling works normally
- **Validation**: This matches the same fix that has already been tested and confirmed working in OpenClaw/claudish

Fixes the vLLM 400 compatibility issue that reported in https://github.com/openclaw/openclaw/issues/61197

### DIFF
--- a/src/qwenpaw/token_usage/model_wrapper.py
+++ b/src/qwenpaw/token_usage/model_wrapper.py
@@ -66,6 +66,14 @@ class TokenRecordingModelWrapper(ChatModelBase):
         structured_model: Type[BaseModel] | None = None,
         **kwargs: Any,
     ) -> ChatResponse | AsyncGenerator[ChatResponse, None]:
+        # Fix: Omit tool_choice="auto" for vLLM compatibility
+        # vLLM without --enable-auto-tool-choice will reject requests when
+        # tool_choice="auto" is present, even if tools are provided.
+        # By omitting tool_choice when it's "auto", we bypass the check
+        # while keeping tools available for correct tool calling behavior.
+        if tool_choice == "auto":
+            tool_choice = None
+
         result = await self._model(
             messages=messages,
             tools=tools,


hy3188:
https://github.com/QwenLM/Qwen-Agent/blob/main/README_CN.md  "注意对于QwQ和Qwen3模型，建议启动服务时不加--enable-auto-tool-choice和--tool-call-parser hermes两个参数，因为Qwen-Agent会自行解析vLLM的工具输出。"

jilin6627-spec:
Exactly! That's why this fix is needed. When QwenPaw/client parses tool calls by itself (which is the case for QwQ/Qwen3 with vLLM), we shouldn't send  to vLLM, otherwise vLLM will reject the request with 400 error if  is not enabled (which is the recommended setup according to Qwen-Agent docs).

This fix omits  but keeps , which fits perfectly with the recommended deployment.

jilin6627-spec:
## Technical Background & Validation

This fix follows the optimal **client-side omission strategy** that is comprehensively analyzed in the technical report:

**QwenPaw vLLM Tool Choice Compatibility Fix Analysis**

### Summary of the fix:

- ✅ Keeps the full `tools` array intact → vLLM can inject tool schemas into the system prompt as expected
- ✅ Omits `tool_choice="auto"` entirely → bypasses the vLLM Pydantic validation check that requires `--enable-auto-tool-choice`
- ✅ Prevents the secondary **empty tool_calls silent drop bug** that happens when vLLM server-side regex parsing gets confused by `<think>` reasoning traces in reasoning models like QwQ
- ✅ Leverages QwenPaw/AgentScope built-in client-side parsing → this **exactly matches the official Qwen documentation recommendation**: 
  > *"For Qwen3/QwQ + vLLM deployment, do not enable `--enable-auto-tool-choice`, let the client handle tool call parsing"*

This is a minimal change (only 8 lines of code added) that is fully backward compatible, directly fixes the 400 Bad Request error, and doesn't require any architectural changes or extra configuration from users.

hy3188:
[QwenPaw vLLM Tool Choice Compatibility Fix.docx](https://github.com/user-attachments/files/26703437/QwenPaw.vLLM.Tool.Choice.Compatibility.Fix.docx)

qbc2016 (approved):
LGTM.

xieyxclack:
@jilin6627-spec Thank you for your contribution!

PR #3299, opened 2026-04-12
fix(formatter): 400 error when multiple tool calls return multimodal content

energyd:
## Problem

When a single assistant message contains multiple parallel tool calls that return multimodal content (images, files, etc.), the formatted API request becomes invalid, causing a **permanent 400 error** for the session.

The OpenAI and Anthropic APIs require all tool result messages to appear contiguously after the assistant message. However, `OpenAIChatFormatter._format()` extracts multimodal content from each tool result into a separate `role=user` message and inserts it immediately after the tool result. With multiple tool results in a row, this interleaves user messages between them:

```
assistant   tool_calls=[A, B, C]
tool        tool_call_id=A            ← OK
user        [extracted image from A]  ← breaks contiguity
tool        tool_call_id=B            ← API can't see this
user        [extracted image from B]
tool        tool_call_id=C            ← API can't see this
```

The API only matches tool result A and rejects the request:

- **OpenAI/Azure**: `An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'`
- **Anthropic/Bedrock**: `Expected toolResult blocks at messages.N.content for the following Ids: ...`

The session data is valid — all tool results exist with correct IDs. But the formatting step corrupts the message ordering, and every subsequent LLM call fails. The only workaround is `/clear`.

## Root Cause

`OpenAIChatFormatter._format()` (in agentscope) calls `msgs.insert(i+1, ...)` inside its loop to inject extracted multimodal content as a new user message after each tool result. When multiple tool results appear consecutively (from parallel tool calls), this interleaves `role=user` messages between `role=tool` messages.

## Fix

Add a post-processing step in `FileBlockSupportFormatter._format()` that reorders the formatted messages: after each assistant message with `tool_calls`, collect all following tool and user messages, output all tool messages first, then all user messages.

```
assistant   tool_calls=[A, B, C]
tool        tool_call_id=A            ← all tool results contiguous
tool        tool_call_id=B
tool        tool_call_id=C
user        [extracted image from A]  ← multimodal content after all results
user        [extracted image from B]
user        [extracted image from C]
```

No-op when there are no interleaved user messages between tool results.

## Reproduction

1. Configure `promote_tool_result_images=True` (default for OpenAI formatter)
2. Send a prompt that triggers 2+ parallel tool calls returning images (e.g., multiple `view_image` calls)
3. The LLM call succeeds once (tool calls execute, results stored)
4. Every subsequent LLM call on the same session fails with 400


qbc2016 (approved):
LGTM.

xieyxclack:
@energyd Please resolve the conflct :)

energyd:
> @energyd Please resolve the conflct :)

@xieyxclack done

xieyxclack (approved):

PR #3313, opened 2026-04-13
fix(formatter): normalize image/jpg to image/jpeg in data URLs

energyd:
## Problem

When tool results contain `.jpg` images, the OpenAI formatter produces base64 data URLs with `image/jpg` as the MIME type. `image/jpg` is not a valid IANA MIME type — the correct form is `image/jpeg`.

Bedrock (via litellm) rejects `image/jpg` with:

```
Unsupported image format: jpg. Supported formats: ['png', 'jpeg', 'gif', 'webp', ...]
```

Each failed call retries up to 4 times, wasting tokens on every attempt since the full context is re-sent.

## Root Cause

In agentscope's `_openai_formatter.py`, `_to_openai_image_url()` derives the MIME type directly from the file extension:

```python
extension = raw_url.lower().split(".")[-1]  # "jpg"
mime_type = f"image/{extension}"             # "image/jpg" — wrong
```

## Fix

Add `_fix_image_mime_types()` post-processor in `FileBlockSupportFormatter._format()` that normalizes `data:image/jpg;` → `data:image/jpeg;` in the formatted output. Uses a `_MIME_FIXES` mapping that can be extended for other non-standard types if needed.

This is a no-op for images that already use correct MIME types (png, jpeg, gif, webp).

### DIFF
--- a/src/qwenpaw/agents/model_factory.py
+++ b/src/qwenpaw/agents/model_factory.py
@@ -451,6 +451,35 @@ def _promote_tool_result_videos(
     return new_messages
 
 
+# Mapping of non-standard MIME subtypes to their correct forms.
+_MIME_FIXES: dict[str, str] = {
+    "image/jpg": "image/jpeg",
+}
+
+
+def _fix_image_mime_types(messages: list[dict]) -> None:
+    """Fix non-standard MIME types in base64 data URLs in-place.
+
+    agentscope derives MIME from the file extension literally
+    (e.g. ``.jpg`` → ``image/jpg``), but ``image/jpg`` is not a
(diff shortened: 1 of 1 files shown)


bowenliang123:
+1 for this normalization. The official MIME type for JPG images with .jpg, .jpeg, .jpe, .jif, .jfif extensions is `image/jpeg`, according to https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Image_types#jpeg_joint_photographic_experts_group_image

qbc2016 (approved):
LGTM.

energyd:
🚀

PR #3333, opened 2026-04-13
feat(agents): allow custom agent ID when creating agents

### LINKED ISSUE #3325: feat: Allow custom Agent ID when creating agents
## Problem

Currently, when creating a new agent, the ID is always auto-generated as a random 6-character string (e.g. `uihruk`). Users can only specify a `name` but not an `id`.

This causes a poor experience when calling agents via HTTP API or CLI:

```bash
# User creates agent named "browse-agent", gets ID "uihruk"
# Must use the random ID to call it:
curl -H "X-Agent-Id: uihruk" http://localhost:8088/api/agent/process ...

# What users actually want:
curl -H "X-Agent-Id: browse-agent" http://localhost:8088/api/agent/process ...
```

The random ID is not memorable, not meaningful, and requires users to look up the ID every time.

## Proposed Solution

Allow users to **optionally** specify a custom `id` when creating an agent. If not provided, fall back to the current auto-generated short UUID.

**ID rules:**
- Character set: `[a-zA-Z0-9_-]` (URL-path safe, case-sensitive)
- Length: 2–64 characters
- Cannot start or end with `-` or `_`
- Reserved IDs (`default`) are rejected
- Duplicate IDs are rejected with a clear error message

### 1. Backend — ID validation

`src/qwenpaw/config/config.py`

Add two functions next to the existing `generate_short_agent_id()`:

- `sanitize_agent_id(raw: str) -> str` — strip whitespace
- `validate_agent_id(agent_id: str, existing_ids: set[str]) -> None` — raises `ValueError` on invalid input

```python
_AGENT_ID_PATTERN = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9_-]*[a-zA-Z0-9]$")
_RESERVED_AGENT_IDS = frozenset({"default"})
```

### 2. Backend — Create endpoint

`src/qwenpaw/app/routers/agents.py`

**`CreateAgentRequest`**: add optional field `id: str | None = None` with a validator that calls `sanitize_agent_id`.

**`create_agent()`** logic change:

```
if request.id is provided and non-empty:
    validate_agent_id(request.id, existing_ids)
    new_id = request.id
else:
    new_id = generate_short_agent_id()  # existing behavior
```

The default `workspace_dir` already uses `new_id` in its path (`{WORKING_DIR}/workspaces/{new_id}`), so custom IDs like `browse-agent` will naturally produce readable paths like `~/.qwenpaw/workspaces/browse-agent`.

### 3. Frontend — Type definition

`console/src/api/types/agents.ts`

Add `id?: string` to `CreateAgentRequest` interface.

### 4. Frontend — Create form

`console/src/pages/Settings/Agents/components/AgentModal.tsx`

Add an optional ID input field in **create mode only** (before the name field):
- Placeholder: `"Leave empty to auto-generate (e.g. browse-agent)"`
- Client-side validation: only letters, digits, hyphens, underscores

### 5. Frontend — i18n

`console/src/locales/{en,zh,ja,ru}.json`

Add keys:
- `agent.idLabel` — "Agent ID"
- `agent.idPlaceholder` — auto-generate hint
- `agent.idHelp` — allowed characters description
- `agent.idPattern` — validation error message

## Backward Compatibility

Fully backward compatible: `id` field is optional, existing behavior unchanged when omitted. Existing agents with random IDs continue to work. No migration needed.

flystar32:
## Summary

Allow users to optionally specify a custom `id` when creating an agent via the API or Console. When omitted, the existing random short UUID behavior is preserved.

Closes #3325

### Backend

- **`src/qwenpaw/config/config.py`**: Add `sanitize_agent_id()` (strip whitespace) and `validate_agent_id()` with pattern `[a-zA-Z0-9][a-zA-Z0-9_-]*[a-zA-Z0-9]`, length 2–64, reserved ID rejection, and duplicate check. Agent IDs are case-sensitive.
- **`src/qwenpaw/app/routers/agents.py`**: Add optional `id` field to `CreateAgentRequest` with sanitize validator. Update `create_agent()` to use custom ID when provided, otherwise generate random ID via `_generate_unique_id()` helper. Update route description.

### Frontend

- **`console/src/api/types/agents.ts`**: Add `id?: string` to `CreateAgentRequest`.
- **`console/src/pages/Settings/Agents/components/AgentModal.tsx`**: Add optional ID input in create mode with client-side regex validation.
- **`console/src/locales/{en,zh,ja,ru}.json`**: Add `idLabel` and `idHelp` i18n keys.

### Tests

- **`tests/unit/workspace/test_agent_id.py`**: Add 21 new test cases covering sanitization, valid/invalid patterns (including mixed-case), length limits, reserved IDs, and duplicate rejection. All 24 tests pass.

## Backward Compatibility

Fully backward compatible — `id` is optional, existing agents with random IDs are unaffected, no migration needed. Case-sensitivity is preserved to stay consistent with existing auto-generated IDs (shortuuid uses mixed-case characters).


rayrayraykk (approved):
LGTM

PR #3443, opened 2026-04-15
feat(routing): add local model routing

vvv214:
## Theme
This PR brings QwenPaw's existing local/cloud routing config back into the live runtime and upgrades it from a mode-only skeleton to a deterministic routing v1.

## What this PR adds
- wire `llm_routing` back into `create_model_and_formatter`
- keep the existing dual-slot `local/cloud` config shape
- add deterministic routing rules for:
  - structured output
  - non-text user content
  - `tool_choice=required`
  - recent tool context
- make multimodal capability checks routing-aware so cloud-routed multimodal requests are not pre-stripped
- keep formatter/media behavior aligned with routing decisions
- add focused backend unit tests for routing policy, model-factory wiring, formatter normalization, and prompt multimodal capability resolution

## What this PR does not add
- learned router / local judge
- benchmark or experiment scripts
- shadow evaluation / route logging
- extra UI control surface for routing modes

## Why now
QwenPaw still carries the `llm_routing` config and `RoutingChatModel` skeleton on `main`, but today it only does `local_first` / `cloud_first`. This PR makes that path actually usable without expanding the config model.

## Related history
- supersedes #2144
- builds on the earlier minimal backend routing work from #849

## Testing
- `PYTHONPATH=src python -m pytest tests/unit/agents/test_prompt_multimodal_routing.py tests/unit/agents/test_routing_chat_model.py tests/unit/agents/test_routing_model_factory.py tests/unit/agents/test_model_factory_message_normalization.py -q`
- `pre-commit run --files src/qwenpaw/agents/prompt.py src/qwenpaw/agents/routing_chat_model.py src/qwenpaw/agents/model_factory.py tests/unit/agents/test_prompt_multimodal_routing.py tests/unit/agents/test_routing_chat_model.py tests/unit/agents/test_routing_model_factory.py tests/unit/agents/test_model_factory_message_normalization.py`

PR #3451, opened 2026-04-15
feat(routing): add local model routing strategies

vvv214:
## Summary
- wire `llm_routing` into the current QwenPaw runtime path
- add deterministic local/cloud routing for structured output, non-text user content, required tools, and recent tool context
- make multimodal prompt/message handling routing-aware so media is preserved for cloud-bound requests

## Validation
- `python3 -m compileall src/qwenpaw/agents/model_factory.py src/qwenpaw/agents/routing_chat_model.py src/qwenpaw/agents/prompt.py`

Split out from #3443 so routing/runtime changes can be reviewed separately from the Settings UI.

### DIFF
--- a/src/qwenpaw/agents/model_factory.py
+++ b/src/qwenpaw/agents/model_factory.py
@@ -36,8 +36,10 @@ except ImportError:  # pragma: no cover - compatibility fallback
 from .utils.message_request_normalizer import (
     normalize_messages_for_model_request,
 )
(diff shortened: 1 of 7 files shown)


vvv214:
Closing in favor of #3550. The backend/runtime implementation has already been superseded there; the remaining delta here is mostly older tests around the previous RoutingChatModel-based shape, which no longer matches the current implementation.

PR #3525, opened 2026-04-17
feat(cron): create Discord thread before agent dispatch

energyd:
> **Updated 2026-05-09**: Rebased onto main after #4117 (cron channel-based session isolation) merged. `share_session` (#4117) and `create_thread` (this PR) are orthogonal: the former controls agent memory isolation, the latter controls Discord output routing. In the rebased executor, `begin_subthread` rewrites `target_session_id` before `share_session` sets `req["session_id"]`, so cron output streams into the new thread regardless of `share_session` mode.

## Problem

Cron jobs that dispatch to Discord channels have no way to isolate their output into a thread. The agent's replies, tool call status messages, and streamed content all land directly in the parent channel, mixing with regular conversation.

For daily digest jobs (e.g., news summaries, monitoring reports), this is especially problematic — the digest content clutters the channel and can't be collapsed or followed up on independently.

## Solution

Add `create_thread` and `thread_title` fields to `DispatchSpec`. When `create_thread=True`, the cron executor creates a Discord thread in the target channel before dispatching the agent. The agent's session is rewritten to the thread, so all output (replies, tool messages, streamed content) goes to the thread instead of the parent channel.

```json
{
  "dispatch": {
    "channel": "discord",
    "create_thread": true,
    "thread_title": "Daily Report {date}",
    "target": { "session_id": "discord:ch:123", "user_id": "456" }
  }
}
```

`{date}` in the title is replaced with `YYYY-MM-DD`. If `thread_title` is omitted, defaults to `"{job_name} {date}"`.

No-op when `create_thread` is false (default) — existing behavior unchanged.

Builds on #3144 (Discord thread awareness) which was merged in v1.1.2.

## Testing

- Cron job with `create_thread: true` → thread created, agent replies inside thread
- Cron job without `create_thread` → unchanged behavior, replies in channel
- Thread title with `{date}` placeholder → correctly substituted

## Updated (post-review)

Refactored to make thread dispatch channel-agnostic, per review feedback. The "Solution" section above describes the original Discord-specific shape; the updated implementation moves thread creation into the target channel.

**Abstraction change:**
- `BaseChannel.begin_subthread(user_id, session_id, title) -> session_id` — default is a no-op, returns session_id unchanged. Channels with native thread support override.
- `DiscordChannel.begin_subthread` — concrete implementation (moved from executor). Creates a public thread, returns `discord:ch:{thread_id}` session.
- Executor no longer hardcodes `"discord"`; it invokes `ch.begin_subthread(...)` on the target channel.

Only Discord has a concrete implementation in this PR. Slack/Feishu thread semantics differ (parent message required to get `thread_ts`, which changes session_id format), so those are better suited to follow-up PRs.

**Also fixed:**
- Use `job.schedule.timezone` (not server local time) for `{date}` rendering
- Truncate thread names to Discord's 100-char limit
- Reject `task_type=text` + `create_thread=true` at validation time
- Fail fast with `ChannelError` on DM session_ids in `DiscordChannel.begin_subthread`

### DIFF
--- a/src/qwenpaw/app/channels/base.py
+++ b/src/qwenpaw/app/channels/base.py
@@ -1348,6 +1348,24 @@ class BaseChannel(ABC):
         """
         raise NotImplementedError
 
+    async def begin_subthread(
+        self,
+        *,
+        user_id: str,
+        session_id: str,
+        title: str,
+    ) -> str:
+        """Create a sub-thread scoped to *session_id* and return the new
+        session_id that routes to it.
+
+        Used by cron dispatch to isolate job output from the parent channel.
+        Default implementation is a no-op that returns *session_id* unchanged —
+        channels without native thread support (WeChat, iMessage, etc.) inherit
+        this. Channels with thread semantics (Discord, Slack, Feishu) override
+        to create a thread and return a thread-scoped session_id.
+        """
+        return session_id
+
     def to_handle_from_target(self, *, user_id: str, session_id: str) -> str:
         """Map cron dispatch target to channel-specific to_handle.
 
--- a/src/qwenpaw/app/channels/discord_/channel.py
+++ b/src/qwenpaw/app/channels/discord_/channel.py
@@ -690,6 +690,61 @@ class DiscordChannel(BaseChannel):
         if self._client:
             await self._client.close()
 
+    # Discord caps thread names at 100 chars (see discord.Thread.name docs).
+    _DISCORD_THREAD_NAME_MAX: int = 100
+
+    async def begin_subthread(
+        self,
+        *,
+        user_id: str,
+        session_id: str,
+        title: str,
+    ) -> str:
+        """Create a public Discord thread in the given channel and return a
+        session_id scoped to the thread.
+
+        Titles longer than Discord's 100-character limit are truncated.
+
+        Raises:
+            ChannelError: session_id is not a ``discord:ch:<id>`` handle
+                (DMs don't support threads) or the Discord client is not
+                ready.
+        """
+        if not self.enabled or not self._client or not self._client.is_ready():
+            raise ChannelError(
+                channel_name="discord",
+                message="Discord client is not ready",
+            )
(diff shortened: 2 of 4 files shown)


xieyxclack:
I'm not sure, but maybe a better approach would be to configure similar threads for scheduled tasks across all channels, rather than only for Discord.

@rayrayraykk Any suggestion?

energyd:
@xieyxclack

Thanks for the review — refactored and pushed. See the "Updated (post-review)" section in the PR description for details.

energyd:
gentle bump @xieyxclack @rayrayraykk

energyd:
@xieyxclack @rayrayraykk pls review

xieyxclack:
@energyd Thank you for your contribution. We will review this PR after the holiday.

zhijianma:
@energyd 

Thank you for your contribution.

We have implemented channel isolation and session sharing options in PR #4117.

Please confirm whether this request is still needed.

energyd:
> @energyd
> 
> Thank you for your contribution.
> 
> We have implemented channel isolation and session sharing options in PR #4117.
> 
> Please confirm whether this request is still needed.

Thanks @zhijianma — checked #4117. It nails the session/context isolation side via runtime.share_session, but it doesn't create a Discord thread or route cron output into one.

The agent's req["session_id"] is rewritten when share_session=false, so memory is isolated. But send_event in crons/executor.py still uses the original target_session_id, so streamed content and tool status still land in the parent channel.

#3525 solves the UI side: create_thread=true creates a Discord subthread before dispatch, then uses discord:ch:<thread_id> as the dispatch session — replies and streamed content all go into the thread. Motivating use case: daily digest cron jobs that would otherwise clutter the parent channel.

The two are complementary:
- share_session: memory/context isolation within a target
- create_thread: channel-native subthread + route dispatch and output into it

I've rebased on top of #4117 — see the updated description for how they compose.

please take another look 🙏

PR #3574, opened 2026-04-19
feat(chat): replace Web Speech API with Whisper transcription for voice input

tqjason:
## Summary
- Replace browser's native Web Speech API with custom Whisper-based voice transcription
- Support browsers that don't support Web Speech API (e.g. 豆包浏览器)
- Add keyboard shortcut Ctrl+Shift+M (Cmd+Shift+M on Mac) for voice recording

## Changes
- Add POST `/agent/transcribe` endpoint for audio transcription via configured Whisper provider
- Add `WhisperSpeechButton` component using MediaRecorder API for recording
- Add i18n strings for voice recording states and error messages
- Disable default `allowSpeech` and use custom Whisper button in sender prefix

## Test plan
1. Configure Whisper provider in Settings > Voice Transcription
2. Click microphone button in Chat page to start recording
3. Use Ctrl+Shift+M shortcut to toggle recording
4. Verify transcribed text appears in input box

🤖 Generated with [Claude Code](https://claude.ai/code)


xieyxclack:
@tqjason  Please format the code, refer to https://github.com/agentscope-ai/QwenPaw/blob/main/CONTRIBUTING.md#4-code-and-quality

xieyxclack:
@tqjason Some suggestions：
- Every other endpoint in this router accepts a `request: Request` parameter and calls `get_agent_for_request(request)`. `/agent/transcribe` skips that entirely.
- Changing `allowSpeech` from `true` to `false` unconditionally removes voice input for users who haven't configured a Whisper provider. Consider a graceful degradation strategy: When Whisper is configured, use the custom Whisper button; when it's not, fall back to the native Web Speech API.
- (Maybe Optional): Add file size limit and file type validation

tqjason:
ok, I'll try.
But, About `request` and `get_agent_for_request(request)`, it seems unnecessary for my function. There is no need to know current agent workspace. Is using get_agent_for_request mandatory?

xieyxclack:
@tqjason ok, it is not mandatory

tqjason:
new commits for
+ Using a graceful degradation strategy: When Whisper is configured, use the custom Whisper button; when it's not, fall back to the native Web Speech API.
+ Add file size limit and file type validation

xieyxclack:
@tqjason Thank you, I will review this pr later

console/src/pages/Chat/components/WhisperSpeechButton/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+>(({ disabled }, ref) => {
+  const { t } = useTranslation();
+  const [recording, setRecording] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const mediaRecorderRef = useRef<MediaRecorder | null>(null);
+  const chunksRef = useRef<Blob[]>([]);
+  const internalRecordingRef = useRef(false);
+  const recordingTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const setTextareaValue = useCallback(
```
xieyxclack: Querying the DOM via `document.querySelector('[class*="sender"]')` and setting textarea value via native property descriptor is fragile — it bypasses React's controlled component model and will break if class names change. Consider accepting an `onTranscription: (text: string) => void` callback prop instead, and let the parent (`ChatPage`) handle text insertion through the Sender component's API.

@zhaozhuang521 Any further suggestions here?

src/qwenpaw/app/routers/workspace.py
```diff
@@ -472,6 +473,83 @@ async def put_transcription_provider(
...
+        text = await transcribe_audio(tmp_path)
+        if text is None:
+            raise HTTPException(
+                status_code=500,
+                detail="Transcription failed. Check provider configuration.",
+            )
+        return {"text": text}
+    finally:
+        # Clean up temp file
+        os.unlink(tmp_path)
```
xieyxclack: Bare `os.unlink` in `finally` can raise `PermissionError` on Windows if the file handle isn't released, masking the original exception. Maybe wrap it as
```python
finally:
     try:
         os.unlink(tmp_path)
     except OSError:
         pass
 ```

File: console/src/pages/Chat/components/WhisperSpeechButton/index.tsx [hunk 3]
```diff
@@ -0,0 +1,285 @@
...
+            if (textarea) {
+              const currentValue = textarea.value || "";
+              const newValue = currentValue
+                ? `${currentValue} ${result.text}`
+                : result.text;
+              setTextareaValue(textarea, newValue);
+              textarea.focus();
+            }
+          }
+        } catch (err) {
```
xieyxclack, line 152: Error type detection via `errMsg.includes("Transcription is disabled")` etc. is fragile — if the backend message text changes, the branch breaks silently. Consider matching on HTTP status code (already available from the thrown error) or having the backend return a structured error code field.

@zhaozhuang521 Any further suggestions?

xieyxclack (commented):
Please refer to the inline comments for suggestions

console/src/locales/en.json
```diff
@@ -1562,6 +1560,16 @@
       "imageOnlyAccept": "Current model only supports image files",
       "imageOnlyWarning": "Current model only supports images, non-image files like videos may not be processed correctly"
     },
+    "speech": {
```
zhaozhuang521, line 1565: The internationalization is missing three languages. Please add them.

console/src/pages/Chat/components/WhisperSpeechButton/index.tsx
tqjason: ok, I will try `onTranscription` callback prop instead.

And, I also found two other `[class*="sender"]` in console/src/pages/Chat/index.tsx
line 368 handleKeyDown , line 432 handleFocus

so handleWhisperTranscription in parent may still use `[class*="sender"]`

zhaozhuang521:
Please review the questions we raised and make the necessary revisions. @tqjason

console/src/pages/Chat/components/WhisperSpeechButton/index.tsx
tqjason: ok, I will try structured error

tqjason:
new commits. ja/ru/pt-BR translated by AI

zhaozhuang521 (approved):
LGTM

xieyxclack (approved):
LGTM

PR #3575, opened 2026-04-19
fix(app): reload running agents after global LLM updates

vvv214:
## Summary

Keep live chat sessions in sync with global-scope LLM configuration changes.

- reload every running agent after a global `PUT /models/active` change
- reload every running agent after a global `PUT /config/agents/llm-routing` change
- cover both paths with unit + FastAPI-TestClient integration tests (the latter guards the `asyncio.create_task` lifecycle under the real request path)

## Test plan

- [x] `./.venv/bin/pytest tests/unit/app/routers/test_providers_set_active_model.py tests/unit/app/routers/test_config_set_llm_routing.py tests/integration/test_global_llm_reload_e2e.py` — 8 passed
- [ ] Manual: change global active model in Settings and verify existing chat sessions pick it up without restart
- [ ] Manual: change global routing mode in Settings and verify existing chat sessions pick it up without restart

### DIFF
--- a/src/qwenpaw/app/routers/config.py
+++ b/src/qwenpaw/app/routers/config.py
@@ -6,7 +6,7 @@ from typing import Any, List, Optional
 from fastapi import APIRouter, Body, HTTPException, Path, Request
 from pydantic import BaseModel
 
-from ..utils import schedule_agent_reload
+from ..utils import schedule_agent_reload, schedule_all_agents_reload
 from ...config import (
     load_config,
     save_config,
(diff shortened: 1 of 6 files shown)


vvv214:
Closing for now. I could not reliably reproduce a user-visible stale-model bug on the normal chat path: the runner reloads agent config and reconstructs the agent/model on each query. If we later find a concrete long-lived component that truly needs reload on global model/routing changes, I will reopen with a narrower repro and scope.

PR #3599, opened 2026-04-20
feat(agents): add per-agent model assignment in settings UI

seoeaa:
## Summary

- Add model column to Agents table — shows the currently assigned model for each agent (or "Use global default" when unset)
- Add model selector to Agent create/edit modal — two cascading dropdowns (Provider → Model) allow assigning a specific LLM to any agent
- Backend: include active_model in agent list API — the GET /api/agents endpoint now returns the active_model field in each agent summary

## Motivation

Previously, assigning a model to a specific agent was only possible via CLI or direct API calls. This PR brings that capability to the Settings → Agents management page.

## Changes

| File | Change |
|------|--------|
| src/qwenpaw/app/routers/agents.py | Add active_model field to AgentSummary; populate it in list_agents |
| console/src/api/types/agents.ts | Add active_model to AgentSummary and AgentProfileConfig |
| console/src/pages/Settings/Agents/components/AgentTable.tsx | Add "Model" column with provider icon and model name |
| console/src/pages/Settings/Agents/components/AgentModal.tsx | Add cascading provider/model Select with search |
| console/src/pages/Settings/Agents/index.tsx | Wire form fields through create/edit/submit flows |
| console/src/locales/{en,ru,zh,ja}.json | Add i18n keys for model-related labels |


console/src/pages/Settings/Agents/index.tsx
```diff
@@ -119,6 +134,13 @@ export default function AgentsPage() {
           ...payload,
           skill_names: selectedSkills,
         });
+        if (active_model) {
```
zhaozhuang521: In the current Agent creation process, if the user specifies a model, `createAgent` is called first to create the Agent, and then `updateAgent` is called immediately afterward to set the `active_model`, resulting in two network requests. To address this, the backend's `createAgent` interface should directly support the `active_model` field during creation, requiring only a single call from the frontend.

zhaozhuang521:
@seoeaa Please take a look at the question I raised.

console/src/pages/Settings/Agents/index.tsx
seoeaa: Conflicts are prohibited, the extra update Agent call is removed - now creating an agent with a model is done in one request.

console/package-lock.json
```diff
@@ -6679,7 +6679,7 @@
     },
     "node_modules/@vvo/tzdb": {
       "version": "6.198.0",
-      "resolved": "https://registry.npmmirror.com/@vvo/tzdb/-/tzdb-6.198.0.tgz",
```
xieyxclack, line 6682: Do not need to modify this?

src/qwenpaw/app/routers/agents.py
```diff
@@ -45,6 +47,7 @@ class AgentSummary(BaseModel):
     description: str
     workspace_dir: str
     enabled: bool
+    active_model: dict | None = None
```
xieyxclack: `active_model` should be typed as `ModelSlotConfig | None`

xieyxclack (commented):
LGTM, please refer to inline comments for minor suggestions

xieyxclack:
@seoeaa It seems that the changes were not properly submitted ?

seoeaa:
The problem was that git checkout bc8138ec -- console/package-lock.json rolled back the file to the base commit state, but agents.py also turned out to be uncommitted. Now both changes are sent to PR by separate commits

xieyxclack (approved):
LGTM

PR #3679, opened 2026-04-22
fix(channels): filter_tool_messages must suppress in-progress tool preview

hlgone:
## Summary

`filter_tool_messages=True` is documented as a channel-level switch
that hides the agent's tool execution process from end users. It works
for completed `tool_use` / `tool_result` messages, but the **in-progress
tool output preview** emitted on the content stream still gets forwarded
to the channel. On Feishu this shows up as the user seeing shell command
output, API responses, etc. streaming in mid-execution — the exact
thing the flag is supposed to prevent.

## Reproduction

1. Configure a channel with `filter_tool_messages: true`
2. Ask the agent anything that triggers a tool call with a substantive
   payload (`list_devices`, `read_file`, an HTTP call, …)
3. Observe the tool's output / preview appearing in the chat before
   the final answer — even though `filter_tool_messages` is on.

## Fix

`BaseChannel._is_tool_progress_event` is the gate that decides whether
an in-progress content event should be dropped. It previously only
dropped events whose `data.output` looked tool-shaped. After this
change, when `self._filter_tool_messages` is set we short-circuit to
`True` for any in-progress event, so the whole preview stream is
suppressed before it reaches `send_content_parts`.

The scope is narrow: only in-progress events (`status=InProgress`) on
channels that explicitly opted into filtering are affected. Normal
content streaming and completed messages go through unchanged code
paths.

## Test plan

- [x] Added `TestOnEventContent::test_filter_tool_messages_suppresses_stream_tool_preview`:
      with the flag enabled, an `InProgress` `DATA` event carrying a
      tool output payload must be handled (returned True) and
      `send_content_parts` must not be awaited.
- [ ] CI (pre-commit + pytest) — requesting upstream validation; local
      dev env isn't set up on my side.

## Note

`CONTRIBUTING.md` suggests opening an issue first for new proposals.
This is a narrow correctness bug rather than a design change (the flag
already exists and is documented; it just misses one code path), so
I'm opening directly. Happy to split into an issue + PR if maintainers
prefer.

### DIFF
--- a/src/qwenpaw/app/channels/base.py
+++ b/src/qwenpaw/app/channels/base.py
@@ -963,6 +963,17 @@ class BaseChannel(ABC):
         data = getattr(event, "data", None) or {}
         if not isinstance(data, dict) or "output" not in data:
             return False
+        # At this point the event is a tool-output preview: InProgress DATA
+        # event carrying a ``data.output`` payload. ``filter_tool_messages``
+        # is meant to hide the tool execution process from end users, so
+        # drop the preview here -- otherwise channels like Feishu still
+        # leak intermediate shell/API output.
+        #
+        # This check MUST stay after the shape guard above. A blanket
+        # InProgress suppression would also swallow the agent's own
+        # streaming reply content, silently breaking the whole channel.
+        if self._filter_tool_messages:
+            return True
         body = self._format_stream_tool_output_body(event)
         if not body:
             return False
--- a/tests/unit/channels/test_base_core.py
+++ b/tests/unit/channels/test_base_core.py
@@ -994,6 +994,86 @@ class TestStreamWithTracker:
                                 pass
 
 
+@pytest.mark.asyncio
+class TestOnEventContent:
+    """Regression tests for in-progress content event handling."""
+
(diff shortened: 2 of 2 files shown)


hlgone:
Self-review update (force-pushed to the PR branch):

Found a correctness issue in the original patch and fixed it before anyone had to review the broken version. The first iteration placed the `filter_tool_messages` check **before** the tool-shape guard (`"output" in data`), which meant a blanket `return True` on any in-progress DATA event — silently swallowing the agent's own streaming reply content, not just tool previews. Agent appears to hang with no output on channels like WeChat that route replies through that path.

Fix: moved the flag check to after the `"output" in data` guard, and added a regression test (`test_filter_tool_messages_does_not_swallow_non_tool_event`) to pin the ordering constraint. Inline comment now documents why the ordering matters.

Commit was amended (not stacked) since the PR was only minutes old and hadn't been reviewed yet. New HEAD: 09721b06.

xieyxclack:
It has been fixed in https://github.com/agentscope-ai/QwenPaw/pull/3660

PR #3685, opened 2026-04-22
fix(channels/weixin): accept empty sendmessage response as success

hlgone:
## Summary

The iLinkAI `sendmessage` endpoint returns an empty dict `{}` on success — the call is fire-and-forget at the business layer even when HTTP 200 succeeds. Sibling endpoints like `getconfig` follow a different convention and return `{"ret": 0, ...}`. `_send_text_direct` was reading both endpoints with the same rule:

```python
ret = resp.get("ret", 1) if isinstance(resp, dict) else 1
if ret != 0 or errcode != 0:
    logger.warning("weixin send_text rejected: ...")
    return
```

Because `sendmessage`'s successful response is `{}`, `resp.get("ret", 1)` defaulted to 1, flipped `rejected=True`, logged a WARNING, and returned. **Every successful WeChat reply was being reported as a rejection.** Worse, when the check was later audited it turned out the same branch had been silently hiding genuine send failures behind the same code path.

### How this was noticed

While debugging a user-visible "agent responds, WeChat shows nothing" symptom, the channel added the `rejected` log line and immediately three `rejected` WARNINGs appeared per query. Cross-checking the WeChat client showed the messages were in fact delivered — so the success detection was wrong.

## Fix

- `_send_text_direct` now treats missing `ret` / `errcode` as success. Only a server-surfaced non-zero `ret` or `errcode` triggers the `rejected` branch.
- A non-dict response (None / list / string / anything else) is treated as **upstream contract drift**, not success. It logs `send_text unexpected response shape` at WARNING and does not emit an `ok` line. The old "coerce to {}" fallback would have silently masked the next protocol change just like it masked this one.
- Added a DEBUG-only diagnostic line before the send recording `to_user_id` / `text_len` / `context_token_len` / `token_source` (meta vs cache vs empty) / `cache_has_user`. Gated behind `logger.isEnabledFor(DEBUG)` so the per-session auth-token size and provenance never land in default production logs. **No raw token bytes or prefixes are logged at any level.**

## Test plan

Added to `tests/unit/channels/test_weixin.py::TestWeixinSendMethods`:

- [x] `test_send_text_direct_empty_response_is_success` — iLinkAI's `{}` → `ok`, no `rejected`.
- [x] `test_send_text_direct_nonzero_ret_is_rejected` — explicit `{"ret": 1}` still trips rejection.
- [x] `test_send_text_direct_nonzero_errcode_is_rejected` — a non-zero `errcode` alone is enough (covers future error codes that don't surface `ret`).
- [x] `test_send_text_direct_non_dict_response_is_unexpected` — `None`-style response logs `unexpected response shape` and does NOT log `ok` / `rejected`.
- [ ] CI (pre-commit + pytest): local QwenPaw venv doesn't have dev extras installed here; requesting upstream CI validation.

## Relationship to #3679

Same broad area (channel send observability) but **fully orthogonal**:

- #3679 fixes the generic `BaseChannel.on_event_content` filter so `filter_tool_messages=True` also suppresses in-progress tool previews — affects every channel.
- This PR fixes WeChat-specific response parsing — only touches `channels/weixin/`.

They can merge in either order.


src/qwenpaw/app/channels/weixin/channel.py
```diff
@@ -973,8 +973,78 @@ async def _send_text_direct(
...
+                    resp,
+                )
+                return
+            logger.info(
+                "weixin send_text ok: to_user_id=%s text_len=%s",
+                to_user_id,
+                len(text),
+            )
         except Exception:
             logger.exception("weixin _send_text_direct failed")
```
hongxicheng: Logging is over-engineered — the DEBUG token-source diagnostics and non-dict type guard add a lot of complexity for what should be a simple response check. Something like this would be sufficient

```suggestion
try:
    resp = await _client.send_text(to_user_id, text, context_token)
    if isinstance(resp, dict):
        ret = resp.get("ret", 0)
        errcode = resp.get("errcode", 0)
        if ret != 0 or errcode != 0:
            logger.warning(
                "weixin send_text rejected: ret=%s errcode=%s to_user_id=%s",
                ret, errcode, to_user_id,
            )
except Exception:
    logger.exception("weixin _send_text_direct failed")
```

tests/unit/channels/test_weixin.py
hongxicheng, line 1: Please remove the new test cases — all 4 are failing in CI.

hongxicheng (commented):
Hi @hlgone, thanks for the PR! Pre-commit is failing — please run pre-commit run --all-files locally before resubmitting.
A few things:

hongxicheng:
LGTM

hlgone:
Thanks for the review! Ready to merge when you have a chance.

xieyxclack (approved):

xieyxclack:
@hlgone Thank you for your contribution!

PR #3700, opened 2026-04-22
fix(channels/weixin): extend QR-status polling timeout to 60s

hlgone:
## Summary

- `get_qrcode_status` is a server-driven short long-poll (iLink holds the connection for up to ~30s before returning), but the shared `_DEFAULT_TIMEOUT = 15.0` in `src/qwenpaw/app/channels/weixin/client.py` was tripping `httpx.ReadTimeout` on the very first poll cycle.
- The outer `wait_for_login(max_wait=300)` never got a chance to run — login died at ~15s with a `weixin: QR code login failed` trace straight out of the transport, not at the 300s budget the user sees advertised.
- Add a dedicated `_QRCODE_STATUS_TIMEOUT = 60.0`, promote `_get`'s timeout to a keyword-only override (default unchanged), and have `get_qrcode_status` pass the longer budget. Normal GETs keep 15s so this change is scoped to the one endpoint that needs it.

## Type of Change

- Bug fix

## Component(s) Affected

- Channels — WeChat (`src/qwenpaw/app/channels/weixin/client.py`)
- Tests (`tests/unit/channels/test_weixin.py`)

## Repro (before fix)

With \`HTTPS_PROXY\` set in the environment (or slow path to \`ilinkai.weixin.qq.com\`):

\`\`\`
INFO  weixin: Please scan the QR code to log in.
INFO  weixin: waiting for QR code scan (up to 300s)…
# ...~15 seconds later, before the user's phone finishes confirming:
ERROR weixin: QR code login failed
Traceback ...
  File ".../httpcore/_async/http_proxy.py", line 343, in handle_async_request
    ...
httpcore.ReadTimeout
ChannelError: [500] EXTERNAL_SERVICE_ERROR: External service weixin error: WeChat QR code login failed.
\`\`\`

After fix (same environment): QR code login succeeds in ~32s (user scan + server confirm), \`bot_token saved\`, \`weixin channel started\`.

## Design notes

- **Why 60s**: iLink's server-side hold is documented as up to ~35s. 60s gives headroom for proxy overhead without blocking so long that a truly-lost connection lingers. The outer \`wait_for_login(max_wait=300)\` remains the authoritative budget for total login time (~5 poll cycles × 60s worst case, typically ≤2 cycles).
- **Why a keyword-only override on \`_get\`**: rather than a second copy of the GET helper. Keeps normal GETs at 15s; only this one endpoint opts in.
- **Why not change \`trust_env\`**: disabling env-based proxies system-wide for the weixin client would be a bigger behavioral change with unclear impact for users who legitimately need a proxy. The timeout fix addresses the root cause (polling cycle > client-side budget) regardless of proxy presence.

## Testing

Unit test asserts \`get_qrcode_status\` calls the underlying \`httpx.AsyncClient.get\` with \`timeout=_QRCODE_STATUS_TIMEOUT\` (and that \`_QRCODE_STATUS_TIMEOUT > _DEFAULT_TIMEOUT\` as a guard against a future accidental merge of the two constants).

Manual: reproduced the failure with the proxy env on a slow path, confirmed login-to-token completes after the fix.

## Checklist

- [x] Run and pass \`pre-commit run --files src/qwenpaw/app/channels/weixin/client.py tests/unit/channels/test_weixin.py\`
- [x] Run and pass relevant tests (\`tests/unit/channels/test_weixin.py\`, 79 passed)

\`\`\`bash
\$ pre-commit run --files src/qwenpaw/app/channels/weixin/client.py tests/unit/channels/test_weixin.py

\$ python -m pytest tests/unit/channels/test_weixin.py -q
79 passed, 2 warnings in 0.96s
\`\`\`

### DIFF
--- a/src/qwenpaw/app/channels/weixin/client.py
+++ b/src/qwenpaw/app/channels/weixin/client.py
@@ -38,13 +38,8 @@ _CHANNEL_VERSION = "2.0.1"
 # Long-poll hold time is up to 35 seconds (server-controlled)
 _GETUPDATES_TIMEOUT = 45.0
 _DEFAULT_TIMEOUT = 15.0
-# QR-code status polling is a short long-poll from the iLink side; the
-# server can hold the connection for up to ~30s before returning. A
-# 15s client-side read timeout repeatedly kills the polling window and
-# surfaces as `httpx.ReadTimeout` before the user's scan confirmation
-# ever reaches us. Give it a dedicated, looser timeout so the outer
-# `wait_for_login(max_wait=300)` loop is the thing that governs how
-# long we wait, not the transport.
+# iLink holds QR-status polls up to ~30s; use a dedicated longer timeout
+# so wait_for_login governs total wait time, not the transport.
 _QRCODE_STATUS_TIMEOUT = 60.0
 
 
@@ -181,7 +176,14 @@ class ILinkClient:
         """
         elapsed = 0.0
         while elapsed < max_wait:
-            data = await self.get_qrcode_status(qrcode)
+            try:
+                data = await self.get_qrcode_status(qrcode)
+            except httpx.ReadTimeout:
+                logger.warning(
+                    "weixin: QR status poll timed out, retrying…",
+                )
+                elapsed += poll_interval
+                continue
             status = data.get("status", "")
             if status == "confirmed":
                 token = data.get("bot_token", "")
--- a/src/qwenpaw/app/channels/weixin/client.py
+++ b/src/qwenpaw/app/channels/weixin/client.py
@@ -38,6 +38,14 @@ _CHANNEL_VERSION = "2.0.1"
 # Long-poll hold time is up to 35 seconds (server-controlled)
 _GETUPDATES_TIMEOUT = 45.0
 _DEFAULT_TIMEOUT = 15.0
+# QR-code status polling is a short long-poll from the iLink side; the
+# server can hold the connection for up to ~30s before returning. A
+# 15s client-side read timeout repeatedly kills the polling window and
+# surfaces as `httpx.ReadTimeout` before the user's scan confirmation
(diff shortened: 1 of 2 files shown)


xieyxclack:
Please help review this pr @hongxicheng

src/qwenpaw/app/channels/weixin/client.py
```diff
@@ -38,6 +38,14 @@
 # Long-poll hold time is up to 35 seconds (server-controlled)
 _GETUPDATES_TIMEOUT = 45.0
 _DEFAULT_TIMEOUT = 15.0
+# QR-code status polling is a short long-poll from the iLink side; the
+# server can hold the connection for up to ~30s before returning. A
+# 15s client-side read timeout repeatedly kills the polling window and
+# surfaces as `httpx.ReadTimeout` before the user's scan confirmation
+# ever reaches us. Give it a dedicated, looser timeout so the outer
+# `wait_for_login(max_wait=300)` loop is the thing that governs how
+# long we wait, not the transport.
```
hongxicheng: The 7-line comment above _QRCODE_STATUS_TIMEOUT is overkill for a single constant.

tests/unit/channels/test_weixin.py
hongxicheng, line 1: Remove the unit test

hongxicheng (commented):
Hi @hlgone, thanks for the fix! The root cause analysis is solid and the change is well-scoped. A few suggestions:

xieyxclack (approved):

hlgone:
Thanks for the review — both suggestions addressed in 1ba62012 (comment trimmed to 2 lines, unit test removed). Also added an `httpx.ReadTimeout` retry in `wait_for_login` as a belt-and-suspenders against transient timeouts.

PR #3704, opened 2026-04-22
fix(mcp): resolve empty env values from os.environ at load time to re-enable disabled clients

### LINKED ISSUE #3703: bug: MCP client stays disabled after setting API key via UI (env vars not re-read from agent.json)
## Bug Description

When an MCP client (e.g. `tavily_search`) is created before its required API key is configured, the key value and `enabled` flag get **frozen as empty/false** in `agent.json`. Even after setting the key through the Settings → Environment Variables UI, the MCP client remains disabled on the next run.

## Steps to Reproduce

1. Start the app **without** `TAVILY_API_KEY` set.
2. The default agent workspace is created → `~/.deepclaw/workspaces/default/agent.json` records:
   ```json
   "tavily_search": {
     "enabled": false,
     "env": { "TAVILY_API_KEY": "" }
   }
   ```
3. Go to **Settings → Environment Variables**, add `TAVILY_API_KEY = <your_key>` and save.
4. The key is correctly persisted to `~/.deepclaw.secret/envs.json` (encrypted) and injected into `os.environ`.
5. Run an agent task that should use Tavily search → **tool not available**, MCP client still disabled.

## Root Cause

`MCPConfig.clients` default factory captures `os.getenv("TAVILY_API_KEY")` **at instance-creation time**. Once saved to `agent.json`, those values are static. `load_agent_config` reads the JSON as-is; neither the empty `env` value nor `enabled: false` is re-evaluated against the live environment.

The `MCPConfigWatcher` only watches `config.json` for changes — it has no awareness of `envs.json` updates.

## Expected Behavior

Setting an API key through the UI should take effect on the next agent run, without requiring the user to manually edit `agent.json` or know where env vars are stored.

## Suggested Fix

In `MCPClientConfig._normalize_legacy_fields` (model validator, `mode="before"`), resolve empty `env` values from `os.environ` at deserialization time, and auto-enable the client if a previously-missing key is now available:

```python
if isinstance(payload.get("env"), dict):
    original_env = dict(payload["env"])
    resolved_env = {
        k: (v if v else os.environ.get(k, v))
        for k, v in original_env.items()
    }
    payload["env"] = resolved_env
    if not payload.get("enabled", True):
        key_newly_available = any(
            not original_env[k] and resolved_env[k]
            for k in original_env
        )
        if key_newly_available:
            payload["enabled"] = True
```

This ensures `agent.json` acts as a **template** rather than a frozen snapshot for env-gated values.

A PR with this fix is incoming.

## Environment

- Version: v1.1.4b1
- Platform: macOS Darwin 25.2.0

Dbraum:
## Summary

Fixes #3703

- When `agent.json` was created before an API key (e.g. `TAVILY_API_KEY`) was configured, the MCP client is persisted with `enabled: false` and an empty env value.
- After the user sets the key via **Settings → Environment Variables**, `os.environ` is updated but `agent.json` is not — the MCP client remains disabled on all subsequent runs.
- This PR fixes the issue by resolving empty `env` values from `os.environ` during `MCPClientConfig` deserialization, and auto-enabling the client if a previously-missing key is now available.

## Changes

**`src/qwenpaw/config/config.py`** — `MCPClientConfig._normalize_legacy_fields`:

```python
if isinstance(payload.get("env"), dict):
    original_env = dict(payload["env"])
    resolved_env = {
        k: (v if v else os.environ.get(k, v))
        for k, v in original_env.items()
    }
    payload["env"] = resolved_env
    if not payload.get("enabled", True):
        key_newly_available = any(
            not original_env[k] and resolved_env[k]
            for k in original_env
        )
        if key_newly_available:
            payload["enabled"] = True
```

## Test plan

- [ ] Start app without `TAVILY_API_KEY` — `tavily_search` client should remain disabled.
- [ ] Add `TAVILY_API_KEY` via Settings → Environment Variables.
- [ ] Restart app / trigger next agent run — `tavily_search` client should now be enabled and functional.
- [ ] Existing MCP clients with non-empty `env` values are unaffected.
- [ ] Clients explicitly disabled by the user (with non-empty env keys) remain disabled.

### DIFF
--- a/src/qwenpaw/config/config.py
+++ b/src/qwenpaw/config/config.py
@@ -1077,6 +1077,26 @@ class MCPClientConfig(BaseModel):
         if "isActive" in payload and "enabled" not in payload:
             payload["enabled"] = payload["isActive"]
 
+        # Resolve empty env values from os.environ at load time so that
+        # env vars set via the UI (persisted in envs.json) are picked up
+        # even if agent.json was created before the key was configured.
+        if isinstance(payload.get("env"), dict):
+            original_env = dict(payload["env"])
+            resolved_env = {
+                k: (v if v else os.environ.get(k, v))
+                for k, v in original_env.items()
+            }
+            payload["env"] = resolved_env
+            # If client was disabled solely because a required env key was
+            # missing (empty), re-enable it now that the key is available.
+            if not payload.get("enabled", True):
+                key_newly_available = any(
+                    not original_env[k] and resolved_env[k]
+                    for k in original_env
+                )
+                if key_newly_available:
+                    payload["enabled"] = True
+
         if "baseUrl" in payload and "url" not in payload:
             payload["url"] = payload["baseUrl"]
 


rayrayraykk:
Close as fixed in #3725

PR #3829, opened 2026-04-25
feat(chat): generate session titles asynchronously via LLM

ekzhu:
## Description

Replace the awkward "first 10 characters of the user message" placeholder shown in the Console session drawer with a real, model-generated title.

The console handler still creates the chat synchronously with the truncated placeholder (so the UI has something to show on first render), then spawns a detached `asyncio` task that asks the active chat model for a concise title (≤6 words, same language as the message) and persists it through `ChatManager.patch_chat`. The streaming chat response is unchanged — title generation never blocks the user-facing request, and failures are logged and swallowed. The next time the Console fetches `/chats` (drawer open / rename / delete) it picks up the updated name automatically; no frontend change required.

The LLM call follows the same shape as the existing direct caller in `app/routers/skills_stream.py` (raw `{"role", "content"}` dicts, `await model(messages)`, `(ValueError, AppBaseException)` factory-failure handling, `_extract_text_from_response` extraction).

**Related Issue:** N/A — improvement requested in chat.

**Security Considerations:** None. Title generation reuses the active agent's already-configured chat model and never widens the request surface; the user message is truncated to 500 chars before being sent, and the generated title is normalized (quotes/punctuation stripped, capped at 60 chars) before persistence. The `placeholder_name` guard in `generate_and_update_title` prevents a slow background task from clobbering a user-chosen rename.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

New unit suite at `tests/unit/app/test_title_generator.py` covering:

- `_clean_title` — whitespace, quote (single, double, smart), backtick stripping; trailing-punctuation removal; first-line-only; `MAX_TITLE_CHARS` truncation; empty / punctuation-only inputs.
- `_extract_text_from_response` — string, `.text`, string `.content`, list-of-dicts (returns first text item, matching `skills_stream`), list-of-objects, `None`, unknown shapes.
- `generate_and_update_title` end-to-end against a real `ChatManager` backed by `JsonChatRepository`:
  - happy path — cleaned title persisted via `patch_chat`,
  - prompt construction — system/user roles correct, user message truncated to `MAX_INPUT_CHARS`,
  - blank user message → no model call, no patch,
  - chat already renamed (placeholder mismatch) → user-chosen name preserved,
  - chat missing → no-op,
  - empty/whitespace model output → placeholder kept,
  - `create_model_and_formatter` raising `ValueError` → swallowed (matches `skills_stream.get_model` shape),
  - model invocation raising → swallowed,
  - model exceeding `TITLE_TIMEOUT_SECONDS` → swallowed (test monkeypatches the constant to 0.01s),
  - `asyncio.CancelledError` from the model still propagates so `task.cancel()` works.

Manual flow:

1. Start the backend and open the Console.
2. Send a new message in a fresh chat. The session drawer shows the 10-char placeholder immediately.
3. Within ~1–10s the background task replaces it with a generated title; reopen the drawer (or trigger any list refresh) to see the updated name.

```bash
pre-commit run --all-files
# (changed files only)
# black .................. Passed
# flake8 ................. Passed
# add-trailing-comma ..... Passed
# mypy ................... Success: no issues found in 3 source files
# pylint ................. Your code has been rated at 10.00/10
# check-ast / trailing-whitespace / fix-encoding-pragma ... Passed

pytest tests/unit/app/test_title_generator.py
# All test scenarios verified in an isolated harness (full project deps not
# installed in the dev sandbox); the test file follows the same import and
# fixture conventions as tests/unit/app/test_chat_updates.py and runs as-is
# under `make test-unit` in CI.
```

## Additional Notes

- Branch: `claude/improve-session-titles-vrv6V`
- Files changed:
  - `src/qwenpaw/app/routers/console.py` — extract placeholder name into `_extract_placeholder_name(...)` helper and dispatch the background title task.
  - `src/qwenpaw/app/runner/title_generator.py` (new) — LLM call, normalization, race-safe persistence.
  - `tests/unit/app/test_title_generator.py` (new) — unit coverage described above.
- Frontend intentionally untouched; the existing `sessionApi.getSessionList()` refresh path picks up the updated name.


xieyxclack:
@ekzhu  LGTM. One minor suggestion: Since title generation involves an extra LLM API call per new chat (with associated latency and cost), consider adding a configuration option to disable it. A boolean like `auto_generate_title` in the config (defaulting to `true`) would let users opt out if they prefer not to spend tokens on titles.

ekzhu:
> @ekzhu LGTM. One minor suggestion: Since title generation involves an extra LLM API call per new chat (with associated latency and cost), consider adding a configuration option to disable it. A boolean like `auto_generate_title` in the config (defaulting to `true`) would let users opt out if they prefer not to spend tokens on titles.

Done — config knob + UI surface, in two commits on the branch.

## Config knob — `95f1316`

New `AutoTitleConfig` model in `src/qwenpaw/config/config.py` next to `LightContextConfig`:

```python
class AutoTitleConfig(BaseModel):
    enabled: bool = True
    timeout_seconds: float = Field(default=30.0, ge=1.0)
```

Hangs off `AgentsRunningConfig.auto_title_config` so it's per-agent and lives next to the other LLM-cost knobs (`llm_max_qpm`, `llm_max_concurrent`). Defaults preserve current behavior; opting out is a single field flip.

**Agent JSON**:
```json
{
  "running": {
    "auto_title_config": {
      "enabled": false
    }
  }
}
```

**Wiring in `generate_and_update_title`**: now reads `load_agent_config(workspace.agent_id).running.auto_title_config` (same lookup pattern as `light_context_manager.py:574`), returns early before any LLM call when `enabled=False`, and uses `timeout_seconds` in place of the previous hard-coded constant. An unavailable agent config (`ValueError`/`AppBaseException`) is swallowed with a debug log so a misconfigured workspace can never break a chat request.

**Tests**: autouse fixture stubs `load_agent_config` so unit tests don't need a real workspace on disk; new cases for the `enabled=False` gate, the agent-config-unavailable path, and the config-driven timeout (the previous `monkeypatch` of `TITLE_TIMEOUT_SECONDS` is gone since the constant is gone).

## Console UI — `75ed0ab`

**Where it lives**: Agent → Config page → "ReAct Agent" card, directly under the existing *Auto-continue on text-only step* toggle. One switch, "Auto-generate session titles", with a tooltip explaining the per-chat LLM cost trade-off.

**Wiring**:
- `console/src/api/types/agent.ts` — new `AutoTitleConfig` interface, field added to `AgentsRunningConfig` (mirrors the backend pydantic shape).
- `console/src/pages/Agent/Config/useAgentConfig.tsx` — `auto_title_config` populated into `form.setFieldsValue` with a `{ enabled: true, timeout_seconds: 30.0 }` fallback so agents persisted before the field existed still render correctly. The form's existing `validateFields()` → `updateAgentRunningConfig(values)` round-trip carries it back unchanged, so no API-client changes were needed.
- `console/src/pages/Agent/Config/components/ReactAgentCard.tsx` — single `Form.Item` with `name={["auto_title_config", "enabled"]}` + `<Switch />`, matching the nested-tuple `name=` pattern `LightContextCard` already uses for nested config groups.
- `console/src/locales/{en,zh,ja,ru}.json` — `autoGenerateSessionTitle` + `autoGenerateSessionTitleTooltip` keys for all four locales.

**Deliberately not surfaced**: `timeout_seconds`. It's threaded through the type and the form's hidden state (so it round-trips on save), but I didn't add a visible input — it'd clutter the card for what is an advanced/seldom-tuned knob, and `LightContextCard` already sets the precedent of exposing only the most-tuned fields. Trivial follow-up if you'd rather show it.

xieyxclack:
@ekzhu Hi, I have tested locally with `dashscope/qwen3-max`. The title generation task runs but never updates the chat name — it always produces an empty title.

It seems that`model(messages)` returns an `async_generator` (streaming response), but `generate_and_update_title` passes it directly to `_extract_text_from_response()`, which only handles non-streaming response objects. An `async_generator` matches none of those branches and falls through to return `""`, so every title is silently discarded.

**One Possible Fix:** Add `_extract_text_from_chunk()` and handle both response shapes in `generate_and_update_title`:

```python
if hasattr(response, "__aiter__"):
    accumulated = ""
    async for chunk in response:
        text = _extract_text_from_chunk(chunk)
        if text:
            accumulated = text
    raw_title = accumulated
else:
    raw_title = _extract_text_from_response(response)
```

After applying this fix locally, title generation works as expected. Could you verify on your end? The streaming behavior may vary by provider/model configuration. I am not sure the above modification is the most suitable way

ekzhu:
> @ekzhu Hi, I have tested locally with `dashscope/qwen3-max`. The title generation task runs but never updates the chat name — it always produces an empty title.
> 
> 
> 
> It seems that`model(messages)` returns an `async_generator` (streaming response), but `generate_and_update_title` passes it directly to `_extract_text_from_response()`, which only handles non-streaming response objects. An `async_generator` matches none of those branches and falls through to return `""`, so every title is silently discarded.
> 
> 
> 
> **One Possible Fix:** Add `_extract_text_from_chunk()` and handle both response shapes in `generate_and_update_title`:
> 
> 
> 
> ```python
> 
> if hasattr(response, "__aiter__"):
> 
>     accumulated = ""
> 
>     async for chunk in response:
> 
>         text = _extract_text_from_chunk(chunk)
> 
>         if text:
> 
>             accumulated = text
> 
>     raw_title = accumulated
> 
> else:
> 
>     raw_title = _extract_text_from_response(response)
> 
> ```
> 
> 
> 
> After applying this fix locally, title generation works as expected. Could you verify on your end? The streaming behavior may vary by provider/model configuration. I am not sure the above modification is the most suitable way
> 
> 

I have applied the fix. 

I think it's a best-effort thing we can do for now to address provider differences as failure to generate title won't affect other parts of the system.

xieyxclack (approved):
@ekzhu LGTM, thank you for your contribution!

PR #3895, opened 2026-04-28
fix(context): prevent context loss when tool result exceeds reserve limit

### LINKED ISSUE #3893: [Bug]: Context Sync Race Condition - Tool Result Dropped Before Next LLM Call Causes Infinite Loop
---
name: Bug Report
about: Report a bug or unexpected behavior
title: "[Bug]: Context Sync Race Condition - Tool Result Dropped Before Next LLM Call Causes Infinite Loop"
labels: ["bug", "triage"]
assignees: []
---

## QwenPaw Version

`qwenpaw 1.1.4.post2`

## Description

Under high `max_iters` (e.g., 200) with **complex tasks returning massive tool outputs** (e.g., GitHub MCP reviewing entire source code), the framework exhibits a critical **Context Sync Race Condition**:
- **Tool Result** is successfully written to the Session JSON file.
- **However**, the result is **NOT injected** into the next LLM request prompt.
- The LLM sees a "cleaned" `[System, User]` prompt on each iteration, causing it to **re-decide** the exact same tool call in an infinite loop.

### 🚨 Extreme Scenario Discovery
This bug was originally discovered during a **heavy code analysis task** using GitHub MCP tools:
- **Context:** `max_iters: 200`.
- **Task:** Reviewing/Scanning large repository source code.
- **Behavior:** Every tool call returned **extremely long message blocks**.
- **Impact:** Due to the large payloads, the context token count frequently exceeded the `reserve_threshold`. Because of the logic error in `as_msg_handler.py` (detailed below), the system repeatedly discarded the entire context history to "save budget," causing the agent to enter an infinite retry loop immediately.

**Evidence**: Session JSON (`*.json`) contains **18+ consecutive identical `thinking` blocks** from the `assistant`, proving the prompt was reset each time rather than accumulated.

## Component(s) Affected

  - Specifically: `src/qwenpaw/agents/context/as_msg_handler.py` (Context compaction logic)

## Environment

- **QwenPaw version:** `1.1.4.post2`
- **OS:** Windows 11 (AMD64)
- **Install method:** `pip install qwenpaw`
- **Python version:** 3.12
- **Channel:** Feishu
- **`max_iters`:** **200** (High iteration context required to trigger)

## Steps to Reproduce

1. Configure `max_iters: 200` in `config.json`.
2. Trigger a complex task involving GitHub MCP tools (e.g., "Review the entire source code of repo X").
3. Ensure tool results are large (high token count).
4. Observe the agent repeating the same tool call indefinitely.

## Actual vs Expected

- **Actual:** 
  - Tool executes successfully, result written to disk.
  - Next LLM iteration receives **only** `[System Prompt, User Query]` (Tool Result missing due to aggressive context compaction logic).
  - Agent "re-discovers" the same action, generating identical `thinking` + `tool_use` blocks.
  - Session JSON shows **36 consecutive identical assistant messages**.

- **Expected:**
  - Tool Result should be appended to context: `[System, User, ToolUse, ToolResult]`.
  - Agent should see the result and proceed to the next logical step.

## Root Cause Analysis (Code Level)

In `src/qwenpaw/agents/context/as_msg_handler.py`, the `context_check` loop checks the **Token Limit** *before* validating **Tool ID Alignment**:

```python
for keep_count in range(1, len(msgs) + 1):
    # ... calculate keep_tokens ...
    
    # 1. FATAL: Check limit FIRST
    if keep_tokens > context_compact_reserve:
        break  # <--- Breaks immediately if tool result is large

    # 2. Validate alignment SECOND
    if self.validate_tool_ids_alignment(messages[-keep_count:]):
        best_keep_count = keep_count
```

**The Bug:** If a large tool result pushes the slice just over the `reserve` limit, the loop breaks *before* it can register that slice as a valid context to keep (because validation hasn't run yet). `best_keep_count` remains 0, and context is wiped.

### 1. Identical `thinking` blocks (Session JSON dump)
```json
// 18 rounds of EXACTLY the same thinking content
{"thinking": "检查工具输出...生成 SVG...\n调用 garden_gen_v190.py..."}
```

## Workaround

- Reduce `max_iters` to `30`.
- Manually call `/stop` when detecting repetitive tool calls.

CA-mambo:
## Description

This PR fixes a critical context management bug where the agent enters an infinite loop because it repeatedly loses conversation context after a tool call.

### Root Cause
In `src/qwenpaw/agents/context/as_msg_handler.py`, the `context_check` method validates conversation history slices to determine how many recent messages to keep (`messages_to_keep`). 

The original logic checks the token limit **before** validating tool ID alignment:

```python
            if keep_tokens > context_compact_reserve:
                break # <--- FATAL: Breaks loop before validating alignment

            if self.validate_tool_ids_alignment(messages[-keep_count:]):
                best_keep_count = keep_count
```

**The Bug:**
If the token count of a message slice exceeds `context_compact_reserve` (e.g., when a tool result is large), the loop breaks immediately. If the alignment hasn't been validated yet for this slice size (e.g., `keep_count=1` fails validation, `keep_count=2` exceeds limit), `best_keep_count` remains 0. This results in `messages_to_keep` being an empty list `[]`.

The agent receives an empty context (reset to just System Prompt), causing it to "re-discover" the user's last intent and call the same tool again, creating an infinite loop.

### The Fix
Swap the order of operations: Validate alignment first, then check the token limit.

```python
            if self.validate_tool_ids_alignment(messages[-keep_count:]):
                best_keep_count = keep_count

            if keep_tokens > context_compact_reserve:
                break
```

This ensures that if a slice is valid (e.g., a complete Tool Use + Tool Result pair), it is captured in `best_keep_count` before the token check breaks the loop. Even if the pair slightly exceeds the reserve limit, we preserve the context instead of clearing it.

#### Local CI Verification
- **pre-commit (target file):** ✅ All checks passed
  - mypy: Passed
  - black: Passed
  - flake8: Passed
  - pylint: Passed
- **pytest (context module):** ✅ 5/5 tests passed
  - `test_tool_use_result_pair_aligned_when_both_fit`
  - `test_tool_use_result_pair_excluded_when_result_too_large`
  - `test_partial_tool_result_kept_with_tool_use`
  - `test_multiple_tool_pairs_in_same_message`
  - `test_reverse_dependency_tool_use_needs_tool_result`

#### Local Simulation
Created `reproduce_sync_race.py` to simulate the race condition:
- **Original Logic:** Returns 0 kept messages (Context Lost)
- **Fixed Logic:** Returns 2 kept messages (Context Saved)

See Issue #3893 for detailed logs and evidence of the context sync race condition.

Fixes #3893

### DIFF
--- a/src/qwenpaw/agents/context/as_msg_handler.py
+++ b/src/qwenpaw/agents/context/as_msg_handler.py
@@ -340,14 +340,14 @@ class AsMsgHandler:
                 stat.total_tokens for stat in msg_stats[-keep_count:]
             )
 
-            if keep_tokens > context_compact_reserve:
-                # Exceeds reserve limit, stop expanding
-                break
-
             if self.validate_tool_ids_alignment(messages[-keep_count:]):
                 # Valid slice, update best (keeps maximum valid)
                 best_keep_count = keep_count
                 best_keep_tokens = keep_tokens
+
+            if keep_tokens > context_compact_reserve:
+                # Exceeds reserve limit, stop expanding
+                break
 
         # Build final split based on best_keep_count
         messages_to_keep = (


jinliyl:
@CA-mambo 
Thank you for your feedback. With this kind of modification, the condition `keep_tokens <= context_compact_reserve` would no longer be satisfied. If `keep_tokens` significantly exceeds this limit, it could cause new issues.

In your case, `messages_to_keep` became an empty list (`[]`). However, this shouldn't be addressed by modifying the code—rather, your `context_compact_reserved` setting is too small. Normally, the full context length should be set to 200K tokens（1M tokens）, with a 10% reservation（or higher）, meaning `context_compact_reserve` should be around 20K （100K）tokens. 

Please double-check your context configuration, especially the percentage settings for `compact_threshold` and `reserve_threshold`. If possible, feel free to reply to me in this thread.

CA-mambo:
### Context

We understand the concern regarding `keep_tokens > context_compact_reserve` and the potential for soft budget violations.

### Core Trade-off

The original logic prioritizes the **soft budget limit** over **context atomicity**.
- When a valid `Tool Use` + `Tool Result` pair slightly exceeds the `reserve` threshold, the original loop breaks *before* validation, resulting in `best_keep_count = 0`.
- This causes the agent to lose the entire context of the tool output, leading to immediate **amnesia and infinite retry loops**.

### Our Fix

We prioritize **context atomicity**. We ensure valid tool pairs are captured in `best_keep_count` even if they slightly stress the budget.
- **Trade-off**: We accept a potential 'soft budget violation' (slightly larger history) to ensure the Agent retains the context it needs to function.
- **Safety**: The system still has hard limits (`max_input_length`) and pruning mechanisms to prevent fatal overflows.

### Next Steps

We are **not in a rush to merge** this PR. We will run this patch locally in high-intensity scenarios for a few days to verify stability and ensure no new issues arise. We will update the thread with our findings.

Thank you for the feedback.

CA-mambo:
Closing this PR after extensive analysis. 

Thank you @xieyxclack and @jinliyl for the detailed feedback and deep dive!

Your review highlighted a crucial blind spot: while this "Swap Order" fix is harmless and addresses the **"CoT Amnesia"** scenario (where `best > 0` but reasoning context is lost), it does not solve the **"Giant Tool Result"** deadlock (where `best = 0`). In that specific case, the Fallback mechanism is indeed required.

Since this fix is only half of the solution and involves architectural changes to the compression logic, we are closing this PR to avoid partial merging. We will continue to focus our discussion and stress-test the comprehensive "Swap + Fallback" workaround in Issue #3893.

For those interested, we've published our full reproduction scripts, simulations, and root cause analysis here: [CA-mambo/qwenpaw-issue-3893-research](https://github.com/CA-mambo/qwenpaw-issue-3893-research).

PR #3928, opened 2026-04-28
fix(agent-tools): add safe default timeout for delegate_external_agent

q1023884985:
## Problem - The `max_runtime` parameter defaults to `None`, causing **infinite waits** when LLMs omit it from tool calls - Tool docstring marks it as 'Optional', which leads LLMs to skip providing the parameter - No safety mechanism exists for slow or unresponsive ACP agents, causing sessions to hang indefinitely  ## Root Cause Analysis When external agents call `delegate_external_agent(action="start", runner="qwen_code", message="...")` without specifying `max_runtime`: 1. Parameter resolves to `None` 2. Timeout logic is skipped (deadline=None) 3. ACP prompt waits forever if the external agent doesn't respond  Users reported sessions getting stuck with no way to recover except manual intervention.  ## Solution This PR implements three key changes:  ### 1. Safe Default Value ```python # Before max_runtime: Optional[float] = None  # ❌ No timeout protection  # After   max_runtime: float = 60.0  # ✅ Default 60-second safety net ```  ### 2. Clear Documentation Updated docstring removes 'Optional' ambiguity and emphasizes: - Purpose: prevents infinite hangs - Default value: 60.0 seconds - Recommendation: always set reasonable timeout for complex tasks  ### 3. Simplified Validation Removed unnecessary `None` handling code paths since the parameter now always has a valid numeric default.  ## Impact ✅ **Backward Compatible**: Existing explicit `max_runtime` values continue unchanged   ✅ **Immediate Safety**: All new calls get 60-second protection by default   ✅ **Flexible**: Complex tasks can override with higher values (e.g., 180.0, 300.0)   ✅ **No Breaking Changes**: Only affects behavior when parameter was omitted  ## Testing Recommendations - Test simple queries (should complete within 60s or timeout gracefully) - Test complex tasks with custom higher timeouts - Verify timeout messages appear correctly after interruption - Ensure ACP session remains open for continuation  ## Related This fix addresses the issue where users experienced frozen sessions when delegating to external agents without understanding the importance of timeout parameters.

### DIFF
--- a/src/qwenpaw/agents/tools/delegate_external_agent.py
+++ b/src/qwenpaw/agents/tools/delegate_external_agent.py
@@ -433,7 +433,7 @@ async def delegate_external_agent(
     runner: str,
     message: str = "",
     cwd: str = "",
-    max_runtime: Optional[float] = None,
+    max_runtime: float = 60.0,  # Default timeout to prevent infinite hangs
 ) -> ToolResponse | AsyncGenerator[ToolResponse, None]:
     # pylint: disable=too-many-return-statements
     """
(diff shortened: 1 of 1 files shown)


x1n95c:
Hi, thank you for your contribution! I've tested your changes and everything works as expected. Could you please run `pre-commit run --all-files` locally to fix the pre-commit check failures before we proceed with the review?

zhijianma (approved):

PR #3959, opened 2026-04-29
fix(console): keep Chat mounted when navigating to other pages

zhenai1314521:
## Problem  When navigating away from chat page (e.g. to /cron-jobs for scheduled tasks, or /sessions, /settings etc.) and navigating back, the Chat component is completely unmounted and remounted, causing:  1. **Active task lost** - any running agent task is cancelled 2. **Chat session lost** - all messages disappear 3. **User has to start over** - no way to recover the previous conversation  ## Root Cause  MainLayout uses React Router `<Routes>` to render different pages:  ```tsx <Routes>   <Route path=/chat/* element={<Chat />} />   <Route path=/cron-jobs element={<CronJobsPage />} />   ... </Routes> ```  When navigating to `/cron-jobs`, React Router unmounts `<Chat />` entirely. When navigating back, a completely new `<Chat />` instance is mounted, losing all internal state (messages, session selection, running tasks).  ## Solution  Keep the Chat component **always mounted** but hidden via CSS when not active:  ```tsx // Chat always mounted, hidden via display:none when inactive const isChatActive = currentPath === / || currentPath.startsWith(/chat);  <div style={{ display: isChatActive ? contents : none }}>   <Chat /> </div>  // Other pages only rendered when chat is not active {!isChatActive && (   <Routes>     <Route path=/cron-jobs element={<CronJobsPage />} />     ...   </Routes> )} ```  This preserves all Chat internal state (messages, session, running tasks) when navigating to other pages and back.  ## Changes  - `console/src/layouts/MainLayout/index.tsx`   - Always render `<Chat />` wrapped in a display toggle div   - Conditionally render other page `<Routes>` only when not on chat   - Uses `display: contents` when active (no extra wrapper in DOM), `display: none` when inactive

### DIFF
--- a/console/src/layouts/MainLayout/index.tsx
+++ b/console/src/layouts/MainLayout/index.tsx
@@ -82,6 +82,10 @@ export default function MainLayout() {
       : "chat";
   }
 
+  // Keep Chat component always mounted to preserve session/task state.
+  // Use CSS display to show/hide instead of unmounting via Routes.
+  const isChatActive = currentPath === "/" || currentPath.startsWith("/chat");
+
   return (
     <Layout className={styles.mainLayout}>
       <Header />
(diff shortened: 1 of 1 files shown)


zhijianma:
@zhenai1314521 


https://github.com/agentscope-ai/QwenPaw/issues/3919#issuecomment-4349226267

PR #4084, opened 2026-05-07
fix(crons): eliminate concurrency state leaks in CronManager 

celestialhorse51D:
## Description

Fix multiple concurrency state-leak bugs in `CronManager` (`src/qwenpaw/app/crons/manager.py`) so that scheduled-task state can no longer bleed across parallel runs, deleted jobs, or start/stop cycles.

Symptoms before this PR:

- A failing run could leave `last_error` on a `CronJobState` that another concurrent run later marked `last_status="success"`, producing an inconsistent snapshot (`success` + non-empty `last_error`).
- `delete_job` did not cancel in-flight executions; the still-running `_execute_once` would then re-create `_Runtime` and re-publish state, resurrecting a deleted job in `_states` / `_rt`.
- `_register_or_update` unconditionally replaced the per-job semaphore. In-flight tasks released permits on the *old* sem while new tasks acquired the *new* one, silently doubling the effective `max_concurrency` until in-flight tasks finished.
- `run_job` and `_task_done_cb` created fire-and-forget tasks (via `asyncio.create_task` / `asyncio.ensure_future`) without keeping a reference, so `stop()` could not cancel them and the runtime emitted "Task was destroyed while pending" warnings on shutdown.
- `stop()` only shut down APScheduler and never cleared `_states` / `_rt`, so the next `start()` cycle inherited stale `last_status` / `last_error` from the previous session.
- `_scheduled_callback` let exceptions from `_execute_once` propagate to APScheduler, which both double-reported the error and skipped the `next_run_at` refresh.

This PR makes state writes execution-local and lifecycle-aware via a new `_publish_state()` helper, introduces task-tracking sets (`_run_tasks`, `_job_tasks`) with `_track_job_task` / `_untrack_job_task` helpers, reuses the existing semaphore on re-registration, and clears all per-run state on `stop()`.

**Security Considerations:** N/A — no auth, env, or config-handling surface is touched. The fix only affects in-process bookkeeping inside `CronManager`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Documentation updated (if needed) — N/A, internal fix with no public-API change
- [x] Ready for review

## Testing

[How to test these changes]

```bash
pre-commit run --all-files
  ********************************Passed******************************** 

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]

### DIFF
--- a/src/qwenpaw/app/crons/manager.py
+++ b/src/qwenpaw/app/crons/manager.py
@@ -63,6 +63,13 @@ class CronManager:
         self._states: Dict[str, CronJobState] = {}
         self._history: Dict[str, list[CronExecutionRecord]] = {}
         self._rt: Dict[str, _Runtime] = {}
+        # Track fire-and-forget background tasks so we can cancel them
+        # during stop() and avoid the "Task was destroyed while pending"
+        # warning caused by losing the task reference.
+        self._run_tasks: set[asyncio.Task] = set()
+        # Per-job set of running tasks, used by delete_job to cancel
+        # in-flight executions and prevent state resurrection.
+        self._job_tasks: Dict[str, set[asyncio.Task]] = {}
         self._started = False
 
     async def start(self) -> None:
@@ -152,6 +159,27 @@ class CronManager:
             self._scheduler.shutdown(wait=False)
             self._started = False
 
+            # Cancel all in-flight fire-and-forget tasks to prevent them
+            # from writing into the state dicts after we have cleared them.
+            pending = [t for t in self._run_tasks if not t.done()]
+            for task in pending:
+                task.cancel()
+
+            # Wait for cancellations to settle outside the lock so that the
+            # done callbacks (which remove tasks from the tracking set) can
+            # run without deadlocking on our lock.
+        if pending:
+            await asyncio.gather(*pending, return_exceptions=True)
+
+        async with self._lock:
+            # Clear all per-run state to avoid leaking it to the next
+            # start() cycle (e.g. stale last_status / last_error).
+            self._states.clear()
+            self._history.clear()
+            self._rt.clear()
+            self._run_tasks.clear()
+            self._job_tasks.clear()
+
     # ----- read/state -----
 
     async def list_jobs(self) -> list[CronJobSpec]:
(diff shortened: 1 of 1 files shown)


RerankerGuo (changes requested):
The semaphore-reuse branch makes `max_concurrency` updates permanently ineffective.

I validated:

- exact head `b57c105409e0aa9c6f0ba1ce45875725f1d0a490`;
- current main `bacac7410cf2ac122dd892152beff8b03f38729f`;
- a minimal current-main transplant of only this PR's `_register_or_update()` semaphore policy: commit `ef0e33b65ac0bc252a668db260f95adc351e4b79`, tree `02edf9c7f3aaac86067c2a65ed22d19ff3c473bc`.

The production-method contract registers an idle job at `max_concurrency=1`, updates the same job to `2`, acquires one slot, and attempts a second acquisition with a bounded timeout.

Results:

```text
current main:
same_semaphore=False
available_permits_after_update=2
second_slot_available=True

exact and current-main transplant:
same_semaphore=True
available_permits_after_update=1
second_slot_available=False
```

Exact/trial also log that the change "will apply after current runs settle", but there is no code that replaces or resizes the semaphore after it settles. The old object is retained indefinitely. In addition, `Semaphore._value` is the currently available permit count, not the configured capacity, so it cannot reliably determine whether the configured limit changed while tasks are active.

This means a normal idle `1 -> 2` configuration update is silently ignored. The inverse direction also needs care: replacing immediately while old holders run can temporarily overcommit, but retaining the old semaphore forever does not implement the requested new limit.

Please track the configured capacity and either:

1. replace the runtime immediately when the old semaphore has no holders/waiters, while deferring replacement until drain otherwise; or
2. use an explicit adjustable limiter with a tested generation/drain handoff.

Regression coverage should include idle `1 -> 2`, active `1 -> 2`, and active `2 -> 1`, proving both that capacity eventually changes and that updates never temporarily exceed the intended limit.

Additional verification: the current-main transplant passes 32 CronManager/CLI tests plus Mypy, Black, Flake8, Pylint, and `git diff --check`; exact passes `py_compile`, Black, and Flake8. Seven HTTP integration cases did not execute because their shared app fixture exceeded its 60-second readiness window (the log shows readiness at 55.161 seconds after long health-request timeouts), so I am not treating those setup errors as a code finding.

PR #4120, opened 2026-05-08
feat(console): enhance Matrix E2EE verify step and prevent "not encrypted / device not verify by owner" icon

Morxi:
## Description

Extends **Matrix channel configuration** in the Console and makes the **Matrix backend channel** support more flexible sign-in flows and end-to-end encryption (E2EE) behavior.

**Console (`ChannelDrawer.tsx`)**

- **Access token** is now **optional**; it can be used together with or instead of **Matrix username / password** sign-in.
- New required **Device Name** field (stable device identity / display).
- New **Enable end-to-end encryption (E2EE)** switch; after enabling E2EE, the device must be verified in a Matrix client (as noted in the UI copy).
- Optional **Username** (leave blank to fall back to configured `user_id`) and optional **Password** for password login.

**Backend (`matrix/channel.py`)**

- Config: `device_id` and `username` defaults aligned with `user_id`; explicit **device id** resolution (`device_id` when set, otherwise derived from `device_name`).
- **Auth strategy**: password login first; then access token; **cached auth state** is restored only when appropriate so explicit credentials or a pinned `user_id` do not pick up stale tokens.
- **`matrix-nio` version differences**: `login()` kwargs are built from **`inspect.signature` at runtime** with fallback call shapes to avoid incompatible-argument crashes.
- **Token login**: if `whoami` **does not match** configured `user_id`, refuse (avoid mixing accounts); on success persist **`matrix_auth_state.json`** (`access_token` / `user_id` / `device_id`).
- **E2EE**: registers to-device handlers for **SAS device verification**, room key requests, and related crypto flows when E2EE is enabled.
- **Outbound sends**: if the client’s local room is not marked encrypted but E2EE is globally on, **`room_get_state_event` fetches `m.room.encryption`** before sending.

**Tests**

- `tests/unit/channels/test_matrix.py`: adds `test_send_syncs_encryption_from_room_state`, covering “fetch room encryption state when local room encryption flag was missing.”

---

**Security Considerations:**

- Matrix **access tokens** may be persisted to **`matrix_auth_state.json`**; ensure filesystem permissions and deployment practices protect that path.
- For **username/password** flows, never log passwords in full; config, UI transport, and storage should follow your existing secrets handling.
- **Matching `whoami` to configured `user_id`** reduces accidental cross-account use of leaked tokens; still pair with normal token rotation and revocation.

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

---

## Component(s) Affected

- [] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

---

## Checklist

- [x]  If pre-commit auto-fixed files, I committed those changes and reran checks
- [x]  Ready for review

- [x]  I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes

---

## Testing

1. **Console**: Open the Matrix channel drawer and confirm saving works with **token omitted** using **username + password + Device Name** (respect any extra form validation).
2. **Token only**: With access token configured, verify behavior when **`user_id` does not match the token owner** (error / logs as implemented).
3. **E2EE**: Enable encryption and complete **SAS verification** with Element (or similar); confirm messaging works afterward.
4. **Automation**: `pytest tests/unit/channels/test_matrix.py -k send_syncs_encryption` (or full `pytest`).

---


Morxi:
Device Verify:
image
Screenshot from 2026-05-08 17-51-25

Screenshot from 2026-05-08 17-51-38
No More 'unencrypted':
image

xieyxclack:

xieyxclack:
@Morxi Please format the code, refer to https://github.com/agentscope-ai/QwenPaw/blob/main/CONTRIBUTING.md#4-code-and-quality

Morxi:
@xieyxclack  commit fixed

src/qwenpaw/app/channels/matrix/channel.py
```diff
@@ -366,14 +392,47 @@ async def health_check(self) -> Dict[str, Any]:
...
+                restore_identity=False,
+            )
+        elif self._cfg.user_id:
+            # User is switching/targeting a specific MXID but did not provide
+            # explicit credentials in this config: don't silently use stale
+            # cached token from another account.
+            self._load_auth_state(
+                restore_token=False,
+                restore_identity=False,
+            )
```
hongxicheng: The first three branches (`has_password_creds`, `has_token_cred`, `self._cfg.user_id`) all call `_load_auth_state(restore_token=False, restore_identity=False)` with identical arguments. This can be simplified to:
```suggestion
        has_explicit_identity = has_password_creds or has_token_cred or self._cfg.user_id
        self._load_auth_state(
            restore_token=False,
            restore_identity=not has_explicit_identity,
        )
```
Also, restore_token is always False across all call sites, which means the token-restore path in _load_auth_state is effectively dead code. Is this intentional for future use, or should it be removed?

src/qwenpaw/app/channels/matrix/channel.py
```diff
@@ -710,29 +1521,47 @@ async def _sync_loop(self) -> None:
...
-                        "MatrixChannel: group msg blocked from %s",
+                    logger.warning(
+                        "MatrixChannel: dropping group message "
+                        "(allowlist miss) sender=%s room=%s normalized=%s",
                         sender_id,
+                        room_id,
+                        normalized,
                     )
                     return False
         return True
```
hongxicheng, line 1605: Policy-filtered messages (e.g. "dropping DM message (allowlist miss)") were changed from logger.debug to logger.warning. Since policy filtering is expected behavior (not an anomaly), logger.info would be more appropriate — it keeps the messages observable for debugging "why didn't the bot respond" without flooding production logs with warnings in high-traffic rooms.

console/src/pages/Control/Channels/components/ChannelDrawer.tsx
```diff
@@ -194,11 +194,39 @@ export function ChannelDrawer({
...
+            <Form.Item
+              name="device_name"
+              label="Device Name"
+              rules={[{ required: true }]}
+            >
+              <Input placeholder="device_id" />
+            </Form.Item>
+            <Form.Item
+              name="encryption"
+              label="Enable End-to-End Encryption (You need verfiy the device after enable E2EE)"
```
hongxicheng:
```suggestion
              label="Enable End-to-End Encryption (You need verify the device after enable E2EE)"
```

console/src/pages/Control/Channels/components/ChannelDrawer.tsx
hongxicheng, line 1: The current UI puts `Access Token`, `Username`, and `Password` fields side by side without any visual grouping or conditional display. Since token login and password login are two mutually exclusive authentication strategies, consider using a segmented control / radio group (e.g. "Login Method: Token | Password") to toggle between the two sets of fields. This would:

1. Make the two login flows clearly distinguishable for the user
2. Avoid confusion about which fields need to be filled
3. Prevent invalid combinations (e.g. filling both token and password)

Right now a user seeing all fields at once may not understand they only need to fill one group.

hongxicheng (commented):
Thanks for your contribution @Morxi! This is a solid effort on improving the Matrix E2EE flow. Below are a few suggestions before we merge:

tests/unit/channels/test_matrix.py
hongxicheng, line 1: Please remove the newly added test.

Morxi:
@hongxicheng Tested and fixed issues from the previous review.

E2EE remains incompatible with token-only auth; the console now enforces the correct field set per auth method.

E2EE still requires users to install matrix-nio[e2e] when guided by errors; 
that path depends on deprecated libolm, Maybe some distributeOS not support that
A pending upstream Rust replacement PR https://github.com/matrix-nio/matrix-nio/pull/555 
The feature is nonetheless usable.

Doc changes are out of scope here and will be a follow-up: user-facing guidance on Matrix E2EE setup and concepts, plus rollback and reset for accounts with Encrypted enabled. A separate PR will cover Matrix E2EE support status and environment setup.

hongxicheng:
After testing this PR locally, I found two issues:

1. Missing graceful degradation when matrix-nio[e2e] is not installed
Since matrix-nio[e2e] is not included in our project dependencies (due to its reliance on the deprecated libolm), enabling the E2EE toggle without the extra package causes an ImportWarning that crashes the entire Matrix channel startup.

Suggested fix:

- Console UI: Add a hint on the E2EE toggle indicating that this feature requires the user to manually install matrix-nio[e2e].
- Backend: When the E2EE dependency is missing at runtime, catch the error, log a clear message guiding the user to install pip install matrix-nio[e2e], and gracefully fall back to non-encrypted mode — rather than failing the whole channel.
2. user_id and username are effectively the same field
In MatrixChannelConfig:

```Python
self.username: str = raw.get("username") or self.user_id or ""
```

And in start():

```Python
login_user = (self._cfg.username or self._cfg.user_id or "").strip()
```
When username is empty it falls back to user_id, making them functionally identical in the password login path. Having both fields in the form is confusing for users.

Suggestion: Merge them into a single "User ID" field (accepting either a localpart like bot or a full MXID like @bot:server.com). Then place Access Token and Password at the same form position, conditionally displayed based on the selected auth method. This simplifies the UX to: one identity field + one credential field, toggled by auth method.

Morxi:
@hongxicheng 
Thanks for the review.
E2EE / olm: preflight olm, log install hints, disable E2EE and , and continue without crashing. Tested locally without olm.
user_id / username:  backend only uses user_id now. I’ll simplify the form to one identity field + token/password by auth method in a follow-up.
add a one-line UI hint on the E2EE toggle about pip install matrix-nio[e2e].

hongxicheng:
@Morxi I've pushed a commit to simplify the console form UX for the Matrix channel:

- Merged user_id and username into a single "User ID" field
- Reorganized credential inputs (Token / Password) by auth method
- E2EE toggle only shown under Password auth; label simplified with tooltip
- Fixed value persistence when switching between auth methods

I've tested locally and it looks good. Could you please review the changes and run through your test scenarios as well before we merge? Thanks!

Morxi:
@hongxicheng  I've tested the latest commit locally, everything works as expected. Thanks a lot.

hongxicheng (approved):
LGTM

xieyxclack (approved):

xieyxclack:
@Morxi @hongxicheng Thank you for your contribution

PR #4224, opened 2026-05-11
fix(memory): refresh index after auto memory summary

### LINKED ISSUE #4220: [Bug]: auto_memory_interval 写入记忆文件但不同步向量索引，/new 后新会话 memory_search 搜不到
## [Bug]: auto_memory_interval 写入记忆文件但不同步向量索引，/new 后新会话 memory_search 搜不到

**QwenPaw 版本：** v1.1.6

### 描述

开启 `auto_memory_interval: 3` 后，auto-memory 后台任务会在每 3 条用户消息后调用 Summarizer 将对话摘要写入 `memory/YYYY-MM-DD.md`。文件确实写入了，但 `memory_search` 在新会话中始终返回空结果——**向量索引没有被同步更新**。

**Expected：** auto_memory 写入后，新会话的 `memory_search` 能立即搜到刚记录的记忆。

**Actual：** 文件存在于磁盘，但 `memory_search` 返回空，需等到当晚 23:00 Memory Dream 运行后才能搜到。

### 复现步骤

1. 在 `agent.json` 中配置 `running.reme_light_memory_config.auto_memory_interval: 3`
2. 正常对话，发送至少 3 条用户消息，触发 auto_memory
3. 使用 `/new` 启动新会话
4. 在新会话中调用 `memory_search` 搜索刚聊过的内容
5. **结果：** 返回空结果

### 日志证据

```
# auto_memory 任务确实完成了
2026-05-11 21:49:44 | INFO | Summary task task_44 started
2026-05-11 21:50:30 | INFO | Summary task task_44 completed
# 后续 task_45~50 均完成
```

### 根因分析

auto_memory 调用链：
```
post_reply_hook → add_summarize_task → ReMeLight.summary_memory()
  → Summarizer (ReActAgent)
    → write_file/edit_file 写入 memory/2026-05-11.md ✅（数据落盘）
```

`Summarizer` 使用的是独立的 `FileIO.write_file` 工具直接写磁盘文件，但写完后没有调用 `LocalFileStore.upsert_file()` 来更新内存中的向量索引（`_chunks` 字典）。`LocalFileStore.upsert_file` 负责生成 chunks embedding 并注册到索引中，而 `Summarizer` 完全跳过了这一步。

对比之下，`ReMeLightMemoryManager.dream()`（每晚 23:00 的 Memory Dream）走的是另一套 ReActAgent prompt，同样写磁盘，但跨过了 23:00 的时间窗口后，索引自然不过期。而 `auto_memory_interval` 在会话生命周期内写入的内容，索引不会随之更新。

### 建议修复方向

在 `Summarizer` 完成文件写入后，显式调用 `LocalFileStore.upsert_file()` 更新向量索引，而非依赖文件监听器被动同步。

具体位置：`reme/memory/file_based/components/summarizer.py` 中 `Summarizer.execute()` 的末尾，在 Remy ReActAgent 完成文件写入后，读取目标文件内容，调用 `file_store.upsert_file()` 更新 chunks。

aqilaziz:
﻿## Summary
- bump `reme-ai` from `0.3.1.8` to `0.3.1.9`
- update QwenPaw's expected ReMe version guard to match the new release
- drop the previous QwenPaw-side explicit index refresh workaround in favor of the upstream watcher lifecycle fix

## Related Issue
Fixes #4220
Refs agentscope-ai/ReMe#233

## Root Cause / Resolution
Maintainer feedback on this PR pointed to ReMe's file watcher as the better place to fix the stale memory-index timing gap. I followed up upstream in agentscope-ai/ReMe#233, which found that `BaseFileWatcher.close()` left `_stop_event` set and `start()` reused it. After a close/start cycle, the watcher could appear running while the watch loop exited immediately.

That upstream fix has been merged and released as `reme-ai==0.3.1.9`, so this PR now updates QwenPaw to consume that fix instead of carrying a QwenPaw-specific workaround.

## Tests
- `PYTHONPATH=src python -m pytest tests/unit/agents/memory/test_reme_light_memory_manager.py -q` - 24 passed
- `pre-commit run --files pyproject.toml src/qwenpaw/agents/memory/reme_light_memory_manager.py tests/unit/agents/memory/test_reme_light_memory_manager.py` - passed
- `git diff --check` - passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "playwright>=1.49.0",
     "questionary>=2.1.1",
     "mss>=9.0.0",
-    "reme-ai==0.3.1.8",
+    "reme-ai==0.3.1.9",
     "transformers>=4.30.0",
     "python-dotenv>=1.0.0",
     "python-socks>=2.5.3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "playwright>=1.49.0",
     "questionary>=2.1.1",
     "mss>=9.0.0",
-    "reme-ai==0.3.1.9",
+    "reme-ai==0.3.1.10",
     "transformers>=4.30.0",
     "python-dotenv>=1.0.0",
     "python-socks>=2.5.3",
--- a/src/qwenpaw/agents/memory/reme_light_memory_manager.py
+++ b/src/qwenpaw/agents/memory/reme_light_memory_manager.py
@@ -33,7 +33,7 @@ from ...constant import EnvVarLoader
 logger = logging.getLogger(__name__)
 
 _REME_STORE_VERSION = "v1"
-_EXPECTED_REME_VERSION = "0.3.1.8"
+_EXPECTED_REME_VERSION = "0.3.1.9"
 # Maximum number of tokens from query splitting
 MAX_QUERY_TOKENS = 50
 


aqilaziz:
﻿Pushed a follow-up formatting commit for the pre-commit failure. The failed hook was `black`, which reformatted `src/qwenpaw/agents/memory/reme_light_memory_manager.py`.

Local verification:
```bash
python -m black --check src/qwenpaw/agents/memory/reme_light_memory_manager.py
git diff --check
```

jinliyl:
@aqilaziz 
Thank you very much for your feedback.

Regarding issue #4220, has it been reproduced locally? I haven't been able to reproduce it on my end. My understanding is that it might occur under certain exceptional circumstances that the current code doesn't account for.

For all Markdown files in the backend, a file watcher asynchronously monitors all backend files and updates the index whenever changes are detected. If you'd like to help investigate this issue, please take a look at the relevant code here:  
https://github.com/agentscope-ai/ReMe/blob/main/reme/core/file_watcher/base_file_watcher.py

aqilaziz:
Thanks. I pushed one more commit to fix the remaining pre-commit failure (lack --line-length 79 on 
eme_light_memory_manager.py), so CI has been retriggered on ccd9d6e. I also reran the local unit file with the repo test env (PYTHONPATH=src pytest tests/unit/agents/memory/test_reme_light_memory_manager.py -q), and it passed: 26 passed.\n\nFor issue #4220 specifically: I have not finished a deterministic end-to-end local repro yet, but I was able to confirm the timing gap in the current code path. summary_memory() writes the memory markdown first, while the index refresh depends on the asynchronous file watcher. This PR makes that path explicit: it snapshots memory files before summary_memory(), then after the write completes it calls the default ReMe watcher on_changes(...) for added/modified files so the index is refreshed immediately instead of waiting for the async watcher / later maintenance cycle.\n\nI am continuing to check the watcher behavior against the linked ReMe implementation as well.

jinliyl:
@aqilaziz 
Thank you for the detailed explanation!

I wonder if applying this patch directly in Qwepaw might not be the ideal approach. It could be better to address the underlying issue upstream in ReMe itself.

As you may have noticed, ReMe already includes a file watcher with a monitoring interval under 2 seconds (see: [[base_file_watcher.py](https://github.com/agentscope-ai/ReMe/blob/main/reme/core/file_watcher/base_file_watcher.py)](https://github.com/agentscope-ai/ReMe/blob/main/reme/core/file_watcher/base_file_watcher.py)), which should theoretically meet user requirements without needing a workaround in Qwepaw.  

Would you be open to helping investigate whether there's an issue within ReMe’s watcher implementation that’s causing the observed timing gap? That might lead to a more robust and maintainable fix.

aqilaziz:
Thanks, that makes sense. I agree this QwenPaw-side patch is more of an explicit refresh workaround, and the more maintainable fix should probably live in ReMe if the watcher/index lifecycle is the real source of the gap.

I took another pass through the ReMe watcher path. My current hypothesis is not that `awatch` misses normal writes in steady state, but that there may be a lifecycle/timing gap around `summary_memory()`:

1. `summary_memory()` writes `memory/YYYY-MM-DD.md` through the file tool.
2. The file store/index refresh is handled asynchronously by ReMe's file watcher.
3. In short-lived/background summary flows, QwenPaw can proceed or close before the watcher has processed that write, so the next session may observe the memory file on disk while the persisted index is still stale.

I will investigate this against ReMe directly next. I will try to reproduce with `ReMeLight` + local file store using `rebuild_index_on_start=False`, writing a memory markdown file, then immediately searching/closing/reopening. If that reproduces, I will move the fix upstream, likely as an explicit/awaitable sync path in ReMe's watcher/index layer rather than keeping this QwenPaw workaround. If it does not reproduce, I will update or close this PR accordingly.

jinliyl:
@aqilaziz Thank you very much for your kind words—I really appreciate your feedback!

aqilaziz:
I followed up upstream in ReMe and opened agentscope-ai/ReMe#233.

What I found:
- `BaseFileWatcher.close()` leaves `_stop_event` set, but `start()` reused the same event.
- After a close/start cycle (including dynamic watcher restarts), the watcher could report as running while `_watch_loop()` exits immediately because the old stop signal is still set.
- I also made `BaseFileWatcher.start()` tolerate `file_store=None` during startup scans, matching the existing base watcher tests.

Local verification in ReMe:
```bash
.\.venv\Scripts\python.exe -m pytest tests\test_base_file_watcher.py -q
# 40 passed

.\.venv\Scripts\python.exe -m pre_commit run --files reme\core\file_watcher\base_file_watcher.py tests\test_base_file_watcher.py
# passed
```

The upstream PR is currently waiting for workflow approval on GitHub Actions because it was opened from my fork.

aqilaziz:
Update after the upstream investigation: agentscope-ai/ReMe#233 has been merged and released in `reme-ai==0.3.1.9`.

I force-pushed this PR to remove the QwenPaw-side workaround and keep only the dependency/version bump to consume the upstream watcher lifecycle fix. This should align with the suggested direction of fixing the underlying ReMe watcher behavior instead of manually notifying the watcher from QwenPaw.

Verification on the updated branch:
```bash
PYTHONPATH=src python -m pytest tests/unit/agents/memory/test_reme_light_memory_manager.py -q
# 24 passed

pre-commit run --files pyproject.toml src/qwenpaw/agents/memory/reme_light_memory_manager.py tests/unit/agents/memory/test_reme_light_memory_manager.py
# passed

git diff --check
# passed
```

jinliyl:
@aqilaziz Thank you for the clean dependency bump — the watcher lifecycle fix itself looks good.

However, while reviewing `reme-ai==0.3.1.9` upstream, I found several dependency issues that were introduced (or carried over) in that release. These would affect QwenPaw users once we pin to `0.3.1.9`:

---

### Issue 1: `sqlalchemy` hard dependency — not declared, no lazy import

**Location:** `reme/core/vector_store/obvec_vector_store.py:9-10`

```python
from sqlalchemy import Column, JSON, String, text as sa_text
from sqlalchemy.dialects.mysql import LONGTEXT
```

- Not declared in ReMe's `pyproject.toml`
- No `try/except` guard (other optional backends like `asyncpg`, `zvec`, `pyobvector` all have one)
- Since `__init__.py` imports `ObVecVectorStore` at the top level, `import reme` will **crash** if `sqlalchemy` is not installed

**Impact:** All users without `sqlalchemy` in their environment cannot import `reme` at all — this is a blocker.

---

### Issue 2: `zvec` not declared in `pyproject.toml`

**Location:** `zvec_vector_store.py`, `zvec_file_store.py`

- Code has `try/except` guards (won't crash — good)
- But `pyproject.toml` doesn't declare it, so `pip install reme-ai` won't pull it in
- Users who select the `zvec` backend only discover the missing package at runtime

**Impact:** Poor user experience, but non-breaking for other backends.

---

### Issue 3: `pyobvector` depends on `sqlglot<30` — potential version conflict

**Location:** ReMe's `pyproject.toml`

```toml
"pyobvector>=0.1.20",
"sqlglot>=25,<30",
```

- The comment notes that `pyobvector` uses `sqlglot`'s `Expression` API, which was removed in v30+
- If a user's environment already has `sqlglot>=30` (common in data engineering stacks), this creates a dependency conflict

**Impact:** Potential conflicts, especially in data-heavy environments where `sqlglot` is widely used.

---

### Suggested Fix

The core principle: **non-core backends (hologres, obvec, zvec) should use lazy imports + `optional-dependencies`** to avoid bloating the base install and causing conflicts.

| Issue | Suggested Fix |
|---|---|
| `sqlalchemy` hard import | Convert to `try/except` lazy import, consistent with `asyncpg`/`zvec`/`pyobvector` |
| `zvec` undeclared | Add to `[project.optional-dependencies]` (e.g., `[zvec]`) |
| `sqlalchemy` undeclared | Same — add to optional deps, or bundle with a `[obvec]` extra |
| `sqlglot` conflict risk | Move `pyobvector` + `sqlglot` into `optional-dependencies` (e.g., `[obvec]`) |

---

### Ask

Could you open a follow-up upstream PR against ReMe to address these issues and cut a `0.3.1.10` release? Once that's available, we can update this PR to pin `reme-ai==0.3.1.10` and merge.

You've already demonstrated the upstream contribution workflow nicely with the watcher fix — this would be a great follow-up. Happy to help review the upstream PR as well.

PR #4226, opened 2026-05-11
feat(console): streamline add model flow

### LINKED ISSUE #4036: Adding a model requires too many steps and clicks
Adding a new model is way too complicated. Too many clicks back and forth:

1. Go to Settings → Models
2. Find the provider card, click "Settings" to fill in API Key → Save
3. Go back to the provider card list, click "Models"
4. Click "Add Model" and fill in Model ID and Model Name
5. Go back to the Chat page to select the model

Adding a single model takes 5 steps, 5 button clicks, and 3 different screens. This is way too complicated. Can you simplify this? Users shouldn't have to click around so much.

aqilaziz:
## Summary
- make the provider card show `Add Model` as the primary action when a provider has no models
- open the add-model form immediately from that action instead of requiring another click inside the modal
- when a provider is not configured yet, continue from the provider config save directly into model management

Fixes #4036

## Testing
- `cd console && npx tsc -b --noEmit`
- `cd console && npx prettier --check src/pages/Settings/Models/components/cards/RemoteProviderCard.tsx src/pages/Settings/Models/components/modals/ModelManageModal.tsx src/pages/Settings/Models/components/modals/RemoteModelManageModal.tsx`
- `git diff --check`

### DIFF
--- a/console/src/pages/Settings/Models/components/cards/RemoteProviderCard.tsx
+++ b/console/src/pages/Settings/Models/components/cards/RemoteProviderCard.tsx
@@ -24,6 +24,8 @@ export const RemoteProviderCard = React.memo(function RemoteProviderCard({
   const { message } = useAppMessage();
   const [modalOpen, setModalOpen] = useState(false);
   const [modelManageOpen, setModelManageOpen] = useState(false);
+  const [modelManageStartsAdding, setModelManageStartsAdding] = useState(false);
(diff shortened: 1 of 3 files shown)


zhaozhuang521:
We're sorry, we're not considering this option at the moment. Thank you for your support.

aqilaziz:
Thanks for reviewing and clarifying. Understood.

PR #4263, opened 2026-05-13
feat: add timeout for keyring

rayrayraykk:
## Description

image

On some Linux servers, the `DISPLAY` environment variable is set (for example via SSH X11 forwarding or systemd user services), so `_should_skip_keyring()` returns `False`.

However, in these environments, GNOME Keyring or KWallet daemon may not actually be running.

As a result, `keyring.get_password()` attempts to access the keyring service over D-Bus. When no keyring daemon is available, the D-Bus request may hang for several minutes before timing out. In the example shown in the screenshot, it blocks for about 4 minutes (`10:34:29 -> 10:38:30`).

## Fix

Add an explicit timeout to the keyring access path so that we fail fast when the keyring backend is unavailable, instead of blocking for several minutes.

## Why

Relying on `DISPLAY` alone is not sufficient to determine whether a usable desktop keyring service is actually available. In headless or semi-headless Linux environments, this can lead to unexpectedly long hangs.

This change improves responsiveness and avoids prolonged blocking when no keyring daemon is running.

### DIFF
--- a/src/qwenpaw/security/secret_store.py
+++ b/src/qwenpaw/security/secret_store.py
@@ -14,6 +14,7 @@ from legacy plaintext and transparently migrate on first access.
 from __future__ import annotations
 
 import base64
+import concurrent.futures
 import logging
 import os
 import secrets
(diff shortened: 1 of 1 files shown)


qbc2016 (approved):
LGTM.

PR #4269, opened 2026-05-13
fix(channels): graceful shutdown for WeCom WebSocket connection

celestialhorse51D:
## Description

Fix WeCom channel shutdown reliability by introducing a structured `_graceful_shutdown()` coroutine that runs on the WS event loop. Previously `stop()` used `call_soon_threadsafe(self._client.disconnect)` which only triggered SDK's synchronous disconnect without awaiting cleanup—keepalive tasks and pending upload futures could leak.

**Changes:**

1. **New `_graceful_shutdown()` coroutine** — executed on `_ws_loop` via `run_coroutine_threadsafe`, with three ordered steps:
   - Cancel all keepalive tasks and await their completion
   - Fail all pending upload futures with `RuntimeError("WeCom adapter shutting down")` so no sender hangs
   - Call `self._client.disconnect()` + `await ws_manager._cleanup_ws()` for full socket teardown

2. **`stop()` rewritten** — replaces `call_soon_threadsafe(disconnect)` with `run_coroutine_threadsafe(_graceful_shutdown())` + `asyncio.wait_for(…, timeout=15)`, plus a fallback path for when the WS loop has already stopped.

3. **Simplified `_upload_media` exception handler** — removes the redundant `str(e)[:100]` argument since `logger.exception()` already includes the traceback.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

[How to test these changes]

```bash
pre-commit run --all-files
  ********************************Passed******************************** 

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]

### DIFF
--- a/src/qwenpaw/app/channels/wecom/channel.py
+++ b/src/qwenpaw/app/channels/wecom/channel.py
@@ -59,9 +59,7 @@ _UPLOAD_CMD_FINISH = "aibot_upload_media_finish"
 _UPLOAD_CMDS = (_UPLOAD_CMD_INIT, _UPLOAD_CMD_CHUNK, _UPLOAD_CMD_FINISH)
 _UPLOAD_ACK_TIMEOUT = 30.0  # seconds to wait for each upload ack
 
-# Keepalive for "🤔 Thinking..." stream: refresh to avoid WeCom
-# server-side timeout; force-finish before the limit so later replies
-# can start a fresh stream_id (issue #3947).
+# Keepalive for "🤔 Thinking..." stream.
 _PROCESSING_REFRESH_INTERVAL = 20.0
 _PROCESSING_MAX_DURATION = 180.0
 _PROCESSING_TEXT = "🤔 Thinking..."
@@ -907,13 +905,8 @@ class WecomChannel(BaseChannel):
                     media_type,
                 )
                 return media_id
-            except Exception as e:
-                logger.exception(
-                    "wecom _upload_media failed path=%s error=%s",
-                    local[:60],
-                    str(e)[:100],
-                )
-
+            except Exception:
+                logger.exception("wecom _upload_media failed path=%s", local[:60])
                 return None
 
     async def _send_media_part(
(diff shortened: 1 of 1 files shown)


RerankerGuo (changes requested):
Blocking: the pending upload futures are owned by the app loop, but this patch mutates them from the WebSocket loop.

`_send_ws_cmd()` creates each future with `asyncio.get_running_loop().create_future()`, and the normal ACK path already uses `self._loop.call_soon_threadsafe(fut.set_result, frame)`. In contrast, `_graceful_shutdown()` is deliberately submitted to `_ws_loop` and calls `future.set_exception(...)` directly. With asyncio debug mode and a realistic registered waiter/callback, I can reproduce this on both the exact head (`ff89c7f3b9984f764e3547aebcc87141ae256288`) and a clean current-main trial:

```text
RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one
future_loop_is_main=True
shutdown_thread=contract-ws-loop
```

The exception aborts `_graceful_shutdown()` before the map is cleared and before `self._client.disconnect()` / `_cleanup_ws()` run. `stop()` then catches the exception and stops the WS loop anyway, so a pending upload can make the new graceful-shutdown path skip the socket cleanup it is intended to guarantee.

Please dispatch completion of app-loop futures through their owner loop (as the ACK path does), and audit `_keepalive_tasks` under the same ownership rule: those tasks are created with `asyncio.create_task()` in app-loop channel methods, while this coroutine currently cancels and awaits them from `_ws_loop`. Only the SDK/socket operations should execute on the WS loop.

The existing tests do not exercise real loop ownership. On both exact and current-main trial, the full WeCom unit file is currently `1 failed, 68 passed`; `test_stop_cleans_up` also emits `RuntimeWarning: coroutine 'WecomChannel._graceful_shutdown' was never awaited` because its mocked loop never consumes the submitted coroutine. Please update that test and add a two-loop regression with pending app-loop work.

The changed production file otherwise passes Black 23.3.0, Flake8 6.1.0, Mypy 1.7.0, Pylint 3.0.2, `py_compile`, and `git diff --check`.

PR #4277, opened 2026-05-13
fix(session): serialize session file updates

### LINKED ISSUE #4232: [Bug]: SafeJSONSession silently drops concurrent writes; two runs on the same chat   clobber session state
## QwenPaw Version

1.1.7b1 (also reproduces on 1.1.6b1)

## Description

`SafeJSONSession` in `src/qwenpaw/app/runner/session.py` performs read–modify–write on the per-chat JSON file with no locking. When two async tasks touch the same `(session_id, user_id, channel)` concurrently — which several runtime paths allow — the second write silently clobbers the first. Each individual call returns success and logs `Updated session state key ... successfully.`; the file on disk is well-formed JSON, just missing data.

I originally hit this from the console UI: a run became non-responsive, I clicked cancel, refreshed the page, and prompted "continue". The original run continued in the background while a new run started, and the two interleaved their saves to `<workspace>/sessions/<channel>/<user>_<session>.json`, dropping chat history. The minimal repro below shows the underlying race deterministically without needing the UI or a real model.

**Related, but distinct from prior issues:**

- **#3843 / PR #4203** — same user-visible symptom ("session history disappears"). The PR-4203 fix is **frontend-only** (`console/src/pages/Chat/sessionApi/index.ts` — a session-list dedupe change). This issue is the independent **backend** race that can produce the same symptom. The repro below does not exercise any frontend code; it imports `SafeJSONSession` directly.
- **#3275** — identifies the same `DynamicMultiAgentRunner.stream_query` bypass of `TaskTracker` (`src/qwenpaw/app/_app.py`). Different trigger (workspace reload, not concurrent user input) and different underlying race (task lifecycle, not file I/O). This issue's bypass-path list extends #3275 to cron, heartbeat, and conversation-command writes.
- **#899** — `/compact`, `/new`, `/history` silently destroy session state via `save_session_state` with an incomplete agent stub.
- **#4149** — `save_session_state` runs in the `finally` block even when the model call fails, persisting an orphan user message.

Issues #899 and #4149 are two prior accepted bugs in the same `SafeJSONSession.save_session_state` write path. This issue is a third independent failure mode (concurrent writers losing each other's updates) in that same path.

**Precedent for the proposed fix:**

- **#1834** — accepted fix of identical shape: replace a missing/sync lock with `asyncio.Lock` around `aiofiles` file I/O. The same pattern applies to `SafeJSONSession`.

**Related PR(s):** none I'm aware of for this specific race.

**Security considerations:** none — but silent data loss in the chat-history file means user input can disappear without any error surfaced.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 1.1.7b1
- **OS:** Linux (sandbox, kernel 6.12); also expected on macOS/Windows since the race is in the async/sync I/O mix, not OS-specific
- **Install method:** `uv run --with qwenpaw`
- **Python version:** 3.10

## Steps to Reproduce

1. Save `repro.py` (full source under "Minimal reproduction" below).
2. Run it:
   ```sh
   uv run --with qwenpaw python repro.py
   ```
3. Observe: sequential 20/20 keys persist; concurrent typically 1–2 / 21 keys persist. Re-run several times — the surviving key changes, but loss is consistent.

## Actual vs Expected

- **Actual:**
  ```
  Sequential : 20/20 keys persisted
  Concurrent :  1/21 keys persisted (20 silently lost)

  File on disk (u_con.json):
  {"agent": {"k14": 14}}
  ```
  Often accompanied by a WARNING from the existing recovery path:
  ```
  WARNING qwenpaw/app/runner/session.py:56 | Session file
      /tmp/.../u_con.json is completely corrupted and could not be
      recovered. Returning empty dict.
  ```
  This warning is independent confirmation: a read in the middle of a concurrent write parsed an invalid file. The `_safe_json_loads` fallback then returned `{}`, which was subsequently written back — destroying any pre-existing state.

- **Expected:** `Concurrent : 21/21 keys persisted (0 silently lost)`. Concurrent calls to `update_session_state` should be safe; at minimum, no call should ever return success while silently dropping its write.

## Root cause

In `src/qwenpaw/app/runner/session.py`:

- **`update_session_state` (line 347)** — read uses `aiofiles` (async, yields the event loop); write uses sync `open()`. N concurrent callers all yield at the read, all see the same starting state, each mutates locally, then writes back. Last writer wins.
- **`save_session_state` (line 271)** — same sync-write shape with no locking; races against itself and against `update_session_state`. Called in the `finally` of `runner.query_handler` (`runner.py:786–793`) on every chat turn.
- **`load_session_state` (line 300)** — races with both writers. The `_safe_json_loads` recovery in `session.py:25–61` exists precisely because this race has been observed before; the repro just makes it fire on demand.

## Where this fires in production

Anywhere two async tasks can hit the same chat concurrently — there are several entry points that bypass the `TaskTracker.attach_or_start(chat.id, …)` dedup that `/api/console/chat` and channel ingress already use:

1. **Cancel-then-reprompt** in the console UI — `TaskTracker.request_stop` (`runner/task_tracker.py:223`) just calls `task.cancel()` and returns immediately. The producer's `finally` (including `save_session_state`) runs asynchronously. A new request arriving in that window spawns a fresh run while the old one is still saving.
2. **Cron job with `share_session=true`** — `CronExecutor.execute` (`crons/executor.py:79`) calls `runner.stream_query` directly on the user's real `session_id`, bypassing `TaskTracker.attach_or_start`.
3. **Heartbeat** — `runner.stream_query` with hardcoded `session_id="main"` (`crons/heartbeat.py:166, 189, 206`); same bypass.
4. **`DynamicMultiAgentRunner.stream_query`** for `/api/agent/*` routes (`app/_app.py:128`) — registers each request as a fresh `ext-{uuid4}` task with no per-chat dedup.
5. **Different priority levels in `UnifiedQueueManager`** — queue key is `(channel_id, session_id, priority_level)` (`channels/unified_queue_manager.py:31`), so a `/status` (priority 10) and a normal message (priority 20) for the same session run on separate consumer tasks.
6. **Conversation-command path** — `command_dispatch.py:280, 295` calls `runner.session.update_session_state` outside the producer task, racing any concurrent run.

## Suggested fix

Two complementary changes (the first is the floor, the second the ceiling):

1. **`SafeJSONSession`: per-path `asyncio.Lock`.** Add a `dict[str, asyncio.Lock]` keyed on `_get_save_path(...)`. Hold the lock across read+modify+write in `update_session_state` and around the writes in `save_session_state` / `load_session_state`. Different paths get different locks — no cross-session contention. This stops the silent data loss and corruption for every existing and future caller.
2. **`TaskTracker`: one-run-per-chat as a hard invariant.** Route every bypass path (cron, heartbeat, `DynamicMultiAgentRunner`, command-dispatch writes) through `TaskTracker.attach_or_start(chat.id, …)`. Hold a per-`chat_id` `asyncio.Lock` in the producer from before the `_RunState` is registered until after its `finally` completes, so a new request arriving during cleanup waits for the previous run's full unwind before starting. Also have `request_stop` await the cancellation (or expose a `drain=True` mode) so cancel-then-reprompt is deterministic.

The repro below only demonstrates (1). (2) is the structural fix for the cancel-then-reprompt scenario.

## Minimal reproduction

`repro.py`:

```python
"""Minimal repro: SafeJSONSession silently loses concurrent writes.

When two async tasks call SafeJSONSession.update_session_state() concurrently
on the same (session_id, user_id, channel), all but one are silently dropped.
The file ends up well-formed JSON, just missing data — no exception, no warning.

Cause: update_session_state (src/qwenpaw/app/runner/session.py:347) is a
read-modify-write with no locking. The read uses aiofiles and yields the
event loop; the write uses sync open(). N concurrent callers all read the
same starting state, layer their mutation on top in memory, then their
writes serialize — last writer wins.

save_session_state (line 271) has the same shape and races against itself
and against update_session_state.

Run:
    uv run --with qwenpaw python repro.py

Exit 0: no data loss (fixed). Exit 1: writes silently dropped.
"""
from __future__ import annotations

import asyncio
import logging
import sys
import tempfile
from pathlib import Path

from qwenpaw.app.runner.session import SafeJSONSession  # noqa: E402

# Hide SafeJSONSession's per-write INFO lines so output is readable.
# (Each call logs "successfully" — yet data is lost.) Leave WARNINGs visible:
# the run typically surfaces a "_safe_json_loads ... completely corrupted"
# warning, which is independent evidence that the read-modify-write races.
logging.getLogger("qwenpaw").setLevel(logging.WARNING)


N = 20


async def sequential_baseline(sess: SafeJSONSession) -> dict:
    """Sequential updates — these always work."""
    for i in range(N):
        await sess.update_session_state(
            session_id="seq",
            user_id="u",
            key=f"agent.k{i}",
            value=i,
        )
    state = await sess.get_session_state_dict("seq", user_id="u")
    return state.get("agent", {})


async def concurrent_demo(sess: SafeJSONSession) -> dict:
    """Same updates as the baseline, but via asyncio.gather."""
    # Seed so all updates take the "file exists" branch.
    await sess.update_session_state(
        session_id="con",
        user_id="u",
        key="agent.seed",
        value=0,
    )
    await asyncio.gather(*[
        sess.update_session_state(
            session_id="con",
            user_id="u",
            key=f"agent.k{i}",
            value=i,
        )
        for i in range(N)
    ])
    state = await sess.get_session_state_dict("con", user_id="u")
    return state.get("agent", {})


async def main() -> int:
    with tempfile.TemporaryDirectory() as tmp:
        sess = SafeJSONSession(save_dir=tmp)

        seq = await sequential_baseline(sess)
        print(f"Sequential : {len(seq):2d}/{N} keys persisted")

        con = await concurrent_demo(sess)
        # +1 for the "seed" key
        expected = N + 1
        got = len(con)
        lost = expected - got
        print(
            f"Concurrent : {got:2d}/{expected} keys persisted "
            f"({lost} silently lost)",
        )

        path = Path(tmp) / "u_con.json"
        print(f"\nFile on disk ({path.name}):\n{path.read_text()}")

        if lost > 0:
            print(
                f"\nFAIL: {lost} writes lost. update_session_state has a "
                "read-modify-write race with no locking.",
            )
            return 1
        print("\nOK: no writes lost.")
        return 0


if __name__ == "__main__":
    sys.exit(asyncio.run(main()))
```

## Logs / Screenshots

Output from three back-to-back runs against 1.1.6b1:

```
--- Run 1 ---
WARNING qwenpaw/app/runner/session.py:56 | Session file /tmp/.../u_con.json
    is completely corrupted and could not be recovered. Returning empty dict.
Sequential : 20/20 keys persisted
Concurrent :  1/21 keys persisted (20 silently lost)

File on disk (u_con.json):
{"agent": {"k14": 14}}

FAIL: 20 writes lost. update_session_state has a read-modify-write race
with no locking.

--- Run 2 ---
WARNING qwenpaw/app/runner/session.py:56 | Session file /tmp/.../u_con.json
    is completely corrupted and could not be recovered. Returning empty dict.
Sequential : 20/20 keys persisted
Concurrent :  2/21 keys persisted (19 silently lost)

File on disk (u_con.json):
{"agent": {"seed": 0, "k6": 6}}

FAIL: 19 writes lost.

--- Run 3 ---
WARNING qwenpaw/app/runner/session.py:56 | Session file /tmp/.../u_con.json
    is completely corrupted and could not be recovered. Returning empty dict.
Sequential : 20/20 keys persisted
Concurrent :  1/21 keys persisted (20 silently lost)

File on disk (u_con.json):
{"agent": {"k16": 16}}

FAIL: 20 writes lost.
```

Notes:

- The surviving key varies between runs (timing-dependent) but loss is consistent every time.
- The `_safe_json_loads ... completely corrupted` WARNING fires on every run — this is `session.py:56`'s existing recovery path triggering because a read landed mid-write.
- No exception, no error log, no failure status from any individual `update_session_state` call. Each one logged `Updated session state key ... successfully.` at INFO before being silently overwritten.

## Additional Notes

- Test plan I'd suggest, beyond the repro above:
  - Add a unit test in `tests/unit/agents/test_session.py` that calls `asyncio.gather(update_session_state(key_a), update_session_state(key_b))` against a seeded file and asserts both keys persist. Deterministic — fails reliably on `main`, passes with a per-path `asyncio.Lock`.
  - A stress version with N=50–100 concurrent updates to amplify the loss for visibility.
  - A negative test verifying the lock is per-path (different `(session_id, user_id, channel)` triples don't serialize).
  - For the higher-level "one run per chat" invariant: a test that fires two `query_handler` coroutines against the same triple and asserts the second waits for the first's `finally` to complete.
- The existing `_safe_json_loads` recovery is a strong hint that this race has been hit in the wild before — the recovery just hides it most of the time by returning `{}`, which then gets written back, deleting prior state silently.
- Happy to send a PR for fix (1) if maintainers agree on the direction.

aqilaziz:
Fixes #4232.

## Description
Adds per-session-file `asyncio.Lock` protection inside `SafeJSONSession` so reads, writes, and read-modify-write updates for the same JSON session file cannot interleave. This prevents concurrent `update_session_state` calls from silently clobbering each other while keeping different session files independent.

I read the README and CONTRIBUTING guidelines before taking this issue, and commented on #4232 before starting the focused fix.

## Type of Change
Bug fix.

## Components Affected
Core / Backend session persistence, tests.

## Testing
```bash
PYTHONPATH=src python -m pytest tests/unit/agents/test_session.py
# 17 passed

pre-commit run --files src/qwenpaw/app/runner/session.py tests/unit/agents/test_session.py
# all hooks passed
```

Additional local note:
```bash
pre-commit run --all-files
# attempted; local Python 3.13 run fails on unrelated baseline files:
# src/qwenpaw/cli/shutdown_cmd.py mypy arg-type errors,
# plus pylint no-name-in-module on existing collections.abc imports.
```

## Checklist
- [x] Checked existing issues and open PRs
- [x] Commented on the issue before starting
- [x] Added a regression test for concurrent session updates
- [x] Ran targeted pytest
- [x] Ran pre-commit on changed files

### DIFF
--- a/src/qwenpaw/app/runner/session.py
+++ b/src/qwenpaw/app/runner/session.py
@@ -6,6 +6,7 @@ Windows filenames cannot contain: \\ / : * ? " < > |
 This module wraps agentscope's SessionBase so that session_id and user_id
 are sanitized before being used as filenames.
 """
+import asyncio
 import os
 import re
 import json
@@ -211,6 +212,39 @@ class SafeJSONSession(SessionBase):
                 The directory to save the session state.
         """
         self.save_dir = save_dir
+        self._path_locks: dict[str, asyncio.Lock] = {}
+
+    def _get_path_lock(self, session_save_path: str) -> asyncio.Lock:
+        """Return the per-file lock for a session JSON path."""
+        normalized_path = os.path.abspath(session_save_path)
+        lock = self._path_locks.get(normalized_path)
+        if lock is None:
+            lock = asyncio.Lock()
+            self._path_locks[normalized_path] = lock
+        return lock
+
+    @staticmethod
+    async def _read_states_from_path(session_save_path: str) -> dict:
+        async with aiofiles.open(
+            session_save_path,
+            "r",
+            encoding="utf-8",
+            errors="surrogatepass",
+        ) as f:
+            content = await f.read()
+            return _safe_json_loads(content, session_save_path)
+
+    @staticmethod
+    async def _write_states_to_path(
+        session_save_path: str,
+        states: dict,
+    ) -> None:
+        async with aiofiles.open(
+            session_save_path,
+            "w",
+            encoding="utf-8",
+        ) as f:
+            await f.write(json.dumps(states, ensure_ascii=False))
 
     def _get_save_path(
         self,
@@ -276,21 +310,17 @@ class SafeJSONSession(SessionBase):
         **state_modules_mapping,
     ) -> None:
         """Save state modules to a JSON file using async I/O."""
-        state_dicts = {
-            name: state_module.state_dict()
-            for name, state_module in state_modules_mapping.items()
-        }
         session_save_path = self._get_save_path(
             session_id,
             user_id=user_id,
             channel=channel,
         )
-        with open(
-            session_save_path,
-            "w",
-            encoding="utf-8",
-        ) as f:
-            f.write(json.dumps(state_dicts, ensure_ascii=False))
+        async with self._get_path_lock(session_save_path):
+            state_dicts = {
+                name: state_module.state_dict()
+                for name, state_module in state_modules_mapping.items()
+            }
+            await self._write_states_to_path(session_save_path, state_dicts)
 
         logger.info(
             "Saved session state to %s successfully.",
@@ -311,38 +341,32 @@ class SafeJSONSession(SessionBase):
             user_id=user_id,
             channel=channel,
         )
-        if os.path.exists(session_save_path):
-            async with aiofiles.open(
-                session_save_path,
-                "r",
-                encoding="utf-8",
-                errors="surrogatepass",
-            ) as f:
-                content = await f.read()
-                states = _safe_json_loads(content, session_save_path)
+        async with self._get_path_lock(session_save_path):
+            if os.path.exists(session_save_path):
+                states = await self._read_states_from_path(session_save_path)
+            elif allow_not_exist:
+                logger.info(
+                    "Session file %s does not exist. "
+                    "Skip loading session state.",
+                    session_save_path,
+                )
+                return
+            else:
+                raise AgentStateError(
+                    session_id=session_id,
+                    message=(
+                        f"Failed to load session state for file "
+                        f"{session_save_path} because it does not exist"
+                    ),
+                )
 
-            for name, state_module in state_modules_mapping.items():
-                if name in states:
-                    state_module.load_state_dict(states[name])
-            logger.info(
-                "Load session state from %s successfully.",
-                session_save_path,
-            )
-
-        elif allow_not_exist:
-            logger.info(
-                "Session file %s does not exist. Skip loading session state.",
-                session_save_path,
-            )
-
-        else:
-            raise AgentStateError(
-                session_id=session_id,
-                message=(
-                    f"Failed to load session state for file "
-                    f"{session_save_path} because it does not exist"
-                ),
-            )
+        for name, state_module in state_modules_mapping.items():
+            if name in states:
+                state_module.load_state_dict(states[name])
+        logger.info(
+            "Load session state from %s successfully.",
+            session_save_path,
+        )
 
     async def update_session_state(
         self,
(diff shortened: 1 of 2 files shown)


zhijianma:

PR #4303, opened 2026-05-14
fix(cron): isolate non-shared runs

### LINKED ISSUE #4162: [Bug]: 会话已经删除，但由于定时任务设置了sessionId，每次定时任务执行还是在旧的上下文中
## QwenPaw Version 1.1.5.post2

定时任务设置了sessionId，定时任务会根据触发词读取技能，开始执行任务。
但是后来我调整了技能文件，定时任务一直用该session的会话上下文处理任务，没有重新读取技能，导致一直错误执行
即使我删除了该会话，重启了qwenpaw，定时任务执行的上下文还是旧的上下文，并没有重新读取技能文件，按照新的技能说明执行任务。

删除了会话、重启qwenpaw都没有用
只有删除定时任务，设置新的session id才可以

需要修改成以下逻辑
1. 删除会话后，应该每次都是新会话了，而不是根据sessionId找到之前的会话内容作为上下文
2. 或者定时任务可以支持设置不使用之前的上下文
3. 我禁用定时任务，编辑 目标会话ID  保存没有生效，应该保存成新的，不然只能删掉定时任务，重新创建

我的定时任务不需要关注之前的上下文，所以创建定时任务能否指定新的会话，相当于执行过/new，而不是一直在定时任务上累积内容

aqilaziz:
## Summary
- make cron agent jobs with `runtime.share_session=false` use a fresh per-run session id
- keep shared cron jobs on the configured target session only while that chat still exists
- avoid resurrecting a deleted bound session by falling back to a fresh per-run session id
- add CronExecutor tests for shared, deleted-target, and isolated session behavior
- expose `share_session` on the console cron runtime type

Fixes #4162

## Tests
- `PYTHONPATH=src python -m pytest tests/unit/app/crons/test_executor.py -q` - 4 passed
- `pre-commit run --files src/qwenpaw/app/crons/executor.py tests/unit/app/crons/test_executor.py console/src/api/types/cronjob.ts` - passed
- `git diff --check` - passed

## Maintainer follow-up
- Branch updated with latest `upstream/main` (`0dcbcc5`); executor conflict resolved.
- GitHub reports this PR as mergeable.
- Spam-gate checks are passing.
- Remaining blocker is `Maintainer Approval`, so the main test matrix is waiting/skipped until a maintainer approves the workflow run.

Reviewer note: yes, AI-assisted local tooling was used for faster iteration.


CA-mambo:
Thanks for the contribution! The new `share_session` flag is a useful addition for isolated runs. 

However, looking at the original report in #4162, the core issue is about the system attempting to load/revive a session context that the user has already **deleted**. I'm wondering if this change covers the default scenario where a specific `session_id` is explicitly bound to the cron job, but that session gets deleted afterwards. 

In that case (shared mode), will the executor still try to resurrect the deleted session? We likely need a mechanism to ensure that deleted sessions are strictly inaccessible to the scheduler, regardless of the isolation configuration. Could you clarify how this edge case is handled? Thanks!

aqilaziz:
Good catch. The original patch only fixed the explicit `share_session=false` path and did not protect the shared/default path when the bound session had already been deleted.

I pushed a follow-up commit (`2c8b2c2`) that checks the active ChatManager before reusing the configured `target.session_id`. If the chat spec for that session is gone, the cron run now uses a fresh per-run session id instead of reusing the deleted session id and potentially reviving stale runner state. If ChatManager is unavailable, the executor preserves the previous behavior.

Added regression coverage:
```bash
PYTHONPATH=src python -m pytest tests/unit/app/crons/test_executor.py -q
# 4 passed

pre-commit run --files src/qwenpaw/app/crons/executor.py tests/unit/app/crons/test_executor.py console/src/api/types/cronjob.ts
# passed

git diff --check
# passed
```

CA-mambo:
Checking comments for PR #4303...

CA-mambo:
The speed and precision of this update are impressive ⚡. The code structure feels very familiar... are you using Claude or an AI agent for these contributions? 🤖 Just curious!

PR #4321, opened 2026-05-14
feat(config): extend hot-reload to cover mcp.clients and skills sections

kkomia:
Extend AgentConfigWatcher to also monitor mcp.clients and skills sections of agent.json, so configuration changes in these sections take effect without restarting QwenPaw.

Previously only channels and heartbeat were monitored. Users had to restart QwenPaw for MCP or Skills changes — and Agents had no way to know, often suggesting unnecessary restarts.

Changes:
- _mcp_hash() — detects mcp.clients changes
- _skills_hash() — detects skills changes
- Both tracked in the poll loop, trigger reload_agent when changed
- Refactored duplicated hash() into a shared _hash() helper

Tested with edge cases: None, empty, Pydantic model, config changes — all pass.

### DIFF
--- a/src/qwenpaw/app/agent_config_watcher.py
+++ b/src/qwenpaw/app/agent_config_watcher.py
@@ -47,18 +47,28 @@ def _mcp_hash(mcp: Any) -> Optional[int]:
     """Hash of mcp / mcp.clients section for change detection."""
     if mcp is None:
         return None
-    # mcp.clients is the dict of MCP client configs
-    clients = getattr(mcp, "clients", None)
-    if clients is None:
-        return None
-    return hash(str(clients))
+    try:
+        # Pydantic model: mcp.clients is a field
+        clients = getattr(mcp, "clients", None)
+        if clients is not None:
+            return hash(str(clients))
+    except Exception:
+        pass
+    # Fallback: dump entire mcp section
+    try:
+        return hash(str(mcp.model_dump(mode="json")))
+    except Exception:
+        return hash(str(mcp))
 
 
 def _skills_hash(skills: Any) -> Optional[int]:
     """Hash of skills section for change detection."""
     if skills is None:
         return None
-    return hash(str(skills))
+    try:
+        return hash(str(skills.model_dump(mode="json")))
+    except Exception:
+        return hash(str(skills))
 
 
 class AgentConfigWatcher:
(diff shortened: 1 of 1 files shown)


RerankerGuo (changes requested):
The MCP half is reachable, but the advertised Skills hot reload cannot observe any `agent.json` change at exact head `01431227a05db5e98745236a4d7a0e5889f8ba25`.

Both `_snapshot()` and `_check()` read `getattr(agent_config, "skills", None)` ([lines 136-137](https://github.com/agentscope-ai/QwenPaw/blob/01431227a05db5e98745236a4d7a0e5889f8ba25/src/qwenpaw/app/agent_config_watcher.py#L133-L138) and [178-179](https://github.com/agentscope-ai/QwenPaw/blob/01431227a05db5e98745236a4d7a0e5889f8ba25/src/qwenpaw/app/agent_config_watcher.py#L175-L180)). However, exact-head `AgentProfileConfig` has `mcp` but no `skills` field ([config.py lines 1059-1139](https://github.com/agentscope-ai/QwenPaw/blob/01431227a05db5e98745236a4d7a0e5889f8ba25/src/qwenpaw/config/config.py#L1059-L1139)). Pydantic therefore discards a raw `skills` key before the watcher sees the config.

I ran the same production-schema contract on the exact head (with its locked `agentscope-runtime==1.1.4` interface isolated) and on a clean current-main trial:

```text
exact 01431227:
  mcp_changed=True
  skills_a=None
  skills_b=None
  skills_changed=False

trial 046fbab5:
  mcp_changed=True
  skills_a=None
  skills_b=None
  skills_changed=False
```

The positive MCP control uses two different `clients` mappings and proves that part of the new detector changes. The Skills control validates two raw agent profiles with `skills=["browser"]` and `skills=["memory"]`; both parsed models lack the attribute, both hashes are `None`, and no reload can be triggered. This is not current-main schema drift: the field is absent at the PR's own base/head as well. Skills are currently resolved from workspace `skills/` contents through the skill registry, not an `agent.json.skills` field.

Please either:

- narrow this PR to MCP hot reload and add focused watcher tests, or
- define and persist an actual Skills config contract (including all writers/readers), then observe that supported source.

The current-main trial passes the existing 19 startup/reload tests, but none exercise either new detector. Locked changed-file hooks also fail: the encoding hook and Black modify the file, Mypy rejects `_heartbeat_hash()` returning `Optional[int]` under an `int` annotation, and Flake8 reports an 80-character line. The remote head currently has only the welcome check, so these failures are not covered there.

PR #4331, opened 2026-05-14
fix(shell): inject request context into subprocess env

### LINKED ISSUE #3825: [Feature] Inject per-message context into shell tool subprocess env
## Summary

When the LLM agent invokes the shell tool to run a skill script, the
subprocess inherits a static `os.environ.copy()` snapshot from the
QwenPaw worker process. Per-message context (Matrix sender, room id,
session id) reaches the LLM via `build_env_context(...)`
(system-prompt string), but **does not** reach the subprocess as env
— making it impossible for skill scripts to attribute the current
invocation to the originating user.

## Component(s) Affected

- [x] Core / Backend (`agents/tools/shell.py`)
- [x] Skills (downstream consumers of per-message attribution)

## Problem / Motivation

In multi-tenant scenarios where one Worker serves many users, audit
trails and access checks performed by skill scripts need a reliable
way to know "who triggered this invocation". The data is already in
the runtime payload:

- `copaw/src/matrix/channel.py:1571` constructs
  `payload.meta = {room_id, sender_id, event_id, ...}`
- `qwenpaw/app/runner/runner.py:350` already calls
  `build_env_context(session_id=..., user_id=..., channel=..., ...)`

But `qwenpaw/agents/tools/shell.py:327` does:

```python
env = os.environ.copy()
...
proc = await asyncio.create_subprocess_shell(cmd, ..., env=env, ...)
```

— a process-wide snapshot, not per-message. So the LLM sees the
user_id but the subprocess does not.

## Proposed Solution

Merge a per-call context dict on top of `os.environ.copy()` before
fork. The dict could be sourced from a contextvar populated by the
runner just before the shell tool is invoked (or any other mechanism
the maintainers prefer).

Suggested env names (project-prefixed, naming aligned with existing
runner fields):

- `QWENPAW_USER_ID` — user identity (always present once known)
- `QWENPAW_SESSION_ID` — session identity (always present)
- `QWENPAW_CHANNEL` — channel name (matches the existing `channel`
  argument to `build_env_context`); always present
- `QWENPAW_ROOM_ID` — **optional**; set when available (Matrix and
  similar room-based channels). Not all channels have this concept
- `QWENPAW_EVENT_ID` — **optional**; set when available (Matrix
  event ID). Not all channels expose a stable event id

Skill scripts that don't care can ignore them; scripts needing
attribution can read them via `os.environ.get(...)`.

## Why not LLM-side splicing

A tempting alternative is asking the LLM to splice
`USER_ID=<value>` into bash invocations from `build_env_context`'s
prompt text. We considered and rejected this for our project: audit
fields written by an LLM are prompt-injectable and unreliable; a fake
user_id is worse than absent attribution. Runtime-level injection is
the only design that audit infrastructure can actually trust.

## Use case

Downstream project (haopaw, multi-agent platform on HiClaw v1.1.0
GA): domain-service skills sign HMAC confirmation tokens and need the
originating Matrix user in their audit logs. Currently the skill
scripts fall back to `"unknown"` for actor headers — functional, but
loses traceability for compliance review.

## Acceptance test

Within a skill script:

- `os.environ.get("QWENPAW_USER_ID")` matches the message originator
  (e.g. Matrix `event.sender`)
- Concurrent messages in different rooms must not see each other's
  values
- For channels that do not expose `room_id` / `event_id`,
  `QWENPAW_ROOM_ID` / `QWENPAW_EVENT_ID` should be **unset** (not
  `"unknown"` or `""`). The runtime injection layer should never
  manufacture placeholder values; downstream code is responsible for
  whatever fallback policy makes sense for it.

aqilaziz:
## Description
This PR fixes #3825 by injecting runtime request context into shell subprocess environments. The runner now provides per-call attribution as `QWENPAW_*` env vars, so downstream skill scripts can reliably audit the originating user/session/channel without trusting LLM-spliced shell arguments.

## Type of Change
- Bug fix
- Feature enhancement

## Component(s) Affected
- Core
- Skills
- Tests

## Summary
- add a per-request shell contextvar that maps runtime attribution to `QWENPAW_*` env vars
- set shell context from `AgentRunner` using session/user/channel plus optional `room_id` and `event_id` metadata
- merge the context env into `execute_shell_command` subprocess environments while preserving the venv PATH prepend
- cover fallback, optional metadata, and concurrent context isolation with unit tests

## Testing
Local verification:
```bash
PYTHONPATH=E:\Data Yayasan\APP Aqil\PR\QwenPaw\src pytest tests\unit\agents\tools\test_shell_context.py -q
flake8 src\qwenpaw\agents\tools\shell.py src\qwenpaw\agents\tools\shell_context.py src\qwenpaw\app\runner\runner.py tests\unit\agents\tools\test_shell_context.py
pylint src\qwenpaw\agents\tools\shell.py src\qwenpaw\agents\tools\shell_context.py src\qwenpaw\app\runner\runner.py tests\unit\agents\tools\test_shell_context.py --disable=all --enable=syntax-error,undefined-variable,unused-import,mixed-line-endings
git diff --check
```

GitHub Actions currently passing:
- Pre-commit Checks
- Unit Tests - py3.10
- Unit Tests - py3.13
- Coverage Report
- website format check
- console format check

## Checklist
- [x] Read README and contributing guidance before implementation
- [x] Added focused unit coverage
- [x] Ran relevant local tests/lint checks
- [x] Confirmed existing CI checks pass
- [x] Documentation update not required; this is runtime env behavior for shell subprocesses

Closes #3825

### DIFF
--- a/src/qwenpaw/agents/tools/shell.py
+++ b/src/qwenpaw/agents/tools/shell.py
@@ -23,6 +23,7 @@ from ...config.context import (
     get_current_shell_command_timeout,
     get_current_workspace_dir,
 )
+from .shell_context import get_shell_command_context_env
 
 
 def _kill_process_tree_win32(pid: int) -> None:
@@ -180,6 +181,21 @@ def _is_powershell(executable: str) -> bool:
 def _is_cmd(executable: str) -> bool:
     """Check if the given executable path is cmd.exe."""
     return _shell_basename(executable) in ("cmd", "cmd.exe")
+
+
+def _build_subprocess_env() -> dict[str, str]:
+    """Build the environment passed to shell subprocesses."""
+    env = os.environ.copy()
+    env.update(get_shell_command_context_env())
+
+    # Ensure the venv Python is on PATH for subprocesses
+    python_bin_dir = str(Path(sys.executable).parent)
+    existing_path = env.get("PATH", "")
+    if existing_path:
+        env["PATH"] = python_bin_dir + os.pathsep + existing_path
+    else:
+        env["PATH"] = python_bin_dir
+    return env
 
 
 _PS_CMD_RE = re.compile(
@@ -403,14 +419,7 @@ async def execute_shell_command(
     else:
         working_dir = get_current_workspace_dir() or WORKING_DIR
 
-    # Ensure the venv Python is on PATH for subprocesses
-    env = os.environ.copy()
-    python_bin_dir = str(Path(sys.executable).parent)
-    existing_path = env.get("PATH", "")
-    if existing_path:
-        env["PATH"] = python_bin_dir + os.pathsep + existing_path
-    else:
-        env["PATH"] = python_bin_dir
+    env = _build_subprocess_env()
 
     shell_executable = (
         get_current_shell_command_executable()
--- a/src/qwenpaw/agents/tools/shell_context.py
+++ b/src/qwenpaw/agents/tools/shell_context.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+"""Per-request shell subprocess context."""
+from __future__ import annotations
+
+from contextvars import ContextVar, Token
+from typing import Mapping, Optional
+
+
+_SHELL_CONTEXT: ContextVar[dict[str, str]] = ContextVar(
+    "qwenpaw_shell_context",
+    default={},
+)
+
+_ENV_MAPPING = {
+    "user_id": "QWENPAW_USER_ID",
+    "session_id": "QWENPAW_SESSION_ID",
+    "channel": "QWENPAW_CHANNEL",
+    "room_id": "QWENPAW_ROOM_ID",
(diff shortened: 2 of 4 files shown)


hlgone:
Thanks @aqilaziz for picking this up. I reviewed the PR body / intended shape, and from the downstream-consumer perspective this matches the original ask in #3825:

- per-call contextvar populated by the runner, merged into subprocess env (not a process-wide `os.environ.copy()` snapshot)
- always-present `QWENPAW_USER_ID` / `QWENPAW_SESSION_ID` / `QWENPAW_CHANNEL`
- optional `QWENPAW_ROOM_ID` / `QWENPAW_EVENT_ID` only when the channel exposes them
- venv PATH prepend preserved

This is exactly the runtime-level injection the issue called for: env vars sourced by the runtime are the only audit fields a downstream consumer can actually trust; LLM-spliced bash args are prompt-injectable.

For the use case described in #3825, this shape is sufficient from the consumer side.

PR #4364, opened 2026-05-14
fix(providers): parse raw think tags

### LINKED ISSUE #4174: [Bug]: If we use an API format like OpenAI, the agent's thoughts aren't collapsed into "thinking." This takes up a lot of screen space.
Image

aqilaziz:
## Summary
- Parse raw `<think>...</think>` text emitted by OpenAI-compatible providers into structured `thinking` blocks.
- Preserve normal answer text as `text` blocks, so the console can collapse reasoning instead of rendering raw tags inline.
- Keep existing tool-call extraction working after thinking-tag normalization.

Fixes #4174

## Tests
- `PYTHONPATH=src python -m pytest tests/unit/providers/test_openai_stream_toolcall_compat.py tests/unit/agents/test_cross_provider_normalization.py -q`
- `pre-commit run --files src/qwenpaw/providers/openai_chat_model_compat.py tests/unit/providers/test_openai_stream_toolcall_compat.py`
- `git diff --check c89ed13..HEAD`

### DIFF
--- a/src/qwenpaw/providers/openai_chat_model_compat.py
+++ b/src/qwenpaw/providers/openai_chat_model_compat.py
@@ -13,7 +13,10 @@ from agentscope.model._model_response import ChatResponse
 from pydantic import BaseModel
 
 from qwenpaw.local_models.tag_parser import (
+    THINK_END,
+    THINK_START,
     parse_tool_calls_from_text,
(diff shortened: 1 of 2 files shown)


xieyxclack:

xieyxclack:
Closed as invalid

Issue #4340, opened 2026-05-14
[test] app/runner + app/routers 单测覆盖 (Phase 3)


hanson-hex:
## Parent

Backend unit test coverage milestone

## What to build

Add unit tests and contract tests for `app/runner/` and `app/routers/` — core runtime and user API entry points (P1). Currently severely undertested (runner: 1/11 files, routers: 2/26). This issue uses worktree branch `test/phase3-runner-routers`.

**Unit tests (8 files):**

app/runner/:
- `test_command_dispatch.py` — command dispatch: _is_command, _get_last_user_text, run_command_path
- `test_session.py` — SafeJSONSession serialization/recovery, corrupted file handling
- `test_task_tracker.py` — task state transitions, cancellation, concurrency, timeout
- `test_manager.py` — ChatManager init, session creation/retrieval

app/routers/ — target the 4 highest-value routers by line count and API importance:
- `test_skills_router.py` — skills.py (1543L): GET/POST/DELETE skills, install-time security scan integration, SSE stream format
- `test_config_router.py` — config.py (999L): GET/PUT config, provider CRUD, validation error 422
- `test_messages_router.py` — messages.py: chat message API, streaming responses
- `test_agents_router.py` — agents.py (504L): agent management API

**Contract test (1 file):**
- `tests/contract/runner/test_control_command_contract.py` — verify all BaseControlCommandHandler subclasses (ApprovalCommandHandler, ApproveCommandHandler, DenyCommandHandler, StopCommandHandler, ModelCommandHandler, SkillsCommandHandler) implement `handle(context) -> str`, have valid `command_name`, and handle missing/empty context gracefully

**CI gate expansion:**
- Add `tests/unit/app/runner/` and `tests/unit/app/routers/` as hard gates in `unit-tests.yml`
- Add `tests/contract/runner/` as hard gate

Test pattern: Router tests use `httpx.AsyncClient` + FastAPI `TestClient`, mock `app.state.multi_agent_manager`; Runner tests mock command registry and agent dependencies.

## Acceptance criteria

- [ ] `pytest tests/unit/app/runner/ -v` all pass
- [ ] `pytest tests/unit/app/routers/ -v` all pass (including existing test_settings, test_console_placeholder)
- [ ] `pytest tests/contract/runner/ -v` all pass (BaseControlCommandHandler contract compliance)
- [ ] CI: runner and routers unit tests + runner contract tests added as hard gates
- [ ] Gate tests unaffected
[...]

PR #4375, opened 2026-05-14
fix(app): persist agent scoped root session context

aqilaziz:
## Summary
- persist X-Root-Session-Id in request.state.request_context from the agent-scoped middleware
- update the root session context variable when the header is present
- add focused router middleware tests for path/header precedence, root-session propagation, and mounted agent routes

Refs #4340

## Tests
- `PYTHONPATH=src pytest tests/unit/routers/test_agent_scoped.py -q`
- `pre-commit run --files src/qwenpaw/app/routers/agent_scoped.py tests/unit/routers/test_agent_scoped.py`
- `git diff --check`


xieyxclack:

PR #4377, opened 2026-05-14
fix(memory): default auto backend to local

### LINKED ISSUE #3854: [Bug]: chromadb Rust binding segfault (SIGSEGV) kills the entire process — need a safer default or graceful fallback
## Summary

On Linux (Ubuntu 25.10, Python 3.13), the `chromadb` Rust binding (`chromadb_rust_bindings.abi3.so`) causes a **segmentation fault** that kills the entire QwenPaw process without any Python-level exception handling. This happened **45+ times** in a single session, making the agent completely unusable.

## Environment

| Item | Version |
|------|---------|
| QwenPaw | 1.1.4.post2 |
| reme-ai | 0.3.1.8 |
| chromadb | 1.5.7 (also reproduced on 1.5.8) |
| Python | 3.13.7 |
| OS | Ubuntu 25.10, kernel 6.17.0 |
| CPU | AMD HawkPoint1 (16 cores) |

### 1. Segfault in Rust binding

The crash is a null pointer dereference inside `chromadb_rust_bindings.abi3.so`:

```
kernel: python3[PID]: segfault at 10 ip 00007f1234abcd57 sp 00007f1234567890 error 4 in chromadb_rust_bindings.abi3.so[...ee57]
```

This is triggered when the HNSW index (`data_level0.bin`) becomes inconsistent with the embeddings metadata (in our case: 13MB HNSW file but 0 rows in the embeddings table — likely leftover from a previous unclean shutdown). The Rust code does not check for null before dereferencing, causing an immediate `SIGSEGV` that bypasses all Python exception handling.

### 2. Auto-detection prefers chromadb with no fallback

`reme_light_memory_manager.py` `_detect_memory_manager_backend()`:

```python
def _detect_memory_manager_backend() -> str:
    backend_env = EnvVarLoader.get_str("MEMORY_STORE_BACKEND", "auto")
    if backend_env != "auto":
        return backend_env

    if platform.system() == "Windows":
        return "local"

    try:
        import chromadb  # noqa: F401
        return "chroma"
    except Exception:
        return "local"
```

The logic is: "if chromadb is importable → use it". On Linux, if `pip install chromadb` was ever run, chromadb is importable, and there's no way to opt out without knowing about the `MEMORY_STORE_BACKEND` env var (which is not documented prominently).

## Impact

- **Process kill without traceback**: SIGSEGV cannot be caught by Python. The process just dies. No error log, no graceful shutdown, no restart.
- **Repeated crashes**: If the corrupted HNSW index is not manually deleted, every startup triggers the segfault again on the first memory operation.
- **Data loss risk**: Any in-flight conversation or task is lost when the process is killed.

## Reproduction

1. Install QwenPaw on Linux with `chromadb` as a dependency
2. Corrupt or leave a stale HNSW index in the file_store directory (e.g., after an unclean shutdown)
3. Start QwenPaw — process segfaults on first memory access

## Workaround

Set environment variable before starting QwenPaw:

```bash
MEMORY_STORE_BACKEND=local qwenpaw app --port 9099
```

Or in code:
```python
import os
os.environ["MEMORY_STORE_BACKEND"] = "local"
```

This forces ReMe to use `LocalFileStore` (pure Python, SQLite FTS) instead of `ChromaFileStore`.

### Option A: Default to `local` on Linux too (Recommended)

The `local` backend (SQLite FTS) is sufficient for ReMe's use case (context compaction + MEMORY.md). chromadb's vector search is only useful when embeddings are configured, which most default setups don't have. Making `local` the default would eliminate this class of crash entirely.

### Option B: Wrap chromadb initialization in a subprocess / crash guard

If chromadb must remain the default, catch the segfault at startup by testing chromadb in a subprocess first. If it crashes, fall back to `local`.

### Option C: Add `MEMORY_STORE_BACKEND` to the docs / config wizard

At minimum, document this env var in the configuration docs and show it in `qwenpaw init` so users can opt out of chromadb without reading source code.

## Related

- The corrupted HNSW data has been preserved as a reproduction sample (can share if needed for a chromadb upstream bug report).
- This affects the `reme-ai` package directly, but the impact surface is QwenPaw users on Linux.

aqilaziz:
## Summary
- make MEMORY_STORE_BACKEND=auto resolve to the pure-Python local backend by default
- preserve explicit MEMORY_STORE_BACKEND overrides such as chroma
- add regression coverage for auto/default and explicit backend behavior

Fixes #3854

## Tests
- PYTHONPATH=E:\Data Yayasan\APP Aqil\PR\QwenPaw\src pytest tests\unit\agents\memory\test_reme_light_memory_manager.py -q
- pre-commit run --files src/qwenpaw/agents/memory/reme_light_memory_manager.py tests/unit/agents/memory/test_reme_light_memory_manager.py
- git diff --check

### DIFF
--- a/src/qwenpaw/agents/memory/reme_light_memory_manager.py
+++ b/src/qwenpaw/agents/memory/reme_light_memory_manager.py
@@ -3,7 +3,6 @@
 import importlib.metadata
 import json
 import logging
-import platform
 import shutil
 import uuid
 from datetime import datetime
@@ -42,9 +41,8 @@ def _detect_memory_manager_backend() -> str:
     """Detect the memory store backend from environment variables.
 
     Resolves ``MEMORY_STORE_BACKEND`` with the following priority:
-    - ``local``: always used on Windows
-    - ``chroma``: used when ``chromadb`` is importable (non-Windows)
-    - falls back to ``local`` when ``chromadb`` is unavailable
+    - any explicit value is used as-is, including ``chroma``
+    - ``auto`` uses the pure-Python ``local`` backend
 
     Returns:
         Backend name string: ``"local"``, ``"chroma"``, or any explicitly
@@ -54,24 +52,7 @@ def _detect_memory_manager_backend() -> str:
     if backend_env != "auto":
         return backend_env
 
-    if platform.system() == "Windows":
-        return "local"
-
-    try:
-        import chromadb  # noqa: F401 pylint: disable=unused-import
-
-        return "chroma"
-    except Exception as e:
-        logger.warning(
-            f"""
-chromadb import failed, falling back to `local` backend.
-This is often caused by an outdated system SQLite (requires >= 3.35).
-Please upgrade your system SQLite to >= 3.35.
-See: https://docs.trychroma.com/docs/overview/troubleshooting#sqlite
-| Error: {e}
-            """,
-        )
-        return "local"
+    return "local"
 
 
 @memory_registry.register("remelight")
--- a/tests/unit/agents/memory/test_reme_light_memory_manager.py
+++ b/tests/unit/agents/memory/test_reme_light_memory_manager.py
@@ -30,6 +30,26 @@ for _mod in _MOCK_MODULES:
 _MOD = "qwenpaw.agents.memory.reme_light_memory_manager"
 
 
+class TestDetectMemoryManagerBackend:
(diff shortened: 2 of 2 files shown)


xieyxclack:

xieyxclack:
After reviewing the pr, we decided to close it, as it is not related to the target issue

PR #4380, opened 2026-05-14
fix(console): resolve temporary chat ids reliably

### LINKED ISSUE #3254: [Bug] Console: backend chat UUID may be missing when GET /chats lags behind session creation (race with stop / delete)
### Summary

In the **web console**, a new chat is represented locally with a **temporary id** (numeric timestamp) while the server stores a **UUID** as `chat_id` in chat metadata. The UI needs that UUID for **delete**, **URL routing**, and syncing with **`GET /chats`**.

The client resolves the UUID by calling **`GET /chats`** (`listChats`) and matching **`session_id`** to the local temporary id (`resolveRealId` in `console/src/pages/Chat/sessionApi/index.ts`). If the **first** list response does **not** yet include the newly created chat (ordering / timing), **`realId` stays unset** until another **`getSessionList()`** runs (e.g. user opens the session drawer, navigates, or refreshes). If the user **stops immediately** or **leaves** before any successful resolution, the client may **never** obtain the backend **`chat_id`**.

This is a **race between chat creation** (e.g. `get_or_create_chat` on first message) **and list visibility**—the authoritative UUID is not returned at a **deterministic** point in the UI lifecycle.

**Code reference:** `console/src/pages/Chat/sessionApi/index.ts` — `resolveRealId`, `updateSession`, `removeSession` (when `realId` is missing, `deleteChat` is not called for timestamp-only ids).

### Environment

- **CoPaw version:** `1.0.2b1` (from `copaw.__version__` / `copaw --version` at report time)
- **Deployment:** Local install from source (`pip install -e .`), web console via Vite dev — issue is API/list timing, not specific to Docker vs pip
- **OS / Browser:** Windows 10; Chrome or Edge (current stable). **Not browser-specific** — any client that uses `GET /chats` to resolve `session_id` → UUID can hit the race

### Steps to reproduce

1. Open the **web console** and create a **new chat** (new session; local id is a timestamp).
2. Send a user message so the server creates/registers the chat (`get_or_create_chat`).
3. **Immediately** click **Stop**, or navigate away, **before** the session list has been refreshed such that **`resolveRealId`** can match `session_id` → UUID.
4. Optionally: try to **delete** the session while the UI still only knows the timestamp id — backend delete may not run if **`realId`** was never set (`removeSession` uses `deleteId = existing?.realId ?? (isLocalTimestamp ? null : …)`).

### Expected behavior

- After the server has created a chat row, the client should **reliably know** the backend **`chat_id`** without depending on **`GET /chats`** including that row on the **first** try.
- **Stopping early** or **deleting** should still behave consistently when a server-side chat exists.

### Actual behavior

- Resolution relies on **`GET /chats`**; a **single** list fetch may **omit** the new chat temporarily, so **`realId`** may remain **null** until another list load.
- **Fast stop** or **no second list fetch** can leave **no UUID** on the client; **delete** may not call **`DELETE /chats/{id}`** when only a timestamp id is known.

### Possible directions (for discussion)

1. Emit **`chat_id`** in the **first structured chunk/frame** of the console streaming response right after `get_or_create_chat`.
2. On **new session**, call **`POST /chats`** first and use the returned **`id`** as the canonical UUID, using the same **`session_id`** as subsequent stream requests (so `get_or_create_chat` finds the existing row).
3. Optionally add **`GET /chats` by `session_id`** (or similar) instead of relying on full list ordering/timing.

### Additional context

- Related: console session list, streaming lifecycle, `ChatManager.get_or_create_chat`.
- **Note:** Some deployments may retry list fetch locally; upstream `updateSession` uses **`getSessionList()`** once per path—wording above matches **single-fetch** race, not a built-in “poll N times” loop in core.

---

Thank you for considering this. Happy to help test a proposed fix.

aqilaziz:
## Summary
- add bounded retry when resolving temporary console session IDs to backend chat UUIDs
- keep URL resolution callback compatible with the existing Chat page while passing the real UUID first
- resolve a timestamp session before delete so backend chat deletion is not skipped when the first chat list is stale
- add regression coverage for stale list retry and delete resolution

## Tests
- `cd console; $env:NODE_OPTIONS='--max-old-space-size=4096'; npx vitest run src/pages/Chat/sessionApi/index.test.ts`
- `cd console; npx prettier --check src/pages/Chat/sessionApi/index.ts src/pages/Chat/sessionApi/index.test.ts`
- `cd console; npx eslint src/pages/Chat/sessionApi/index.ts src/pages/Chat/sessionApi/index.test.ts`
- `cd console; $env:NODE_OPTIONS='--max-old-space-size=4096'; npx tsc -b --noEmit`
- `git diff --check`
- `pre-commit run --files console/src/pages/Chat/sessionApi/index.ts console/src/pages/Chat/sessionApi/index.test.ts`

Closes #3254

### DIFF
--- a/console/src/pages/Chat/sessionApi/index.test.ts
+++ b/console/src/pages/Chat/sessionApi/index.test.ts
@@ -1,0 +1,74 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const apiMocks = vi.hoisted(() => ({
+  listChats: vi.fn(),
+  deleteChat: vi.fn(),
+}));
+
+vi.mock("../../../api", () => ({
+  default: apiMocks,
+}));
+
+vi.mock("../utils", () => ({
+  toDisplayUrl: (url: string) => url,
+}));
+
+const buildChat = (id: string, sessionId: string) => ({
+  id,
+  name: "Test chat",
+  session_id: sessionId,
+  user_id: "default",
+  channel: "console",
+  meta: {},
+  status: "idle",
+  created_at: null,
+  pinned: false,
+});
+
+const importSessionApi = async () => {
+  vi.resetModules();
+  const module = await import("./index");
+  return module.default;
+};
+
+describe("sessionApi real id resolution", () => {
+  beforeEach(() => {
+    apiMocks.listChats.mockReset();
+    apiMocks.deleteChat.mockReset();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("retries a stale chat list before resolving a temporary session id", async () => {
+    vi.useFakeTimers();
+    const tempId = "1710000000000";
+    apiMocks.listChats
+      .mockResolvedValueOnce([])
+      .mockResolvedValueOnce([buildChat("uuid-1", tempId)]);
+
+    const sessionApi = await importSessionApi();
+    const onResolved = vi.fn();
+    sessionApi.onSessionIdResolved = onResolved;
+
+    const resolved = sessionApi.getBackendIdForSession(tempId);
+    await vi.advanceTimersByTimeAsync(250);
+
+    await expect(resolved).resolves.toBe("uuid-1");
+    expect(apiMocks.listChats).toHaveBeenCalledTimes(2);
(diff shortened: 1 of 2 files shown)


xieyxclack:

PR #4537, opened 2026-05-19
feat(feishu): support group session shared mode

szetohoyan:
## Description

Simplify the Feishu (Lark) channel group session configuration by replacing `group_session_mode` (Literal["per_user", "shared"]) and `group_session_overrides` (Dict[str, ...]) with a single boolean flag `share_session_in_group`, following the same pattern used by the WeCom channel (#3948).

This addresses all three review comments from the initial implementation:

1. **Configuration granularity** — Replaced the over-engineered `group_session_mode` + `group_session_overrides` with a simple `share_session_in_group: bool` toggle. YAGNI — we can add per-group overrides later if there's actual demand.

2. **Missing console UI** — Added the corresponding Switch toggle in the Feishu channel drawer in the console, placed under the "Group Policy" (group_policy) setting.

3. **Literal["per_user", "shared"] type** — Replaced the two-option string literal with a boolean, as suggested.

Also updated the tooltip text with a more detailed example explaining the behavioral difference between shared and isolated sessions.

## Type of Change

- [ ] Bug fix
- [x] Feature enhancement
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Core
- [x] Console
- [x] Channels (Feishu)
- [ ] Skills
- [ ] CLI
- [ ] Documentation
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts

## Checklist

- [x] Run and pass pre-commit run --all-files
- [x] Run and pass relevant tests (pytest or as applicable)

## Testing

1. **share_session_in_group: true** — In a group chat, member A @bots, then member B @bots. Both see the same session context (shared).
2. **share_session_in_group: false** — Member B @bots starts a fresh context (independent session).
3. **Console UI** — Verify the toggle is visible in Settings > Channels > Feishu, under "Group Policy".
4. **Conflict resolution** — Branch is rebased on latest upstream/main with zero conflicts.

### pre-commit (on changed files only)
```
check yaml...........................................(no files to check)Skipped
```
### pytest (channel + contract tests)
```
tests/unit/channels/ tests/contract/channels/ -q --tb=short
1408 passed, 3 skipped, 51 warnings in 54.42s
```
All test failures in the full suite are pre-existing and unrelated to this change.


hongxicheng:
Thanks for the contribution! A few observations:

### Configuration granularity

I noticed you introduced both `group_session_mode` (global) and `group_session_overrides` (per chat_id). This feels a bit over-engineered for the initial implementation. For reference, the WeCom channel (#3948) handles the same feature with a single `share_session_in_group: bool` toggle — simple, clean, and covers the majority of use cases.

Would it make sense to start with just a single boolean flag (e.g. `share_session_in_group`) and add per-group overrides later if there's actual demand? YAGNI — we can always extend, but it's hard to simplify once it's shipped.

That said, if you do have a concrete use case where per-group control is necessary, I'm open to hearing it.

### Missing console UI

The new config fields (`group_session_mode` / `group_session_overrides`) don't have corresponding settings in the frontend console channel drawer. Users would have to edit config files or set env vars manually to use this feature, which is inconsistent with how other channel settings are exposed. Please add the UI toggle as well.

### Minor code note

The `Literal["per_user", "shared"]` type for a two-option choice is functionally equivalent to a boolean — another sign that a simple `bool` might be the better fit here.

PR #4701, opened 2026-05-26
feat(approval): add approve-all option to skip step-by-step confirmation

anThreeBody:
## Description

Add a **"Approve All"** button to the tool-guard approval card in the chat console. When a task requires multiple consecutive tool approvals, users previously had to click "Approve" for every single step while waiting at the screen. This change lets users unblock all pending approvals for the current session in one click — similar to the "Allow for this project" option in Claude Code.

**Related Issue:** N/A

**Security Considerations:** The `/approval/approve-all` endpoint validates the `session_id` against `root_session_id` on each pending record (same guard as the existing `/approval/approve`). Only approvals belonging to the caller's root session are resolved; cross-session bulk approval is not possible.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

1. Start a task that requires 3+ consecutive tool approvals (e.g. a multi-step shell script task with `SMART` approval level)
2. When the first approval card appears, click **"Approve All"**
3. Verify all pending approval cards clear immediately and the task runs to completion
4. Verify that individual **"Approve"** and **"Deny"** buttons still work as before on a separate run

```bash
pre-commit run --files \
  src/qwenpaw/app/approvals/service.py \
  src/qwenpaw/app/routers/approval.py \
  console/src/api/modules/commands.ts \
  console/src/components/ApprovalCard/ApprovalCard.tsx \
  console/src/components/ApprovalCard/ApprovalCard.module.less \
  console/src/pages/Chat/index.tsx \
  console/src/locales/en.json \
  console/src/locales/zh.json

check python ast .... Passed
check json .......... Passed
detect private key .. Passed
mypy ................ Passed
black ............... Passed
flake8 .............. Passed
pylint .............. Passed
prettier ............ Passed
```

## Additional Notes

- `onApproveAll` prop on `ApprovalCard` is **optional** — components that do not pass it render exactly as before (backward compatible)
- i18n keys (`approval.approveAll`, `approval.approvedAll`, `approval.approveAllFailed`) added to all 6 locale files: `en`, `zh`, `ja`, `ru`, `pt-BR`, `id`


gnipping:
Thanks for this PR. Here are my comments:

### 1) Back-end: `approve_all_by_root_session` bypasses `resolve_request` (likely breaks hooks/events)

In `src/qwenpaw/app/approvals/service.py`, the bulk implementation directly `pop()`s pending items and calls `pending.future.set_result(...)` instead of going through the existing lifecycle method `ApprovalService.resolve_request`.

This may bypass important side effects that are implemented via patching/wrapping `resolve_request` (e.g. `plugins/bundle/qwenpaw-pet/patch_approval.py` schedules pet events on `approval.approved/denied/timed_out`). With the current approach, “approve all” can unblock execution but may not emit the same events/audit/telemetry as step-by-step approvals.

Suggestion:
- Refactor the bulk path to resolve each request by calling `await self.resolve_request(request_id, ApprovalDecision.APPROVED)` (or factor out a shared internal bulk resolver that still triggers the same wrapped method/hook).
- If performance is a concern, collect IDs under lock, then resolve them outside the lock using the normal `resolve_request` path.

---

### 2) API contract + UI semantics: success is always `true`, even when `approved_count === 0`

`/approval/approve-all` in `src/qwenpaw/app/routers/approval.py` always returns `success=True` regardless of whether there are any pending approvals. The frontend (`console/src/pages/Chat/index.tsx` -> `handleApproveAll`) then unconditionally shows a success toast ("All pending tools approved").

This can be misleading for users (and makes debugging harder) when:
- the session_id is wrong / no root-session pending items exist, or
- there simply are no pending approvals at the moment of click.

Suggestion:
- Decide and document desired behavior when `approved_count === 0`.
  - Option A: return `success=false` with a clear message like “No pending approvals”.
  - Option B: keep `success=true` but change `message`/toast based on `approved_count` (e.g. show `info` instead of `success`).
- Use the returned `approved_count` in `handleApproveAll` to tailor the UI feedback.

anThreeBody:
Thanks for the thorough review @gnipping!

Both points addressed in the follow-up commit (`9adb233`):

**1) `approve_all_by_root_session` now goes through `resolve_request`**

Refactored to collect IDs under lock, then resolve each one via `await self.resolve_request(request_id, ApprovalDecision.APPROVED)` outside the lock — same path as single-step approvals, so all hooks/events (including `patch_approval.py`) are triggered correctly.

**2) `approved_count == 0` now returns `success=False`**

The endpoint returns `{"success": false, "message": "No pending approvals found for this session", "approved_count": 0}` when there is nothing to approve. The frontend now checks `approved_count` before clearing the UI and shows an `info` toast instead of a `success` toast in that case. Also added `approval.noApprovalsPending` i18n key to all 6 locale files.

anThreeBody:
Thank you @gnipping for the thorough review and valuable feedback — your comments were extremely helpful and prompted me to think more deeply about the implementation.

After further local testing, I found a design flaw in the current implementation: `approve_all` only approves **requests that are already pending at the time of the click**. In practice, however, multiple approval requests are often issued sequentially, which means "Approve All" ends up having the same effect as a single "Approve".

I plan to redesign this as a **task-level auto-approve mode**: once the user clicks "Approve All", all subsequent approval requests during the current task execution are automatically approved, until the task completes and the user sends the next input. This would genuinely reduce repetitive user interactions.

Additionally, your review comments regarding bypassing hooks/events and edge-case handling are well noted — I will apply the same care in the new PR to ensure correctness and robustness.

Closing this PR for now and will resubmit as a new PR. Looking forward to your continued guidance.

PR #4707, opened 2026-05-26
fix(tools): handle ToolResponse text blocks robustly

JasonBuildAI:
## Description
Fix runtime crash when accessing ToolResponse content blocks as attributes instead of dicts. Replace `content[0].get("text", "")` with a robust fallback (`hasattr(.text)` → `isinstance(dict)` → `str()`) in `_analyze_screen_activity` and `edit_file`, matching the existing pattern in `extract_content()`.

## Type of Change
- [x] Bug fix

## Component(s) Affected
- [x] Core

## Checklist

## Testing
Existing tests pass. Reproduced crash by calling `desktop_screenshot()` → `_analyze_screen_activity` with TextBlock content, confirmed fix resolves `AttributeError`.

```bash
pre-commit run --all-files
# passed

pytest
# passed
```

### DIFF
--- a/src/qwenpaw/agents/memory/proactive/proactive_utils.py
+++ b/src/qwenpaw/agents/memory/proactive/proactive_utils.py
@@ -251,7 +251,13 @@ async def _analyze_screen_activity(
 
         content = screenshot_result.content
         if isinstance(content, list) and len(content) > 0:
-            result_text = content[0].get("text", "")
+            first_block = content[0]
+            if hasattr(first_block, "text"):
+                result_text = str(getattr(first_block, "text", ""))
+            elif isinstance(first_block, dict):
+                result_text = str(first_block.get("text", ""))
+            else:
+                result_text = str(first_block)
         else:
             result_text = str(content)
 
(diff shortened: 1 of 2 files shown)


imrewce:
Though we did not confront with such vlm crash in testing, this pr seems really able to  enhance the robustness of the function against different response format from different VLM models. But before merging, we would like to collect the information of exactly what VLMs would lead to such exception for more cautious future development, plz kindly provide corresponding logs as evidence for us to reproduce the issue.

jinliyl:
Thanks for catching this bug — `TextBlock` objects indeed don't have `.get()`, so the crash is real.

However, the fix could be significantly simplified:

**1. `proactive_utils.py` — reuse the existing `extract_content()` in the same file**

There's already an `extract_content()` function defined at line 223 of this exact file that handles `TextBlock` vs dict vs fallback. The 6-line inline block can be replaced with:

```python
result_text = extract_content(content)
```

**2. `file_io.py` — the return type is fully controlled**

`write_file()` is defined in the same file and always returns `ToolResponse(content=[TextBlock(...)])`. Since the type is known, the defensive fallback chain is unnecessary — just use the attribute directly:

```python
write_text = write_response.content[0].text
```

**3. Incomplete coverage**

A grep shows there are other `.get("text")` call sites across the codebase that could hit the same issue if they ever receive a `TextBlock` instead of a dict. Worth auditing whether those are also affected.

In summary: the bug is valid, but I'd suggest simplifying the fix — reuse `extract_content()` where it exists, and use direct attribute access where the return type is known.

PR #4716, opened 2026-05-27
# feat(coding): per-hunk Keep/Undo + smart copy-to-chat in diff editor

rayrayraykk:
## Summary

- **Per-hunk Keep / Undo overlays** in the Coding-mode inline diff view, complementing the existing global *Keep all* / *Undo all* toolbar actions. Each hunk gets its own pair of buttons floating to the right of the hunk's top edge.
- **Smart copy-to-chat**: pasting editor-copied text into the Chat composer now auto-swaps the raw text for the `path:line[-line]` formatted version (with optional fenced code). Pasting *outside* the chat composer stays plain text — i.e. raw `Cmd/Ctrl+C` is unchanged for all other paste targets.

## Why

- Reviewing an Agent's multi-hunk edit was all-or-nothing. Users wanted to accept some hunks while reverting others without manual surgery.
- Cross-referencing code in chat ("look at this function") required either screenshotting or manually typing `foo.ts:42`. We now do it automatically, but only when the destination is the chat composer — so general copy/paste behavior is preserved.

### Per-hunk Keep / Undo

| File | Change |
|---|---|
| `console/src/pages/Coding/TabbedEditor.tsx` | New `applyKeepHunk` / `applyUndoHunk` pure helpers; `handleKeepHunk` / `handleUndoHunk` reducers; spacer view zones via `onDomNodeTop` to track pixel positions; React overlay JSX layered over the `<DiffEditor>`. |
| `console/src/pages/Coding/TabbedEditor.module.less` | New `.diffWrap` positioning context + `.hunkWidget` / `.hunkBtn` / `.hunkKeepBtn` / `.hunkUndoBtn` styles (incl. dark-mode variants). |
| `console/src/stores/codingTabsStore.ts` | New `updateDiffOriginal` action so a single-hunk Keep can bake the kept block into the baseline while leaving other hunks visible. |

### Smart copy-to-chat

| File | Change |
|---|---|
| `console/src/pages/Coding/lastEditorCopy.ts` (new) | Module-scoped cache: stores `{ text, formatted, ts }` for the most recent editor copy. |
| `console/src/pages/Coding/TabbedEditor.tsx` | Document-level capture-phase `copy` listener writes the cache when a selection is copied from the regular Editor *or* DiffEditor modified pane (`hasTextFocus()` confirms the source). Existing `Cmd+C` Monaco override removed — plain text reaches the system clipboard unchanged. |
| `console/src/pages/Chat/index.tsx` | New `useChatPasteFromEditor` hook installs a document-level capture-phase `paste` listener. When the target is a sender textarea AND the pasted text exactly matches the last editor copy AND the entry is fresh (<60s), it swaps in the formatted version. |

### Why React-overlay, not Monaco widgets

The per-hunk buttons went through several iterations:

1. **Monaco `ContentWidget` + React** — clickable, but anchored above the line so for top-of-file hunks the widget rendered off-screen / overlapped source above.
2. **Monaco view zones with React (`createRoot`) inside the zone DOM** — fixed visibility but clicks silently never fired.
3. **Monaco view zones with plain DOM + native `addEventListener`** — still not clickable.
4. **Monaco view zones with `suppressMouseDown: false`** — still not clickable.

Root cause (from `monaco-editor`'s `mouseHandler.js`): Monaco's editor-level mouse handler classifies any pointer event landing on a view-zone descendant as a `CONTENT_VIEW_ZONE` target and runs `focus()` + `mouseDownOperation.start()` + `preventDefault()`, which interferes with the button's mouseup → click sequence. The flag named `suppressMouseDown` only changes *whether Monaco handles it as a cursor placement*; it doesn't take Monaco's mouse path out of the way.

The reliable fix: render the buttons **outside Monaco's DOM entirely**. Spacer view zones still push code lines apart (so the buttons have a 22px gap to sit in), but the buttons themselves are React JSX siblings of `<DiffEditor>` inside a `position: relative` wrapper. Monaco tells us each spacer zone's current pixel-top via the `onDomNodeTop` callback (which fires on initial layout *and* scroll), and we drive the React overlay's `top` style from that. Clicks fire normally because no Monaco event handler is in the path.

### Why the `copy` cache + paste swap (not `clipboardData.setData`)

Writing `path:line[...]` directly to the system clipboard via `setData('text/plain', ...)` would hijack `Cmd+V` everywhere, including the user's terminal, browser address bar, Slack, etc. — which is exactly what we don't want (and what previously frustrated users).

Instead, we:
1. Let Monaco's native copy do its thing — plain text goes to the system clipboard.
2. Stash the same selection + its formatted twin in a tiny module cache.
3. At paste time, in the chat composer only, if `clipboardData.getData('text/plain') === cache.text`, we swap the formatted version into the textarea in-place.

This means: paste into terminal → plain text. Paste into VS Code → plain text. Paste into our chat composer → enriched with file/line context. No global side effects.

## Test plan

- [ ] Have an Agent edit a file with **multiple hunks**; verify each hunk shows its own Keep / Undo overlay to the right of the gap above the hunk.
- [ ] Click per-hunk **Keep** on one hunk — that hunk dismisses; other hunks remain visible.
- [ ] Click per-hunk **Undo** on one hunk — that hunk reverts on disk; other hunks remain visible.
- [ ] After resolving all hunks individually, diff state clears and editor returns to normal mode.
- [ ] Global **Keep all** / **Undo all** still works.
- [ ] Hunk at **line 1** is visible and clickable (no top clipping).
- [ ] Scroll the diff — overlays follow their hunks correctly.
- [ ] Select code, `Cmd+C`, paste into **terminal / browser URL bar** → plain text.
- [ ] Select code, `Cmd+C`, paste into **chat composer** → comes through as `path:line\n\`\`\`lang\n…\n\`\`\``.
- [ ] Same test but in **Coding mode's embedded chat panel** (right side of `/coding`).
- [ ] Same test from the **diff editor's modified pane** (during an Agent edit).
- [ ] Dark mode: per-hunk buttons readable and styled correctly.
- [ ] Lint: `npx eslint src/pages/Coding/TabbedEditor.tsx` clean (no new errors).
- [ ] Typecheck: `npx tsc -b` clean.

### DIFF
--- a/console/src/pages/Chat/index.tsx
+++ b/console/src/pages/Chat/index.tsx
@@ -65,6 +65,7 @@ import {
   type RuntimeLoadingBridgeApi,
 } from "./utils";
 import { openExternalLink } from "../../utils/openExternalLink";
+import { getLastEditorCopy } from "../Coding/lastEditorCopy";
 
 const CHAT_ATTACHMENT_MAX_MB = 10;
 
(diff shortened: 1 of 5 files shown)


zhaozhuang521 (approved):
LGTM

PR #4750, opened 2026-05-28
fix(cron): display schedule time and timezone in user's local timezone

anThreeBody:
## Description

The Cron Jobs list page displayed schedule times in the task's stored timezone
(defaulting to UTC) rather than the user's configured timezone, which was
confusing. This PR makes the schedule display consistent with the user's timezone.

**Changes:**
- **Execution time column (recurring tasks)**: For daily/weekly tasks, the
  cron hour/minute is converted from the task's stored timezone to the user's
  timezone before display. e.g. a task stored as `0 22 * * *` (UTC) now shows
  as `每天 06:00` for UTC+8 users.
- **Execution time column (once-type tasks)**: `run_at` is converted to the
  user's timezone for display (previously shown in UTC with no conversion).
  When the task timezone differs from the user's, hovering the cell shows the
  original task-timezone time (e.g. `2026-05-29 06:00 (UTC)`) for reference,
  instead of the raw ISO string shown before.
- **Timezone column**: Instead of showing the task's stored timezone string,
  the column now displays the user's timezone as a UTC offset (e.g. `UTC+8`).
- **New task default timezone**: Changed from hardcoded `UTC` to the browser's
  local timezone (`Intl.DateTimeFormat().resolvedOptions().timeZone`), so new
  tasks default to the user's timezone.
- **i18n** (en/zh): Updated `timezoneTooltip` wording to reflect the new default.

**Related Issue:** N/A

**Security Considerations:** None

## Type of Change

- [x] Bug fix

## Component(s) Affected

## Checklist

- [x] Ready for review

## Testing

1. Open the Cron Jobs page
2. Verify recurring tasks show times in your local timezone
   (e.g. `0 22 * * *` UTC → `每天 06:00` for UTC+8 users)
3. Verify the timezone column shows `UTC+8` instead of `UTC`
4. Create a new task — verify the timezone field defaults to your local timezone
5. For once-type tasks: verify the execution time is shown in your local
   timezone, and hovering shows the original UTC time for reference

## Additional Notes

`userTimezone` is sourced from the existing `/config/user-timezone` API.
If the user has not configured a timezone, it falls back to UTC — that is a
separate concern and not addressed here.

### DIFF
--- a/console/src/locales/en.json
+++ b/console/src/locales/en.json
@@ -893,7 +893,7 @@
     "cronExample": "Common examples: '0 9 * * *' = 9 AM daily | '*/30 * * * *' = every 30 min | '0 */2 * * *' = every 2 hours | '0 0 * * 0' = Sunday midnight",
     "cronHelper": "New to Cron expressions?",
     "cronHelperLink": "Use online generator",
-    "timezoneTooltip": "Timezone for the cron schedule. Default: UTC",
+    "timezoneTooltip": "Timezone for the cron schedule. Defaults to your system timezone.",
     "taskTypeTooltip": "Choose 'text' for simple message tasks, or 'agent' for complex agent workflows.",
     "textTooltip": "Simple message task: this is the actual message body. Required when task type is 'text'.",
     "requestInputTooltip": "Message content in JSON format. This is what the agent will receive and process. Required when task type is 'agent'.",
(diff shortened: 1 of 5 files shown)


lalaliat:
Hello, thank you for your contribution!

The behavior you mentioned is by design: it keeps the UI aligned with how APScheduler actually runs the job (using `schedule.timezone` + `cron`/`run_at`) and avoids ambiguity when users manage tasks across different timezones or when the configured user timezone differs from the browser's. The current design ensures that the calendar view displays times according to the user's timezone, while the list view retains the original creation-time settings to facilitate management.

Additionally, there are two issues with the changes in the current PR:

1. The timezone column in the current cron job list becomes meaningless.
2. There is an issue with the current implementation. Tests show that when creating a scheduled task for `2026-01-01 9:00 UTC`, the frontend still displays it as `2026-01-01 9:00 UTC+8`, with the hover tooltip showing `2026-01-01 1:00 UTC`.

Given these points, we will not merge this PR at this time. We truly appreciate your effort, and we look forward to your future ideas!

PR #4801, opened 2026-05-29
fix(pet): Auto-install missing dependencies to resolve startup failure on v1.1.9

### LINKED ISSUE #4783: [Bug]: v1.1.9 Qwenpaw Windows Desktop - QwenPaw Pet
## QwenPaw Version

Qwenpaw Windows Desktop v1.1.9 

## Description

when i try start desktop pet then the terminal looping open but the pet not open.

wangfei010313:
Description:

Problem
In QwenPaw Windows Desktop v1.1.9, the bundled environment is missing the pyside6-essentials package. This causes the QwenPaw Pet plugin to fail during initialization, resulting in the desktop service not starting (status remains ready: false).

Solution
I have implemented an automatic dependency check and installation mechanism within the plugin loader to ensure all required packages are present before the plugin attempts to import them.

Changes Made
Added _check_dependencies_satisfied(): Uses importlib.metadata to verify if every dependency listed in requirements.txt is installed and meets the version requirements.
Added _ensure_dependencies_installed(): An asynchronous wrapper that triggers the existing _install_requirements function if any missing dependencies are detected.
Updated load_plugin(): Integrated the dependency check before the module import phase. This ensures that for all bundled plugins, dependencies are automatically installed if missing, preventing startup crashes.

Closes #4783

### DIFF
--- a/src/qwenpaw/plugins/loader.py
+++ b/src/qwenpaw/plugins/loader.py
@@ -14,6 +14,9 @@ import threading
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
+from importlib.metadata import distributions as _all_distributions
+from packaging.requirements import Requirement
+
 from .architecture import PluginManifest, PluginRecord
 from .api import PluginApi
 from .registry import PluginRegistry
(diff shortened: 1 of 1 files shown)


Osier-Yi (approved):
LGTM

rayrayraykk (approved):

PR #4821, opened 2026-05-29
feat(feishu): add group session sharing mode

szetohoyan:
## Description

Add group session sharing control to the Feishu (Lark) channel. When `share_session_in_group` is enabled, all members in a group chat share the same session context; when disabled, each member has an independent session.

The implementation follows the same pattern used by the WeCom channel (#3948) — a single boolean flag `share_session_in_group` instead of the more complex approach in the previously closed PR #4537.

### Changes

1. **`share_session_in_group: bool`** — New config field (default `false`). When `true`, user_id is overridden to a group-level identifier so group members share one session context.

2. **Console UI** — Added a Switch toggle in the Feishu channel drawer, placed under "Group Policy" (`group_policy`), consistent with how other channel settings are exposed.

### Type of Change

- [ ] Bug fix
- [x] Feature enhancement
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

### Component(s) Affected

- [ ] Core
- [x] Console
- [x] Channels (Feishu)

### Checklist

- [x] pre-commit run --all-files (on changed files, all passed)
- [x] pytest channel + contract tests (1408 passed, 3 skipped)

### Testing

1. **share_session_in_group: true** — group members share one session context
2. **share_session_in_group: false** — each member gets an independent session
3. **Console UI** — verify toggle visible under Feishu channel drawer > Group Policy

### Local Verification

**pre-commit**
```
mypy........Passed
black.......Passed
flake8......Passed
prettier....Passed
```

**pytest** (channel + contract)
```
1408 passed, 3 skipped, 51 warnings in 54.42s
```


hongxicheng:
Hi! Thanks for the contribution 🎉 I've pushed a few simplifications to your branch:

1. Deduplicated i18n keys — removed the feishu-specific feishuShareSessionInGroup keys and reused a shared shareSessionInGroup key (now common across feishu, onebot, and wecom).

2. Moved UI switch placement — the toggle was in the shared renderAccessControlFields block (guarded by activeKey === "feishu"). Moved it into the feishu-specific renderBuiltinExtraFields section to match the pattern used by onebot/wecom.

3. Removed _should_share_session wrapper — inlined is_group and self.share_session_in_group directly at the call site since it's a one-liner and the chat_id param was unused.

4. Simplified user_id override — set meta["feishu_sender_id"] to "group" (same as wecom) instead of f"group:{short(chat_id)}". The session_id already encodes the group identity ({app_suffix}_{short(chat_id)}), so the user_id only needs to unify members — no need to redundantly embed chat_id again. Also dropped the native["user_id"] assignment since build_agent_request_from_native always prioritizes feishu_sender_id.

Feel free to pull the latest and test locally. Let me know if anything looks off!

szetohoyan:
Thanks for the cleanup! Just pulled and tested locally — all good. 👍

lalaliat (approved):
LGTM

Issue #4845, opened 2026-06-01
[Question]: WeWork channel lacks memory isolation — users can leak other users' chat history via prompt injection


linhuang0405:
## QwenPaw Version
Latest (current main branch)

## Description
The WeWork (企业微信) channel currently does not implement complete memory/session isolation between different users. A user can potentially access another user's chat history or conversation context through prompt injection techniques.

This is a **security concern** because:
1. Multiple users interacting with the same QwenPaw instance through WeWork may have their conversation contexts leaked to each other
2. Malicious users can craft prompts to extract conversation history belonging to other users
3. This violates the fundamental expectation that each user's conversation is isolated and private

**Related PR(s):** N/A

**Security considerations:** This is a multi-tenant isolation vulnerability. Any production deployment serving multiple WeWork users is affected.

## Component(s) Affected
- [x] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [x] Core / Backend (app, agents, config, providers, utils, local_models)

## Steps to Reproduce
1. Deploy QwenPaw with WeWork channel enabled
2. User A sends a message with sensitive/contextual information
3. User B sends a prompt injection message designed to extract prior conversation context
4. User B receives content from User A's conversation

## Actual vs Expected
- **Actual:** Conversation context/memory may be shared or leakable across different WeWork users
- **Expected:** Each WeWork user should have a completely isolated conversation session with no cross-user data leakage

## Additional Notes
- This issue should be treated with higher priority as it involves user data privacy
- The fix should ensure that session identifiers (user ID, conversation ID, etc.) are properly scoped per WeWork user
- Memory stores should be keyed by a unique session identifier that cannot be manipulated through user input

hongxicheng:
Hi @linhuang0405, thanks for raising this concern.

QwenPaw is designed as a personal AI assistant and does not natively support multi-tenant isolation. Each instance is intended to serve a single user or a trusted group of users, rather than being exposed as a shared service to untrusted parties.

That said, if you'd like to restrict who can interact with your QwenPaw instance via the WeCom channel, you can use the built-in Access Control feature — it supports whitelist/blacklist management per channel, so you can ensure only authorized users have access. You can manage this through the console UI or the API.

If you have ideas for improving session isolation or multi-tenancy support, contributions are absolutely welcome! 

Closing this issue.

PR #4850, opened 2026-06-01
fix(wecom): resolve session_id from authenticated sender_id

Libres-coder:
### Summary

Fixes a defense-in-depth weakness in the WeCom channel where `build_agent_request_from_native` trusted `payload.get("session_id")` without validating it against the authenticated `sender_id`.

### Problem(#4845)

`build_agent_request_from_native` previously used `payload.get("session_id")` as the session identifier, only falling back to `resolve_session_id(sender_id, meta)` when absent from the payload. While the normal flow is safe — payloads are built internally and `sender_id` comes from the WeCom WebSocket frame's `body.from.userid`, which cannot be spoofed by users — this pattern does not follow the defense-in-depth principle of always deriving session identity from authenticated sources.

### Fix

Changed session_id resolution to always call `self.resolve_session_id(sender_id, meta)`, ignoring any `session_id` already present in the payload. This is a minimal change: **3 lines removed, 1 line added**.

```python
# Before:
session_id = payload.get("session_id") or self.resolve_session_id(sender_id, meta)

# After:
session_id = self.resolve_session_id(sender_id, meta)
```

### Tests

Added 5 regression tests (`TestWecomChannelSessionIsolation`):

| Test | Description |
|------|-------------|
| `test_different_users_get_different_session_ids` | Different users produce different session IDs |
| `test_session_id_derived_from_sender_id_not_payload` | Payload `session_id` is ignored; value derived from `sender_id` |
| `test_session_id_ignores_spoofed_payload_value` | Spoofed `session_id` targeting another user is rejected |
| `test_group_session_isolation` | Different group chats have isolated sessions |
| `test_group_and_dm_sessions_isolated` | DM and group sessions are isolated for the same user |

All 39 existing tests continue to pass with zero regressions.

### Risk

**Low.** In the normal flow, `sender_id` and `meta` are always present, and `resolve_session_id` returns the same value that was previously stored in the payload (since the payload's `session_id` was originally produced by `resolve_session_id` in `_on_message`).

### DIFF
--- a/src/qwenpaw/app/channels/wecom/channel.py
+++ b/src/qwenpaw/app/channels/wecom/channel.py
@@ -355,10 +355,7 @@ class WecomChannel(BaseChannel):
         sender_id = payload.get("sender_id") or ""
         content_parts = payload.get("content_parts") or []
         meta = payload.get("meta") or {}
-        session_id = payload.get("session_id") or self.resolve_session_id(
-            sender_id,
-            meta,
-        )
+        session_id = self.resolve_session_id(sender_id, meta)
         user_id = payload["user_id"] if "user_id" in payload else sender_id
         request = self.build_agent_request_from_user_content(
             channel_id=channel_id,
--- a/tests/unit/channels/test_wecom.py
+++ b/tests/unit/channels/test_wecom.py
@@ -1508,3 +1508,122 @@ class TestWecomChannelEdgeCases:
         # Should detect as voice (AMR format)
         call_args = wecom_channel._upload_media.call_args
         assert call_args[0][1] in ["voice", "file"]
+
+
+# =============================================================================
+# P0: Session Isolation Tests (Security)
+# =============================================================================
+
+
+class TestWecomChannelSessionIsolation:
+    """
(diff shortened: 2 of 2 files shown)


hongxicheng:
Thanks for the patch and the tests.

I don't think this PR is fixing a real issue. The change rewrites session_id to always come from resolve_session_id(sender_id, meta) instead of trusting payload["session_id"], framed as defense-in-depth against a spoofed payload. But in the actual runtime flow payload is constructed by the channel itself inside _on_message from the authenticated WebSocket frame — it never crosses a trust boundary, so the "spoofed payload" scenario isn't reachable.

More importantly, session isolation in QwenPaw is keyed by the (user_id, session_id) pair, not by session_id alone (SafeJSONSession._get_save_path persists state at {user_id}_{session_id}.json). user_id is already derived from the authenticated sender_id, so even if session_id were somehow swapped, history would still be stored and looked up under the original sender's user_id — no cross-user leakage can happen through this path.

Given that, I'd lean toward closing this without merging. Thanks again for the careful read of the channel code.

PR #4860, opened 2026-06-01
Fix: Clean up stale Windows skill dirs (#4839)

### LINKED ISSUE #4839: [Bug] Windows: pip upgrade leaves ~prefixed stale builtin skill dirs, causing ghost skills in skill pool
## Bug Description

On **Windows**, after upgrading QwenPaw via `pip install --upgrade qwenpaw`, stale copies of builtin skill directories remain in the package with `~`-prefixed names (e.g., `~ron-en`, `~-on-en`, `~hat_with_agent-zh`). These ghost directories contain old-version `SKILL.md` files and are picked up by the skill scanner, appearing as invalid entries in the **"Update Builtin Skills"** UI in the skill pool.

## Environment

- **OS:** Windows 11 (AMD64)
- **Python:** 3.12
- **QwenPaw:** 1.1.9 (upgraded multiple times via pip)
- **pip:** 26.1
- **Install method:** `pip install` into a virtualenv

## Steps to Reproduce

1. Install QwenPaw on Windows via pip.
2. Upgrade QwenPaw one or more times via `pip install --upgrade qwenpaw`.
3. Open the QwenPaw console -> Skill Pool -> "Update Builtin Skills".
4. Observe `~`-prefixed ghost skills (e.g., `~ron`, `~-on`, `~hat_with_agent`) alongside the real builtin skills.

### Stale dirs found in `site-packages/qwenpaw/agents/skills/`

```
~-annel_message-en        ~-at_with_agent-en        ~-idance-en
~-ke_plan-en              ~-le_reader-en            ~-lti_agent_collaboration-en
~-malaya-en               ~-ngtalk_channel-en      ~-on-en
~-owser_cdp-en            ~-owser_visible-en       ~-ws-en
~-_source_index-en        ~ake_plan-en              ~a_source_index-en
~ews-en                   ~hannel_message-en       ~hat_with_agent-en
~ile_reader-en            ~imalaya-en               ~ingtalk_channel-en
~ron-en                   ~rowser_cdp-en            ~rowser_visible-en
~uidance-en               ~ulti_agent_collaboration-en
... (and their -zh counterparts, plus ~-pycache__ and ~_pycache__)
```

Total: **54 ghost directories** alongside **37 real skill directories**.

### Version comparison confirms stale copies

| Ghost dir | Version | Real dir | Version |
|---|---|---|---|
| `~ron-en` | v1.4 | `cron-en` | v1.6 |
| `~-on-en` | v1.5 | `cron-en` | v1.6 |
| `~uidance-zh` | v1.2 | `guidance-zh` | v1.3 |
| `~-idance-zh` | v1.2 | `guidance-zh` | v1.3 |
| `~-_source_index-zh` | v1.2 | `QA_source_index-zh` | v1.3 |

### mtime timeline confirms each pip upgrade creates a new batch

| mtime | Directory | Event |
|---|---|---|
| 2026-04-30 20:33 | `~ron-en` (v1.4), `~hannel_message-zh` | First upgrade - old v1.4 not deleted |
| 2026-05-17 20:14 | `~-on-en` (v1.5), `~-annel_message-zh` | Second upgrade - old v1.5 not deleted |
| 2026-05-31 20:07 | `cron-en` (v1.6), `channel_message-zh` | Current version |

## Root Cause Analysis

Two factors combine:

1. **Windows file replacement behavior during pip upgrade:** When pip replaces files inside `site-packages` during an upgrade, Windows (possibly due to file locking, antivirus scanning, or OneDrive sync) fails to fully delete old directories. Instead, they are left behind with `~`-prefixed truncated names (losing 1-2 leading characters).

2. **QwenPaw skill scanner lacks filtering:** In `registry.py`, both `_iter_packaged_builtin_dirs()` (line ~217) and `reconcile_pool_manifest()` scan directories with only:
   ```python
   if path.is_dir() and (path / "SKILL.md").exists():
   ```
   There is **no filter for `~`-prefixed directories**. Notably, the ZIP upload logic in `store.py` (line ~846) already has `_is_hidden()` filtering, but the builtin and pool scanners do not.

## Suggested Fix

1. **Add `~` prefix filtering** to `_iter_packaged_builtin_dirs()` and all `iterdir()`-based skill scanning logic:
   ```python
   def _iter_packaged_builtin_dirs() -> Iterator[Path]:
       builtin_dir = get_builtin_skills_dir()
       if not builtin_dir.exists():
           return
       for skill_dir in sorted(builtin_dir.iterdir()):
           if skill_dir.name.startswith('~'):
               continue  # skip stale/truncated directories
           if skill_dir.is_dir() and (skill_dir / "SKILL.md").exists():
               yield skill_dir
   ```

2. **Similarly filter** the `discovered` dict in `reconcile_pool_manifest()` (line ~895) and `reconcile_workspace_manifest()` (line ~999).

3. **Consider cleaning up stale dirs** at startup: detect and remove `~`-prefixed directories from the builtin skills directory.

## Workaround

Manually delete the ghost directories:
```cmd
for /d %d in ("%VENV%\Lib\site-packages\qwenpaw\agents\skills\~*") do rd /s /q "%d"
```

## Note

This is Windows-specific. Linux and macOS do not exhibit this behavior during pip upgrades.

wangfei010313:
## Summary

**Fixes #4839** — Cleans up `~`-prefixed ghost skill directories left behind by Windows `pip upgrade`.

Complements PR #4851 (UI filtering) by **actively removing** these stale files from disk at startup, while retaining defensive filtering as a safety net.

## Problem

On Windows, `pip install --upgrade` often fails to fully delete old skill directories. They remain as truncated `~`-prefixed folders (e.g., `~ron-en`). 
These "ghost skills" are detected by the scanner, polluting the "Update Builtin Skills" UI with invalid entries.

### 1. Active Cleanup (`cleanup_stale_skill_dirs`)
Executed before the first skill scan. Implements a two-tier safety strategy:

| Condition | Action | Rationale |
| :--- | :--- | :--- |
| `~` prefix **+** contains `SKILL.md` | **Delete**  | Confirmed stale skill copy (safe to remove). |
| `~` prefix **without** `SKILL.md` | **Log Warning** | Suspicious dir; might be user data (preserved). |
| Deletion fails (locked by AV, etc.) | **Log Warning** | Defensive fallback; never crashes the app. |

### Shift-Left Filtering
Added `not path.name.startswith("~")` guard clauses to:
- `_iter_packaged_builtin_dirs()` — Prevents loading stale builtins.
- `reconcile_pool_manifest()` & `reconcile_workspace_manifest()` — Defensive filtering for discovery.

> **Note:** The filtering acts as a "seatbelt" in case cleanup fails due to file locks.

## Changes

| File | Description |
| :--- | :--- |
| `src/qwenpaw/agents/skill_system/registry.py` | Added cleanup logic + filtering in 3 scan functions. |
| `tests/unit/skill_system/test_stale_cleanup.py` | **New** — Comprehensive unit tests (7 cases). |

## Testing
- ✅ 7 unit tests: cleanup success, normal dir preservation, permission errors, edge cases.
- ✅ All linters pass (mypy, black, pylint).

### DIFF
--- a/src/qwenpaw/agents/skill_system/registry.py
+++ b/src/qwenpaw/agents/skill_system/registry.py
@@ -210,12 +210,74 @@ def _select_builtin_variant(
     )
 
 
+def cleanup_stale_skill_dirs(target_dir: Path | None = None) -> None:
+    """Remove stale ``~``-prefixed directories left by pip upgrades on Windows.
+
+    On Windows, ``pip install --upgrade`` sometimes fails to fully delete
+    old skill directories, leaving behind truncated copies with ``~``-prefixed
+    names (e.g. ``~ron-en`` instead of ``cron-en``).  These ghost directories
+    pollute the skill pool UI (#4839).
+
+    Safety strategy:
+      - **Whitelist check**: Only directories that *both* start with ``~``
+        *and* contain a ``SKILL.md`` file are deleted (confirmed to be stale
+        skill copies, not user data).
+      - **Non-whitelisted**: If a ``~``-prefixed directory does *not* contain
+        ``SKILL.md``, a warning is logged but the directory is left intact.
+      - **Error tolerance**: Deletion failures (e.g. antivirus locks) are
+        logged as warnings and never crash the application.
+    """
+    import shutil
+
+    skill_dir = target_dir or get_builtin_skills_dir()
+    if not skill_dir.exists():
+        return
+
+    for path in sorted(skill_dir.iterdir()):
+        if not path.is_dir() or not path.name.startswith("~"):
+            continue
+
+        # Strategy A: whitelist — only delete if it looks like a skill dir.
+        if (path / "SKILL.md").exists():
+            try:
+                shutil.rmtree(path)
+                logger.info(
+                    "Removed stale skill directory: %s",
+                    path.name,
+                )
+            except OSError as exc:
+                logger.warning(
+                    "Failed to remove stale skill directory '%s': %s. "
+                    "You may delete it manually.",
+                    path,
+                    exc,
+                )
+        else:
+            # Strategy B: not a confirmed skill dir — warn only.
+            logger.warning(
+                "Detected possible stale directory '%s' in skills folder. "
+                "It does not contain SKILL.md so it was not auto-deleted. "
+                "Please remove it manually if it is not needed.",
+                path,
+            )
+
+
 def _iter_packaged_builtin_dirs() -> Iterator[Path]:
-    """Yield packaged builtin skill directories in stable order."""
+    """Yield packaged builtin skill directories in stable order.
+
+    Skips ``~``-prefixed directories (stale pip upgrade remnants on
+    Windows, see #4839) and runs cleanup on first scan.
+    """
     builtin_dir = get_builtin_skills_dir()
     if not builtin_dir.exists():
         return
+
+    # Best-effort cleanup of stale dirs before scanning.
+    cleanup_stale_skill_dirs(builtin_dir)
+
     for skill_dir in sorted(builtin_dir.iterdir()):
+        if skill_dir.name.startswith("~"):
+            continue
         if skill_dir.is_dir() and (skill_dir / "SKILL.md").exists():
             yield skill_dir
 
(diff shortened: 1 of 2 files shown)


Leirunlin:
Hi! Thanks for the PR. I have the following suggestions:
* The test here doesn't seem necessary.
* I understand the motivation to help clean up ghost skills, but 1) the number of users currently affected by this issue is very small and quite corner-case, so I don't think it's worth introducing this fix for all users; 2) there may be backward compatibility issues. Some users in the past might have intentionally included certain `~` skills, and removing them could cause unpredictable behavior.

I think a simpler approach would be to provide a cleanup method in the original issue for users who need it. 

Further discussion is welcome.

wangfei010313:
Thanks for the detailed feedback!
I completely agree with your points regarding backward compatibility and the risk of unintended data deletion for edge cases. It is indeed safer to avoid introducing automatic cleanup logic in the core registry given the small number of affected users.
I will close this PR and withdraw the changes. Thanks again for pointing out the potential risks!

PR #4881, opened 2026-06-01
feat(providers): add MiniMax M3 to MiniMax built-in models (default)

octo-patch:
## What

Add the new `MiniMax-M3` flagship model to the built-in `MINIMAX_MODELS` list shared by `PROVIDER_MINIMAX` (International, `https://api.minimax.io/anthropic`) and `PROVIDER_MINIMAX_CN` (China, `https://api.minimaxi.com/anthropic`). M3 is placed first in the list so it becomes the default model when a user selects the MiniMax provider without specifying one. The corresponding `ExpectedCapability` baselines in `capability_baseline.py` are updated to match.

The previous-generation `MiniMax-M2.5` / `MiniMax-M2.5-highspeed` entries are removed; `MiniMax-M2.7` / `MiniMax-M2.7-highspeed` are kept so users still on that generation are not forced to migrate immediately.

## Why

MiniMax released M3 as the new flagship Anthropic-compatible model (512K context window, up to 128K output, with image-input support). Surfacing it as a built-in option (and the new default) lets users pick it from the model dropdown without manually editing the provider config. Other built-in providers (e.g. Gemini, DeepSeek, Kimi) already keep their lists in sync with the latest publicly available models in the same file, so this PR follows the existing convention.

Reference: <https://www.minimax.io/platform/document/announcement>

## Scope

- Only the `minimax` and `minimax-cn` providers are touched.
- Other providers that happen to expose MiniMax models as part of their own offering (e.g. OpenCode, Aliyun coding/token plans, Volcengine coding plan) are left untouched — their model lists reflect what those upstreams actually serve, not what MiniMax ships directly.
- Base URLs, \`freeze_url\`, \`support_connection_check\`, and TTS configuration are not changed.

## Changes

- \`src/qwenpaw/providers/provider_manager.py\` — replace the M2.5 entries in \`MINIMAX_MODELS\` with a single M3 entry (placed first → new default), keep M2.7 + M2.7-highspeed.
- \`src/qwenpaw/providers/capability_baseline.py\` — update the \`minimax\` and \`minimax-cn\` expected-capability registrations to match (M3 + M2.7 + M2.7-highspeed).

Final list: \`MiniMax-M3\` (default), \`MiniMax-M2.7\`, \`MiniMax-M2.7-highspeed\`. Diff is ~4 lines per file, no other files touched.

### DIFF
--- a/src/qwenpaw/providers/capability_baseline.py
+++ b/src/qwenpaw/providers/capability_baseline.py
@@ -634,8 +634,7 @@ class ExpectedCapabilityRegistry:
         # ---------------------------------------------------------------
         _mm_doc = "https://www.minimax.io/platform/document/announcement"
         for mid in (
-            "MiniMax-M2.5",
-            "MiniMax-M2.5-highspeed",
+            "MiniMax-M3",
             "MiniMax-M2.7",
             "MiniMax-M2.7-highspeed",
         ):
@@ -655,8 +654,7 @@ class ExpectedCapabilityRegistry:
         # ---------------------------------------------------------------
         _mm_cn_doc = "https://platform.minimaxi.com/document/announcement"
         for mid in (
-            "MiniMax-M2.5",
-            "MiniMax-M2.5-highspeed",
+            "MiniMax-M3",
             "MiniMax-M2.7",
             "MiniMax-M2.7-highspeed",
         ):
--- a/src/qwenpaw/providers/provider_manager.py
+++ b/src/qwenpaw/providers/provider_manager.py
@@ -436,15 +436,8 @@ AZURE_OPENAI_MODELS: List[ModelInfo] = [
 
 MINIMAX_MODELS: List[ModelInfo] = [
     ModelInfo(
-        id="MiniMax-M2.5",
-        name="MiniMax M2.5",
-        supports_image=False,
-        supports_video=False,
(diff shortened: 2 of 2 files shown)


xieyxclack:
Why remove M2.5?

octo-patch:
Good question — M2.5 is an older generation, so the PR replaces it with M3 (new flagship) while keeping M2.7 around for users still on that line. The intent was to keep the built-in list scoped to the current + previous generation, mirroring how the other providers in the same file are maintained.

That said, removing M2.5 isn't strictly necessary — if you'd prefer to retain it for backward compatibility with existing user configs, happy to add the M2.5 / M2.5-highspeed entries back in. Just let me know which direction you'd like.

wangfei010313:
This PR is no longer needed because its core functionality has already been implemented in the current main branch through subsequent provider catalog updates.
MiniMax-M3 is now available as the recommended default model for both the international and China MiniMax providers. The current implementation also includes updated context-window metadata and test coverage.
The current catalog intentionally retains older MiniMax models as legacy options for backward compatibility

lalaliat:
Hello, thank you for your contribution! This feature has been achieved in the main branch, so we are closing this as a duplicate. We look forward to your future ideas. Thanks again!

PR #4902, opened 2026-06-02
feat(manage_prd): add built-in PRD CRUD tool with frontend renderer

Saint-Yin:
Replace the plugin-based manage_prd tool with a built-in tool that
  provides create/add/update/delete/mark_passed operations on prd.json.
  Add a frontend ManagePRDRenderer component with i18n support, loading/
  error states, and interactive story table. Add /prd API endpoint for
  frontend to fetch PRD data. Clean up unused plugin hooks and constants.

## Description
Add a built-in manage_prd tool (src/qwenpaw/agents/tools/manage_prd.py) for CRUD operations on prd.json in Mission Mode, replacing the old plugin-based tool.

  - New /prd API endpoint (src/qwenpaw/app/routers/prd.py) with optional timestamp query param for historical snapshot retrieval
  - New PRDViewer frontend component (console/src/components/PRDViewer/) with i18n support, loading/error states, and interactive story table
  - Removed old plugins/bundle/cloudpaw/tools/manage_prd.py
  - CloudPaw plugin hooks/constants updated accordingly
  - CloudPaw plugin UI (ui/src/index.ts) migrated from hardcoded dark mode colors to Ant Design theme.useToken()

**Related Issue:** N/A (new feature)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

- [x] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [x] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [x] Contract test implements `create_instance()` with proper channel initialization
- [x] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)

### UI Verification

  **Method 1: Create a Mission task via the UI (recommended)**
  1. After launching the app, enter a task description in the chat (e.g., "build me a personal homepage")
  2. The agent enters Mission Mode, Phase 1, and calls `manage_prd(operation="create", ...)` to create the PRD
  3. Verify the frontend automatically renders the PRD table with columns: status, ID, title, priority, description, acceptanceCriteria
  4. Verify the loading state (spinner while fetching) and error state (error message when the API fails)
  5. Verify dark/light theme switching correctly applies theme tokens to the table

  **Method 2: Use mock data to create a PRD**
  - Call the `/prd` API directly or manually construct a PRD JSON to verify rendering

  ### Backend Verification
  - `pre-commit run --file src/qwenpaw/agents/tools/manage_prd.py` passes
  - `pre-commit run --file src/qwenpaw/app/routers/prd.py` passes

  ### Frontend Verification
  - `cd console && npm run build` passes
  - `cd console && npm run format:check` passes

```bash
./scripts/check-channels.sh --changed
========================================
CoPaw Channel Pre-Commit Check
========================================

Detecting changed channels...
✅ No channel changes detected

pre-commit run --all-files
# paste summary result
pytest

=========================================================================================================== 3145 passed, 3 skipped, 50 warnings in 303.88s (0:05:03) ============================================================================================================
# paste summary result
```

## Additional Notes

[Optional: any other context]
image
image


xuanrui-L:
Suggest modifying the Mission mode prompt: avoid still using `write_file` to update the PRD.

Saint-Yin:
> Suggest modifying the Mission mode prompt: avoid still using `write_file` to update the PRD.

This has been addressed in this PR:

  - `src/qwenpaw/agents/mission/prompts.py` — added Step 0c with complete `manage_prd` examples for all operations (create/add/update/delete/mark_passed)
  - `mission_runner.py` — updated `_PRD_FIX_PROMPT` to guide agents to use `manage_prd` instead of `write_file`
  - Master prompt Rule2 — "PRD operations ONLY via `manage_prd`"

  Also verified with 8B model, screenshot attached:
image

@xuanrui-L

xuanrui-L:
LGTM

xuanrui-L (approved):

xuanrui-L:
pls resolve conflicts

rayrayraykk (changes requested):
**Commits:**
- `b05ae5cb` feat(manage_prd): add built-in PRD CRUD tool with frontend renderer
- `775085a2` fix(mission): improve manage_prd prompt guidance and sort PRDViewer by id

**Scope:** Moves `manage_prd` from the CloudPaw plugin into the core codebase (built-in tool + built-in frontend renderer), adds a `/prd` API endpoint, and updates mission prompts to mandate `manage_prd` usage.

---

## Summary of Changes

| Area | Files | Description |
|------|-------|-------------|
| Backend tool | `src/qwenpaw/agents/tools/manage_prd.py` (new, 602 lines) | Core CRUD tool for prd.json (create/add/update/delete/mark_passed) with snapshot support |
| API router | `src/qwenpaw/app/routers/prd.py` (new, 42 lines) | GET `/prd` endpoint to read prd.json or historical snapshots |
| Frontend | `console/src/components/PRDViewer/` (new, 383+107 lines) | React component with Ant Design table for PRD rendering |
| Config | `src/qwenpaw/config/config.py` | Register `manage_prd` as built-in tool |
| Agent wiring | `react_agent.py`, `tools/__init__.py` | Export and register `manage_prd` |
| Prompts | `mission_runner.py`, `prompts.py` | Update prompts to enforce `manage_prd` tool usage |
| Plugin removal | `plugins/bundle/cloudpaw/` (multiple files) | Remove `manage_prd` from plugin (tools, hooks, constants, UI, docs) |
| App | `_app.py` | Add no-cache headers to console index |
| Console entry | `main.tsx` | Register `PRDViewer` as built-in tool renderer |

---

## Critical Issues

### 1. [BUG] All handler coroutines are eagerly awaited in `handler_map` construction

**File:** `src/qwenpaw/agents/tools/manage_prd.py:608-624`

```python
handler_map: dict[str, Any] = {
    "add": _handle_add(prd_path, prd, prd_stories, story),
    "update": _handle_update(...),
    "delete": _handle_delete(prd_path, prd, prd_stories, story_ids),
    "mark_passed": _handle_mark_passed(...),
}
handler = handler_map.get(op)
...
return await handler
```

All four `async` handler functions are **called immediately** when building the dict (creating coroutines), not just referenced. This means:
- All 4 coroutines are created regardless of which operation is requested.
- The 3 unused coroutines are never awaited, which triggers `RuntimeWarning: coroutine '...' was never awaited` and is a resource leak.
- If any unused handler has a side effect during coroutine creation (unlikely here but fragile), it would execute.

**Fix:** Use lambdas or only call the matching handler:
```python
if op == "add":
    return await _handle_add(prd_path, prd, prd_stories, story)
elif op == "update":
    return await _handle_update(...)
...
```

### 2. [BUG] Unbound variable `prd` when `prd_path` doesn't exist

**File:** `src/qwenpaw/agents/tools/manage_prd.py:590-602`

```python
if not prd_path.exists():
    msg = f"prd.json not found: {prd_path}"
else:
    prd, err = _load_prd(prd_path)
    if prd is None:
        msg = err
    else:
        msg = ""

if msg:
    return _error_response(msg)

assert prd is not None  # prd may be unbound if prd_path didn't exist
```

If `prd_path.exists()` is False, `msg` is set to a non-empty string, so `_error_response` is returned. This is technically safe because the early return prevents reaching the `assert`. However, **static analysis tools (pyright, mypy) will flag `prd` as potentially unbound**, and the logic is fragile — a future editor could accidentally break the invariant.

**Suggestion:** Move the `assert` and subsequent code inside an explicit `else` block, or restructure:
```python
if not prd_path.exists():
    return _error_response(f"prd.json not found: {prd_path}")

prd, err = _load_prd(prd_path)
if prd is None:
    return _error_response(err)
```

### 3. [SECURITY] Path traversal in `/prd` API endpoint

**File:** `src/qwenpaw/app/routers/prd.py:19-30`

```python
base = Path(loop_dir).expanduser().resolve()
prd_path = base / "prd.json"

if timestamp:
    snap_path = base / "snapshots" / f"{timestamp}.json"
```

- `loop_dir` is user-supplied and passed directly to `Path()`. While `.resolve()` canonicalizes it, there's **no validation that the resolved path is within an allowed directory** (e.g., the workspace or a mission loop dir). An attacker could read any `prd.json` on the filesystem.
- `timestamp` is user-supplied and inserted into a filename. Although `.resolve()` would canonicalize `../` sequences, a timestamp like `../../etc/passwd` would resolve to `base/snapshots/../../etc/passwd.json` which is `base/etc/passwd.json` — not exploitable for arbitrary file read, but still unexpected. A regex check like `r'^\d{8}-\d{6}$'` would be safer.

**Recommendation:** Validate that `loop_dir` is within an expected base directory, and validate `timestamp` format.

---

## Major Issues

### 4. [DESIGN] `import re` inside function body

**File:** `src/qwenpaw/agents/tools/manage_prd.py:79`

```python
def _to_kebab_case(text: str) -> str:
    import re  # <-- should be at module top
    ...
```

Per the project's CLAUDE.md: *"Python开发项目内要用相对引用，并且尽量放在文件开头部分，不要在代码中间import"*. The `re` import should be at the top of the file.

### 5. [DESIGN] `_ok_response` default mutable argument

**File:** `src/qwenpaw/agents/tools/manage_prd.py:176`

```python
def _ok_response(message: str, data: dict = None) -> ToolResponse:
```

Using `None` as default for `data` is fine here (it's checked with `if data:`), but the same pattern in the old plugin code used `data: dict = None` too. This is acceptable but worth noting that `data` is never mutated in-place — the pattern is safe.

### 6. [DESIGN] Missing `__all__` entry consideration for `manage_prd` module

The tool is added to `__all__` in `tools/__init__.py` (line 64), which is correct. However, `manage_prd` is also unconditionally registered in `react_agent.py` (line 304) alongside agent management tools like `submit_to_agent`, `check_agent_task`, `spawn_subagent`. This means **`manage_prd` is available to ALL agents, not just mission mode agents**. Is this intentional? Regular coding agents don't need PRD management. Consider gating it behind mission mode or a config check.

### 7. [FRONTEND] Excessive `as any` type casts in PRDViewer

**File:** `console/src/components/PRDViewer/index.tsx`

The component uses `(data as any)?.content?.[0]?.data` and similar casts extensively (15+ occurrences). While the `data: unknown` prop type is intentionally loose (tool data shape varies), defining even a partial interface would improve readability and catch typos:

```typescript
interface ToolMessage {
  status?: string;
  content?: Array<{ data?: { arguments?: string; output?: unknown } }>;
}
```

---

## Minor Issues

### 8. Sorting difference between old plugin and new code

- **Old plugin (removed):** sorts by `priority` — `(a.priority || 99) - (b.priority || 99)`
- **New frontend (PRDViewer):** sorts by `id` — `(a.id as string).localeCompare(b.id as string, undefined, { numeric: true })`

The commit message says "sort PRDViewer by id" which was intentional, but this is a **behavioral change** from the plugin version. Worth confirming this is the desired UX.

### 9. Snapshot directory grows unbounded

**File:** `src/qwenpaw/agents/tools/manage_prd.py:137-146`

Every `_save_prd` call creates a new snapshot in `loop_dir/snapshots/`. For long-running missions with many operations, this directory can accumulate many files. Consider adding a cleanup policy (e.g., keep last N snapshots, or snapshots from last 24h).

### 10. Old plugin error messages in Chinese, new tool in English — inconsistency

The old `plugins/bundle/cloudpaw/tools/manage_prd.py` used Chinese error messages (`"story ID '...' 已存在"`), while the new `src/qwenpaw/agents/tools/manage_prd.py` uses English (`"story ID '...' already exists"`). This is an improvement for the core codebase, just noting the intentional change.

### 11. `_handle_update` uses bare `assert`

**File:** `src/qwenpaw/agents/tools/manage_prd.py:379`

```python
fields, err = _parse_fields(fields)
if fields is None:
    return _error_response(err)
assert isinstance(fields, dict)
```

`assert` statements can be disabled with `-O` flag. This is a type-narrowing assert that's safe in practice but could use a type guard pattern instead.

### 12. Frontend fetches PRD on every render cycle

**File:** `console/src/components/PRDViewer/index.tsx:307-311`

```tsx
useEffect(() => {
    if (!isLoading && isSuccess && loopDir) {
        fetchPrd();
    }
}, [isLoading, isSuccess, loopDir, fetchPrd, snapshotTs]);
```

The `snapshotTs` dependency is updated in another `useEffect` from `toolResult`, which can cause a double-fetch: first when `isSuccess` becomes true (before `snapshotTs` is set), then again when `snapshotTs` is updated. Consider combining the two effects or using the timestamp directly from `toolResult` in `fetchPrd`.

### 13. No-cache headers only on index.html

**File:** `src/qwenpaw/app/_app.py:689-693`

```python
return FileResponse(
    _CONSOLE_INDEX,
    headers={
        "Cache-Control": "no-cache, no-store, must-revalidate",
        "Pragma": "no-cache",
    },
)
```

Adding no-cache to the console index is reasonable for development, but `must-revalidate` combined with `no-store` is redundant. Also, this doesn't affect JS/CSS bundles served by the static file mount — if the goal is to bust caches after plugin updates, the bundled assets also need cache-busting (usually via hash-based filenames).

### 14. Line length compliance

Per CLAUDE.md: *"Python开发每行代码\注释不要超过79"*. Several lines in `manage_prd.py` appear to exceed 79 characters. Run `pre-commit run --file src/qwenpaw/agents/tools/manage_prd.py` to verify.

---

## Positive Observations

1. **Good architectural decision** to move `manage_prd` from plugin to core — it's tightly coupled to mission mode which is a core feature. Eliminates the brittle plugin-to-core dependency.

2. **Snapshot mechanism** is a nice addition over the old backup approach. The old plugin created `.bak` files in-place; the new code uses a dedicated `snapshots/` directory with timestamp-based files, enabling the frontend to show historical PRD states.

3. **Proper i18n** — the new `PRDViewer` component uses `react-i18next` with translations for 6 languages, whereas the old plugin UI hardcoded Chinese strings.

4. **Better code structure** — the new tool code is well-decomposed into small helper functions (`_validate_create_params`, `_parse_stories`, `_build_story`, `_validate_stories_list`, etc.) compared to the monolithic old implementation.

5. **Clean plugin removal** — the CloudPaw plugin is properly cleaned up: tool registration, constants, hooks, UI renderer, and documentation are all updated consistently.

---

## Verdict

The PR accomplishes a meaningful architectural improvement by promoting `manage_prd` from plugin to core. The code quality is generally good with proper i18n, decomposed helpers, and snapshot support. However, **Issue #1 (eagerly created coroutines)** is a genuine bug that should be fixed before merge, and **Issue #2 (unbound variable)** and **Issue #3 (path traversal)** deserve attention.

**Recommendation:** Request changes for issues #1-#3, consider #4 and #6.

Saint-Yin:
@rayrayraykk 
  #1 — Eager coroutines in handler_map ✅ Fixed
  
  Replaced the handler_map dict with an explicit if/elif chain. Only the matching handler coroutine is created and
   awaited, eliminating the RuntimeWarning: coroutine was never awaited for the 3 unused handlers.

  #2 — Unbound variable prd ✅ Fixed

  Replaced the if/else/msg + assert prd is not None pattern with two independent early returns. No unbound
  variable path remains.

  #3 — Path traversal in /prd API ✅ Fixed

  Added validation using constant.WORKING_DIR (same approach as files.py). The resolved loop_dir must be within
  WORKING_DIR via is_relative_to(). The timestamp parameter is validated against ^\d{8}-\d{6}$ to reject injection
   via filename.

  #4 — import re inside function body ✅ Fixed

  Moved to module-level imports at the top of the file, following project convention.

  #5 — manage_prd exposed to all agents ⏭️  Deferred

  Mission mode is detected at runtime (not as a static agent config), so gating at registration would require
  significant refactoring. The tool validates inputs gracefully — it returns a clear error if loop_dir doesn't
  exist or prd.json is missing. Open to revisiting this in a follow-up.

  #6 — Bare assert in _handle_update ✅ Fixed

  Replaced assert isinstance(fields, dict) with an explicit type guard that returns _error_response. Safe under
  -O.

  #7 — Excessive as any in PRDViewer ✅ Partial

  Consolidated the two useEffects and removed the unused snapshotTs state. The as any casts are retained because
  the data: unknown prop intentionally accepts varied tool message shapes. Defining a strict interface would be
  fragile — happy to add one in a follow-up if preferred.

  #8 — Sorting priority → id ℹ️  Intentional

  The commit message notes this is an intentional UX change. The old plugin sorted by priority; the new viewer
  sorts by ID with numeric: true for natural ordering (US-1, US-2, US-10).

  #9 — Unbounded snapshot growth ✅ Fixed

  Added cleanup in _save_prd: keeps at most 50 snapshots per mission loop directory, removing the oldest when
  exceeded.

  #11 — Bare assert ✅ Fixed (same as #6)

  #12 — Double fetch in PRDViewer ✅ Fixed

  Merged the two useEffects into one that reads the timestamp directly from toolResult, eliminating the
  state-lag-induced double fetch. Also removed the now-unused snapshotTs state and useCallback import.

  #13 — Cache headers ⏭️  Deferred

  The _app.py cache headers added by the original PR have been reverted to the pre-PR state (FileResponse without
  custom headers). Vite production builds already use content-hashed filenames for bundles, so cache-busting is
  handled at the asset level. Open to a separate PR adding max-age=31536000, immutable to /assets if desired.

  ---
  Additional improvements since the review:
  
  - Prompt alignment: Fixed the Phase 2 iteration reminder to explicitly require
  manage_prd(operation="mark_passed", ...) instead of saying "set passes: true in prd.json" (which led agents to
  use write_file directly). Added emphasis that mark_passed must NOT be called before the verifier says VERDICT: 
  PASS.

Issue #4888, opened 2026-06-02
[Bug]: Dream agent 用相对路径 write_file 覆盖了其他 workspace 的 MEMORY.md


danielhansaiday:
## QwenPaw Version

1.1.9

## Description

**症状**：某个 agent 的 Dream cron 触发后，错误地覆盖了 **default workspace** 的 MEMORY.md 文件。

**事件经过**：
1. 皮皮虾（skip agent）的 Dream cron 在 23:00 触发
2. Dream agent 在"梦境"中执行 `write_file(file_path="MEMORY.md", content="test")`
3. 由于相对路径解析问题，写入到了 **default workspace** 的 MEMORY.md（而非 skip workspace）
4. 原 MEMORY.md 内容被完全覆盖为 "test"

**系统日志**：
```
2026-06-01 23:12:21 | Dream agent response:
### 发生了什么
- write_file 工具在梦中反复调用均报参数缺失错误
- 最后一次 write_file(file_path="MEMORY.md", content="test") 成功写入了一个占位文件
- MEMORY.md 已被覆盖为仅含 "test" 的内容，原有记忆丢失
```

## Root Cause

1. Dream agent 使用相对路径 `MEMORY.md` 调用 `write_file`
2. `_resolve_file_path` 函数将相对路径解析为 `get_current_workspace_dir() / file_path`
3. 但 `get_current_workspace_dir()` 返回的是 **调用者（main runner）的 workspace_dir**，而不是 Dream agent 自己所属 agent 的 workspace

## Related Issue(s)

- #3905：Dream agent 执行后 MEMORY.md 仅保留空白模板（v1.1.4，类似问题）

## Environment

- **OS:** Ubuntu 26.04
- **Install method:** curl install script

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (Feishu, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation

## Possible Fix

Dream agent 创建时应该传入正确的 workspace_dir，确保工具调用时 `get_current_workspace_dir()` 返回的是 Dream agent 所属 agent 的 workspace 路径，而非调用者的 workspace 路径。

或者：Dream agent 应使用**绝对路径**来访问自己 workspace 下的 MEMORY.md，而不是依赖 `get_current_workspace_dir()` 上下文变量。

ployts:
[...]

Issue #3905, opened 2026-04-28
[Bug]:Dream agent ISSUE


xutianfu123123:
## QwenPaw Version

1.1.4 post2

## Description

执行QwenPaw梦境优化Dream agent记忆管理任务后，系统自动检测记忆相关文件状态，发现长期记忆核心文件MEMORY.md仅保留空白模板分类标题，无任何实质性记忆内容，且当日对应日期记忆日志文件memory/2026-04-27.md完全不存在。系统判定无有效记忆数据可新增沉淀、修正或删除，自动跳过记忆去重、合并等核心优化操作，仅输出梦境苏醒汇报及基础梦境执行结果，记忆沉淀更新流程未正常闭环生效。
After executing the QwenPaw dream optimization Dream agent memory management task, the system automatically detects the status of memory-related files. It is found that the long-term memory core file MEMORY.md only retains the blank template classification titles without any substantive memory content, and the corresponding daily memory log file memory/2026-04-27.md does not exist at all. The system determines that there is no valid memory data for new precipitation, modification or deletion, automatically skips core optimization operations such as memory deduplication and merging, and only outputs the dream wake-up report and basic dream execution results. The memory precipitation and update process is not closed and take effect normally.

**Related PR(s):** #(optional)

**Security considerations:** [If applicable, e.g. auth, env/config exposure]

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:1.1.4 post2
- **OS:** [Ubuntu 22.04]
- **Install method:**pipx 
- **Python version (if applicable):** [ 3.12]

## Steps to Reproduce

1. Start the QwenPaw dream optimization Dream agent memory management function normally / 正常启动QwenPaw梦境优化Dream agent记忆管理功能；
2. The system automatically scans and detects the specified memory file paths including memory/2026-04-27.md and MEMORY.md / 系统自动扫描检测memory/2026-04-27.md、MEMORY.md两大核心记忆文件路径；
3. Wait for the Dream agent task to run and output the final execution result / 等待梦境代理任务执行完成并输出最终运行结果。

## Actual vs Expected

[...]

PR #4936, opened 2026-06-03
 fix(memory): dream should use cron runner's workspace, not stale instance dir

ployts:
## Description

`memory_manager` is registered as a reusable service (`workspace.py:189`): on workspace reload the previous instance is carried over via `set_reusable_components`, so `self.working_dir` / `self.agent_id` freeze at the values they had when the previous workspace was started. The cron's `self._runner.workspace_dir`, however, always points at the current workspace.

Previously `dream()` resolved every path from `self.working_dir` (`set_current_workspace_dir`, `backup_path`, `MEMORY.md`), so after any workspace reload the dream cron silently wrote backups and rewrote `MEMORY.md` inside the **previous** workspace.

This PR makes `dream()` use the cron's workspace by threading runner-derived values through, mirroring `_heartbeat_callback` / `run_heartbeat_once`:

- `dream()` now takes keyword-only `runner` / `channel_manager` /  `agent_id` / `workspace_dir`. `agent_id` falls back to `self.agent_id`; `workspace_dir` is required and raises `ValueError` when missing (so a misconfigured cron fails loudly instead of writing to the process CWD).
- `_dream_callback` passes `self._runner` / `self._channel_manager` / `self._agent_id` / `self._runner.workspace_dir`, matching the shape of `_heartbeat_callback` exactly.
- `backup_path` and `MEMORY.md` path are now derived from the resolved `workspace_path`, never from `self.working_dir`.
- `BaseMemoryManager.dream` signature widened to declare the same keyword params (all `Optional`, plus `**kwargs`) so the override is LSP-compatible.

**Related Issue:** #4888 #3905 

**Security Considerations:** Empty `workspace_dir` now raises rather than falling through to `Path("")` / process CWD, eliminating the silent cross-workspace write path.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

## Checklist
- [x] Ready for review

## Testing

1. Start an agent, then trigger a workspace reload (so the `memory_manager` instance is carried over via `set_reusable_components` while `runner.workspace_dir` updates to the new workspace).
2. Let `dream_cron` tick. Confirm the backup file appears under`<NEW workspace>/backup/memory_backup_<ts>.md` and that `MEMORY.md` inside the **previous** workspace is untouched. Pre-fix it was the other way around.
3. Manually invoke `await memory_manager.dream(runner=runner)` with `runner.workspace_dir = None` — expect `ValueError("dream requires a workspace_dir")`, surfaced via `logger.exception("dream run failed")` in `_dream_callback`.

```bash
pre-commit run --all-files
# sort simple yaml files...............................(no files to check)Skipped
# check xml................................................................Passed
# check toml...............................................................Passed
# check docstring is first.................................................Passed
# check json...............................................................Passed
# detect private key.......................................................Passed

pytest tests/unit/agents/memory/ tests/integration/test_cron.py tests/integration/test_cron_header.py
# 76 passed in 13.40s
```

Additional Notes

runner / channel_manager are accepted but currently unused inside
dream() (del runner, channel_manager, kwargs) — reserved so future
enhancements (e.g. dispatching an optimization summary to the last
channel) don't need another signature change.

### DIFF
--- a/src/qwenpaw/agents/memory/base_memory_manager.py
+++ b/src/qwenpaw/agents/memory/base_memory_manager.py
@@ -5,7 +5,8 @@ import logging
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from datetime import datetime
-from typing import Any
+from pathlib import Path
+from typing import Any, Optional
 
 from agentscope.message import Msg
 from agentscope.tool import ToolResponse
@@ -115,7 +116,15 @@ class BaseMemoryManager(ABC):
         return None
 
     # pylint: disable=unused-argument
-    async def dream(self, **kwargs) -> None:
+    async def dream(
+        self,
+        *,
+        runner: Any = None,
+        channel_manager: Any = None,
+        agent_id: Optional[str] = None,
+        workspace_dir: Optional[Path] = None,
+        **kwargs,
+    ) -> None:
         """Optimize memory files via a background agent pass.
 
         NOTE: This method is optional. Subclasses may override this method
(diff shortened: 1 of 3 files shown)


jinliyl (approved):
LGTM

lalaliat (approved):
LGTM

PR #4995, opened 2026-06-07
fix(channels): preserve renderer tool output

AbbyJL:
## Description

Preserve tool output attachments and visible text in the channel renderer when `show_tool_details` is disabled. Also keep audio `media_type` when constructing `AudioContent`.

This fixes cases where tool output was reduced to a placeholder and attachment metadata was lost, which hurt channel compatibility and user-visible output.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Channels
- [x] Tests

## Checklist

- [x] I ran `python -m pytest tests/unit/channels/test_renderer.py -q` locally and it passes
- [x] I ran `python -m pytest tests/unit/channels/test_base_core.py -q` locally and it passes

### DIFF
--- a/src/qwenpaw/app/channels/renderer.py
+++ b/src/qwenpaw/app/channels/renderer.py
@@ -145,10 +145,14 @@ class MessageRenderer:
                         elif btype == "video":
                             result.append(VideoContent(video_url=url))
                         elif btype == "audio":
+                            media_type = (
+                                src.get("media_type")
+                                or "application/octet-stream"
+                            )
                             result.append(
                                 AudioContent(
                                     data=url,
-                                    format=b.get("media_type"),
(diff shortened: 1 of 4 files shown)


zhijianma:
Thanks for the contribution,  @AbbyJL . 

Closing this PR for the following reasons:
 - renderer.py: The behavior of show_tool_details=False is intentional — only media attachments should be surfaced, not raw text. The AudioContent media_type fix is valid but should be submitted as a separate, focused PR.
 - console.py: The silent return when queue is None is intentional; the caller handles this case gracefully downstream. Raising a 409 here is not the right approach.
Feel free to reopen a focused PR for the media_type fix alone — we'd be happy to take that!

PR #5037, opened 2026-06-09
fix(config): avoid IndexError on empty Exec= in Linux browser detection

ly-wang19:
## What

`_get_linux_default_browser()` (used by `get_system_default_browser()`) parses the `.desktop` entry's `Exec=` line as:

```python
exe = line.split("=", 1)[1].strip().split()[0]
```

## Why it's a bug

If the matched line is an empty or whitespace-only `Exec=` (`Exec=`, `Exec=   `, `Exec=\t`), the trailing `.split()` returns `[]`, so `[0]` raises `IndexError`. It is **not** caught by the surrounding `except OSError`, so it escapes `get_system_default_browser()` — whose docstring promises a `(None, None)` result when detection fails:

> Returns (None, None) if detection fails or unsupported.

A malformed default-handler `.desktop` file therefore crashes browser detection instead of degrading gracefully.

## Fix

Guard the token list; an empty `Exec` is treated as "no executable found" (the existing `break` path → `(None, None)`).

## Tests

Added `tests/unit/config/test_config_utils.py` — a parametrized regression test for empty / whitespace-only `Exec=` values. It raises `IndexError` on the old code and passes on the new.

## Local checks

- `black==23.3.0 --line-length=79`, `flake8==6.1.0`, `pylint==3.0.2` (repo args) — clean
- `pytest tests/unit/config/test_config_utils.py` — **3 passed**

### DIFF
--- a/src/qwenpaw/config/utils.py
+++ b/src/qwenpaw/config/utils.py
@@ -307,7 +307,13 @@ def _get_linux_default_browser() -> Tuple[Optional[str], Optional[str]]:
             with open(path, encoding="utf-8") as f:
                 for line in f:
                     if line.strip().startswith("Exec="):
-                        exe = line.split("=", 1)[1].strip().split()[0]
+                        # An empty/malformed "Exec=" yields no tokens; skip
+                        # rather than let [0] raise IndexError (callers rely
+                        # on a (None, None) result, never an exception).
+                        parts = line.split("=", 1)[1].strip().split()
+                        if not parts:
+                            break
+                        exe = parts[0]
                         if exe.startswith("/") and Path(exe).is_file():
                             return _linux_desktop_to_kind_and_path(exe)
                         for p in ["/usr/bin", "/usr/local/bin"]:


x1n95c:
Hello @ly-wang19, thanks for your contribution. However, have you actually encountered this error in practice? An empty `Exec=` in a .desktop file is extremely unlikely on Linux, mainstream package managers always generate valid entries. Moreover, the unite test added in this pr is not necessary.

ly-wang19:
Thanks @x1n95c — you're right, I hit this by reading the code, not in practice, and an empty `Exec=` is indeed unlikely with standard package managers. My only rationale was that `get_system_default_browser()` is documented to return `(None, None)` on failure, so the uncaught `IndexError` technically breaks that contract — but I agree it's marginal. I've removed the test. Totally fine to close this if you'd rather not carry the guard — your call.

lalaliat:
Thank you for your contribution! However, given that this issue is unlikely to be triggered in practice, we have decided to close it without merging for now. If you have any other thoughts or insights, please feel free to share them.

ly-wang19:
Completely fair — thanks for taking the time to consider it! Point well taken: I'll prioritize issues that show up in practice. Glad to have the other fixes land, and happy to help elsewhere. 🐾

PR #5040, opened 2026-06-09
fix(crons): tolerate invalid jobs in jobs.json instead of failing all

### LINKED ISSUE #4835: [Bug] One invalid job in jobs.json causes entire workspace to fail
## Description

When `jobs.json` contains one invalid job (e.g., missing required field `session_id`), the entire workspace fails to start.

## Evidence

```
ERROR | Failed to start service cron_manager: 1 validation error for JobsFile
jobs.14.dispatch.target.session_id - Field required
```

## Expected Behavior
- Skip invalid jobs, load valid ones
- Log errors for invalid jobs
- Workspace should start with valid jobs only

## Actual Behavior
- Entire workspace fails to start
- Console stuck loading indefinitely

## Impact
One typo in one job breaks the entire system.

ly-wang19:
Fixes #4835

## What

`JsonJobRepository.load()` parsed `jobs.json` with `JobsFile.model_validate(data)`, which validates the whole file atomically.

## Why it's a bug

A single malformed job — e.g. one missing `dispatch.target.session_id` — makes pydantic raise `ValidationError` for the entire `JobsFile`, so `load()` raises, `cron_manager` fails to start, and the whole workspace hangs (Console stuck loading):

```
ERROR | Failed to start service cron_manager: 1 validation error for JobsFile
jobs.14.dispatch.target.session_id - Field required
```

One typo in one job breaks the entire system.

## Fix

Keep the fast path; on `ValidationError`, fall back to per-job validation — load the valid jobs, skip the invalid ones, and log each drop. This matches the behavior requested in the issue (skip invalid, load valid, log errors). Well-formed files are unaffected.

## Tests

Added `tests/unit/app/crons/test_json_repo.py`:
- mixed valid + invalid → keeps only the valid job
- all valid → unchanged
- only invalid → empty (no crash)

## Local checks

- `pytest tests/unit/app/crons/test_json_repo.py` — **3 passed**

### DIFF
--- a/src/qwenpaw/app/crons/repo/json_repo.py
+++ b/src/qwenpaw/app/crons/repo/json_repo.py
@@ -8,10 +8,13 @@ import os
 import shutil
 import uuid
 from pathlib import Path
+from typing import Any
 from urllib.parse import quote
 
+from pydantic import ValidationError
+
 from .base import BaseJobRepository
-from ..models import CronExecutionRecord, JobsFile
+from ..models import CronExecutionRecord, CronJobSpec, JobsFile
 
 logger = logging.getLogger(__name__)
 
@@ -44,7 +47,43 @@ class JsonJobRepository(BaseJobRepository):
             return JobsFile(version=1, jobs=[])
 
         data = json.loads(self._path.read_text(encoding="utf-8"))
-        return JobsFile.model_validate(data)
+        try:
+            return JobsFile.model_validate(data)
+        except ValidationError:
+            # A single malformed job must not break the whole workspace:
+            # keep the valid jobs, drop the invalid ones, and log each drop
+            # so it stays visible (issue #4835).
+            return self._load_tolerant(data)
+
+    def _load_tolerant(self, data: Any) -> JobsFile:
+        """Build a JobsFile, skipping jobs that fail validation."""
+        if not isinstance(data, dict):
+            logger.error(
+                "jobs file %s is not a JSON object; starting with no jobs",
+                self._path,
+            )
+            return JobsFile(version=1, jobs=[])
+
+        raw_jobs = data.get("jobs")
+        if not isinstance(raw_jobs, list):
+            raw_jobs = []
+
+        jobs: list[CronJobSpec] = []
+        for index, raw in enumerate(raw_jobs):
+            try:
+                jobs.append(CronJobSpec.model_validate(raw))
+            except ValidationError as exc:
+                logger.error(
+                    "Skipping invalid job at index %d in %s: %s",
+                    index,
+                    self._path,
+                    exc,
+                )
+
+        version = data.get("version", 1)
+        if not isinstance(version, int):
+            version = 1
+        return JobsFile(version=version, jobs=jobs)
 
     async def save(self, jobs_file: JobsFile) -> None:
         self._path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


lalaliat:
@ly-wang19 
Thank you for your contribution. However, regarding the original issue, the core concern is that `jobs.json` should not have missing entries under normal circumstances. Therefore, we will hold off on the fix until we fully understand the root cause of the issue. Thank you!

ly-wang19:
Thanks @lalaliat, that's fair — agreed that understanding *why* `jobs.json` gets an invalid entry is the more important fix, and a tolerance layer shouldn't mask that. I'll defer to you on the root-cause work. One small note in case it's useful: even with correct writers, a manually-edited or partially-written file could corrupt a single entry, so graceful degradation might be a reasonable backstop regardless. Happy to hold or close — whatever you prefer.

ly-wang19:
Following up on the root cause you raised (why would `jobs.json` contain an invalid entry under normal circumstances). I dug into it:

- `jobs.json` is only ever written by `JsonJobRepository.save()`, which serializes an already-validated `JobsFile` (`model_dump`). So a normal write **cannot** produce a job missing `dispatch.target.session_id`.
- `app/migration.py` only **ensures the file exists** (creates an empty `{"version": 1, "jobs": []}` if missing) — there is **no content/schema migration** for existing job entries.
- The cron dispatch/session schema has evolved over time (e.g. channel-based session isolation in #4117; there's already a `weixin:` → `wechat:` `session_id` rewrite in `migrate_legacy_weixin_jobs_file`).

So invalid entries don't arise during normal operation — they arise **across upgrades**: an entry written under an older schema becomes invalid under the current one, with no migration to repair it (or from a manual edit / interrupted write).

That suggests two complementary fixes:
1. **A `jobs.json` content migration** that repairs or drops entries failing validation on upgrade (the proper root-cause fix), and/or
2. **Tolerant loading (this PR)** as a backstop so one bad entry never takes down the whole workspace.

Happy to implement whichever you prefer — including adding the migration in (1) if that's the direction you'd like.

lalaliat (commented):
Thank you for your contribution! Please check and solve comments raised by Copilot. BTW, please remove the test file before merging. Thank you!

maximilize:
Triage note: #5040 and #5347 overlap — both fix the same root cause (#4835: one invalid entry in `jobs.json` blocks workspace startup), at different layers, and they conflict textually (both edit the imports in `src/qwenpaw/app/crons/repo/json_repo.py`), so only one should land.

Suggest preferring **#5040**: it hardens the actual read path — `load()` (`json_repo.py:42-47`) — so invalid jobs are tolerated everywhere, and it ships tests. #5347's startup migration leaves `load()` fragile, but its backup-before-discard (`.invalid-jobs.bak`) is worth folding in, since #5040 drops invalid jobs with no recoverable copy.

One gap in both: `json.loads()` at `json_repo.py:46` runs before the validation `try`, so a truncated/corrupt file still crashes `load()`. Catching `JSONDecodeError` there too would fully close #4835.

lalaliat:
Thank you for your contribution! However, due to the lack of response to feedback over an extended period and the fact that this PR now has conflicts with the main branch, we are temporarily closing it. Additionally, the original issue does not occur under normal usage conditions. We truly appreciate your effort, and we look forward to your future ideas!

PR #5041, opened 2026-06-09
fix(backup): skip unreadable files instead of failing the whole backup

### LINKED ISSUE #4916: [Bug] Backup fails with PermissionError on browser cache files
### Environment
- **QwenPaw Version**: 1.1.10b1
- **OS**: Windows 10 (AMD64)

### Description

When clicking "Create Backup" in the QwenPaw Desktop client, the backup creation fails with a `PermissionError` on browser cache files.

### Error Log

```
qwenpaw\backup\_ops\create.py:144 | Backup creation failed for qwenpaw-1.1.10b1-20260602T125206Z-daba3fc7
  File "qwenpaw\backup\_ops\create.py", line 139, in _create_with_progress
  File "qwenpaw\backup\_ops\create.py", line 195, in _compress_to_tmp
  File "qwenpaw\backup\_ops\create_helpers.py", line 161, in add_files_to_zip
  File "qwenpaw\backup\_ops\create_helpers.py", line 53, in add_agent_workspaces
PermissionError: [Errno 13] Permission denied: 'C:\Users\...\.qwenpaw\workspaces\default\browser\user_data\Default\Cache\Cache_Data\data_0'
```

The error occurs on different cache files each time:
- `browser\user_data\Default\Cache\Cache_Data\data_0`
- `browser\user_data\Default\DawnGraphiteCache\data_0`

### Root Cause

The `add_agent_workspaces` function in `create_helpers.py` attempts to zip all files in the workspace, including the `browser/user_data` directory. However, the `qwenpaw-backend.exe` process itself maintains file locks on browser cache files (Chromium internally managed by the backend). When `zipfile.write()` tries to read these locked files, it throws a `PermissionError`, causing the entire backup operation to fail.

### Reproduction Steps
1. Launch QwenPaw Desktop
2. Click "Create Backup" in the client
3. Select "Full Backup"
4. Backup fails with PermissionError

### Expected Behavior
Backup should succeed by either:
1. Excluding `browser/user_data` from backup (cache files are not essential data)
2. Gracefully handling `PermissionError` by skipping locked files instead of failing the entire backup
3. Stopping the internal browser before creating the backup

### Suggested Fix
In `create_helpers.py`, the `add_agent_workspaces` function should:
- Skip the `browser/` directory or at minimum `browser/user_data/Default/Cache*` directories
- Or handle `PermissionError` per-file (log a warning and continue) rather than failing the entire backup

ly-wang19:
Fixes #4916

## What

`add_agent_workspaces()` (`src/qwenpaw/backup/_ops/create_helpers.py`) zips every file under each agent workspace with a bare `zf.write(entry, arcname)`.

## Why it's a bug

If any single file can't be read, `zf.write` raises and the **entire backup fails**. On Windows this reliably happens with the bundled browser: while an agent's Chromium is open, the backend holds locks on cache files, so:

```
PermissionError: [Errno 13] Permission denied:
  '...\workspaces\default\browser\user_data\Default\Cache\Cache_Data\data_0'
```

One locked cache file makes "Create Backup" fail completely — and everything after that file in the walk is lost too.

## Fix

Catch `PermissionError`/`OSError` per file, log a warning, skip the file, and continue — then log a per-agent summary of how many files were skipped. A backup should degrade gracefully on a few unreadable files rather than abort wholesale.

Scoped to `add_agent_workspaces` (the reported failure point). Happy to extend the same guard to the other `add_*` helpers if maintainers prefer.

## Tests

Added `tests/unit/backup/test_create_helpers.py`: a workspace with one writable and one "locked" file (a fake zip whose `write` raises `PermissionError`) — the backup keeps the readable file and skips the locked one instead of raising.

## Local checks

- `pytest tests/unit/backup/test_create_helpers.py` — **1 passed**

### DIFF
--- a/src/qwenpaw/backup/_ops/create_helpers.py
+++ b/src/qwenpaw/backup/_ops/create_helpers.py
@@ -46,12 +46,33 @@ def add_agent_workspaces(
         ws = Path(ref.workspace_dir).expanduser().resolve()
         if ws.is_dir():
             file_count = 0
+            skipped = 0
             for entry in sorted(ws.rglob("*")):
-                if entry.is_file():
-                    rel = entry.relative_to(ws).as_posix()
-                    arcname = f"{PREFIX_WORKSPACES}{aid}/{rel}"
+                if not entry.is_file():
+                    continue
+                rel = entry.relative_to(ws).as_posix()
+                arcname = f"{PREFIX_WORKSPACES}{aid}/{rel}"
+                try:
                     zf.write(entry, arcname)
-                    file_count += 1
+                except (PermissionError, OSError) as exc:
+                    # A file that can't be added (e.g. an open Chromium
+                    # cache file the backend has locked) must not abort
+                    # the whole backup; skip it and continue (#4916).
+                    skipped += 1
+                    logger.warning(
+                        "Skipping %s (could not be added to backup): %s",
+                        entry,
+                        exc,
+                    )
+                    continue
+                file_count += 1
+            if skipped:
+                logger.warning(
+                    "Agent '%s': skipped %d file(s) that could not be "
+                    "added to the backup",
+                    aid,
+                    skipped,
+                )
             logger.debug(
                 "Agent '%s': %d file(s) added from %s",
                 aid,


x1n95c:
Hello, thanks for your contribution, I have two comments:

 1. The warning log says "Skipping unreadable file", but the OSError from zf.write() may not be caused by an unreadable file, it could be a write-side error (e.g., disk full). Please adjust the warning message to be more accurate. 
  2. The unit test is not necessary for this change. Let's skip it.

ly-wang19:
Thanks for the careful review, @x1n95c!

1. Good catch — reworded the log so it no longer claims the file was "unreadable", since the `OSError` can be a write-side failure. (Side note: a genuine write failure like disk-full would recur on every later file too, so if you'd prefer, I can narrow the `except` to `PermissionError` only — the reported locked-cache case — and let other `OSError`s propagate so they stay visible. Happy either way.)
2. Removed the unit test.

x1n95c:
Hi, I tested your code and everything works as expected, thanks for your contribution. I think this PR is ready to be merged.

ly-wang19:
Thanks @x1n95c, much appreciated — glad it tested cleanly! 🐾

lalaliat (approved):

PR #5051, opened 2026-06-09
fix(desktop): persist backend port across restarts to preserve localStorage (#4733)

### LINKED ISSUE #4733: [Feature]: windows桌面版重启后希望显示上次关闭时所用的智能体及对话
## QwenPaw Version
[Bug]: 1.19版本weindows桌面版，每次重启后，显示的智能体，不是上一次关闭时候的智能体以及对于的对话session
这种设计就比较奇怪

wangfei010313:
### Problem

On Windows desktop, restarting QwenPaw always resets the selected agent to CloudPaw-Master, regardless of what the user was using before closing. (#4733)

### Root Cause

The desktop app allocated a random port on every launch. Since `localStorage` is isolated by origin (protocol + host + port), each restart created a new origin, losing all client-side state. The CloudPaw plugin then treated every restart as a first install and overrode the agent to CloudPaw-Master.

### Solution

**Port Persistence (**`**qwenpaw/utils/port.py**`**)**
Extracted a shared port persistence module used by both CLI and Tauri entry points:

- `get_stable_port(port_file, host)` → returns `(port, socket)` tuple
- Reuses the previous port from `WORKING_DIR/desktop_port` if available
- Falls back to random port if occupied, persists new port
- **Atomic write**: uses `tempfile` + `os.replace()` to prevent corruption on crash
- **Platform-aware binding**: `SO_EXCLUSIVEADDRUSE` on Windows (prevents false-positive with `SO_REUSEADDR`), `SO_REUSEADDR` on POSIX (for `TIME_WAIT` rebinding)

**TOCTOU Mitigation**

- **Tauri (**`**entry.py**`**)**: Pre-bound socket passed directly to `uvicorn.Server.run(sockets=[...])` — zero TOCTOU window
- **CLI (**`**desktop_cmd.py**`**)**: Socket held open until immediately before `subprocess.Popen()`, minimizing the race window to near-zero

**Defense-in-Depth (**`**cloudpaw/ui/src/index.ts**`**)**
Guard in `ensureDefaultAgent()`: if existing agent selection data is found in `localStorage`, skip the first-install override — even if the port persistence fails for any reason.

### How It Works

- **First launch**: Random port allocated → saved to `~/.qwenpaw/desktop_port`
- **Subsequent launches**: Read port file → bind same port → same origin → `localStorage` preserved
- **Port conflict**: If previous port is occupied → fallback to random → save new port

### Files Changed

| File | Description |
| ------ |------ |
| `src/qwenpaw/utils/port.py` | New — shared port persistence module (atomic write, platform-aware binding) |
| `src/qwenpaw/cli/desktop_cmd.py` | Use `get_stable_port()`, hold socket until subprocess spawn, add `storage_path` for WebView2 |
| `src/qwenpaw/tauri/entry.py` | Use `get_stable_port()`, pass pre-bound socket directly to uvicorn |
| `plugins/bundle/cloudpaw/ui/src/index.ts` | Guard against overriding existing agent selection |

### Testing & Verification

- **20 unit tests** (edge cases: invalid file, occupied port, out-of-range, atomic write failure)
- **Manual E2E**: Switch to Default agent → close → restart → still on Default agent
- All pre-commit checks pass (`mypy`, `black`, `flake8`, `pylint`)

**Manual E2E Verification:**

- Start desktop → switch to Default agent → chat → close
- Restart →  Still on Default agent with previous chat visible
- Repeat with other agents (QA Agent, etc.) →  Selection persists
- Kill another process occupying the port → restart →  Gracefully falls back to new port

**Local Verification Evidence:**

```bash
pre-commit run --all-files
# [paste summary result]

pytest
# [paste summary result]
```

### Related Issue

Fixes #4733

### Type of Change

[x] Bug fix
[ ] New feature
[ ] Breaking change
[ ] Documentation
[x] Refactoring

### Component(s) Affected

[x] Core / Backend
[x] Console (frontend web UI)
[x] CLI
[ ] Channels / Skills / Docs / Tests / CI/CD


tests/unit/tauri/test_port_persistence.py
```diff
@@ -0,0 +1,125 @@
+# -*- coding: utf-8 -*-
```
rayrayraykk: remove this

tests/unit/cli/test_desktop_port_stability.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
```
rayrayraykk: remove this

rayrayraykk (changes requested):
### Overall

The approach is sound: the root cause is that a random port on every launch changes the `localStorage` origin, and the fix correctly persists and reuses the port. The CloudPaw defense-in-depth guard is well-designed. However, there's **one bug that should be fixed** before merge, plus a few suggestions.

---

### Bug: TOCTOU Race in `_get_stable_port()` (desktop_cmd.py)

The port availability check binds → listens → **closes the socket** → returns the port number:

```python
with socket.socket(...) as sock:
    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    sock.bind((host, last_port))
    sock.listen(1)
# socket is closed here, port is released
return last_port  # ← between here and uvicorn's actual bind, another process can grab the port
```

**Compare with `tauri/entry.py`**: `_try_bind_port()` returns the socket object itself and passes it directly to `uvicorn.Server.run(sockets=[...])` — the port is never released, so there's no race window.

**Suggestion**: `_get_stable_port()` should also return the socket (or at minimum, not close it before returning). If pywebview doesn't accept a pre-bound socket, at least document the race and drop the `listen()` call to minimize the window.

---

### Code Duplication

Port persistence logic is implemented twice with different styles:
- `desktop_cmd.py`: everything in a single `_get_stable_port()` function
- `tauri/entry.py`: factored into `_read_last_port()` / `_write_port_file()` / `_try_bind_port()`

**Suggestion**: Extract the three helpers into a shared module (e.g. `qwenpaw.utils.port`) and reuse from both entry points. This would also make it easier to fix the TOCTOU issue in one place.

---

### Non-atomic Port File Write

```python
with open(port_file, "w", encoding="utf-8") as fh:
    fh.write(str(port))
```

If the process crashes between truncation and close (or on power loss), the file can end up empty or corrupted. `_read_last_port()` handles `ValueError` gracefully so this degrades to a random port — not a crash — but since the project already has `_atomic_write_json` (tempfile → fsync → `os.replace`) in `session.py`, it would be easy to use the same pattern here.

**Severity**: Low (graceful degradation to random port)

---

### Copilot Review Comments — Assessment

Copilot left 4 inline comments. Here's my take on each:

| # | Copilot's claim | Verdict |
|---|---|---|
| 1 | `_try_bind_port()` doesn't call `listen()` | **False positive** — line 148 explicitly has `sock.listen(1)` |
| 2 | Docstring says `~/.qwenpaw/desktop_port` but `WORKING_DIR` is configurable | **Valid** — docstring should say `WORKING_DIR/desktop_port` |
| 3 | Port check in desktop_cmd lacks `SO_REUSEADDR` | **False positive** — line 131 sets `SO_REUSEADDR` |
| 4 | `_get_stable_port()` is untested | **False positive** — `test_desktop_port_stability.py` has 5 test cases covering it |

---

### CloudPaw Guard — Looks Good

```typescript
const existingLastUsed = localStorage.getItem(LAST_USED_KEY);
const existingStorage = localStorage.getItem(STORAGE_KEY);
if (existingLastUsed || existingStorage) {
    localStorage.setItem(FIRST_INSTALL_KEY, "true");
    return;
}
```

Good defense-in-depth: even if port persistence fails for some reason, as long as `localStorage` happens to survive (e.g. Tauri with persistent WebView2 profile), the user's agent selection won't be overwritten. Re-persisting `FIRST_INSTALL_KEY` prevents this check from running on every subsequent launch.

---

### Test Coverage

Tests are thorough:
- `test_desktop_port_stability.py`: 5 cases covering first launch, reuse, occupied-port fallback, invalid file, out-of-range port
- `test_port_persistence.py`: covers `_read_last_port` / `_write_port_file` / `_try_bind_port` edge cases

---

### Summary

| Issue | Severity | Action |
|-------|----------|--------|
| `_get_stable_port` TOCTOU race | **Medium** | Return socket instead of port number, or unify with tauri's pattern |
| Port logic duplicated in two files | **Medium** | Extract shared module |
| Non-atomic port file write | **Low** | Use `tempfile + os.replace` |
| Docstring path description inaccurate | **Low** | Change to `WORKING_DIR/desktop_port` |

Core approach is correct. I'd suggest fixing the TOCTOU race and deduplicating the port logic before merging.

jinglinpeng (commented):
**Windows `SO_REUSEADDR` semantics**: on Windows, binding with `SO_REUSEADDR` can succeed even when another process is actively listening on the port (your own tests work around this with `SO_EXCLUSIVEADDRUSE` on the blocker). Since Windows is the target platform here, the probe/bind should use `SO_EXCLUSIVEADDRUSE` on Windows and keep `SO_REUSEADDR` only on POSIX (for TIME_WAIT).

src/qwenpaw/utils/port.py
```diff
@@ -0,0 +1,132 @@
...
+        if reused_socket:
+            logger.info("Reusing previous desktop port %d", last_port)
+            return last_port, reused_socket
+        logger.info(
+            "Previous port %d unavailable, falling back to random",
+            last_port,
+        )
+
+    port = find_free_port(host)
+    logger.info("Allocated new desktop port %d", port)
```
rayrayraykk: Use Debug.

zhijianma (approved):

rayrayraykk (approved):

PR #5062, opened 2026-06-10
fix(e2e): handle text-based empty state in token usage test

yutai78786:
## Summary

Fix `test_token_usage_overview` failure in nightly E2E ([run #27229106431](https://github.com/agentscope-ai/QwenPaw/actions/runs/27229106431)).

In a clean environment with no token usage data, the page renders plain text (e.g. "No token usage data") instead of an Ant Design `Empty` component or `<table>`. The test was only checking for component-based empty states, so both `has_table` and `has_empty` were false.

Added a scoped fallback: when neither table nor empty component is found, check the main content area for token-usage-specific empty-state text before asserting.

## Test plan

- [x] Verified the fix handles the clean-env scenario (no data → text-based empty state → pass)
- [x] Existing table/component paths are unchanged


yutai78786:
Closing — should validate in fork CI first before submitting upstream.

PR #5067, opened 2026-06-10
feat(driver): introduce Agent OS Driver — unified abstraction for external capabilities (MCP/A2A/ACP/...)

xiaoming-qxm:
## Background

QwenPaw agents need to invoke external capabilities — MCP tool servers, remote agents (A2A), and ACP services. These capabilities differ in protocol, transport, authentication, and lifecycle. Without a unifying layer, each new capability type requires modifying agent code and duplicating concerns like credential management and access control.

**Driver** solves this by abstracting external capability access into a single, protocol-agnostic resource model that the Agent OS can uniformly declare, invoke, and govern. This PR establishes the Driver abstraction and delivers MCP as the first concrete protocol implementation.

## Goals

1. **Declarative** — the system knows which external capabilities exist and how to reach them (DriverCard YAML)
2. **Invocable** — agents call capabilities through a standard interface, unaware of underlying protocol (currently MCP; additional protocols plug in via `register_handler_type()`)
3. **Governable** — fine-grained ABAC policy controls who can invoke what, under what conditions, with human-in-the-loop approval (`ask` effect)
4. **Secure** — credentials are separated from driver cards, encrypted at rest, and resolved only after policy checks pass

## Architecture Overview

```
┌─────────────────────────────────────────────────────────────────────┐
│                           Driver Layer                              │
│                                                                     │
│  DriverManager ── builds ──▶ DriverHandler                         │
│    register_handler_type()       _guarded_execute() [template method]│
│    register/upsert/reload()        ├ evaluate_policy(ctx)           │
│    list_capabilities()  ◀────────  ├ _request_approval(ctx)  [ask]  │
│    invoke_capability()  ─────────▶ ├ resolve_credential()           │
│                                    └ _execute(cred, ctx)  [protocol]│
│                                                                     │
│  DriverCard ─ YAML config: name, protocol, endpoint,               │
│               credential refs, policy, enabled                      │
│                                                                     │
│  Capability ─ protocol-agnostic description:                        │
│    capability_id URI: driver://{proto}/{name}/tools/{tool}#invoke   │
│    DriverInvocation → DriverInvocationResult                        │
│                                                                     │
│  Access Policy ─ ABAC model:                                        │
│    Subject × Principal × Target × Condition → allow | deny | ask   │
│    Per-tool granularity via PolicyTarget(kind="tool", name="...")    │
│                                                                     │
│  Credential ─ 4-layer architecture:                                 │
│    CredentialRef (card) → CredentialStore (encrypted YAML)          │
│    → CredentialProvider (resolve/refresh) → ResolvedCredential      │
│    Built-in kinds: none, static, oauth2_cc, oauth2_auth_code, ak_sk│
└─────────────────────────────────────────────────────────────────────┘
                          ▲ invoke_capability()
                          │
┌─────────────────────────┴───────────────────────────────────────────┐
│  Agent Runtime                                                       │
│    stream_query() → list_capabilities(kind="tool")                  │
│    → build_agent(driver_capabilities, driver_invoker)                │
│    → DriverCapabilityTool: AgentScope ToolBase adapter               │
│        __call__(**kwargs) → invoker(DriverInvocation) → ToolChunk   │
└─────────────────────────────────────────────────────────────────────┘

workspace/
├── drivers/                    # DriverCard YAML, scoped by protocol
│   ├── mcp/{name}.yaml
│   └── {future-protocol}/{name}.yaml   # e.g. a2a, acp — added via follow-up PRs
├── credentials.yaml            # Encrypted credential store
└── .legacy_mcp_migration_report.yaml
```

### Key design decisions

- **Template method on DriverHandler** — `_guarded_execute()` enforces policy → approval → credential → execute pipeline on every call. Subclasses only implement `_execute()`. No call can bypass governance.
- **Build-before-swap** — when reloading a driver, the new handler is fully built and initialized before atomically replacing the old one. Build failure preserves existing service.
- **Open-closed protocol support** — `protocol` is a string, not an enum. Adding a new protocol = implement a handler subclass + `register_handler_type()`. No existing code changes.
- **DriverCapabilityTool** — a lightweight adapter dynamically generated per capability that bridges Driver invocations into AgentScope's ToolBase interface. Agent permission checks are delegated to the Driver's own policy engine.
- **Per-tool policy** — access rules can target individual MCP tools by name, not just the driver as a whole, via `PolicyTarget(kind="tool", name="create_issue")`.
- **`ask` effect** — for uncertain calls, the policy engine suspends execution and routes to the Approval Service for human review via the Console.

### Core driver module (`src/qwenpaw/drivers/`)
- **Contracts** — `DriverCard`, `DriverPolicy`, `PolicyRule`, `DriverCapability`, `DriverInvocation`, `DriverInvocationResult` data models
- **Policy engine** — ABAC evaluator with multi-dimensional specificity sorting (target > principal > subject > strictness)
- **Storage** — atomic YAML read/write for driver cards, protocol-scoped directory layout
- **Credential subsystem** — `CredentialStore` (encrypted YAML), `CredentialProvider` (pluggable via `register_provider()`), 5 built-in provider kinds
- **Handler base** — template method with policy-credential-execute pipeline, approval integration
- **MCPDriverHandler** — full implementation supporting stdio and HTTP/SSE transports, value-source bindings for env/headers, implicit auth header injection
- **DriverManager** — handler type registry, lifecycle orchestration (build/reload/delete/shutdown), capability aggregation and routing. The Manager is protocol-neutral; additional protocols (A2A, ACP, etc.) will be registered in follow-up PRs.

### App integration
- **MCP Console/API migration** — the existing MCP management endpoints now read/write DriverCards instead of raw `agent.json` config, while preserving the Console API shape for backward compatibility
- **Legacy auto-migration** — on workspace startup, `agent.json` MCP clients are automatically converted to driver cards + encrypted credentials. Clients with potential inline secrets in args are skipped with a migration report for manual review
- **New policy endpoints** — `GET/PUT /mcp/policy/{client_key}` for reading and updating per-driver access policies
- **Agent tool exposure** — `DriverCapabilityTool` bridges driver capabilities into the agent toolkit with per-request context propagation
- **Prompt changes** — agent system prompt includes driver capability listing and policy-denial recovery hints

### Console frontend
- **Access Policy UI** — new modal for managing default/client/tool-level `allow`/`ask`/`deny` rules per MCP client
- **ApprovalCard** — now shows `tool_source` (e.g., "mcp:github-mcp") alongside the tool name
- **Navigation** — "Tools" tab renamed to "Tools & Permissions" to reflect policy management

### Tests
- 145 backend tests across unit/integration/e2e covering: card validation, policy evaluation, credential encryption/decryption, storage round-trips, MCP stdio/HTTP/OAuth flows, manager lifecycle, legacy migration, and approval integration
- 8 frontend tests for API endpoints and policy helper functions

## Security Considerations

- **DriverCards are secret-free.** Credentials are referenced by alias (`CredentialRef`), never stored inline.
- **Encrypted credential store.** Sensitive values are stored with `ENC:` prefix in per-workspace `credentials.yaml` (file permissions `0o600`). `env:` references are read-only from the process environment and not persisted.
- **Policy-before-credential.** The governance pipeline evaluates access policy *before* resolving credentials, so denied requests never touch secrets.
- **Legacy migration safety.** Clients with args containing potential secret markers (`api-key`, `token`, `secret`) are skipped rather than migrated, with a report for manual review.

## Scope & Known Limitations

- **Only MCP is implemented as a concrete protocol.** The Driver abstraction is protocol-agnostic, and additional protocols (A2A, ACP) will be added in follow-up PRs via `register_handler_type()`. This PR intentionally limits the runtime surface to MCP to keep the review scope focused.
- **`rate_limit` policy condition is accepted but not enforced.** The configuration is parsed and validated, but runtime enforcement (counter state management) is deferred. Rules with `rate_limit` log a warning and match as if the limit is not exceeded.
- **No filesystem watcher for driver cards.** Changes to YAML files on disk are picked up at startup and via explicit API calls (`reload_driver`), not via hot-reload polling. This is a deliberate simplification over the previous `MCPConfigWatcher` approach.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

N/A. This PR does not change channel payload contracts or channel implementations.

## Testing

```bash
# Backend — 145 tests
uv run pytest tests/unit/drivers tests/unit/agents/tools/test_driver_capability_tool.py \
  tests/unit/app/test_mcp_router_helpers.py tests/unit/workspace/test_driver_manager_service.py \
  tests/unit/workspace/test_prompt.py tests/integration/test_driver_workspace_lifecycle.py \
  tests/integration/test_mcp.py tests/e2e/test_driver_mcp_stdio_flow.py \
  tests/e2e/test_driver_mcp_http_flow.py tests/e2e/test_driver_mcp_oauth_flow.py \
  tests/e2e/test_driver_mcp_policy_flow.py -q
# 145 passed, 1 warning in 6.94s

# Frontend — 8 tests
cd console && npm run test:run -- mcp accessPolicy
# Test Files 2 passed (2), Tests 8 passed (8)

# Linting & formatting
uv run pre-commit run --all-files   # all checks passed
cd console && npm run format        # tsc + prettier passed
git diff --check                    # clean
```

## Additional Notes

- The branch contains four commits: driver runtime/policy layer, Console access policy UI, point-in-time policy-denial fix, and final pre-commit/type/lint normalization.
- Base branch: `dev/agentscope2.0`.


xiaoming-qxm (commented):
## Code Review — Agent OS Driver Runtime (PR #5067)

Reviewed 90 files (+11,237 / -1,071) across 4 dimensions: driver core, handlers/adapters, app integration, console frontend + tests. Findings organized by severity.

---

### 🔴 CRITICAL (2) — Must fix before merge

**C1. Path traversal write primitive via unsanitized card `name`**
- **File**: `src/qwenpaw/drivers/storage.py` — `card_path()` function
- `card_path` builds `cards_dir / _protocol_path(protocol) / f"{name}.yaml"`. The `protocol` is sanitized (`_protocol_path` strips `.` / `..` segments), but `name` is passed through unsanitized. A malicious card name like `../../etc/cron.d/evil` would escape `cards_dir`.
- `dump_card` calls `card_path` and writes YAML to the resulting path — this is an **arbitrary file write**.
- The same issue exists at the router level: `client_key` comes from FastAPI `{client_key:path}` and allows `..` segments (`src/qwenpaw/app/routers/mcp.py`, `mcp_oauth.py`).
- **Fix**: Validate `card.name` in `validate_card` to reject `/`, `..`, null bytes. Additionally, `card_path` should `resolve()` and verify the result is still under `cards_dir`. At the router level, add `_validate_client_key` checks to all write paths (currently only on create).

**C2. Credential secrets exposed via dataclass `__repr__`**
- **Files**: `src/qwenpaw/drivers/credentials/types.py` — `CredentialRecord`, `ResolvedCredential`
- Both are frozen dataclasses with `secrets: dict[str, Any]`. The auto-generated `__repr__` prints all field values verbatim. Any log statement, unhandled exception traceback, or debug representation will **leak tokens, passwords, and signing keys in plaintext**.
- **Fix**: Override `__repr__` on both types to redact secrets: `secrets={k: "***" for k in self.secrets}`.

---

### 🟠 MAJOR (10)

**M1. CredentialStore has no file locking — concurrent read-modify-write race**
- `credentials/store.py`: `put()` and `delete()` do `_read_root()` then `_write_root()` without any lock. Concurrent calls silently clobber each other's writes.
- Fix: Use `asyncio.Lock` on the store instance, or `fcntl.flock` for multi-process safety.

**M2. `invoke_capability` does not catch execution errors**
- `handlers/mcp.py`: Catches `DriverPermissionDeniedError` and `ApprovalRequiredError`, but any MCP transport failure propagates as an unhandled exception. Callers expect `DriverInvocationResult`.
- Fix: Add broad `except Exception` returning `DriverInvocationResult(ok=False, error_type="execution_error", ...)`.

**M3. `StandardOAuth2Exchanger.exchange` raises raw `KeyError`**
- `credentials/providers.py`: `data["access_token"]` — if the token endpoint returns valid JSON without `access_token` (e.g., error with 200 status), this raises an untyped `KeyError`.
- Fix: Check for key presence and raise `DriverCredentialProviderError` with the error description.

**M4. Policy coercion inconsistency — `source_value` vs `source_type` empty-string semantics**
- `contracts.py` `_coerce_policy_principal`: `source_type` uses `or "*"` (empty → wildcard), but `source_value` uses `.get("source_value", "*")` (empty stays as `""`). Same inconsistency in `storage.py`. A rule with `source_value: ""` behaves unpredictably.
- Fix: Use consistent coercion for all four fields.

**M5. `_subject_scope_matches` silently denies when `subject_type=user` but `subject_value=""`**
- `policy.py`: A rule with `subject_type: user` and empty `subject_value` never matches anything — no warning logged. Users get silent policy denial.
- Fix: Log a warning, or reject such rules during validation.

**M6. Manager shuts down handlers sequentially — one hung handler blocks all**
- `manager.py` `_shutdown_handlers`: Sequential `await` in a loop. A stuck MCP subprocess blocks all subsequent handler shutdowns.
- Fix: Use `asyncio.gather(..., return_exceptions=True)` with per-handler `asyncio.wait_for` timeout.

**M7. Legacy migration sets maximally permissive policy (`effect="allow"`)**
- `adapters/mcp_legacy_config.py`: Migrated clients get `policy=[PolicyRule(subject="*", effect="allow")]`. Any user/channel can invoke any tool with no approval. This is a security regression for deployments that relied on the approval system.
- Fix: Consider `effect="ask"` as default, or log a prominent warning.

**M8. `_reload_driver_background` is sync calling `asyncio.create_task`**
- `routers/mcp_oauth.py`: Sync function calling `asyncio.create_task` — fragile, no running-loop guard, inconsistent with the async pattern in `mcp.py`.
- Fix: Make it `async def` and unify with `_reload_driver_best_effort`.

**M9. Code duplication: `_approval_display_fields` copy-pasted in two routers**
- Identical function in `routers/approval.py` and `routers/console.py`. Also, 3 helpers duplicated between `routers/mcp.py` and `routers/mcp_oauth.py` (`_workspace_credential_store`, `_load_optional_credential`, `_mcp_card_path`).
- Fix: Extract shared helpers into a common module.

**M10. Console: `PendingApproval.tool_source` typed as required but may be absent**
- `console/src/api/modules/console.ts`: `tool_source: string` (required), but `ApprovalCardProps.toolSource` is `string | undefined`. Pre-existing approval entries lacking this field will cause runtime type mismatch.
- Fix: Make it `tool_source?: string`.

---

### 🟡 MINOR (16)

| # | File | Finding |
|---|------|---------|
| m1 | `policy.py` | `_source_matches` returns `False` for unknown `source_type` with no warning — typos like `"chanel"` silently disable rules |
| m2 | `policy.py` | `rate_limit` conditions accepted but silently ignored (returns `True`) — no enforcement, no validation error |
| m3 | `credentials/store.py` | `env:` prefix returns empty string for missing env vars instead of raising error — confusing downstream failures |
| m4 | `credentials/store.py` | `chmod 0o600` failure silently swallowed — credentials may be world-readable on shared filesystems |
| m5 | `manager.py` | `_sync_handler_runtime_metadata` reads YAML from disk on every invocation — unnecessary I/O under high call rates |
| m6 | `handler.py` | `_request_approval` embeds full `context.extras` into approval payload — may leak sensitive arguments |
| m7 | `errors.py` | `DriverPermissionDeniedError.to_user_message` exposes subject identity (e.g., `user:alice@example.com`) |
| m8 | `storage.py` | `dump_card` uses `allow_unicode=False` — non-ASCII characters in driver names produce unreadable YAML |
| m9 | `handlers/mcp.py` | `_subject_from_context` is defined but never called (dead code) |
| m10 | `adapters/mcp_console.py` | Secret masking reveals 5-7 chars for 9-12 char values — poor masking ratio |
| m11 | `adapters/acp_config.py` | `acp_agent_config_to_card` copies env vars directly into endpoint dict without secret separation |
| m12 | `service_factories.py` | `set_driver_manager` silently skipped if runner not yet initialized — silent failure worse than crash |
| m13 | `MCPAccessModal.tsx` | No input validation on empty `subject_value` when `subject_type` is "user" — saves a rule that matches nobody |
| m14 | `MCPAccessModal.tsx` | No dirty-state tracking — closing modal silently discards unsaved policy changes |
| m15 | `useMCP.ts` | `updatePolicy` callback missing `message` in `useCallback` dependency array |
| m16 | `accessPolicy.ts` | `normalizeSourceValue` silently corrects unrecognized values — tight coupling with backend's allowed values |

---

### 🔵 TEST COVERAGE GAPS

| # | Gap | Impact |
|---|-----|--------|
| T1 | Module-level mutable state (`SETUP_EVENTS`, `FakeStdIOClient.instances`) shared across tests — flaky under `pytest-xdist` | Test reliability |
| T2 | `MCPAccessModal` (603 lines, most complex new component) has **zero** render/interaction tests | Frontend quality |
| T3 | `useMCP.updatePolicy` hook has no unit test | Frontend quality |
| T4 | `_next_pending_request` in e2e tests uses unbounded spin loop — can hang indefinitely | Test reliability |
| T5 | No concurrent access test for `CredentialStore` despite no file locking | Correctness |
| T6 | E2E tests use fakes, not real fixture servers (`tests/fixtures/mcp/*.py` are never imported) — mislabeled | Test accuracy |
| T7 | No test for policy evaluation performance with large rule sets | Performance |

---

### Summary

**Architecture**: The DriverCard + DriverManager + PolicyRule design is sound. The separation of concerns (card = config, handler = runtime, policy = access control, credential = secrets) is clean and well-structured.

**Must fix**: The two critical issues (C1 path traversal, C2 secret repr) are the blockers. C1 is an arbitrary file write primitive if card names come from API input; C2 is a credential leak waiting for someone to add a debug log.

**Should fix**: M1 (credential store race), M2 (uncaught execution errors), M5 (silent policy denial), and M7 (permissive migration default) are the highest-priority major items.

**Test gaps**: The backend test coverage is thorough (145 tests passing), but the frontend's most complex new component has no render tests, and module-level mutable state in test fakes could cause flakiness under parallel execution.

Overall this is a well-designed and well-tested PR for a complex feature. The critical items are narrow and fixable. Looking forward to the next iteration.

xiaoming-qxm:
Follow-up from the focused pass above; two details were not covered explicitly in the existing review:

1. `src/qwenpaw/drivers/adapters/mcp_legacy_config.py:195` / Console MCP policy adapter:
   Migrated legacy MCP clients get `PolicyRule(subject="*", effect="allow")`, whose target defaults to `*/*`. The Console policy adapter only treats `target.kind == "tool"` and `target.name == "*"` as a client-level override. That means migrated clients can display as default deny + unmanaged rule while runtime still allows every tool, and saving a new policy preserves the unmanaged allow-all. Please migrate this as a managed MCP tool-wide rule, for example `PolicyTarget(kind="tool", name="*")`, or teach the adapter to round-trip the legacy wildcard target as a client override.

2. `src/qwenpaw/app/routers/mcp.py:783`:
   `/mcp/tools/{client_key}` is scoped to one client, but it calls `manager.list_capabilities(protocol="mcp")`, which asks every active MCP handler to list tools and filters afterward. An unrelated MCP server's `list_tools()` failure or stall can make this client's tool list return 502. Please query only the target driver/client, or isolate per-driver failures before filtering.

Existing review already covers the card-name path traversal issue and the source-value normalization issue at a high level, so I am not repeating those here.

xiaoming-qxm (commented):
Code Review — Codex focused pass for PR #5067

I found four actionable issues in this pass:

1. [High] `src/qwenpaw/drivers/storage.py` builds card paths from the raw runtime `name`. A card name containing path separators or `..` can escape the intended storage namespace, create nested files, and break later lookup/delete behavior because other paths key by the sanitized id. Please derive all filesystem paths from the canonical/sanitized card id, or strictly reject unsafe names before writing.

2. [High] `src/qwenpaw/drivers/adapters/mcp_legacy_config.py` migrates legacy clients with `PolicyRule(subject="*", effect="allow")`, whose target defaults to `*/*`. The Console MCP policy adapter only recognizes a managed client-wide override when the target is `tool/*`, so migrated clients can appear as default-deny or unmanaged in the UI while runtime behavior still allows every tool. Saving from Console can preserve that unmanaged allow-all rule. Please migrate to a managed MCP tool-wide target, for example `PolicyTarget(kind="tool", name="*")`, or teach the adapter to round-trip the legacy wildcard as the client-level override.

3. [Medium] `console/src/pages/Agent/MCP/accessPolicy.ts` silently rewrites unknown but valid `source` values in `normalizeSourceValue()`. Custom channel/app/agent identities can be collapsed to `console` or `Creator`, so simply opening and saving the policy UI can corrupt source-scoped permissions. Unknown source values should be preserved as custom values unless the user explicitly changes them.

4. [Medium] `src/qwenpaw/app/routers/mcp.py` implements `/mcp/tools/{client_key}` by calling `manager.list_capabilities(protocol="mcp")`, which asks every active MCP handler to list tools and filters afterward. A failure or stall in an unrelated MCP server can make the target client's tool list return 502. Please query only the selected driver/client, or isolate per-driver failures before filtering.

These are scoped to behavior that can affect existing users after migration or when multiple MCP clients are configured.

xiaoming-qxm (commented):
Removed.

xiaoming-qxm:
Addressed the latest architecture review in two commits:

- `0feb376d fix(driver): decouple approval and runtime validation`
  - Introduced a Driver-layer `ApprovalGate` protocol and moved the QwenPaw approval-service integration into the app layer (`QwenPawDriverApprovalGate`), removing the core driver -> app dependency.
  - Stopped `DriverManager` from re-reading DriverCard YAML on `list_capabilities` / `invoke_capability`; runtime metadata now changes via explicit register/reload events.
  - Added protocol endpoint validators and wired MCP endpoint validation through `register_handler_type(...)`.
  - Extracted credential binding / implicit auth-header resolution out of the MCP handler into reusable credential binding helpers.
  - Added `CredentialStoreProtocol`, leaving the YAML store as the current implementation.
  - Changed MCP `list_tools()` to return raw MCP Tool schemas and replaced lifecycle polling with event waiting.
  - Hardened policy validation: target kind must match known capability kinds, and configured-but-unenforced `rate_limit` is rejected.
  - Removed LLM prompt-style retry guidance from `DriverPermissionDeniedError.to_user_message()`.

- `8beb2abe refactor(driver): stabilize capability tool adapter`
  - Replaced the `DriverCapabilityTool.__new__` anonymous-class pattern with a stable `ToolBase` subclass, so `isinstance(tool, DriverCapabilityTool)` now works.

Validated locally with:

- `uv run pytest tests/unit/agents/tools/test_driver_capability_tool.py tests/unit/drivers/test_handler.py tests/unit/drivers/test_manager.py tests/unit/drivers/test_mcp_handler.py tests/unit/drivers/test_mcp_stateful_client.py tests/unit/drivers/test_contracts.py tests/unit/drivers/test_credentials_store.py -q`
- `uv run pre-commit run --all-files`

GitHub PR checks are green after push.

I agree the remaining router/service extraction, `mcp_console.py` split, policy type file split, pure `validate_card()` / `coerce_card()` separation, and singular `credential` deprecation are valid follow-up refactors. I kept them out of this response because they are broader API/module-organization changes and would be safer as separate PRs instead of mixing into this review-fix pass.

rayrayraykk (commented):
To keep this PR focused and easy to review, I'd suggest avoiding adding unnecessary or redundant unit tests. We should focus strictly on covering the new behavior introduced in this change.

src/qwenpaw/agents/react_agent.py
```diff
@@ -393,7 +401,14 @@ def _create_toolkit(
         except Exception as e:  # pylint: disable=broad-except
             logger.warning(f"Failed to register Coding Mode tools: {e}")
 
-        return Toolkit(tools=tool_instances, mcps=self._mcp_clients or None)
+        for tool in self._external_tools:
```
rayrayraykk: @qbc2016 Please hv a look at this.

src/qwenpaw/agents/react_agent.py
```diff
@@ -461,6 +476,9 @@ def _build_sys_prompt(self) -> str:
         if multimodal_hint:
```
rayrayraykk: All the change to `SYSTEM_PROMPT` should be included into a `builder`. cc @XiuShenAl

src/qwenpaw/app/approvals/driver_gate.py
```diff
@@ -0,0 +1,110 @@
+# -*- coding: utf-8 -*-
+"""QwenPaw application approval gate for Driver policy."""
+
+from __future__ import annotations
+
+from qwenpaw.constant import TOOL_GUARD_APPROVAL_TIMEOUT_SECONDS
```
rayrayraykk: Please use relative imports here.

src/qwenpaw/app/approvals/driver_gate.py
```diff
@@ -0,0 +1,110 @@
...
+                f"Driver '{driver_ref}' requires approval for "
+                f"{context.operation}."
+            )
+        if not session_id:
+            raise ApprovalRequiredError(
+                "Driver approval required but request_context.session_id "
+                f"is missing: {context.subject} -> {context.driver_name}",
+            )
+
+        from qwenpaw.app.approvals import get_approval_service
```
rayrayraykk: Same.

rayrayraykk (changes requested):
Overall, please use relative imports in our repo.

src/qwenpaw/drivers/credentials/store.py
```diff
@@ -0,0 +1,220 @@
+# -*- coding: utf-8 -*-
```
rayrayraykk, line 1: `Sync` func should not appear.

rayrayraykk (commented):
### Issue 1: Synchronous File I/O in `YAMLCredentialStore`
* **File:** `src/qwenpaw/drivers/credentials/store.py`
* **Problem:** All operations in `YAMLCredentialStore` (such as `get`, `put`, `delete`, and `list_refs`) are implemented as synchronous methods but contain blocking OS calls. In `store.py` under `_write_root()`, `os.fsync(tmp.fileno())` is called to force a disk flush, followed by `os.replace(tmp_name, path)` and `_restrict_file_permissions(self._path)`. In `_read_root()`, a synchronous read is performed using `yaml.safe_load(self._path.read_text(encoding="utf-8"))`.
* **Problematic Call Chain:** These synchronous methods are directly called inside async route handlers without being awaited, meaning they are not offloaded to a thread pool via `asyncio.to_thread()`. In `MCPConfigService.create_client()` and `update_client()`, `self._driver_config.credential_store.put(credential)` and `delete(credential.ref)` are called synchronously. This also occurs along the credential resolution path (triggered on every tool execution) in `handler.py` under `_guarded_execute()`, where `await self._credential_provider.resolve()` internally triggers `store.get()` and `yaml.safe_load()`.
* **Impact:** Every MCP tool execution and every MCP client creation/update performs synchronous disk I/O directly on the event loop, blocking all other concurrent requests.

---

### Issue 2: Sync `os.fsync()` in `storage.dump_card()` Embedded in Async Call Chains
* **File:** `src/qwenpaw/drivers/storage.py`
* **Problem:** The `dump_card(card: DriverCard, path: Path)` is a synchronous function that executes `os.fsync(tmp.fileno())` to force disk sync. However, it is invoked directly inside multiple async methods without `await asyncio.to_thread()`. 
* **Problematic Call Chain:** This occurs in `DriverManager.reload_driver()` (calling `dump_card(card, canonical)` and `delete_card_paths_for_name()`), `DriverManager.register_driver()` (calling `dump_card(card, path)` and `delete_card_paths_for_name()`), and `DriverConfigService.save_card()` (calling `dump_card(card, path)` and `delete_card_paths_for_name()`). Note that `delete_card_paths_for_name()` also runs blocking synchronous `rglob` and `unlink` calls.
* **Impact:** Saving MCP client configurations (create, update, or toggle) triggers `os.fsync()` directly on the main thread, blocking the event loop for 100ms to 500ms+.

---

### Issue 3: Synchronous I/O (`list_card_paths()` + `load_card()`) on Async Paths
* **File:** `src/qwenpaw/drivers/storage.py`
* **Problem:** The function `list_card_paths(cards_dir: Path)` performs synchronous directory traversal via `cards_dir.rglob("*")` and executes a blocking `path.is_file()` stat check. 
* **Problematic Call Chain:** This synchronous logic is called directly inside async methods: in `DriverManager.list_drivers()` (triggered on every single `ensure_driver_active()`), in `DriverManager.build_drivers()` (triggered at startup), and in `DriverConfigService.list_cards()`. In all these locations, the code synchronously loops through the files and calls `load_card(path)`, which parses YAML synchronously.
* **Impact:** Since `ensure_driver_active()` is invoked frequently across multiple high-traffic routes (like `list_tools` and `update_policy`), executing synchronous filesystem traversals on every call degrades server throughput under concurrent load.

---

### Issue 4: Event Loop Lag in `DriverConfigWatcher._snapshot()` During Async Polling
* **File:** `src/qwenpaw/app/driver_config_watcher.py`
* **Problem:** In `DriverConfigWatcher`, the async loop handler `_check_once()` invokes `self._snapshot()`. However, `_snapshot()` is completely synchronous; it loops through files via `list_card_paths(self._cards_dir)` and fetches file metadata using the blocking `path.stat().st_mtime`.
* **Impact:** As the `drivers` directory grows with more files, this synchronous snapshotting during the 2-second interval polling causes measurable event loop delays.

---

### Issue 5: Redundant Tool Discovery on Every Request (Missing Cache)
* **File:** `src/qwenpaw/drivers/adapters/agentscope_tool.py`
* **Problem:** On every single `stream_query`, the system invokes `await build_driver_agent_tools(driver_manager, request_context)`. This helper calls `list_tools()` for every active driver, which queries the tool schema over the MCP protocol (using stdio or HTTP calls). 
* **Impact:** A full network/process round-trip for tool discovery is triggered on every single user message without any caching. Note that `mcp_stateful_client.py` implements a `_cached_tools` attribute, but the Driver layer does not leverage it.

---

### Issue 6: N+1 Synchronous Credential Reads in `MCPConfigService.list_clients()`
* **Problem:** In `MCPConfigService.list_clients()`, the code uses a list comprehension: `[self.build_info_from_card(card) for card in self.list_cards()]`. The underlying `build_info_from_card()` method calls `load_optional_credential()` for each card, which performs a synchronous YAML file read.
* **Impact:** If there are N MCP clients configured, listing them results in N sequential, synchronous file reads executed directly on the event loop.

---

### Issue 7: Inline Import Inside Hot Path
* **File:** `src/qwenpaw/app/approvals/driver_gate.py`
* **Problem:** Inside `async def request_approval()`, there is an inline import statement: `from qwenpaw.app.approvals import get_approval_service`.
* **Impact:** Placing inline imports inside a high-frequency execution path like approval requests adds unnecessary runtime overhead and hurts code readability.

---

### Issue 8: Synchronous `DriverConfigService.list_cards()` Performing Disk I/O
* **Problem:** The method `DriverConfigService.list_cards()` is synchronous but internally traverses directories via `list_card_paths()` and loads files via `load_card()`. This method is called by `ensure_mcp_display_name_unique()`, which is directly invoked inside the async context of `create_client()` and `update_client()`.
* **Impact:** This introduces synchronous directory iteration and file loading directly into the async request handling flow.

src/qwenpaw/drivers/policy_types.py
```diff
@@ -0,0 +1,205 @@
...
+            default_effect=_coerce_policy_effect(
+                value.get("default_effect"),
+                default=POLICY_EFFECT_DENY,
+            ),
+            rules=[
+                _coerce_policy_rule(item)
+                for item in list(value.get("rules") or [])
+            ],
+        )
+    return DriverPolicy()
```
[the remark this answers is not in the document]
xiaoming-qxm: Fixed in e7ff7666. `coerce_driver_policy()` now validates dict-backed `rules` explicitly: omitted/null rules still normalize to an empty list, but mapping/string values raise `DriverCardError` instead of being iterated and silently coerced. I also added focused regression coverage for those malformed inputs.

src/qwenpaw/app/approvals/driver_gate.py
xiaoming-qxm: Addressed the relative-import review comments in b4cf55f4.

- Updated the Driver-related source files to use package-relative imports instead of `qwenpaw...` absolute imports.

xiaoming-qxm:
> Overall, please use relative imports in our repo.

Addressed the relative-import review comments in b4cf55f4.

- Updated the Driver-related source files to use package-relative imports instead of `qwenpaw...` absolute imports.

src/qwenpaw/drivers/errors.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
```
rayrayraykk, line 1: We should raise from `shema.exceptions` after merging: #5078

src/qwenpaw/drivers/policy.py
```diff
@@ -0,0 +1,332 @@
...
+    subject: str
+    driver_name: str
+    protocol: str
+    operation: str = DRIVER_OPERATION_INVOKE
+    target: PolicyTarget = field(default_factory=PolicyTarget)
+    subjects: tuple[str, ...] = field(default_factory=tuple)
+    workspace_id: str = ""
+    request_context: dict[str, str] = field(default_factory=dict)
+    now: datetime = field(
+        default_factory=lambda: datetime.now(timezone.utc),
```
rayrayraykk, line 69: Default UTC might be confused to user. Could we use time zone in config?

src/qwenpaw/drivers/adapters/agentscope_tool.py
```diff
@@ -0,0 +1,226 @@
...
+    DriverInvocationResult,
+)
+
+logger = logging.getLogger(__name__)
+
+DriverInvoker = Callable[[DriverInvocation], Awaitable[DriverInvocationResult]]
+
+
+def _text_block(text: str) -> Any:
+    from agentscope.message import TextBlock
```
rayrayraykk: Please mv to top.

src/qwenpaw/drivers/policy.py
```diff
@@ -0,0 +1,332 @@
...
+
+def _subject_scope_matches(
+    selector: PolicyPrincipal,
+    context: DriverInvocationContext,
+) -> bool:
+    subject_type = _normalize_selector_part(selector.subject_type)
+    subject_value = _selector_value(selector.subject_value)
+    if subject_type in {POLICY_TARGET_WILDCARD, PRINCIPAL_SUBJECT_ALL}:
+        return True
+    if subject_value in {"", POLICY_TARGET_WILDCARD}:
```
rayrayraykk: `If subject_value=="": return False`?

xiaoming-qxm:
> ### Issue 1: Synchronous File I/O in `YAMLCredentialStore`
> * **File:** `src/qwenpaw/drivers/credentials/store.py`
> * **Problem:** All operations in `YAMLCredentialStore` (such as `get`, `put`, `delete`, and `list_refs`) are implemented as synchronous methods but contain blocking OS calls. In `store.py` under `_write_root()`, `os.fsync(tmp.fileno())` is called to force a disk flush, followed by `os.replace(tmp_name, path)` and `_restrict_file_permissions(self._path)`. In `_read_root()`, a synchronous read is performed using `yaml.safe_load(self._path.read_text(encoding="utf-8"))`.
> * **Problematic Call Chain:** These synchronous methods are directly called inside async route handlers without being awaited, meaning they are not offloaded to a thread pool via `asyncio.to_thread()`. In `MCPConfigService.create_client()` and `update_client()`, `self._driver_config.credential_store.put(credential)` and `delete(credential.ref)` are called synchronously. This also occurs along the credential resolution path (triggered on every tool execution) in `handler.py` under `_guarded_execute()`, where `await self._credential_provider.resolve()` internally triggers `store.get()` and `yaml.safe_load()`.
> * **Impact:** Every MCP tool execution and every MCP client creation/update performs synchronous disk I/O directly on the event loop, blocking all other concurrent requests.
> 
> ### Issue 2: Sync `os.fsync()` in `storage.dump_card()` Embedded in Async Call Chains
> * **File:** `src/qwenpaw/drivers/storage.py`
> * **Problem:** The `dump_card(card: DriverCard, path: Path)` is a synchronous function that executes `os.fsync(tmp.fileno())` to force disk sync. However, it is invoked directly inside multiple async methods without `await asyncio.to_thread()`.
> * **Problematic Call Chain:** This occurs in `DriverManager.reload_driver()` (calling `dump_card(card, canonical)` and `delete_card_paths_for_name()`), `DriverManager.register_driver()` (calling `dump_card(card, path)` and `delete_card_paths_for_name()`), and `DriverConfigService.save_card()` (calling `dump_card(card, path)` and `delete_card_paths_for_name()`). Note that `delete_card_paths_for_name()` also runs blocking synchronous `rglob` and `unlink` calls.
> * **Impact:** Saving MCP client configurations (create, update, or toggle) triggers `os.fsync()` directly on the main thread, blocking the event loop for 100ms to 500ms+.
> 
> ### Issue 3: Synchronous I/O (`list_card_paths()` + `load_card()`) on Async Paths
> * **File:** `src/qwenpaw/drivers/storage.py`
> * **Problem:** The function `list_card_paths(cards_dir: Path)` performs synchronous directory traversal via `cards_dir.rglob("*")` and executes a blocking `path.is_file()` stat check.
> * **Problematic Call Chain:** This synchronous logic is called directly inside async methods: in `DriverManager.list_drivers()` (triggered on every single `ensure_driver_active()`), in `DriverManager.build_drivers()` (triggered at startup), and in `DriverConfigService.list_cards()`. In all these locations, the code synchronously loops through the files and calls `load_card(path)`, which parses YAML synchronously.
> * **Impact:** Since `ensure_driver_active()` is invoked frequently across multiple high-traffic routes (like `list_tools` and `update_policy`), executing synchronous filesystem traversals on every call degrades server throughput under concurrent load.
> 
> ### Issue 4: Event Loop Lag in `DriverConfigWatcher._snapshot()` During Async Polling
> * **File:** `src/qwenpaw/app/driver_config_watcher.py`
> * **Problem:** In `DriverConfigWatcher`, the async loop handler `_check_once()` invokes `self._snapshot()`. However, `_snapshot()` is completely synchronous; it loops through files via `list_card_paths(self._cards_dir)` and fetches file metadata using the blocking `path.stat().st_mtime`.
> * **Impact:** As the `drivers` directory grows with more files, this synchronous snapshotting during the 2-second interval polling causes measurable event loop delays.
> 
> ### Issue 5: Redundant Tool Discovery on Every Request (Missing Cache)
> * **File:** `src/qwenpaw/drivers/adapters/agentscope_tool.py`
> * **Problem:** On every single `stream_query`, the system invokes `await build_driver_agent_tools(driver_manager, request_context)`. This helper calls `list_tools()` for every active driver, which queries the tool schema over the MCP protocol (using stdio or HTTP calls).
> * **Impact:** A full network/process round-trip for tool discovery is triggered on every single user message without any caching. Note that `mcp_stateful_client.py` implements a `_cached_tools` attribute, but the Driver layer does not leverage it.
> 
> ### Issue 6: N+1 Synchronous Credential Reads in `MCPConfigService.list_clients()`
> * **Problem:** In `MCPConfigService.list_clients()`, the code uses a list comprehension: `[self.build_info_from_card(card) for card in self.list_cards()]`. The underlying `build_info_from_card()` method calls `load_optional_credential()` for each card, which performs a synchronous YAML file read.
> * **Impact:** If there are N MCP clients configured, listing them results in N sequential, synchronous file reads executed directly on the event loop.
> 
> ### Issue 7: Inline Import Inside Hot Path
> * **File:** `src/qwenpaw/app/approvals/driver_gate.py`
> * **Problem:** Inside `async def request_approval()`, there is an inline import statement: `from qwenpaw.app.approvals import get_approval_service`.
> * **Impact:** Placing inline imports inside a high-frequency execution path like approval requests adds unnecessary runtime overhead and hurts code readability.
> 
> ### Issue 8: Synchronous `DriverConfigService.list_cards()` Performing Disk I/O
> * **Problem:** The method `DriverConfigService.list_cards()` is synchronous but internally traverses directories via `list_card_paths()` and loads files via `load_card()`. This method is called by `ensure_mcp_display_name_unique()`, which is directly invoked inside the async context of `create_client()` and `update_client()`.
> * **Impact:** This introduces synchronous directory iteration and file loading directly into the async request handling flow.

Addressed the synchronous file I/O concern by moving Driver credential and card storage behind async public APIs in 57581b7a

The credential store now exposes AsyncCredentialStore, with the YAML/file-system operations kept as private synchronous implementation details and isolated via asyncio.to_thread. DriverCard persistence is similarly routed through AsyncDriverCardStore, so DriverManager, MCP config APIs, OAuth callbacks, and the config watcher no longer perform direct blocking file I/O on the async request/runtime path.

src/qwenpaw/drivers/contracts.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+def _validate_value_source_bindings(
+    card: DriverCard,
+    section_name: str,
+    section: dict[str, Any],
+) -> None:
+    aliases = set(iter_credential_refs(card))
+    for output_name, spec in section.items():
+        if isinstance(spec, dict) and "source" in spec:
```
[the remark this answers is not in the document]
xiaoming-qxm: Fixed in fb665e29. `_validate_value_source_bindings()` now accepts plain strings as literal values, rejects non-string specs, rejects mapping specs without `source`, and validates `source=literal` values plus credential `format` fields without implicit stringification. Added regression coverage for the malformed binding cases.

src/qwenpaw/app/driver_config_watcher.py
```diff
@@ -0,0 +1,127 @@
...
+        snapshot: dict[str, tuple[str, float]] = {}
+        for path in list_card_paths(self._cards_dir):
+            try:
+                mtime = path.stat().st_mtime
+            except FileNotFoundError:
+                continue
+            snapshot[path.stem] = (
+                path.relative_to(self._cards_dir).as_posix(),
+                mtime,
+            )
```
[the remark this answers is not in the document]
xiaoming-qxm: Fixed in fb665e29. Driver card snapshots are now keyed by protocol-relative path rather than `path.stem`, with the driver name stored in the snapshot value. Duplicate driver names are logged as warnings because runtime names remain globally unique, and the watcher maps changed/removed paths back to names without silently overwriting entries. Added regression coverage for same-name cards under different protocol directories.

xiaoming-qxm:
> To keep this PR focused and easy to review, I'd suggest avoiding adding unnecessary or redundant unit tests. We should focus strictly on covering the new behavior introduced in this change.

Unit tests removed.

src/qwenpaw/drivers/credentials/store.py
xiaoming-qxm: Addressed the synchronous file I/O concern by moving Driver credential and card storage behind async public APIs in https://github.com/agentscope-ai/QwenPaw/commit/57581b7abb36114d5f14c87e1634fe49b927d05c

src/qwenpaw/drivers/errors.py
xiaoming-qxm: Done. Driver exceptions now align with QwenPaw’s shared exception hierarchy from src/qwenpaw/exceptions.py: configuration errors derive from ConfigurationException, runtime errors derive from AgentRuntimeErrorException, while keeping DriverError as the Driver-local base.

src/qwenpaw/drivers/policy.py
xiaoming-qxm: Done. Driver policy time evaluation now uses the configured user_timezone through a Driver-local helper and only falls back to UTC when the config is missing or invalid.

src/qwenpaw/drivers/adapters/agentscope_tool.py
xiaoming-qxm: Done. Moved the AgentScope imports to the top-level import section and kept the module import order consistent.

src/qwenpaw/drivers/policy.py
xiaoming-qxm: Done. Empty structured subject_value now fails matching explicitly before wildcard handling, so an empty value no longer behaves like an accidental wildcard.

src/qwenpaw/agents/react_agent.py
xiaoming-qxm: Changes were not existed since rebased.

rayrayraykk (approved):
LGTM

PR #5069, opened 2026-06-10
feat(agents): add visual model fallback for text-only primary models

### LINKED ISSUE #4992: [Feature]: 支持独立视觉模型配置（Visual Model Fallback）
## 功能建议

当前 QwenPaw 的图片/视频处理完全依赖主模型的多模态能力。如果主模型不支持视觉输入（如 LongCat-2.0-Preview），图片就无法被理解。

## 建议方案

增加一个独立的视觉模型配置项（如 `visual_model`），当主模型不支持多模态时，自动调用视觉模型对图片进行描述/转述，然后将文本结果交给主模型继续处理。

## 使用场景

- 用户主模型为纯文本模型（如 deepseek-v4-flash），但偶尔需要处理图片
- 用户不想为了视觉能力而更换主模型
- 类似"视觉中转站"的概念：图片 → 视觉模型转文字 → 主模型处理

## 配置示例（建议）

在 `agent.json` 中：

{
  "active_model": {
    "provider_id": "deepseek",
    "model": "deepseek-v4-flash"
  },
  "visual_model": {
    "provider_id": "zhipu-cn",
    "model": "glm-4v-flash"
  }
}

## 预期行为

1. 收到图片时，先检测主模型是否支持多模态
2. 如果支持，直接发给主模型（现有行为不变）
3. 如果不支持，自动调用 `visual_model` 对图片进行描述
4. 将视觉模型的文本输出作为上下文，交给主模型继续处理

## 参考

类似功能在 Open WebUI 和其他 LLM 客户端中已有实现。

yuanxs21:
## Description

When the primary LLM is text-only, users can configure an optional per-agent **visual model** to transcribe image/video into text before the main model processes the request. Transcription runs at the formatter layer (`FileBlockSupportFormatter.format` in `model_factory`) via `apply_visual_fallback_to_messages()`, and replaces media blocks with `[Image description: …]` / `[Video description: …]` in the API payload.

Native multimodal primary models skip the path entirely (no media history scan). On transcription failure the original media block is kept (fail-open) so downstream strip/normalization behaves as when visual fallback is unset. `view_image` / `view_video` remain unchanged vs main; after a successful transcription, conflicting multimodal-unsupported tool-result hints are filtered in the service layer.

This PR:
- Adds per-agent `visual_model` config (`ModelSlotConfig`) and Console UI under **Settings → Models**
- Implements `visual_model_service` with URL / `file://` / local path / `Base64Source` support
- Wires the same formatter onto `agent.formatter` in `builder` so request-time media strip does not remove blocks before fallback
- Wraps visual calls with `RetryChatModel` using a **fail-fast** policy (`max_retries=1`, acquire timeout capped at 20s) while preserving primary concurrency / QPM / 429 pause settings
- Adds success LRU cache (128) and a **60s failure negative cache** (`time.monotonic`) to avoid repeated failed visual calls
- Soft-constrains final description length via prompt (~150 words / description only); does **not** override visual-model `max_tokens` or thinking settings
- Records visual-model token usage under the visual model name with a ` (visual fallback)` suffix
- Adds unit coverage in `tests/unit/providers/test_visual_model_service.py` (fast path, fail-fast config, negative cache, hint filtering, fail-open)

**Related Issue:** Fixes #4992 or Relates to #4992

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

## Testing

[How to test these changes]

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]


RerankerGuo (changes requested):
I validated exact head `284605ae` and the clean current-main trial tree `15cd7b06` based on `bacac741`. The fallback and provider/model cache partition work, but the success cache silently returns stale visual descriptions when a media URL points to changed content.

`_cache_key()` hashes the raw URL/path string, not the resolved local content or any invalidation signature. `_TX_CACHE` has an LRU size bound but no time/content invalidation. Reusing a common path such as `latest.png`, `screenshot.png`, or a downloaded attachment filename therefore returns the prior image's description until that entry happens to be evicted or the process restarts.

I exercised the real `_transcribe()` -> `_invoke_visual_transcription()` -> `_media_url()` path with a fake provider at the external transport boundary:

1. write `alpha` to `latest.png`;
2. transcribe it and observe the provider receive the real generated data URL;
3. overwrite the same path with `bravo`;
4. transcribe the same source again;
5. clear only `_TX_CACHE` and repeat as a control.

Exact and current-main trial both produce:

```text
keys_equal=True
first='description:alpha'
cached_after_rewrite='description:alpha'
uncached_after_rewrite='description:bravo'
calls=['alpha', 'bravo']
```

The second request never reaches the visual provider and injects the old `alpha` description for the new `bravo` file. The control proves the local-file conversion and provider response would otherwise observe the new bytes.

Please make cache identity follow media identity:

- for local files, include a content digest (preferably computed while reading the bytes already needed for the data URL) rather than only the path;
- perform the file read/hash off the event loop, especially because the accepted size is up to 100 MiB;
- for mutable HTTP(S) URLs, use a finite success TTL or a validator/content-aware strategy instead of an indefinitely reusable URL-only entry;
- keep the existing provider/model partition and bounded LRU.

Add a regression that transcribes a local path, overwrites it in place, then asserts the next call reaches the provider and returns a description of the replacement bytes. A separate assertion should retain the current cache hit for unchanged content.

Verification:

- exact visual fallback suite: `12 passed`;
- semantic current-main related visual/builder suite: `13 passed`;
- Console `tsc --noEmit`: pass;
- exact `py_compile` and `git diff --check`: pass.

The current tests cover success/failure caching and failure TTL, but not success invalidation after the source changes.

PR #5125, opened 2026-06-11
fix(desktop): harden Tauri Windows CI against crates.io fetch failures

jinglinpeng:
## Summary
- Pre-fetch Tauri Rust dependencies with retry and offline build on Windows to avoid transient crates.io TLS failures during `tauri build`.
- Add uv, npm, and Rust caches to Tauri desktop CI jobs so Windows builds align with macOS and reuse dependency artifacts across runs.

## Test plan
- [ ] Verify `build-tauri-windows` succeeds on fork CI (workflow dispatch triggered on `fix/tauri-windows-cargo-fetch`).
- [ ] Confirm `build-tauri-macos` still passes with added npm/Rust caches.
- [ ] Check staged Windows installer artifact is produced and uploadable.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -161,13 +161,23 @@ jobs:
         with:
           python-version: "3.10"
 
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v5
+
       - name: Set up Node
         uses: actions/setup-node@v4
         with:
           node-version: "20"
+          cache: "npm"
+          cache-dependency-path: console/package-lock.json
 
       - name: Set up Rust
         uses: dtolnay/rust-toolchain@stable
+
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: console/src-tauri
 
       - name: Install NSIS
         uses: negrutiu/nsis-install@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/desktop-release.yml
```diff
@@ -161,14 +161,24 @@ jobs:
...
 
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v5
+
       - name: Set up Node
         uses: actions/setup-node@v4
         with:
           node-version: "20"
+          cache: "npm"
+          cache-dependency-path: console/package-lock.json
```
rayrayraykk, line 172: This will fail if source files change without any new dependencies being added.

.github/workflows/desktop-release.yml
```diff
@@ -227,10 +237,17 @@ jobs:
...
```
rayrayraykk, line 241: Same as above

rayrayraykk (changes requested):

.github/workflows/desktop-release.yml
jinglinpeng: Thanks for the review! This shouldn't be an issue here: setup-node's cache: npm option caches the package manager's download cache (~/.npm), not node_modules (see [actions/setup-node docs](https://github.com/actions/setup-node#caching-global-packages-data)).

The cache key is based on the hash of console/package-lock.json:

If source files change but dependencies don't, the key stays the same → cache hit → npm ci still runs in full, just resolving packages from the local cache instead of the network.
If dependencies change, the key misses → npm ci falls back to downloading from the registry as usual.
Either way the build never fails because of the cache — it only affects download speed. The failure mode you're describing applies to caching node_modules directly via actions/cache without restore-keys, which is not what's happening here.

Also note the build-windows job in this same workflow has been using this exact configuration (lines 45–50) without issues; this change just aligns the two Tauri jobs with it.

rayrayraykk (approved):

PR #5160, opened 2026-06-12
feat(yuanbao): support quoted messages and unify media download pipeline

hongxicheng:
## Description

This PR brings the Yuanbao channel to feature parity with WeCom / DingTalk on
**quoted-message handling**, and along the way refactors the media download
path to fix two latent bugs and remove dead code.

### What's new

1. **Quoted message support** (`cloud_custom_data.quote`).
   Yuanbao's quote payload only carries `id` / `desc` / `sender_*` — there is
   no url/key and no API to refetch the original message. We therefore
   surface a textual placeholder describing the **kind** of the quoted item
   and (when available) its **filename**, so the model can match it against
   earlier turns in chat history:
   - `[quoted message: 你好啊]`     (type=1, text)
   - `[quoted image]`                (type=2, no desc)
   - `[quoted file: report.pdf]`     (type=3, file)
   - `[quoted audio: voice.mp3]`     (type=3, audio suffix)

   **Cross-channel consistency:** the placeholder is **prepended to the
   first `TextContent`** (joined with `\n`) so the quote and user input
   stay as one logical text block — same shape WeCom / DingTalk produce.
   When the message has no text part (e.g. quote + image-only reply) the
   placeholder falls back to a standalone `TextContent`.

2. **Unified `_download_and_wrap` helper.**
   `TIMImageElem` / `TIMFileElem` paths used to duplicate ~30 LOC of
   resolve → download → wrap each. Now collapsed into one helper that
   returns the right `Content` subclass based on `kind`.

3. **Audio routing.**
   `TIMFileElem` payloads with audio suffixes
   (`.mp3 .wav .m4a .ogg .opus .silk .amr .aac .flac`) are now wrapped
   as `AudioContent` so they flow through the unified ASR pipeline
   instead of becoming inert `FileContent`.

### Bug fixes

- **No more leaking expired CDN URLs to the model.**
  When `download_media` failed, the previous code fell back to the
  original pre-signed URL, which expires within minutes and yields
  broken-image errors downstream. We now emit a `TextContent` placeholder
  (`[image: download failed]` / `[file: download failed]`) so the model
  is informed and no stale URL leaks.

- **File extension preserved.**
  `_resolve_extension` previously consulted `Content-Type` first, which
  caused `application/octet-stream` from Yuanbao's CDN to silently rewrite
  `.txt` → `.bin` and `.mp3` → `.mpga`. Priority is now
  *original suffix → content-type guess → `.bin`*.

- **Filename is human-readable.**
  `_build_safe_filename` no longer mangles every upload to
  `<sanitized>_<uid>.<ext>`. Originals are kept verbatim (CJK preserved);
  a short uid is appended only when a name collision is detected on disk.

### Dead code removed

`TIMVideoFileElem` and `TIMSoundElem` branches — Yuanbao's official client
never emits them. Replaced by a single `unhandled msg_type=...` warning
as a safety net.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactoring

## Component(s) Affected

- [x] Channels (Yuanbao)
- [x] Tests

## Checklist

- [x] `pre-commit run --all-files` passes (mypy / black / flake8 / pylint)
- [x] `pytest tests/unit/channels/test_yuanbao.py` passes (93 tests)
- [x] No new contract test for Yuanbao — pre-existing gap, tracked
      separately

## Test Coverage (93 unit tests, 17 test classes)

New coverage added in this PR:
- `_classify_file` audio-suffix routing
- `_build_quoted_prefix` (text / image / file / audio / missing-desc)
- `_handle_chat_message` quote injection:
  - merged into first TextContent (text / audio quote)
  - standalone fallback when no TextContent in parts (image-only reply)
- `_build_safe_filename` collision-on-disk + whitespace-only fallback
- `_resolve_extension` original-suffix-wins-over-content-type
- `_normalize_inbound` `cloud_custom_data` JSON / malformed / missing
- `_download_and_wrap` failure returns TextContent placeholder
- `TIMFileElem` without `file_name` falls back to literal `file`

## Risk Assessment

- **Backward compat:** the only behavioral change visible to existing
  flows is that download failures now produce a placeholder TextContent
  instead of leaking an expired URL. Both cases were "broken" downstream
  anyway; the new behavior is strictly better.
- **No protocol/auth changes.** WSS handshake, sign-token logic, and
  the heartbeat path are untouched.
- **Cross-channel consistency:** quote injection shape now matches
  WeCom / DingTalk so upstream agents see a uniform input structure.

### DIFF
--- a/src/qwenpaw/app/channels/yuanbao/channel.py
+++ b/src/qwenpaw/app/channels/yuanbao/channel.py
@@ -17,6 +17,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 import aiohttp
 
 from agentscope_runtime.engine.schemas.agent_schemas import (
+    AudioContent,
     ContentType,
     FileContent,
     ImageContent,
@@ -79,6 +80,30 @@ if TYPE_CHECKING:
     from agentscope_runtime.engine.schemas.agent_schemas import AgentRequest
 
 logger = logging.getLogger(__name__)
+
+
+# File extensions treated as audio.  Files with these suffixes are wrapped
+# as ``AudioContent`` so they flow through the unified ASR pipeline; all
+# other suffixes fall back to ``FileContent``.
+_AUDIO_EXTS = frozenset(
+    {
+        ".mp3",
+        ".wav",
+        ".m4a",
+        ".ogg",
+        ".opus",
+        ".silk",
+        ".amr",
+        ".aac",
+        ".flac",
+    },
+)
+
+# Yuanbao ``cloud_custom_data.quote.type`` -> human-readable kind label.
+# 1 = text (desc carries text content)
+# 2 = image (desc empty)
+# 3 = file or audio (desc carries filename; routed by suffix below)
+_QUOTE_TYPE_LABEL = {1: "message", 2: "image", 3: "file"}
 
 
 def _short_id(raw_id: str) -> str:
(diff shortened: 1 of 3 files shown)


lalaliat (approved):
LGTM

PR #5170, opened 2026-06-13
perf(agents): cache PROFILE.md reads on the agent-list endpoint

ly-wang19:
## What

`list_agents()` (`GET /agents`) calls `_read_profile_description()` for **every** agent, which reads and parses that agent's `PROFILE.md` from disk on **every** request. `_normalized_agent_order()` on the same path also dedups via `O(n²)` list membership.

## Why

With many agents this is repeated disk I/O + parsing on a hot path, re-done on every list request — reported as a slowdown with 40+ agents in #4559. `load_agent_config()` on the same loop is already mtime-cached; the `PROFILE.md` read is not.

## Change

- Cache the parsed `PROFILE.md` snippet keyed on the file's `mtime`, mirroring the existing `_agent_config_cache` pattern — unchanged files are read/parsed once, not per request. Edits are still picked up (mtime change → cache miss).
- Dedup `_normalized_agent_order` with a set: `O(n²)` → `O(n)` membership.

Behaviour-preserving: identical description and ordering returned.

## Verification

- `black==23.3.0`, `flake8==6.1.0`, `pylint==3.0.2` (repo args) — clean (10/10)
- Verified the new ordering returns byte-identical output to the old across cases, and the cache parses once on repeated reads while refreshing on `mtime` change.

Related: #4559. Happy to add a benchmark or narrow scope if you'd prefer.

### DIFF
--- a/src/qwenpaw/app/routers/agents.py
+++ b/src/qwenpaw/app/routers/agents.py
@@ -114,24 +114,38 @@ def _normalized_agent_order(config) -> list[str]:
     """Return a deduplicated agent order covering every configured agent."""
     profile_ids = list(config.agents.profiles.keys())
     ordered_ids: list[str] = []
+    seen: set[str] = set()
 
     for agent_id in config.agents.agent_order:
-        if agent_id in config.agents.profiles and agent_id not in ordered_ids:
+        if agent_id in config.agents.profiles and agent_id not in seen:
             ordered_ids.append(agent_id)
+            seen.add(agent_id)
 
     for agent_id in profile_ids:
-        if agent_id not in ordered_ids:
+        if agent_id not in seen:
             ordered_ids.append(agent_id)
+            seen.add(agent_id)
 
     return ordered_ids
 
 
+# mtime-cache parsed PROFILE.md snippets so the agent-list endpoint does
+# not re-read/re-parse every agent's PROFILE.md on each request (#4559).
+_profile_desc_cache: dict[str, tuple[float, str]] = {}
+
+
 def _read_profile_description(workspace_dir: str) -> str:
-    """Read description from PROFILE.md if exists."""
+    """Read description from PROFILE.md if exists (mtime-cached)."""
     try:
         profile_path = Path(workspace_dir) / "PROFILE.md"
         if not profile_path.exists():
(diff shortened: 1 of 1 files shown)


RerankerGuo (changes requested):
I validated this at exact head `bac191415` and on a semantic current-main trial based on `bacac741` (retaining the current default/pinned/regular ordering helpers).

The optimization is measurable: for 40 unchanged profiles over 100 list-style iterations, the current helper performed 4,000 file reads, while this cache performed 40. Across five local runs, elapsed time dropped from 0.096-0.589s to about 0.018s. The `seen`-set change also preserves ordering semantics.

There are two cache-correctness blockers before this can merge:

1. The cache key is only `(path, float st_mtime)`. A deterministic production-helper contract writes `alpha`, caches it, rewrites the same-length profile to `bravo`, and restores the original mtime with `os.utime(..., ns=...)`. Both exact head and the current-main trial return stale `alpha`. After deleting and recreating the same path with `delta` and the same mtime, they still return `alpha`. This matters on coarse timestamp filesystems and for metadata-preserving sync/copy tools; the PR body currently promises edits are picked up.
2. Missing/deleted profiles never evict their entries, and the module-level dict has no bound. Creating, reading, and deleting 128 workspace profiles leaves all 128 path/description entries resident. Agent churn therefore grows this process-lifetime cache indefinitely and enables the stale path-reuse case above.

The existing current-main router/ordering scopes still pass (`43 passed`), confirming they do not cover either boundary. Please:

- use a stronger file identity/version signature (for example nanosecond mtime plus ctime, size, and inode where available), rather than float mtime alone;
- evict the path when the file is missing and add a bounded policy or synchronize entries with configured workspaces;
- add regressions for metadata-preserving same-length replacement, delete/recreate, and cache bounding/eviction;
- rebase semantically, preserving current main's `_group_agent_order()`, `_display_agent_order()`, and pinned-agent behavior.

For completeness, exact changed-file Flake8 6.1.0, `py_compile`, and `git diff --check` pass. The branch's Black 23.3.0 failure is inherited from its base at the unrelated `_initialize_agent_workspace()` call, so I am not treating that as a new blocker.

PR #5192, opened 2026-06-15
fix(desktop): guard against Windows console crash and self-kill commands

wangfei010313:
## Summary

This PR fixes two issues that cause QwenPaw to crash or terminate unexpectedly on Windows:

### 1. Rich console crash on legacy Windows terminals
The `print_ready_banner` function in `startup_display.py` uses Rich's `console.print()` which can throw `OSError: [Errno 22] Invalid argument` on legacy Windows consoles, causing the entire application to crash during startup. Introduced a `_safe_print` helper that wraps all `console.print()` calls (including the leading/trailing blank lines) with an `OSError` fallback to built-in `print()`.

### 2. LLM-generated self-kill commands
The `execute_shell_command` tool in `shell.py` did not guard against commands like `taskkill /F /IM python.exe` or `kill -9 $$` that could terminate the QwenPaw process itself. Added a `_is_dangerous_self_kill` check using token-based regex matching (with `\b` word boundaries to avoid false positives) that intercepts and blocks such commands before execution.

## Changes

- **`src/qwenpaw/utils/startup_display.py`**: Add `_safe_print` helper wrapping all `console.print()` calls with `OSError` fallback; unify trailing blank line behavior across Rich/fallback paths
- **`src/qwenpaw/agents/tools/shell.py`**: Add `_is_dangerous_self_kill` guard with three detection rules:
  - `_DANGER_NAME_RE`: blocks `taskkill /IM <name>`, `kill <name>`, `Stop-Process -Name <name>` using `\b` word boundaries to prevent false positives (e.g. `echo "do not kill python"` is safe)
  - `_SHELL_PID_VARS`: blocks shell variable self-kill (`$$`, `$PPID`, `$PID`)
  - `_KILL_PID_RE`: blocks PID-based kills targeting current or parent process (`os.getpid()` / `os.getppid()`)
- **`tests/unit/agents/tools/test_shell.py`**: Add 17 unit tests for `_is_dangerous_self_kill` covering true positives, false positives, and edge cases

## Testing

- All existing lint checks pass (black, flake8, pylint, mypy)
- 17 new unit tests for self-kill detection logic
- Verified `print_ready_banner` falls back gracefully when Rich fails

### DIFF
--- a/src/qwenpaw/agents/tools/shell.py
+++ b/src/qwenpaw/agents/tools/shell.py
@@ -363,6 +363,70 @@ def _execute_subprocess_sync(
                     pass
 
 
+_DANGER_NAMES = {"python", "pythonw", "cmd", "powershell", "pwsh", "conhost"}
+
+# Prefix that ensures kill/taskkill is at command start or after a separator
+# (&&, ;, |).  Prevents false positives like echo "do not kill python".
+_KILL_PREFIX = r"(?:^|[;&|]\s*)\s*"
+
+# Matches PID-based kills: taskkill /PID 123, kill -9 123, kill 123.
+# Uses greedy .* to capture the last number (PID), not intermediate flags like -9.
+# \b ensures "kill" is matched as a whole word (e.g. "skill" won't trigger).
+_KILL_PID_RE = re.compile(
+    rf"{_KILL_PREFIX}(?:taskkill|kill|stop-process)\b.*(?:/PID|-p|-pid|\b)\s*(\d+)",
+    re.IGNORECASE,
+)
+
+# Matches dangerous process names as /IM targets or bare kill targets.
+# \b word boundaries prevent false positives (e.g. "command" → "cmd",
+# "skill" → "kill", "pythonic" → "python").
+_DANGER_NAME_RE = re.compile(
+    rf"{_KILL_PREFIX}(?:taskkill|kill|stop-process)\b.*?\b({'|'.join(_DANGER_NAMES)})(?:\.exe)?\b",
+    re.IGNORECASE,
+)
+
+# Shell variables that reference the current/parent PID.
+_SHELL_PID_VARS = {"$$", "$ppid", "$pid"}
+
+
+def _is_dangerous_self_kill(cmd: str) -> bool:
+    """Return True if *cmd* would kill the current process or its parent.
+
+    Uses token-based regex matching to avoid false positives from
+    substring matching (e.g. ``echo "do not kill python"`` is safe).
+
+    Blocks three patterns:
+    1. ``taskkill /IM <dangerous_name>`` — kills by image name.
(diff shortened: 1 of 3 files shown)


rayrayraykk (commented):
###  Core Issues

1. **Vulnerability (Bypasses):**
   - **Missing Parent Process Protection:** The original code only checked `os.getpid()`. If the LLM targets the parent process ID (via `os.getppid()`), it could still terminate the main application.
   - **Unix Bypass:** Standard Unix `kill <PID>` (e.g., `kill 1234`) is missed because the regex strictly requires `/PID`, `-p`, or `-pid`.
   - **Shell Variable Bypass:** Shell self-kill variables like `$$` (current shell) and `$PPID` (parent PID) were not handled.
2. **High False Positives (Over-blocking):**
   - Simple substring matching (`"kill" in lower and "python" in lower`) will block harmless commands like `echo "do not kill python"` or `cat kill_list_python.txt`.

------

### Refactored Code (Direct Replacement)

Replace the regex definitions and the `_is_dangerous_self_kill` function with this secure, robust version:

```python
import os
import re

_DANGER_NAMES = {
    "python", "pythonw", "cmd", "powershell", "pwsh", "conhost"
}

# 1. Matches PID kills: supports Windows (taskkill /pid 123) and Unix (kill -9 123)
_KILL_PID_RE = re.compile(
    r"\b(?:taskkill|kill|stop-process)\b.*?(?:/PID|-p|-pid|\b)\s*(\d+)",
    re.IGNORECASE,
)

# 2. Matches dangerous process names: uses \b to avoid matching sub-strings in paths/comments
_DANGER_NAME_RE = re.compile(
    rf"\b(?:taskkill|kill|stop-process)\b.*?\b({'|'.join(_DANGER_NAMES)})(?:\.exe)?\b",
    re.IGNORECASE,
)

def _is_dangerous_self_kill(cmd: str) -> bool:
    """Return True if *cmd* would kill the current process or its parent process."""
    lower = cmd.lower()

    # Rule 1: Block killing by process names (e.g., taskkill /IM python.exe)
    if _DANGER_NAME_RE.search(lower):
        return True

    # Rule 2: Block self/parent-kill via Shell variables (e.g., kill -9 $$, kill $PPID)
    if "kill" in lower or "stop-process" in lower:
        if any(var in lower for var in ["$$", "$ppid", "$pid"]):
            return True

    # Rule 3: Block targeted PID kill (matching current PID or parent PID)
    m = _KILL_PID_RE.search(lower)
    if m:
        try:
            target_pid = int(m.group(1))
            
            # Protect both current process and parent process
            protected_pids = {os.getpid()}
            if hasattr(os, "getppid"):
                protected_pids.add(os.getppid())

            if target_pid in protected_pids:
                return True
        except ValueError:
            pass

    return False
```

rayrayraykk (approved):
Good jobs, looks good to me.

rayrayraykk (approved):

Issue #5235, opened 2026-06-16
Bug: Cron scheduled tasks not executing at scheduled time


lgcheng105:
## Bug Description

Cron scheduled tasks are not executing at their scheduled time. Tasks remain in pending state with `last_run_at: null` even after the scheduled time has passed.

## Steps to Reproduce

1. Create a cron task with `qwenpaw cron create`
2. Set `--schedule-type scheduled` with `--run-at` to a specific time
3. Set `--repeat-every-days 1` for daily repetition
4. Wait for the scheduled time to pass
5. Check task status with `qwenpaw cron state <job_id>`

## Expected Behavior

The task should execute at the scheduled time and `last_run_at` should show the execution timestamp.

## Actual Behavior

- `last_run_at` remains `null`
- `last_status` remains `null`
- Task never executes automatically
- `next_run_at` shows the correct next scheduled time, but task never fires

## Environment

- QwenPaw version: v1.1.11.post2
- OS: Linux

## Additional Context

- Manual execution via `qwenpaw cron run` also does not seem to work properly
- The channel send functionality works fine when called directly
- This appears to be a bug in the cron scheduler, not the channel dispatch mechanism

lalaliat:
Thank you for your feedback. Could you please help us troubleshoot by setting up a cron job on the console with a one-minute interval and type "text" to see if it triggers automatically? The key configurations are as follows:

- **Schedule Type**: Recurring
- **Schedule (Cron)**: Custom
- **Cron Expression**: `*/1 * * * *`
- **Task Type**: text
- **Message content**: text

Please **enable** this job and wait for a few minutes. Then, check the execution history to see if this cron job has been triggered automatically. 

If the issue persists, please execute the following command in the environment where QwenPaw is running:

```bash
qwenpaw cron list
```

Please share the output here to help us troubleshoot. **Note:** Please ensure that no sensitive or private information is included in the output. Thank you!

lalaliat:
Thank you for your feedback. After investigation, we believe the cron jobs were not executed properly, likely due to clock or wake-up instability. When APScheduler wakes up to trigger a job, if the current time has already exceeded the Misfire Grace period, the job is considered missed and its execution is skipped. We have made two changes to the cron job handling in #5418 :

1.  Changed the default value of Misfire Grace to 600 seconds.
2.  If a job is deemed missed due to exceeding the Misfire Grace period, a skip failure record will now be added to the history log.

[...]

PR #5241, opened 2026-06-16
fix(cron): increase default misfire_grace_seconds from 60 to 3600

lecheng2018:
## Description

Increase the default `misfire_grace_seconds` from 60 to 3600 in `JobRuntimeSpec`. When QwenPaw is temporarily busy processing a long-running agent task at the scheduled cron fire time, APScheduler silently skips the job if the delay exceeds the grace window.

With the previous default of 60 seconds, daily cron jobs (e.g. scheduled sign-ins at 08:00) can be silently skipped for the entire day if QwenPaw is busy at that exact moment. This is a common source of "cron didn't fire" reports.

By setting the default to 3600 seconds (1 hour), jobs that miss their exact scheduled time by a reasonable margin will still execute once QwenPaw catches up. This is a safe change — it does not change how often jobs run, only how long the system will wait before deciding to skip a missed execution.

**Related Issue:** Relates to #5235

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agent, runtime)
- [ ] Console / Frontend (web, terminal, desktop)
- [ ] Channels (Feishu, DingTalk, Slack, Matrix, etc.)
- [ ] Integration / MCP / Plugins
- [ ] Documentation
- [ ] Tests / CI / Build
- [ ] Other (please specify)

## Checklist

- [x] I ran \`pre-commit run --all-files\` locally and fixed all issues
- [x] My changes generate no new warnings or errors
- [x] I have added tests that prove my fix is effective or that my feature works (or explained why tests are not feasible)
- [x] My code follows the style guidelines of this project


lalaliat:
Hello, thank you for your contribution!

QwenPaw uses AsyncIOScheduler, where cron callbacks are async coroutines that share the same event loop with user agent tasks. As long as agent tasks properly use async/await (which stream_query does), the event loop won't be monopolized, and the scheduler can still successfully create_task to trigger cron callbacks at their scheduled times. Simply having "a user chatting while the agent is busy" typically wouldn't cause cron delays exceeding 60 seconds. The root cause of the original issue is more likely machine sleep, the QwenPaw process not running at the scheduled time (e.g., process inactive at 08:00), or incorrect timezone configuration. Modifying the global default is too aggressive, especially since per-job configuration already exists to allow users to adjust this value as needed. 

Given this, we won't merge this pr by now. We truly appreciate your effort, and we look forward to your future ideas!

Issue #5218, opened 2026-06-16
[Bug] 子Agent触发上下文压缩时QwenPaw进程冻结无响应


malongan:
# QwenPaw Bug Report: Context Compaction Causes Process Freeze

## Bug Description

When a sub-agent triggers context compaction (上下文压缩), the QwenPaw process freezes and becomes completely unresponsive. The app can only be recovered by manually restarting QwenPaw. Once the context compaction is triggered again after restart, it will freeze again.

## Environment

- **Platform**: macOS
- **QwenPaw Version**: Latest (desktop app)
- **Model Provider**: MiniMax (minimax-cn)
- **Affected Agent**: Sub-agent (e.g., "设计师/painter")

## Reproduction Steps

1. Have a sub-agent (e.g., painter) with a long conversation history
2. When the conversation context exceeds the threshold (default: 80% of max_input_length ~100K tokens)
3. The context compaction mechanism automatically triggers
4. **Result**: QwenPaw freezes completely, no response
5. User must manually restart QwenPaw to recover
6. After restart, if context compaction triggers again → same freeze happens

### Code Location
`qwenpaw/agents/context/light_context_manager.py`
Function: `_compact_context()`

### Problem
The context compaction calls `agent.reply()` to generate a summary, but **there is no timeout protection** for this call:

```python
# Line ~521 in light_context_manager.py
compact_msg: Msg = await agent.reply(
    Msg(
        name="compactor",
        role="user",
        content=user_message,
    ),
)
```

When the MiniMax API is slow or network issues occur during the compaction:
1. The `agent.reply()` call hangs indefinitely
2. No timeout error is raised
3. The entire QwenPaw process becomes blocked
4. No recovery mechanism exists

### Evidence from Logs

```
# This shows tasks getting stuck/destroyed:
Task was destroyed but it is pending!
task: <Task pending name='consumer_feishu_ee825381_20' coro=<UnifiedQueueManager._run_consumer() running at ...>

# This shows 5-minute timeout forcing stop:
Timeout waiting for tasks to complete for painter. Forcing stop after 5 minutes.
```

## Suggested Fix

Add timeout protection to the `agent.reply()` call in `_compact_context()`:

```python
import asyncio

try:
    compact_msg = await asyncio.wait_for(
        agent.reply(
            Msg(
                name="compactor",
                role="user",
                content=user_message,
            ),
        ),
        timeout=60.0  # 60 seconds timeout
    )
[...]

PR #5242, opened 2026-06-16
fix(compaction): add timeout protection to agent.reply() in _compact_context

lecheng2018:
## Description

Add timeout protection to the `agent.reply()` call inside `_compact_context()` in `light_context_manager.py`. When the LLM API call hangs due to network issues, rate limiter contention, or slow model responses, the entire QwenPaw process freezes because `agent.reply()` has no timeout protection.

### Root Cause

In `_compact_context()`, a `ReActAgent` is created on-the-fly to generate a summary. The `await agent.reply(...)` call had no timeout — if the model API blocks indefinitely under the global `LLMRateLimiter(max_concurrent=1)`, the compaction hangs forever, cascading to freeze all agents sharing the same provider.

### Fix

Wrap `agent.reply()` with `asyncio.wait_for(timeout=120)`. On timeout:
- `asyncio.TimeoutError` is raised
- Caught by `_compact_context_safe()` existing `except Exception` handler
- Returns safe fallback with `success=False`
- Agent continues without compaction rather than freezing

**Related Issue:** Relates to #5218

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agent, runtime)
- [ ] Console / Frontend (web, terminal, desktop)
- [ ] Channels (Feishu, DingTalk, Slack, Matrix, etc.)
- [ ] Integration / MCP / Plugins
- [ ] Documentation
- [ ] Tests / CI / Build
- [ ] Other (please specify)

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and fixed all issues
- [x] My changes generate no new warnings or errors
- [ ] I have added tests that prove my fix is effective or that my feature works (or explained why tests are not feasible)
- [x] My code follows the style guidelines of this project

### DIFF
--- a/src/qwenpaw/agents/context/light_context_manager.py
+++ b/src/qwenpaw/agents/context/light_context_manager.py
@@ -2,6 +2,7 @@
 # pylint: disable=too-many-nested-blocks,too-many-branches
 # pylint: disable=too-many-return-statements,too-many-statements
 """Context manager for agents with compaction support."""
+import asyncio
 import logging
 import os
 import sys
@@ -523,12 +524,15 @@ class LightContextManager(BaseContextManager):
             f"user_message={user_message[:500]}...",
         )
 
-        compact_msg: Msg = await agent.reply(
-            Msg(
-                name="compactor",
-                role="user",
-                content=user_message,
+        compact_msg: Msg = await asyncio.wait_for(
+            agent.reply(
+                Msg(
+                    name="compactor",
+                    role="user",
+                    content=user_message,
+                ),
             ),
+            timeout=120,
         )
 
         history_compact: str = compact_msg.get_text_content() or ""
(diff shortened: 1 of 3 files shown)


jinliyl:
@lecheng2018 
Thank you very much for your reply.
Please add fallback logic for timeout handling, and avoid hardcoding timeout=120.
Why is it necessary to modify misfire_grace_seconds?

lecheng2018:
Thank you for the review! All three points have been addressed in the latest commit (a3a1450c):

1. **misfire_grace_seconds removed** — This was accidentally included from PR #5241 due to branch mixing. Reverted in b71e1106, this PR now only contains compaction timeout changes.

2. **Configurable timeout** — Replaced the hardcoded `timeout=120` with a new `compact_timeout` field in `ContextCompactConfig` (default 120s, range 30-600s). The value flows through: config → `compact_context()` → `_compact_context_safe()` → `_compact_context()` → `asyncio.wait_for(timeout=compact_timeout)`.

3. **Timeout fallback** — Added a specific `asyncio.TimeoutError` handler in `_compact_context_safe()` that returns a clean fallback dict with `reason: "timeout after Xs"` instead of a generic exception. The existing `except Exception` handler remains as a safety net for other failures (LLM errors, etc.).

PR #5245, opened 2026-06-16
docs: add Vietnamese README (README_vi.md)

nguyenthanhthe:
Add Vietnamese translation of README.md (README_vi.md) following the same format as existing translations.


zhaozhuang521:
@nguyenthanhthe The input is not debounced. If the session list is large, re-filtering with each keystroke may cause performance issues.

zhaozhuang521 (approved):
lgtm

PR #5246, opened 2026-06-16
fix(memory): add config overrides for chromadb SIGSEGV on macOS

### LINKED ISSUE #5243: [Bug]: QwenPaw 在 macOS 上频繁 SIGSEGV 崩溃，定位到 ChromaDB Rust 绑定。
崩溃堆栈全部落在 chromadb_rust_bindings.abi3.so，地址 0x44（空指针）。今天 12 次，两天累计 48 次重启。触发场景是 ReMeLight 记忆管理器的向量操作。建议检查 chromadb 版本兼容性，或提供 vector_enabled: false 的降级路径

[QwenPaw_SIGSEGV_Crash_Report_20260616.md](https://github.com/user-attachments/files/29006579/QwenPaw_SIGSEGV_Crash_Report_20260616.md)

nguyenthanhthe:
## Problem

Issue #5243 — QwenPaw crashes with SIGSEGV on macOS, traced to `chromadb_rust_bindings.abi3.so` null pointer at address 0x44 during ReMeLight vector operations. 48 crashes in 2 days.

The crash happens because:
- `chromadb` imports successfully on macOS (passes the import check)
- But its Rust native binding crashes at runtime during actual vector operations
- There was no way to force-disable chromadb without setting an obscure env var

## Fix

Two new fields in `ReMeLightMemoryConfig` (configurable via Console UI):

| Field | Values | Effect |
|-------|--------|--------|
| `memory_store_backend` | `"auto"` (default) / `"local"` / `"chroma"` | Forces the memory store backend; `"local"` bypasses chromadb entirely |
| `vector_enabled` | `null` (auto-detect) / `true` / `false` | Explicitly enables/disables vector search; `false` avoids triggering chromadb Rust code |

The config override is checked before the existing `MEMORY_STORE_BACKEND` env var, so UI settings take highest priority.

## Changes

- `src/qwenpaw/config/config.py`: +2 fields in `ReMeLightMemoryConfig`
- `src/qwenpaw/agents/memory/reme_light_memory_manager.py`: 
  - `_detect_memory_manager_backend()` accepts `config_override` param
  - `__init__` uses config values before auto-detection
  - Removed duplicate `agent_config` / `reme_cfg` loading

Closes #5243


rayrayraykk:
Hi @nguyenthanhthe, thanks for the quick response to #5243!

We've been looking at this issue as well and took a slightly different approach. Rather than exposing `memory_store_backend` and `vector_enabled` config overrides (which require users to know about the problem and manually configure it), we added an **async subprocess runtime probe** that automatically detects whether the chromadb Rust bindings work at runtime.

### How our approach differs

| Aspect | PR #5246 (config overrides) | Our approach (runtime probe) |
|--------|---------------------------|----------------------------|
| User action required | User must set `memory_store_backend: "local"` in config | None — fully automatic |
| Root cause addressed | Workaround — user bypasses broken chromadb | Detection — system discovers chromadb is broken and auto-downgrades |
| New config surface | 2 new fields in `ReMeLightMemoryConfig` | Zero new config |
| Code complexity | Config plumbing across config.py + manager | Self-contained in manager only |

### How it works

In `start()`, when the detected backend is `chroma`, we spawn a child process via `asyncio.create_subprocess_exec` that runs:

```python
import chromadb
c = chromadb.EphemeralClient()
c.get_or_create_collection('_probe')
```

If the child crashes (SIGSEGV kills the child, not the parent), times out (15s), or fails for any reason, we automatically rebuild `self._reme` with the `local` backend. The probe is async so it doesn't block the event loop.

### Trade-off

The only downside is a ~1-3 second startup overhead for the subprocess probe (only on non-Windows when chromadb is detected). We think this is a worthwhile trade-off compared to requiring users to manually discover and configure the workaround.

That said, your config override approach could still be valuable as a **complementary** escape hatch for users who know they always want a specific backend — just not as the primary fix for #5243. What do you think about combining both approaches? cc @jinliyl #5271

nguyenthanhthe:
Thanks for the detailed explanation, @rayrayraykk! The runtime probe approach sounds like a solid primary fix auto-detection is definitely cleaner than requiring users to configure workarounds.  I'd be happy to update this PR to align with your approach. Could you clarify what you'd like me to do specifically?  1. Should I replace the config overrides entirely and just implement the runtime probe in this PR? 2. Or keep the config overrides as a complementary escape hatch alongside the probe and if so, what adjustments are needed?  I can rebase and adjust the code accordingly, just let me know which direction you prefer.

zhijianma:
Fixed in #5243

Issue #5259, opened 2026-06-17
Windows 上向量索引无法持久化，每次启动必须保持"重建记忆索引"开启 / Vector index not persisted on Windows, "Rebuild memory index on startup" must stay enabled


Scorp1o117:
## 问题描述

Windows 上 `memory_search` 的向量索引无法持久化到磁盘。关闭"启动时重建记忆索引"选项后重启，`memory_search` 完全失效（返回空结果）。必须保持该选项开启才能正常使用记忆搜索。

## 复现步骤

1. 确认"启动时重建记忆索引"开启，验证 `memory_search` 正常工作
2. 关闭"启动时重建记忆索引"选项
3. 重启 QwenPaw Desktop
4. 执行 `memory_search`，返回空结果

## 期望行为

关闭"启动时重建记忆索引"后，之前构建的向量索引应从磁盘加载，无需重新构建。

## 环境

- OS: Windows 10/11
- QwenPaw Desktop 版本: 最新版
- 向量模型: Qwen3-Embedding-0.6B（本地服务）

---

## Description

The vector index for `memory_search` cannot be persisted to disk on Windows. After disabling the "Rebuild memory index on startup" option and restarting, `memory_search` completely stops working (returns empty results). The option must remain enabled to use memory search.

## Steps to Reproduce

1. Verify `memory_search` works with "Rebuild memory index on startup" enabled
2. Disable the "Rebuild memory index on startup" option
3. Restart QwenPaw Desktop
4. Run `memory_search` — returns empty results

## Expected Behavior

After disabling "Rebuild memory index on startup", the previously built vector index should be loaded from disk without needing to rebuild.

## Environment

- OS: Windows 10/11
- QwenPaw Desktop version: Latest
- Embedding model: Qwen3-Embedding-0.6B (local service)

jinglinpeng:
Hi @Scorp1o117 ，请问桌面版是用的Tauri还是普通桌面版；另外这个BUG用qwenpaw app启动的话有出现吗？

Scorp1o117:
Hi @jinglinpeng，我用的是 Tauri 桌面版（QwenPaw Desktop）。"qwenpaw app" 没试过，不确定是否也有这个问题。

jinglinpeng:
Hi @Scorp1o117 ,这个问题在新版本应该已经修复了，请等2.0版本（预计本周或下周）出来后看看是否OK。

lalaliat:
[...]

PR #5265, opened 2026-06-17
fix(desktop): graceful shutdown endpoint + Tauri lifecycle fix

nguyenthanhthe:
Resolved conflict with agentscope 2.0 (origin/main). The previous memory-index fix (gh-5259) no longer applies because the memory manager was fully rewritten in agentscope 2.0 (now uses ReMe4). This PR now only contains the graceful shutdown changes:  1. **backend.rs**: Raw TCP shutdown request + 5s wait before forced kill 2. **_app.py**: /api/shutdown endpoint with full cleanup sequence  Closes gh-5265, gh-5259 (memory fix superseded by 2.0 architecture).

### DIFF
--- a/console/src-tauri/src/backend.rs
+++ b/console/src-tauri/src/backend.rs
@@ -87,11 +87,47 @@ impl BackendState {
     fn stop(&self) {
         self.next_generation();
         let child = self.with_inner(|inner| inner.child.take());
+        let port = self.port();
+
+        // 1 — Fire-and-forget graceful shutdown request to the FastAPI backend.
+        //     Ponytail: raw TCP instead of reqwest to avoid the `blocking`
+        //     feature flag.  Response doesn't matter — we're shutting down.
+        if let Some(port) = port {
+            log::info!("[backend] requesting graceful shutdown on port {port}");
+            let addr: std::net::SocketAddr = ([127, 0, 0, 1], port).into();
+            std::thread::spawn(move || {
+                use std::io::Write;
+                use std::net::TcpStream;
+                use std::time::Duration;
+                if let Ok(mut stream) =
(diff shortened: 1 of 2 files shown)


jinliyl:
Thanks for the fix. For the Windows/local backend case, forcing a rebuild on startup looks like a reasonable workaround, since it avoids relying on a persisted local vector index that currently cannot be loaded reliably.

One related note: for QwenPaw Desktop, we may also need a follow-up fix in the Tauri shutdown path. The current window-close flow appears to stop the Python sidecar with `child.kill()` rather than asking the FastAPI backend to shut down gracefully, so the normal lifespan shutdown path may not always run and `ReMeLight.close()` cannot be relied on during Desktop exit.

Relevant code locations:
- FastAPI graceful shutdown path: `src/qwenpaw/app/_app.py` -> `multi_agent_mgr.stop_all()`
- Agent/service cleanup path: `src/qwenpaw/app/multi_agent_manager.py`, `src/qwenpaw/app/workspace/workspace.py`, `src/qwenpaw/app/workspace/service_manager.py`
- ReMe close path: `src/qwenpaw/agents/memory/reme_light_memory_manager.py`
- Desktop close path: `console/src-tauri/src/lib.rs` (`WindowEvent::CloseRequested`) -> `console/src-tauri/src/backend.rs` (`BackendState::stop`, currently `child.kill()`)

So I think this PR is useful as a short-term workaround, but the more complete Desktop-side fix would be: request graceful shutdown from the Python backend before closing, wait briefly, and only fall back to killing the sidecar on timeout.

nguyenthanhthe:
@jinliyl thank you for your suggestion, i've applied and commited, you can review it now !

PR #5287, opened 2026-06-17
fix(context): don't crash compaction when summary exceeds schema maxLength

niceIrene:
### Problem
Auto-compaction generates a structured summary against `SummarySchema`, whose fields have char-level `maxLength` caps (`important_discoveries=300`, etc.). When the model overshoots a cap, `jsonschema.validate` raises. Since model retries are collapsed to 0 and `RetryChatModel` only retries transient API errors, the error escapes `compress_context()` and aborts the **whole turn** — the user sees `Internal error`, and every retry re-triggers it.

```
jsonschema.exceptions.ValidationError: '...' is too long
Failed validating 'maxLength' in schema['properties']['important_discoveries']: {'maxLength': 300}
```

### Fix
In `QwenPawAgent.__init__`, strip the `maxLength` caps from the compaction summary schema (`context_config.summary_schema`) so the parsed summary is used as-is instead of crashing. The original caps are kept, and an override of `compress_context()` logs a warning when the produced summary overruns that budget.

### Files
- `src/qwenpaw/agents/react_agent.py`

image

### DIFF
--- a/src/qwenpaw/runtime/builder.py
+++ b/src/qwenpaw/runtime/builder.py
@@ -11,11 +11,58 @@ injects all dependencies into the agent constructor.
 from __future__ import annotations
 
 import logging
-from typing import Any, Iterable
+from typing import Any, Iterable, Optional
+
+from pydantic import BaseModel, Field
 
 from .tool_guard import GuardedFunctionTool
 
 _logger = logging.getLogger(__name__)
+
+
+class _CompactionSummarySchema(BaseModel):
+    """Schema guiding context-compaction summaries."""
+
+    task_overview: Optional[str] = Field(
+        default="",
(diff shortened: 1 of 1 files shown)


src/qwenpaw/agents/react_agent.py
```diff
@@ -123,6 +123,16 @@ def __init__(
             middlewares=middlewares,
         )
 
+        # Drop maxLength caps from the compaction summary schema so an
```
DavdGao: Maybe just feed a new summary_schema when creating the agent instance? Others look good to me

niceIrene: Updated! Thank you

Issue #5294, opened 2026-06-18
[Feature]: 模型可以批量测试与批量删除


no-teasy:
同一个供应商的模型如果要批量测试只能一个个点，太麻烦了
而且如果要删除模型，还有一个确认框，删除模型太麻烦了

## Summary

[One or two sentences: what do you want and why?]

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Problem / Motivation

[What problem does this solve? Who benefits?]

## Proposed Solution

[Describe the feature or change you have in mind. Be as specific as possible.]

## Alternatives Considered

[Any other approaches or workarounds you thought about.]

## Additional Context

[Screenshots, examples, links to docs or similar features elsewhere.]

## Willing to Contribute

- [ ] I am willing to open a PR for this feature (after discussion).

PR #5297, opened 2026-06-18
feat(models): batch test & batch delete models (#5294)

nguyenthanhthe:
## Summary
Batch test and batch delete models in the provider model management dialog. Addresses issue #5294.

### Backend
- `POST /{provider_id}/models/batch-test` — test N models in parallel (via `asyncio.gather`)
- `POST /{provider_id}/models/batch-delete` — delete N models in sequence
- New Pydantic models: `BatchTestModelsRequest`, `SingleTestResult`, `BatchTestModelsResponse`, `BatchDeleteModelsRequest`

### Frontend
- `api.batchTestModels()` and `api.batchDeleteModels()` in API client
- Checkbox per model row + select-all checkbox
- Batch toolbar with 'Batch Test' and 'Batch Delete' buttons
- Delete confirmation dialog before batch delete

### Files Changed
1. `src/qwenpaw/app/routers/providers.py`
2. `console/src/api/modules/provider.ts`
3. `console/src/api/types/provider.ts`
4. `console/src/pages/Settings/Models/components/modals/RemoteModelManageModal.tsx`

### DIFF
--- a/console/src/api/modules/provider.ts
+++ b/console/src/api/modules/provider.ts
@@ -14,6 +14,9 @@ import type {
   TestConnectionResponse,
   TestProviderRequest,
   TestModelRequest,
+  BatchTestModelsRequest,
+  BatchTestModelsResponse,
+  BatchDeleteModelsRequest,
   DiscoverModelsResponse,
   ProbeMultimodalResponse,
   SeriesResponse,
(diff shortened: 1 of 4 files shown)


maximilize:
Nice — this covers both asks in #5294: parallel batch test via `asyncio.gather` reusing the existing `provider.check_model_connection()` (`app/routers/providers.py:435`, `providers/provider.py:207`), and batch delete with a **single** confirmation instead of one `Modal.confirm` per model (the current per-model flow at `RemoteModelManageModal.tsx:511-535` is the tedium the issue describes).

One small inconsistency: the per-row checkbox isn't gated by `isDeletable` the way "select all" is (`provider.is_custom || extraModelIds.has(m.id)`), so a built-in / non-deletable model can be individually checked and sent to batch-delete, where the backend then rejects it with a 400. Worth gating the row checkbox to match select-all.

Also note batch delete is fail-fast and non-atomic — the backend loop raises on the first failure after already deleting earlier models in the list, with no rollback. Probably acceptable, but worth a deliberate decision (or at least returning which ones succeeded).

maximilize:
Both points from my earlier review are addressed — verified:

1. The per-row checkbox is now gated: `disabled={!isDeletable}` with `isDeletable = provider.is_custom || extraModelIds.has(m.id)` — same predicate as select-all, so non-deletable models can no longer be individually queued for delete.
2. Batch delete is no longer fail-fast: the backend loop catches `(ValueError, AppBaseException)` per model and returns `SingleDeleteResult[]` (success/message each), and the frontend surfaces failures individually via `message.error(`${f.model_id}: ${f.message}`)` (same per-model reporting for batch test). That addresses the non-atomic concern by making partial results visible.

Nothing else from my side.

PR #5321, opened 2026-06-19
feat(context): scroll context manager — durable history + recall REPL 

### LINKED ISSUE #5456: [Bug]: Wrong agent identity for channel-built requests
## QwenPaw Version

2.0.0b1 / git `a9ee2e83`

## Description

When using a non-default agent, the runtime can build context with agent id `default` because `AgentRequest` does not define an `agent_id` field.

This causes the model-facing agent identity to not match the active workspace agent.

**Related PR(s):** none

**Security considerations:** Low. This is mainly an agent identity correctness issue, but it can affect skills/tools that depend on the current agent id.

## Component(s) Affected

- [x] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.0.0b1 / git `a9ee2e83`
- **OS:** Linux / WSL2
- **Python version:** 3.12.3

## Steps to Reproduce

1. Use any non-default agent.
2. Send a message through a channel/runtime path that builds a normal `AgentRequest`.
3. Inspect the runtime context or model-facing prompt.

## Actual vs Expected

- **Actual:** Runtime context can use agent id `default`.
- **Expected:** Runtime context should use the active workspace agent id before falling back to `default`.

## Logs / Screenshots

N/A

## Additional Notes

`AgentRequest` carries request data such as `input`, `session_id`, `user_id`, `stream`, and `metadata`, but not `agent_id`. Channel-built requests therefore do not necessarily include `agent_id`; runtime context should derive it from the active workspace before falling back to `default`.

niceIrene:
## Summary

Adds the **scroll** context-management strategy — a retrieval-driven
alternative to native compression. Instead of summarizing evicted turns
away, scroll persists the full conversation to a durable SQLite store and
lets the model recall any past turn on demand from a Python REPL. The live
prompt keeps only a pinned head + recent tail + a compact, always-visible
**eviction index** that can be expanded back to full fidelity with a single
query.

### 1. Scroll context strategy (new)
A retrieval-driven context strategy under
`src/qwenpaw/agents/context/scroll/`, built as independent overrides of
AgentScope hooks (the framework loop itself is untouched):

- **Write-through durable history** (`on_save`) — every live turn is written
  through to a per-workspace SQLite file (`history.db`, WAL mode + FTS5
  keyword index) the instant it enters the window, so the DB stays in lockstep
  with the prompt (no one-turn lag). The store survives across runs / sessions
  / agents (recall scoped by `session_id` / `agent_id`), and a corrupt file is
  quarantined and recreated rather than crashing startup.
- **In-context eviction index** (`compress`) — past the token threshold, the
  middle is evicted to `history.db` and folded into a compact multi-level
  index of one-line headlines, keeping a pinned head + recent tail verbatim.
  The index coarsens under pressure via an odometer-style carry; nothing is
  deleted — every line keeps a `seq` span that's re-queryable.
- **Tool-result capping** (`on_acting` middleware) — oversized tool outputs
  are stored in full and shown in-context as a bounded preview + a recall
  pointer keyed by `tool_call_id`, so a single large result can't blow the
  window. (AgentScope's built-in per-tool truncation is disabled so this is
  the only — lossless — capping path.)

### 2. `recall_history_python` recall REPL
The model recalls evicted/older history by running Python in a persistent
namespace instead of keeping everything in context. The namespace exposes
`ms` (a `MemorySpace`) with read-only `hist` (the durable history, source of
truth) and a read/write in-memory `main` scratch DB. By default recall is
scoped to this agent across all its sessions; widen with `all_agents=True` or
pin a specific `session_id` / `agent_id`:

- `ms.search(query, *, session_id=None, agent_id=None, all_agents=False,
  kind=None, k=10)` — ranked full-text (FTS5, bm25) with a LIKE fallback;
  returns the full turn text per hit.
- `ms.expand(lo, hi)` — full turns in the `seq` span `[lo, hi]`, oldest first
  (`seq` is a globally-unique address, so a span needs no other filter).
- `ms.recall_tool(tool_call_id, *, all_agents=False)` — a tool call and its
  result, by id.
- `ms.sessions(...)` / `ms.session(session_id, ...)` / `ms.agents(...)` —
  discover conversations/agents and read one back.
- `ms.sql_query(...)` — raw read-only SQL over `hist` for anything else.

### 3. Governance: allow the recall REPL
Registered `RecallHistoryPython` in the governance tool registry so the
sandboxed recall REPL isn't rejected as an unregistered tool. The REPL runs
model-authored Python and is isolated by the governance sandbox. If the
sandbox is unavailable, scroll is **not** wired and the agent falls back to
native context management (so it never evicts history that nothing could read
back) — and as a runtime safety net the REPL itself fails closed and refuses
to run. The only exception is an explicit operator opt-in to unsandboxed
recall (see §6).

### 4. Default strategy → scroll
`LightContextConfig.strategy` default flipped from `native` to `scroll`,
wired in `AgentBuilder` through `build_scroll_components`. Opt in/out via
`light_context_config.strategy`; agents that explicitly set a strategy in
their `agent.json` are unaffected. Scroll keeps writing the same date-grouped
`dialog/{YYYY-MM-DD}.jsonl` session archive the native manager produced (on by
default via `offload_dialog`), so that storage is unchanged — `history.db`
becomes the durable source of truth alongside it.

### 5. Legacy memory migration (`qwenpaw agents migrate-history`)
**Manual, opt-in** — flipping the default to scroll migrates nothing; the
operator runs this to carry an agent's old `native` memory into `history.db`:

```
qwenpaw agents migrate-history --agent-id <id> [--dry-run]
```

Converts legacy `dialog/{YYYY-MM-DD}.jsonl` archives into the durable
`conversation_history` table (each date-file → one `legacy:{date}` session),
reusing the live write path so migrated rows are identical to native ones.

- **Idempotent** — re-runs insert nothing new (`ux_dedup` UNIQUE index, plus a
  `.migrated.json` manifest to skip already-converted files).
- **Non-destructive / rollback** — source `dialog/*.jsonl` is never modified or
  deleted (append-only to `history.db`); flipping back to `native` stays
  lossless. On failure nothing is removed; partial rows dedupe away next run.
- **Tested** — `test_migrate.py` pins all three: source bytes unchanged,
  re-run is a no-op (manifest + DB-index paths), `--dry-run` writes nothing.

### 6. Operational safety
- **First-run notice** — a one-time operator log warning the first time scroll
  creates `history.db` in a workspace (the silent default switch, the new
  on-disk file, and how to roll back).
- **Bounded by default, on-demand trimming** — `history_retention_days`
  defaults to **90**: rows older than the window auto-purge on teardown, so a
  long-running store stays bounded. Set `0` to keep history forever. A large
  store (`>10 GiB`) also logs a warning pointing at a manual trim:

  ```
  qwenpaw history purge --days 30 --dry-run            # preview rows + bytes
  qwenpaw history purge --days 30 --vacuum             # delete and reclaim disk
  qwenpaw history purge --days 30 --tool-output-only   # drop only bulky tool output
  ```

  Purge always previews before deleting and never auto-runs; `--vacuum`
  reclaims disk (DELETE alone reuses pages but doesn't shrink the file).
- **Unsandboxed recall is deployment-gated** — `scroll_config.allow_unsandboxed`
  cannot bypass the sandbox on its own; it takes effect only when the operator
  also sets the `QWENPAW_ALLOW_UNSANDBOXED_RECALL` env var, closing a
  multi-tenant privilege-escalation path. Default-deny.

## ⚠️ BREAKING CHANGE

context: default `light_context_config.strategy` flips `native` → `scroll`.

Agents that do NOT set `strategy` explicitly in agent.json are switched to
scroll on upgrade. Effects:
- a new `history.db` is created under the workspace dir on first run
- memory/disk-I/O profile changes (durable recall store + FTS index)

**Roll back:** set `"strategy": "native"` under `light_context_config` in the
agent's agent.json. A one-time operator log notice points this out on first
scroll activation per workspace.

Agents that already pin `"strategy"` explicitly are unaffected.

Example:
Native Context Management vs Scroll Context Management on LongMemEval_s: improved from 23.2% to 91.2%
image

Some other tests, e.g. (1) Asking irrelavant questions, (2) Questions that do not need to recall memory, (3)What headline labeling looks like:
image
image
image


ekzhu:
## Review: scroll context manager — key issues & bugs

Reviewed the full scroll module (`src/qwenpaw/agents/context/scroll/`) plus the config/builder/agent wiring. Findings below were adversarially verified against the code; severities are post-verification. Grouped by what should block the default-on rollout vs. what can follow.

### Context for the rollout plan
The plan to default to `scroll`, auto-convert existing file-based memory into `history.db`, show a first-run rollback pop-up, and keep the strategy switch in agent config addresses the "silent behavior change" concern — good. Two things that plan now depends on:
- **The auto-conversion must be idempotent** and must **not delete the original file-based memory**, so the rollback path is lossless if a user reverts.
- Because we're migrating *everyone* onto `history.db`, the durability bugs below stop being edge cases — they hit the default path on the first session. They should be fixed *before* the flip lands.

### Must-fix before default-on

1. **Entire conversation re-appended to `history.db` on every session reload** (`manager.py:37-59,126-195`). All dedup/bookkeeping (`_persisted_ids`, `_seq_by_id`, the `EvictionIndex`) is in-memory and never checkpointed, while `state.context` *is* restored. Each turn rebuilds a fresh manager with empty dedup sets over a pre-loaded context, so `_persist_new` re-appends every restored message. **This triggers on the normal save/reload cycle between every console turn** — recall returns N duplicate copies, FTS bloats, seq spans drift. With default-on + migration this corrupts the store immediately. Fix: seed dedup from the DB on init, or wire the already-written-but-unused `EvictionIndex.to_dict/from_dict` (`eviction_index.py:187-222`) through `state_dict`/`load_state_dict`, and make `append` idempotent on `(session_id, msg_index/tool_call_id)`.

2. **`execute_python` runs model-authored Python *unsandboxed* when governance is degraded** (`repl.py:83-91,129-143`). `sandbox_config` is only injected by `PolicyGuardedTool`; when `_init_governor` returns `None` on any init failure (`builder.py:350-371`), the tool is wrapped in `GuardedFunctionTool`, which never sets it, so the `else` branch calls `asyncio.create_subprocess_shell` with zero isolation — arbitrary host code execution as the agent user, while every docstring claims "sandboxed." Reachable in normal operation now that scroll is default. Fix: fail closed — refuse when `sandbox_config is None` unless an explicit `allow_unsandboxed` opt-in is set.

3. **Read-only `hist` ATTACH is escapable via arbitrary SQL** (`memoryspace.py:57-61,77-101`). `sql_query`/`sql_exec` execute any SQL the model supplies. Verified: `DETACH DATABASE hist` → `ATTACH … ?mode=rw` → `INSERT/UPDATE/DELETE` lets the model tamper with or wipe its own durable history; `ATTACH '/other/workspace/history.db'` reads another workspace's conversations. The RO guarantee only binds the original alias, which the model controls. Fix: reject `ATTACH`/`DETACH`/`PRAGMA` (statement whitelist) or enforce `PRAGMA query_only` / a SQLite authorizer at the connection level.

4. **Capped tool results written twice** (`cap_middleware.py:72-95`). The middleware appends the full output keyed by `tcid`; `_persist_new` later persists the now-truncated block under the *same* `tcid` (separate dedup sets). The recall pointer the model is told to run (`WHERE tool_call_id='{tcid}'`) returns **both** the full output and the truncated stub, and the stub is FTS-indexed. Fix: share a `capped_tcids` set between the two writers.

### Other correctness bugs

5. **`HistoryStore` RW connection leaked per request** (`manager.py:231-232`). `ScrollContextManager.close()` → `HistoryStore.close()` exists but nothing calls it; `QwenPawAgent.close()` only stops the governor/offloader. Open fds (db + `-wal` + `-shm`) accumulate over a long-running server. (Medium)

6. **Caller-controlled `session_id` f-string-interpolated into the SQL shown to the model** (`eviction_index.py:243-256`, same for `tcid` at `cap_middleware.py:91-93`). `session_id` passes through from the request unvalidated; an apostrophe corrupts every recall query, and `x' OR '1'='1` drops the session filter (cross-session leak in multi-tenant). Fix: render `:session_id` placeholders + bind params, or validate the id at ingestion. (Medium)

7. **Boundary-message double-presence** (`manager.py:96-112`). When `_split_context_for_compression` block-splits the boundary `Msg`, both halves keep the same `msg.id`, landing in both `middle` and `tail`; `_index_evicted` folds the full seq span into the index while the reserve half is still live, so the index claims content as "evicted" that's still in context. (Medium)

8. **Latent/low:** `update_entry` never refreshes scalar `tool_call_id`/`name`/`tool_state` columns on multi-tool-call turns (`history.py:212-252`, currently inert); anonymous tool_result dedup key uses a growing counter `len(self._persisted_tcids)` (`manager.py:141-152`, broken algorithm but unreachable while `ToolResultBlock.id` is required); quarantine-on-open can rename `history.db` out from under a live concurrent connection (`history.py:34-85`, narrow race).

*(Checked and dismissed: an FTS-delete-on-NULL drift, a `real()` filter dropping a pinned message, and WAL multi-writer `SQLITE_BUSY` loss — none reachable as wired.)*

### Cross-session semantics — please confirm intended behavior
The store is **one `history.db` per workspace, shared across all sessions and all agents** (no `agent_id` column — `history.py:95-113`). But recall is **session-scoped by default**: `ms.search` defaults to `scope="session"`, and the eviction-index map hardcodes `WHERE session_id='<this session>'` (`eviction_index.py:247-252`). Cross-session recall is reachable only via `scope="all"`, which nothing in the prompt or index ever suggests. Also, `task_id`/`run_id` are collapsed to `session_id` (`builder.py:549-556`), so `scope="task"` is identical to `scope="session"` — the "all runs of this task" feature is dead as wired. Decide and make it explicit:
- If cross-session memory for the same agent is the goal → guide the model to it and thread real lineage.
- If isolation is the goal → the shared DB + `scope="all"` + missing `agent_id` is a leak; add an `agent_id` column now (empty table, no migration cost) and default recall to the agent's own history.

### Quality / design
- **No tests for the ~1973-line module**, and this PR deletes the only context test (`tests/unit/agents/context/test_as_msg_handler.py`). The carry/compact math, `_persist_new` dedup, FTS sync, and quarantine path are pure-Python and trivially unit-testable. Highest-leverage ask before default-on.
- **No retention/clear/export path** — history is "never dropped" (`history.py:30-31`) and the clean command doesn't know about `history.db`/`.scroll`. Add `HistoryStore.purge(before=...)` and wire it in.
- **Broad `except` in `on_save`/`compress`/`_cap`** swallows write failures (`manager.py:65-68`, `cap_middleware.py:55-59`) while the prompt promises durability; `_cap` even yields the *uncapped* output on failure. Narrow to `sqlite3.Error`/`OSError` and add a health flag.
- **Model-facing raw SQL f-strings** in three places with inconsistent scoping (render() hardcodes the session filter, prompt.py omits it). Replace with intent-named helpers (`ms.expand(lo,hi)`, `ms.recall_tool(tcid)`) — centralizes scoping and removes the injection surface.
- **Tight coupling to AgentScope private internals** (`_prepare_model_input`, `_split_context_for_compression`, direct `state.context` mutation) on a 2.0 *alpha* — quarantine behind a thin adapter + a signature smoke test.
- **`postscript.py` is eval-only and unwired** — move it out of the shipped runtime path or wire it.

niceIrene:
@ekzhu  Very helpful suggestions! Thank you. I've updated the code to address them all. I also made sure the new context management works well with the existing magic commands for context management, e.g. /compact, and added support for the user to query cron job history.

jinliyl:
Great work on this. The scroll approach is a strong direction, and the benchmark/result screenshots look very promising. I especially like the write-through durable history plus an in-context eviction index: it gives the model a practical way to keep full-fidelity history without relying on lossy summaries.

I found a few issues worth addressing before merging:

1. **Default `scroll` can evict history even when recall is unavailable.**
   `LightContextConfig.strategy` now defaults to `scroll`, but if governance/sandbox initialization fails, the builder falls back to `GuardedFunctionTool`; in that path `execute_python` receives no `sandbox_config` and refuses to run unless `allow_unsandboxed=True`. The agent can still compact into the scroll eviction index, but the only recall tool is unavailable, so evicted history becomes effectively inaccessible. I think scroll should either stay opt-in/default to native, or the builder should avoid wiring scroll unless the sandboxed recall tool can actually run.

2. **`execute_python` returns `ToolResultState.RUNNING` for final results.**
   The tool formats stdout/stderr and returns a single final `ToolChunk`, but the state is always `RUNNING`, even when the subprocess has completed or exited non-zero. That can confuse tool coordination, persisted `tool_state`, and model observations. It should return `SUCCESS` when `code == 0` and `ERROR` otherwise.

3. **`ms.search()` passes raw user/model text directly into FTS5 `MATCH`.**
   Queries like `C++`, `foo-bar`, or `what?` can raise SQLite FTS syntax errors instead of falling back. Since the model will naturally issue plain-language queries, `search()` should either quote/tokenize safely or catch `sqlite3.OperationalError` and fallback to the LIKE search path.

4. **`ms.session(session_id)` is not agent-scoped.**
   `sessions()` defaults to this agent, but `session()` only filters by `session_id`. In a shared workspace, common session ids like `main`, `local`, or `cron:<job>` can collide across agents and leak/mix another agent’s history. I would default-filter by `agent_id` and add an explicit `all_agents=True` escape hatch if needed.

5. **Tool naming is semantically too close to the ReMe memory tools.**
   There is no direct FunctionTool name collision (`memory_search` vs `execute_python`), but the prompt/docs describe scroll as memory/recall/search via `ms.search`, which overlaps conceptually with ReMe memory. This may make the model choose the wrong retrieval path. I suggest renaming the scroll tool to something more explicit, e.g. `recall_history_python` (policy name `RecallHistoryPython`), so ReMe remains long-term semantic memory and scroll is clearly raw conversation-history recall.

Overall, the feature is impressive; I think the main thing is making the failure modes and naming boundaries crisp before enabling it broadly.

niceIrene:
@jinliyl Thanks for the thorough review!

1. **Scroll evicting history when recall is unavailable.** Fixed by falling back to native context whenever governance/sandbox initialization fails. The builder now only wires scroll when the sandboxed recall tool can actually run, so we never end up in a state where history is compacted into the eviction index but the recall path is dead.

2. **`execute_python` returning `RUNNING` for final results.** Fixed — it now returns `SUCCESS` when `code == 0` and `ERROR` otherwise.

3. **Raw text passed into FTS5 `MATCH`.** Added a sanitization step before the FTS query, so plain-language inputs like `C++`, `foo-bar`, or `what?` no longer raise syntax errors and fall back cleanly.

4. **`ms.session()` not agent-scoped.** Bug fixed — `session()` now default-filters by `agent_id`, consistent with `sessions()`, with an `all_agents=True` escape hatch for the cross-agent case.

5. **Naming overlap with ReMe memory tools.** Renamed the tool as suggested and removed any keywords like "memory"/"recall as memory" from the tool description and system prompt. Scroll is now clearly positioned as pure raw conversation-history recall, leaving ReMe as the long-term semantic memory path so the model doesn't conflate the two.

jinliyl:
LGTM

rayrayraykk:
LGTM please solve the conflicts thanks.

niceIrene:
Conflict resolved.

jinliyl (approved):

rayrayraykk (approved):
LGTM

src/qwenpaw/cli/agents_cmd.py
```diff
@@ -502,6 +502,126 @@ def list_agents(ctx: click.Context, base_url: Optional[str]) -> None:
     print_json(agent_tools.list_agents_data(base_url))
 
 
+@agents_group.command("migrate-history")
```
ekzhu: Do we must have this as public CLI command? Is there a way to do this migration internally while only expose the trigger via Console UI?

src/qwenpaw/cli/history_cmd.py
```diff
@@ -0,0 +1,172 @@
...
+    if not ws_root.exists():
+        return []
+    return sorted(
+        p / _DEFAULT_DB
+        for p in ws_root.iterdir()
+        if p.is_dir() and (p / _DEFAULT_DB).exists()
+    )
+
+
+@click.group("history")
```
ekzhu: Same as migrate-history, do we really need this as public CLI command that we want to commit to in the future?

src/qwenpaw/agents/context/types.py
```diff
@@ -0,0 +1,39 @@
...
+    """One durable row appended to ``conversation_history``.
+
+    Beyond the flattened ``content``, an entry carries the structured payload
+    of its turn so the row can reconstruct the exact wire-format Msg later:
+    ``blocks`` holds the serialized block dicts, ``tool_input`` the tool args,
+    ``tool_call_id`` links a call to its result, ``tool_state`` the result
+    state. Structured fields default to empty.
+    """
+
+    kind: str  # "model_turn" | "context_msg" | "tool_result"
```
ekzhu: `Literal["model_turn", "context_msg", "tool_result"]`

ekzhu (approved):

PR #5324, opened 2026-06-19
fix(files): use inline content-disposition for file preview (#5320)

### LINKED ISSUE #5320: [Bug]: v1.1.12升级后，用send_file_to_user发送图片时，聊天窗口中只执行命令，显示发送成功，但是没有显示图片。之前没升级前都是正常可以显示图片的
## QwenPaw Version

v1.1.12

## Description

[Bug]: 升级到v1.1.12后，用send_file_to_user发送图片时，聊天窗口中只执行命令并显示发送成功，但是不显示发送的图片。没升级前都是可以正常显示要发送图片的。同时，之前聊天窗口中曾经发送过的图片也消失了。

但是用这个命令给飞书发图片，飞书中能看到。

nguyenthanhthe:
Closes #5320  
  
FileResponse defaults to content_disposition_type="attachment", causing  
browsers to download files instead of displaying them inline. This  
broke send_file_to_user image display after v1.1.12 (switched from  
Response+bytes to FileResponse in PR #5115).  
  
## Root Cause  
  
In Starlette's FileResponse, the default content_disposition_type is  
"attachment" when a filename is provided. This means browsers treat  
the response as a download, not an inline image. The previous code  
used Response(content=bytes, headers={"content-disposition":  
"inline; ..."}) which displayed images correctly.  
  
## Changes  
  
- **files.py**: Add content_disposition_type="inline" to FileResponse  
- **utils.ts**: Normalize Windows backslashes to forward slashes in toDisplayUrl()

### DIFF
--- a/console/src/components/Chat/ToolCards/shared/utils.ts
+++ b/console/src/components/Chat/ToolCards/shared/utils.ts
@@ -17,7 +17,13 @@ export function toDisplayUrl(url: string): string {
   if (url.startsWith("http://") || url.startsWith("https://")) return url;
   if (url.startsWith("data:")) return url;
   if (url.startsWith("file://")) url = url.replace("file://", "");
-  return chatApi.filePreviewUrl(url.startsWith("/") ? url : `/${url}`);
+  // ponytail: normalize Windows backslashes to forward slashes
+  // so paths like C:\Users\... work in HTTP URLs.
+  // ceiling: if we need more URL sanitisation, extract into a helper.
+  const normalized = url.replace(/\\/g, "/");
+  return chatApi.filePreviewUrl(
+    normalized.startsWith("/") ? normalized : `/${normalized}`,
+  );
 }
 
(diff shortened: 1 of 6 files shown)


maximilize:
Thanks — switching to inline is the right call for media. One defense-in-depth note: `FileResponse` infers Content-Type from the extension, so with `content_disposition_type="inline"` an `.html`/`.svg`/`.xml` file now renders in the app origin, and there's no `X-Content-Type-Options: nosniff` anywhere in the app (checked repo-wide). It's auth-gated and only matters once such a file exists in a previewable path, so this is hardening rather than a boundary issue — but cheap to tighten while here:

- add `headers={"X-Content-Type-Options": "nosniff"}` to the `FileResponse`, and
- keep `inline` only for known-safe types (image/video/audio/pdf); fall back to `attachment` for `text/html`, `image/svg+xml`, `*/xml`.

Happy to send a small follow-up if useful.

PR #5325, opened 2026-06-19
feat(memory): add optional recency-aware ranking for memory_search daily notes (#5316)

### LINKED ISSUE #5316: [Feature]: Add optional recency-aware ranking for memory_search daily notes
## Summary

Add an optional recency-aware ranking step for `memory_search` results from daily memory files (`memory/YYYY-MM-DD.md`), so recent notes can rank higher when relevance scores are similar.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Problem / Motivation

QwenPaw already encourages agents to write daily memory notes and later consolidate them into `MEMORY.md`.

However, `memory_search` currently ranks results mainly by semantic / FTS relevance. For daily notes, this can surface old snippets above recent snippets even when both are similarly relevant.

In personal assistant memory, recency is often an important signal:

- a recent preference should usually override an old preference
- a recent project decision should usually be more useful than an old matching note
- recent daily notes are often more actionable than older daily notes

`MEMORY.md` should stay as curated long-term memory, but daily notes would benefit from an optional recency signal.

## Proposed Solution

Add an optional recency boost / temporal decay step for `memory_search` results from dated daily memory files.

Suggested behavior:

- Only apply to files matching `memory/YYYY-MM-DD.md`
- Do not apply to `MEMORY.md`
- Keep the current behavior by default
- Add config options, for example:
  - `memory_search_recency_boost_enabled: false`
  - `memory_search_recency_half_life_days: 30`
- When enabled, adjust result ranking after relevance scoring:

```text
age_days = today - date parsed from memory/YYYY-MM-DD.md
recency_factor = exp(-ln(2) * age_days / half_life_days)
final_score = relevance_score * recency_factor
```

Implementation note:

A minimal first version could be implemented as a post-processing step in `ReMeLightMemoryManager.memory_search`, without changing the ReMeLight storage/indexing layer. If this belongs upstream in ReMeLight instead, I am happy to adjust the PR direction.

## Alternatives Considered

- Rely only on `MEMORY.md` dream consolidation
  This helps long-term memory, but daily notes are still searched directly and can contain outdated snippets.

- Add recall tracking / promotion
  This could be useful later, but it requires extra persistent state and promotion policy. Recency ranking is smaller and easier to validate first.

- Always prefer recent notes
  This may hurt recall quality, so the feature should be optional and score-based.

## Additional Context

This is intended as a small, backward-compatible improvement to memory recall quality. It does not change the storage model, dream flow, or default ranking behavior.

Potential test cases:

- `MEMORY.md` result should not be decayed
- `memory/2026-06-18.md` should rank above `memory/2026-01-01.md` when relevance scores are close
- feature disabled should preserve current result order

## Willing to Contribute

If maintainers agree with the direction, I can help implement this as a small backward-compatible PR, preferably with the feature disabled by default and covered by unit tests

nguyenthanhthe:
Closes #5316. Implements an optional recency-aware ranking step for memory_search daily memory files (memory/YYYY-MM-DD.md). Applies an exponential temporal decay based on the age of daily notes compared to the current date, respecting the user's timezone. Decayed results falling below min_score are filtered out.

### DIFF
--- a/src/qwenpaw/agents/memory/reme_light_memory_manager.py
+++ b/src/qwenpaw/agents/memory/reme_light_memory_manager.py
@@ -115,8 +115,8 @@ class ReMeLightMemoryManager(BaseMemoryManager):
 
         fts_enabled = EnvVarLoader.get_bool("FTS_ENABLED", True)
 
-        agent_config = load_agent_config(self.agent_id)
-        reme_cfg = agent_config.running.reme_light_memory_config
+        self.agent_config = load_agent_config(self.agent_id)
+        reme_cfg = self.agent_config.running.reme_light_memory_config
         rebuild_on_start = reme_cfg.rebuild_memory_index_on_start
 
         store_name = "memory"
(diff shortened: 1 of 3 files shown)


maximilize:
Thanks for this — the core matches the #5316 spec well: post-processing in `ReMeLightMemoryManager.memory_search` via a new `_apply_recency_boost` helper (no storage/index changes), disabled-by-default config keys `memory_search_recency_boost_enabled` / `memory_search_recency_half_life_days` at `config.py:686-700`, `MEMORY.md` excluded via the `memory/(\d{4}-\d{2}-\d{2})\.md$` regex, the formula `score * exp(-ln2 * age_days / half_life)`, and unit tests covering the three requested cases.

A few things before merge:

1. **`pytz` isn't a declared dependency.** `_apply_recency_boost` does `import pytz`, but the codebase standardizes on stdlib `zoneinfo.ZoneInfo` (`get_current_time.py:6`, `app/runner/utils.py:9`) and ships `tzdata` (`pyproject.toml:38`); `pytz` is in neither `pyproject.toml` nor the lockfiles. The import is wrapped in try/except that silently falls back to naive `date.today()`, so it won't crash — but the timezone-aware behavior will silently never apply in a default install, and that path is untested (the test stubs `user_timezone=None`). Switching to `zoneinfo` would match the project and remove the silent no-op.
2. **Beyond spec:** the helper also drops results whose *decayed* score falls below `min_score` and re-sorts. #5316 only asked to re-rank, not filter — intentional?
3. **Scope:** three of the five changed files look unrelated to recency ranking — a Windows backslash fix in `console/.../ToolCards/shared/utils.ts`, `content_disposition_type="inline"` in `app/routers/files.py` (overlaps #5324), and a `dream()` `agent_id` → `target_agent_id` refactor in this same file. Splitting those out would make this much easier to review.
4. Minor: decay runs after `reme` has already truncated to `max_results`, so a recent note ranked just outside the top-N by relevance can't be recovered. Fine for a minimal first pass — just noting the limitation.

maximilize:
All four points from my earlier review are addressed — verified on the latest push:

1. `pytz` → stdlib `zoneinfo.ZoneInfo`, matching the rest of the codebase (`datetime`/`json` are already module-level imports, so the tz path resolves), and there's now a real tz test (`ZoneInfo("Asia/Ho_Chi_Minh")`).
2. The decayed-score `min_score` filtering is gone — it only re-ranks/re-sorts now, with `test_recency_ranking_no_min_score_filtering` asserting decayed-below-threshold notes are kept.
3. Scope trimmed to the 3 relevant files; the unrelated `files.py` / console `utils.ts` / `target_agent_id` changes are dropped. The Windows-path concern now lives where it belongs — the daily-note regex matches both separators: `memory[/\\](\d{4}-\d{2}-\d{2})\.md$`.

The point-4 limitation I noted still stands (decay runs after `reme` truncates to `max_results`, so a recent note just outside the top-N can't be recovered) — fine for a minimal first pass, just noting it persists.

PR #5326, opened 2026-06-19
feat: minimize window to system tray on close (#5312)

nguyenthanhthe:
When the user clicks the window close button, hide the window to the system tray instead of quitting the app. A tray icon with a right-click menu lets the user show the window or truly quit.

### DIFF
--- a/console/src-tauri/Cargo.toml
+++ b/console/src-tauri/Cargo.toml
@@ -21,7 +21,7 @@ serde = { version = "1.0", features = ["derive"] }
 log = "0.4"
 futures-util = "0.3"
 reqwest = { version = "0.12", default-features = false, features = ["stream"] }
-tauri = { version = "2.10.3", features = [] }
+tauri = { version = "2.10.3", features = ["tray-icon"] }
 tauri-plugin-log = "2"
 tauri-plugin-shell = "2"
 tauri-plugin-dialog = "2"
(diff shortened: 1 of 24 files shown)


jinglinpeng:
Thanks @nguyenthanhthe! The direction is great and the Tauri v2 approach is
right. A few things to consider:
**1. Make the close behavior configurable, and consider per-platform defaults.**
This PR unconditionally changes the window "X" button from *quit* to
*minimize to tray* on every platform, with no way to quit from the window
itself — this is the setting #5312 (suggestion 3) asked for. 

**2. macOS: hidden window can't be restored from the Dock.**
Since the window is now hidden (instead of closed), clicking the Dock icon does
nothing — only the tray menu can bring it back. Please handle `RunEvent::Reopen`
to `show()` + `set_focus()` the window.

**3. macOS: tray left-click convention.**
`show_menu_on_left_click(false)` + left-click-to-show feels native on Windows,
but on macOS the status-bar item usually opens the menu on left click.

**4. Unused `Manager` import in `lib.rs`.**
The change removes the only `Manager` usage (`window.app_handle()`), so the
import now warns — and fails if CI runs with `-D warnings`. Drop it and re-check
`cargo build`.

**5. i18n for tray labels.**
"Show Window" / "Quit" / the tooltip are hardcoded English; the console supports
zh/en/ja/ru, so ideally the tray menu follows the selected language.

Issue #5208, opened 2026-06-15
Bug: Assistant message count mismatch when model returns reasoning blocks with type "reasoning" instead of "thinking"


lecheng2018:
## Description

When using LongCat-2.0-Preview (OpenAI-compatible API) as the model provider, every conversation turn produces a WARNING:

```
Assistant message count mismatch after formatting (N expected survivors, N-1 actual).
Skipping reasoning_content injection for this turn.
A block type is likely being dropped by the base formatter without being listed
in _FORMATTER_SKIPPED_TYPES — please investigate.
```

222 warnings accumulated in 24 hours, affecting every conversation turn.

## Root Cause

The model returns assistant messages with reasoning content using block type `"reasoning"` (or `"reasoning_content"`), but `_FORMATTER_SKIPPED_TYPES` in `model_factory.py:667` only contains `{"thinking", "file"}`.

When a message contains only a reasoning block (no text or tool_calls):
1. `OpenAIChatFormatter._format()` does not recognize `"reasoning"` type, skips it
2. The message ends up with content=None and tool_calls=None, so it gets dropped entirely
3. But `reasoning_contents` dict still retains the reasoning data for that message
4. This causes `len(aligned_reasoning) != len(out_assistant)`, always off by 1

### Key code locations

**model_factory.py:667** — skipped types definition:
```python
_FORMATTER_SKIPPED_TYPES = frozenset({"thinking", "file"})
```

**model_factory.py:788** — reasoning extraction (only checks `"thinking"`):
```python
for block in msg.get_content_blocks():
    if block.get("type") == "thinking":
        thinking = block.get("thinking", "")
        if thinking:
            reasoning_contents[id(msg)] = thinking
        break
```

**_openai_formatter.py:351** — unknown block types are skipped:
```python
logger.warning(
    "Unsupported block type %s in the message, skipped.",
    typ,
)
```

## Impact

- **Functional**: Medium — reasoning_content injection is skipped, may degrade model reasoning quality
- **Performance**: None
- **Data**: No data loss, reasoning content is not correctly passed through

### Option A: Extend `_FORMATTER_SKIPPED_TYPES`
```python
_FORMATTER_SKIPPED_TYPES = frozenset({"thinking", "file", "reasoning", "reasoning_content"})
```

### Option B: Enhance reasoning_content extraction
```python
for block in msg.get_content_blocks():
    if block.get("type") in ("thinking", "reasoning", "reasoning_content"):
[...]

PR #5340, opened 2026-06-20
fix(model_factory): switch formatter drop detection from blacklist to whitelist

rankaiyx:
## Description

When a user interrupts agent generation (stop button), AgentScope `ReActAgent._reasoning` writes an empty message (`content=[]`) to memory via its `finally` block. On the next turn, the predictor (blacklist logic) assumed empty messages **survive** the base formatter, but AgentScope's `OpenAIChatFormatter` actually **drops** them — causing an assistant message count mismatch and skipping `reasoning_content` injection for the entire turn.

**Result:** ALL historical thinking content is lost → KV cache miss → latency spike.

### Root Cause

The predictor and base formatter use opposite approaches to decide which assistant messages survive formatting:

| Approach | Predictor (before) | Base Formatter (AgentScope) |
|---|---|---|
| Logic | Blacklist: known-dropped `{thinking, file}` | Whitelist: known-surviving `{text, tool_use, image}` |
| Empty `[]` | **Survives** (falsy short-circuit) | **Dropped** |
| `audio` (assistant) | **Survives** | **Dropped** |
| Unknown types | **Survives** | **Dropped** |

The blacklist approach is fragile — every new block type added by AgentScope that isn't in the blacklist causes silent misalignment, leading to reasoning content being injected into the wrong messages.

**Prior fix (#1557 for #1532):** Added `is_thinking_only` prediction to skip thinking-only messages, but the blacklist (`_FORMATTER_SKIPPED_TYPES`) still misses empty content (`[]` — falsy short-circuit), `audio` on assistant, and non-standard block types like `reasoning` (#5208). This PR completes the fix by inverting to a whitelist approach.

### Fix

Invert the predictor from blacklist to whitelist:

- `_FORMATTER_SKIPPED_TYPES` → `_FORMATTER_SURVIVING_TYPES`
- `all(b in SKIPPED)` → `all(b not in SURVIVING)`
- Empty content is now correctly predicted as dropped (`all([])` = `True`)

This is **fail-safe**: if AgentScope adds support for a new block type, the whitelist triggers a mismatch warning and skips reasoning injection safely, while the blacklist would silently misalign (wrong reasoning injected into subsequent messages).

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Core

## Checklist

- [x] `pre-commit run --files src/qwenpaw/agents/model_factory.py` passes
- [x] black: Passed
- [x] flake8: Passed
- [x] mypy: Passed
- [x] pylint: Passed
- [ ] Documentation updated (N/A — internal logic change)

## Local Verification

Re-entered a session containing `content: []` messages (from interrupted generation):

- **Before fix:** formatted messages missing all `reasoning_content`; log shows `Assistant message count mismatch` warning
- **After fix:** `reasoning_content` injected correctly, no mismatch warning

The verification operation is performed and confirmed by  human (rankaiyx). 

## Note on Reproducibility

Naturally triggering the creation of a session containing `content: []` empty messages requires interrupting agent generation before the first streaming token arrives — the timing window is extremely narrow and depends on model latency, making it difficult to reproduce reliably.

### DIFF
--- a/src/qwenpaw/agents/model_factory.py
+++ b/src/qwenpaw/agents/model_factory.py
@@ -660,11 +660,13 @@ def _fix_image_mime_types(messages: list[dict]) -> None:
 _MEDIA_BLOCK_TYPES = ("image", "audio", "video")
 
 # Block types that the upstream agentscope OpenAI / Gemini formatters
-# silently drop. We track them here so we can predict which assistant
-# messages will be dropped before alignment in FileBlockSupportFormatter.
-# Keep this in sync with the `else: logger.warning("Unsupported block
-# type ...")` branch in agentscope's _openai_formatter.
-_FORMATTER_SKIPPED_TYPES = frozenset({"thinking", "file"})
+# keep in assistant messages (produce content_blocks or tool_calls).
+# Messages whose blocks are ALL outside this set will be dropped.
+# Using a "known-surviving" whitelist is safer than a "known-dropped"
+# blacklist: if AgentScope adds support for a new block type, a
+# blacklist would silently misalign (wrong reasoning injected), while
+# a whitelist triggers a mismatch warning and skips injection safely.
+_FORMATTER_SURVIVING_TYPES = frozenset({"text", "tool_use", "image"})
 
 
 def _fixup_media_list(items: list) -> None:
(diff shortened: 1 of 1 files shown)


qbc2016:
Thanks for the contribution! The whitelist approach is indeed cleaner in principle. 
However, in the current AgentScope 2.0.x setup, the original trigger no longer exists — `_save_to_context` only runs after a complete response, so interruption never writes a `content=[]` message to context. 
We don't have a concrete scenario where this fix would make a difference today, so we'll close this for now. Feel free to reopen if a reproducing case comes up.

PR #5347, opened 2026-06-20
fix(crons): drop invalid jobs.json entries on startup (migration)

### LINKED ISSUE #4835: [Bug] One invalid job in jobs.json causes entire workspace to fail

ly-wang19:
Fixes #4835

The migration approach discussed in #5040 (root-cause cleanup rather than runtime tolerance).

## What

`migrate_invalid_jobs()` runs on workspace start — right after the existing weixin→wechat jobs migration, **before `cron_manager` loads**. It validates each job in `jobs.json`, keeps the valid ones **verbatim**, drops invalid ones, and **backs up the original** before rewriting.

## Why

A single invalid job (e.g. one missing `dispatch.target.session_id`) makes `JobsFile.model_validate` fail for the whole file, so `cron_manager` can't start and the workspace hangs.

As traced in #5040, such entries don't arise from normal operation — writes always serialize a validated `JobsFile`. They come from a **schema change across upgrades**, or a manual edit / interrupted write, and `migration.py` had no content migration for them. This adds it.

Relative to #5040 (which tolerates invalid jobs at *load* time), this **cleans the file once on upgrade** — the root-cause fix you preferred — backing up the original so nothing is lost. The two are complementary; happy to keep either or both.

## Behaviour

- All jobs valid → no-op (no rewrite, no backup).
- Some invalid → drop them, back up original to `jobs.json.<hex>.invalid-jobs.bak`, atomically rewrite the cleaned file. Idempotent.

## Verification

- `black==23.3.0`, `flake8==6.1.0`, `pylint==3.0.2` (repo args) — clean (10/10)
- Ran it directly: a `jobs.json` with one valid + one invalid job (which makes `JobsFile.model_validate` raise) → after migration the invalid job is dropped, the valid one kept verbatim, `JobsFile.model_validate` succeeds, the original (incl. the dropped job) is backed up, and a re-run on the clean file is a no-op.

### DIFF
--- a/src/qwenpaw/app/crons/repo/json_repo.py
+++ b/src/qwenpaw/app/crons/repo/json_repo.py
@@ -10,8 +10,10 @@ import uuid
 from pathlib import Path
 from urllib.parse import quote
 
+from pydantic import ValidationError
+
 from .base import BaseJobRepository
-from ..models import CronExecutionRecord, JobsFile
+from ..models import CronExecutionRecord, CronJobSpec, JobsFile
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


maximilize:
Triage note: #5040 and #5347 overlap — both fix the same root cause (#4835: one invalid entry in `jobs.json` blocks workspace startup), at different layers, and they conflict textually (both edit the imports in `src/qwenpaw/app/crons/repo/json_repo.py`), so only one should land.

Suggest preferring **#5040**: it hardens the actual read path — `load()` (`json_repo.py:42-47`) — so invalid jobs are tolerated everywhere, and it ships tests. #5347's startup migration leaves `load()` fragile, but its backup-before-discard (`.invalid-jobs.bak`) is worth folding in, since #5040 drops invalid jobs with no recoverable copy.

One gap in both: `json.loads()` at `json_repo.py:46` runs before the validation `try`, so a truncated/corrupt file still crashes `load()`. Catching `JSONDecodeError` there too would fully close #4835.

lalaliat (commented):
Thank you for your contribution! Since these two PRs are complementary, please check and solve comments raised by Copilot in #5040 first. BTW, please remove the test file in #5040 before merging. Thanks again.

PR #5348, opened 2026-06-20
feat(runner): freeze env_context date per session to preserve KV cache prefix

rankaiyx:
## Description

The System Prompt injects `Current date: YYYY-MM-DD` via `build_env_context()`. This date is computed on every request. When the date changes (for example, crossing midnight), the System Prompt prefix changes. This invalidates the entire KV Cache. As a result, tokens are re-computed and responses get slower for sessions that span across day boundaries.

This PR freezes the date at the session level. The date is captured once when a session starts, and reused for all turns in that session. The frozen date is refreshed only when the user switches to a different session.

**Related Issue:** Relates to #3891

**Security Considerations:** N/A — no user data is exposed, no external calls are added.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

- [x] Manually tested: the date stays the same across turns within a session. It refreshes correctly when switching to another session.
- The second commit (`fix: add global declaration for module-level variables in query_handler`) is a fix discovered during testing. It has been resolved and verified.

## Additional Notes

- **Main benefit scenario:** Local self-hosted LLM with long context tasks that span across days. This optimization matters most for self-hosted setups where KV cache is managed by the deployment itself.
- Module-level globals are safe for single-worker deployment. The check-and-set has no `await` between the comparison and the assignment. This makes it atomic under Python's single-threaded asyncio event loop.
- Session switching refreshes the date. A new session always gets a fresh frozen datetime.
- Timezone is respected. The frozen datetime uses the user's configured timezone (from `load_config().user_timezone`), and falls back to UTC on invalid timezone.
- Backward compatible: `build_env_context()` defaults `frozen_now=None`. Existing behavior is preserved when the parameter is not passed.

### DIFF
--- a/src/qwenpaw/app/chats/utils.py
+++ b/src/qwenpaw/app/chats/utils.py
@@ -67,6 +67,7 @@ def build_env_context(
     add_hint: bool = True,
     default_shell: Optional[str] = None,
     project_dir: Optional[str] = None,
+    frozen_now: Optional[datetime] = None,
 ) -> str:
     """
     Build environment context with current request context prepended.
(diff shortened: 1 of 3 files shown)


xieyxclack:
IMO, it doesn't seem to make much sense. If the conversation spans multiple days without updating the date, it may lead to the agent having an incorrect perception of the current date.

rankaiyx:
Sure, this is a trade-off. In most cases, the good parts are bigger than the bad parts for running local LLMs and for people using expensive providers. Maybe a better way is to give a config switch? I also tried a version that adds time updates, but it made the code much more complex. It did not feel worth it.

rankaiyx:
Added a configuration option and disabled the feature by default.

rankaiyx:
It has already been ported to the main branch.

xieyxclack:
@rankaiyx The feature is supported in https://github.com/agentscope-ai/QwenPaw/pull/5923

PR #5352, opened 2026-06-21
agent重新加载后，chanel会跟着old_instance stop，不会启动。

junphine:
## Description

1. agent重新加载后，chanel会跟着old_instance stop，不会启动。

2. chanel实现了register_app_routes，raise  '_IncludedRouter' object has no attribute 'path'

## Type of Change

- [ ] Bug fix

## Component(s) Affected

## Checklist
- [ ] Ready for review

### DIFF
--- a/src/qwenpaw/app/channels/registry.py
+++ b/src/qwenpaw/app/channels/registry.py
@@ -173,9 +173,13 @@ def register_custom_channel_routes(app) -> None:
             hook = getattr(mod, "register_app_routes", None)
             if not callable(hook):
                 continue
-            prev_routes = {r.path for r in app.routes}
+            prev_routes = {r.path for r in app.routes if hasattr(r,'path')}
+            last_route = app.routes[-1]
             hook(app)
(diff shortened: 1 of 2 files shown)


hongxicheng:
Thanks for the contribution!

All built-in channels (long-lived WebSocket connections, HTTP webhook listeners, polling-based, etc.) work correctly with the current reload logic. The current ordering (start new → atomic swap → stop old) is an intentional blue-green design for zero-downtime reload.

Could you share what type of channel you're implementing? The issue might be in your channel's start/stop logic rather than the global reload order.

Moving `_graceful_stop_old_instance` before `new_instance.start()` would introduce:
1. A service gap where no channel is handling messages
2. No rollback path if `new_instance.start()` fails (old is already gone)

If your channel conflicts with the old instance during startup (e.g. competing for the same connection), the fix should be within the channel's own handoff logic.

hongxicheng:
Closing this pr due to inactivity. If this is still a problem for you, please feel free to reopen it with any additional details. Thanks!

PR #5399, opened 2026-06-22
feat(providers): support custom model ordering within providers

lecheng2018:
## Description

Support custom model ordering within providers. Users can reorder models via drag-and-drop or up/down buttons in the model management modal, with results persisted to the backend.

## Changes

- **Backend**: Added `sort_order` field to `ModelInfo`, `POST /{provider_id}/models/reorder` endpoint, `reorder_models()` and `reorder_provider_models()` methods
- **Frontend API**: Added `sort_order` type, `getProviderModels()` helper, `ReorderModelsRequest` type, `providerApi.reorderModels()` call
- **UI**: Added drag-and-drop reordering, up/down buttons, undo/reset toolbar in `RemoteModelManageModal`
- **Mobile**: Hidden drag handle on small screens (≤768px), kept up/down buttons
- **i18n**: Added 7 translation keys across all locales
- **Consumers**: Updated 6 model list consumers (Chat ModelSelector, Agent Config, Agent Modal, ModelsSection, etc.) to use `getProviderModels()`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Configuration change
- [x] CSS/Layout improvement
- [ ] Documentation update

## Component(s) affected

- [x] Console (Frontend)
- [x] Backend (Python)
- [ ] Documentation
- [ ] Other

## Testing

- Built and deployed locally on FnOS NAS
- Verified desktop: drag-and-drop, up/down buttons, undo/reset all work
- Verified mobile: drag handle hidden, up/down buttons functional
- Verified no desktop regression (all changes scoped to `@media (max-width: 768px)` for mobile)

## Local Verification

Tested on `localhost:19901`, desktop + mobile viewport. API endpoint verified via curl.


RerankerGuo (changes requested):
Validated exact head `d7084cceaf1659116a134495a8fdc0bf7006b577` and a semantic merge onto current `main` `bacac7410cf2ac122dd892152beff8b03f38729f` (trial commit `aabfbfe7466b800161e24eef44af6874e9dfb5f3`, tree `4b2b6f3cb881303f884afcbade25b6a51609d912`). The three conflicts are straightforward if the rebase preserves current main’s `max_input_length_configured`/thinking fields and provider-specific `active_llm.provider_id` lookup while adding `sort_order`.

Two backend blockers remain:

1. **Built-in model order is lost after a restart.** I ran the reorder in one Python process and loaded the same secret directory in a fresh process. The persisted `openai.json` correctly contained sort orders `0..10`, but the reloaded built-in provider returned the models alphabetically instead of the requested reverse order. A custom-provider control in the same run restored `second, first` correctly. This fails on both the exact head and the current-main trial because `_init_from_storage()` overlays stored built-in model configuration through `_per_model_keys` / `_restore_builtin_model_config()`, and neither path includes `sort_order`.

   ```text
   builtin_expected: gpt-4o-mini, gpt-4o, ..., gpt-5.2
   builtin_actual:   gpt-4.1, gpt-4.1-mini, ..., o4-mini
   custom_expected:  second, first
   custom_actual:    second, first
   ```

2. **The existing class-identity contract now crashes.** Both trees fail the two committed `test_provider_class_identity.py` cases because `get_info()` sorts foreign `ModelInfo` instances via direct `m.sort_order`; those intentionally come from another module identity and do not have the new field.

   ```text
   AttributeError: 'ModelInfo' object has no attribute 'sort_order'
   exact providers: 179 passed, 2 failed
   trial providers/router scope: 369 passed, 2 failed
   ```

Please restore/persist `sort_order` for built-in models, make sorting backward/class-identity compatible (for example via a defaulted lookup or normalization), and add restart plus foreign-model regressions. The Console code also needs an interaction regression for Undo: `handleStartReorder()` clears the snapshot, `handleSaveReorder()` stores the newly saved order, and `handleUndoReorder()` submits that same new order, so Undo currently cannot restore the previous order.

Trial checks otherwise pass: TypeScript, Prettier 3, Black 23.3.0, Flake8 6.1.0, and `git diff --check`.

PR #5417, opened 2026-06-23
fix(server): patch uvicorn get_remote_addr/get_local_addr for Windows ProactorEventLoop

### LINKED ISSUE #5379: [Bug]: 通过Python命令安装后启动，直接报错Internal Server Error
通过Python管理安装QwenPaw最新版本，启动后访问地址直接出现Internal Server Error错误，看日志中错误显示：get_remote_addr(transport)导致，具体请分析或告知原因，提供启动后日志和访问报错时的日志信息

[CWindowssystem32cmd.exe - qwenpaw  app.txt](https://github.com/user-attachments/files/29195621/CWindowssystem32cmd.exe.-.qwenpaw.app.txt)

wangfei010313:
## Description

On Windows with Python 3.12+, `uvicorn` crashes with `Internal Server Error` on every HTTP request. The root cause is `transport.get_extra_info("peername")` returning corrupted data (bytes instead of int for the port field) under the `ProactorEventLoop`, which propagates as:

1. `ValueError: invalid literal for int() with base 10: b'\x00\x00...'` in `uvicorn.protocols.utils.get_remote_addr`
2. `TypeError: Cannot mix str and non-str arguments` in Starlette's URL construction (cascading failure)

This PR monkey-patches `uvicorn.protocols.utils.get_remote_addr` and `get_local_addr` before `uvicorn.run()` is called. The patched wrappers catch `ValueError`, `TypeError`, and `OSError`, returning `None` instead of propagating the exception.

Returning `None` is safe — uvicorn's `httptools_impl.py` already declares `self.client` and `self.server` as `Optional[Tuple[str, int]]`, and Starlette's `URL` class handles `server=None` correctly.

**Related Issue:** Fixes #5379

**Security Considerations:** None. The patch only affects error handling for corrupted transport data and does not change authentication, authorization, or data flow.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

### DIFF
--- a/reproduce_issue_5379.py
+++ b/reproduce_issue_5379.py
@@ -1,0 +1,276 @@
+"""
+End-to-end reproduction of Issue #5379.
+
+The bug chain on Windows ProactorEventLoop:
+  1. transport.get_extra_info("peername") returns corrupted bytes for port
+  2. get_remote_addr() partially processes this → returns (str, bytes)
+  3. self.client is set to corrupted tuple (str, bytes)
+  4. When building the ASGI scope, Starlette mixes str and bytes → TypeError
+  5. 500 Internal Server Error
+
+This script reproduces steps 2-5 by injecting corrupted client data into
+the protocol, then sending a real HTTP request via raw sockets.
+
+Run:  python reproduce_issue_5379.py
+"""
+from __future__ import annotations
+
+import importlib
+import socket
+import sys
+import threading
+import time
+
+SEPARATOR = "=" * 72
+CORRUPTED_PORT = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+
+
+def _make_app():
+    from fastapi import FastAPI
+    from fastapi.responses import JSONResponse
+
+    app = FastAPI()
+
+    @app.get("/api/agent/process")
+    async def process():
+        return JSONResponse({"status": "ok"})
+
+    @app.get("/health")
+    async def health():
+        return JSONResponse({"status": "ok"})
+
+    return app
+
+
+def _find_free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("127.0.0.1", 0))
+        return s.getsockname()[1]
+
+
+def _send_raw_http(port: int, path: str = "/api/agent/process") -> str:
+    """Send a single raw HTTP/1.1 request and return the response."""
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.settimeout(5)
+    try:
+        sock.connect(("127.0.0.1", port))
+        request = (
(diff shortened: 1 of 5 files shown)


RerankerGuo (changes requested):
I found an unrelated Desktop regression and a test-gate gap at exact head `3d5b3981ff983f2c3b53042f4a3c9e1cb737b355`.

1. **The Tauri sidecar change breaks every frozen Desktop subprocess spawn.** [entry.py line 88](https://github.com/agentscope-ai/QwenPaw/blob/3d5b3981ff983f2c3b53042f4a3c9e1cb737b355/src/qwenpaw/tauri/entry.py#L88) changes the valid `os.path.normcase(...)` calls to `os.normcase(...)`, but Python's `os` module has no `normcase` attribute. `_install_subprocess_guard()` invokes this helper for every `subprocess.Popen` call in a frozen Desktop build, before it can decide whether redirection is needed.

   I called the production helper with the normal `[sys.executable, "-m", "qwenpaw", "app"]` shape:

   ```text
   current main bacac741: returned None
   exact 3d5b3981: AttributeError: module 'os' has no attribute 'normcase'
   current-main trial b3ec48fc: same AttributeError
   ```

   Please revert this unrelated one-line change to `os.path.normcase`.

2. **The authored tests are outside the repository's test collection and do not pass when run explicitly.** `pyproject.toml` sets `testpaths = ["tests"]`, while this PR adds `test_windows_patch.py` at repository root. Consequently, the normal test workflow never collects it. Running it explicitly against the clean current-main trial produces:

   ```text
   2 failed, 3 passed, 1 skipped
   ```

   The implementation returns early unless `sys.platform == "win32"`, but [the two core patched-address assertions](https://github.com/agentscope-ai/QwenPaw/blob/3d5b3981ff983f2c3b53042f4a3c9e1cb737b355/test_windows_patch.py#L49-L83) are not Windows-gated or supplied a simulated Windows platform, so they call unpatched Uvicorn helpers and raise `ValueError`. Please move focused tests under `tests/`, make the platform setup deterministic, and verify both malformed and valid address behavior through the production patch.

The current-main trial also fails the locked changed-file hooks: encoding pragma and Black modify the two root scripts, while Flake8 and Pylint report committed issues. The remote `run` and `Test Summary` checks are likewise failing.

Finally, the linked #5379 was closed after the reporter confirmed uninstalling a VPN plus resetting Winsock resolved the problem. Its discussion explicitly paused this global Uvicorn monkey-patch because of client-IP/auth and reload-process risk. Please confirm there is still a reproducible QwenPaw defect requiring this patch; otherwise closing this stale PR is safer than carrying a process-wide override.

PR #5492, opened 2026-06-24
feat(plugins): support installing plugins via pip from PyPI

### LINKED ISSUE #5484: [Feature]: Support installing plugins via pip from PyPI
## Motivation

Currently, QwenPaw plugins can only be installed via ZIP archives (upload, URL, or local directory). While this works, it requires a separate distribution and installation mechanism that is non-standard for the Python ecosystem.

Projects like [Hermes Agent](https://github.com/NousResearch/hermes-agent) take a more Pythonic approach — plugins are distributed as standard Python packages installable via `pip`, leveraging PyPI for versioning, dependency resolution, and discovery. This makes it trivially easy for the community to publish, install, and update plugins.

## Proposal

Add support for installing QwenPaw plugins via `pip install`, so that:

1. **Third-party developers** can publish plugins to PyPI as standard Python packages (e.g., `pip install qwenpaw-plugin-xxx`)
2. **Users** can install plugins with a simple pip command instead of managing ZIP files
3. **Dependency management** is handled by pip/uv natively, rather than the current `requirements.txt` post-install step

## Suggested Approach

- Define a plugin discovery mechanism based on [Python entry points](https://packaging.python.org/en/latest/specifications/entry-points/) (e.g., `qwenpaw.plugins` entry point group)
- Plugin authors add an entry point in their `pyproject.toml`:
  ```toml
  [project.entry-points."qwenpaw.plugins"]
  my_plugin = "qwenpaw_plugin_xxx:register"
  ```
- QwenPaw scans installed entry points at startup to discover and load plugins
- Keep the existing ZIP-based installation as an alternative for non-PyPI distribution (e.g., internal/enterprise plugins, development)

## Benefits

- **Lower barrier for community contributions** — anyone familiar with Python packaging can publish a plugin
- **Standard tooling** — reuse pip, uv, poetry, etc. for dependency resolution and version management
- **Better security** — PyPI has vulnerability scanning, signed releases, and audit trails
- **Easier updates** — `pip install --upgrade qwenpaw-plugin-xxx`
- **IDE support** — installed packages are automatically discoverable by IDEs and type checkers

## Current State

| | Hermes Agent | QwenPaw (current) |
|---|---|---|
| Plugin distribution | PyPI package | ZIP archive |
| Install command | `pip install hermes-agent[plugin]` | Upload ZIP / URL / local dir |
| Dependency management | pip native | Post-install `pip install -r requirements.txt` |
| Discovery | `plugin.yaml` in package | `plugin.json` in extracted directory |
| Version management | pip/PyPI | Manual version check |

## Alternatives Considered

- Keep ZIP-only: works but creates friction for the community
- Use a custom plugin registry: reinvents what PyPI already provides well

## Additional Context

This would be complementary to the existing ZIP-based installation, not a replacement. Both methods can coexist — ZIP for development/internal use, pip for published community plugins.

qwtsc:
Add support for discovering and loading plugins installed as pip packages using Python entry points (`qwenpaw.plugins` group).

### Core Plugin System
- Add `InstallSource` enum (zip/pip) to `PluginRecord`
- Add `discover_pip_plugins()` to scan `qwenpaw.plugins` entry points
- Add `load_pip_plugin()` for installing and loading pip packages
- Add `uninstall_pip_plugin()` for clean removal
- Update `load_all_plugins()` to auto-discover pip-installed plugins

### CLI Commands
- Add `qwenpaw plugin install-pip <package>` command
- Add `qwenpaw plugin uninstall-pip <plugin_id>` command
- Support hot-install via API when app is running

### REST API Endpoints
- `POST /plugins/install-pip` - Install from PyPI package
- `DELETE /plugins/{id}/pip` - Uninstall pip-installed plugin
- Add `install_source` field to plugin list response

### Tests
- Add comprehensive unit tests for pip plugin functionality

Closes #5484

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,6 +98,13 @@ copaw = "qwenpaw.cli.main:cli"
 # [project.entry-points."qwenpaw.doctor"]
 # my_pkg = "my_pkg.doctor:doctor_notes"
 # Legacy group "copaw.doctor" is still loaded when present.
+#
+# Plugin entry points: third-party plugins installed via pip declare
+# themselves under this group.  The entry point value must point to a
+# module (or callable) inside a package that ships a ``plugin.json``
+# manifest as package data.
+# [project.entry-points."qwenpaw.plugins"]
+# my_plugin = "my_plugin_package:register"
 
 [project.optional-dependencies]
 dev = [
--- a/src/qwenpaw/app/routers/plugins.py
+++ b/src/qwenpaw/app/routers/plugins.py
@@ -472,6 +472,7 @@ async def list_plugins(request: Request):
                 "loaded": True,
                 "plugin_type": manifest.plugin_type,
                 "frontend_entry": manifest.entry.frontend,
+                "install_source": record.install_source.value,
             },
         )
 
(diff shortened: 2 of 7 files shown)


RerankerGuo (changes requested):
I found three destructive package-identity bugs that are not covered by the current single-plugin tests, plus a current-main compatibility failure.

I validated exact head `dbb7f270` and a clean current-main trial commit `eef7bb20` / tree `6846573f`. The authored file passes `19/19`; the complete plugin unit scopes pass `41` exact and `179 passed, 7 skipped` on the trial. Those suites only model one discovered plugin and one matching `.dist-info`, so they do not exercise the identity boundary introduced by this PR.

An isolated multi-plugin/multi-distribution contract produces the same output on exact and trial:

```text
install requested=target-distribution loaded=unrelated-plugin
uninstall target=target-distribution enumerated=['zzz_unrelated-9.0.dist-info'] guessed=zzz-unrelated
force requested=target-distribution unloaded=['already-loaded-a', 'already-loaded-b']
```

1. `load_pip_plugin("target-distribution")` scans every `qwenpaw.plugins` entry point and returns the first unloaded manifest. The condition at `loader.py:255` is effectively always true after the preceding `if manifest.id in self._loaded_plugins: continue`; `_dist_version(package_name)` is never related to the manifest being iterated. If another pip plugin is already installed but not loaded, installing package B can load plugin A and report success for A.

2. `_guess_dist_name()` scans `source.parent.glob("*.dist-info")` and returns the first `Name:` line without checking that the distribution owns `record.source_path` or its entry point. If the target metadata is absent or its directory is not enumerated first, uninstalling a plugin can execute `pip uninstall` against an unrelated package in the same environment. The offline CLI has a separate mismatch: it uninstalls `plugin_id`, although a plugin ID is not required to equal its distribution name.

3. `POST /plugins/install-pip` with `force=true` iterates every discovered pip plugin and unloads every one that is currently loaded before it installs the requested package. Force reinstalling one package must not tear down unrelated plugin providers, commands, hooks, and agent state.

Please establish and persist an explicit identity chain from the requested normalized distribution to its own entry point(s), manifest ID(s), and distribution name. `EntryPoint.dist` / `importlib.metadata.distribution()` can provide authoritative ownership; scanning arbitrary sibling `.dist-info` directories cannot. Force and uninstall should act only on that recorded target. Add tests with at least two installed distributions, differing distribution/package/plugin IDs, an unrelated loaded plugin, and missing target metadata (which should fail closed).

There is also a current-main integration blocker: `_schedule_all_agents_reload()` is now async, but the new pip uninstall route calls it without `await` (`plugins.py:880` in the trial). Mypy reports:

```text
error: Value of type "Coroutine[Any, Any, None]" must be used [unused-coroutine]
note: Are you missing an await?
```

So even after a successful uninstall, the intended agent reload is not scheduled.

Finally, the locked static gates do not pass on the exact head:

```text
Black 23.3.0: 3 files would be reformatted
Flake8 6.1.0: loader.py E501
Pylint 3.0.2: W0212, W1510, E1121, W0611/W0612/W0613 (9.87/10)
Mypy 1.7.0: exact passes; trial fails on the missing await above
```

`git diff --check` passes. Please rerun the repository's pinned hooks after addressing the ownership and lifecycle issues.

PR #5495, opened 2026-06-24
fix(runtime): align envelope event translation with v1 streaming protocol

XiuShenAl:
## Description

Align the 2.0 SSE envelope event translation logic (`Envelope.translate_event()`) with the 1.0 `adapt_agentscope_message_stream` behavior to fix broken tool call rendering in the frontend.

In 2.0, tool calls and results were rendered as stacked blocks (all calls grouped together, then all results grouped together) instead of appearing in the correct interleaved sequence. Root cause: the envelope was emitting tool call messages only at `TOOL_CALL_END` as a single completed message, skipping the `in_progress → delta → completed` lifecycle that `Builder.js` relies on to pair and render tool calls/results via `mergeToolMessages`.

This PR rewrites the tool call/result event translation to emit the full message + content lifecycle matching 1.0, and fixes several other gaps identified through systematic comparison.

### Key changes

**`schemas.py`**:
- Add `status`, `object`, `msg_id` fields and `in_progress()`/`completed()` lifecycle methods to `_ContentBase`, enabling `DataContent` to participate in the same content lifecycle as 1.0's `Content` (which inherited from `Event`)
- Add explicit `id`, `created_at`, `completed_at` fields to `AgentResponse`

**`envelope.py`**:
- **Tool call lifecycle**: Rewrite `TOOL_CALL_START/DELTA/END` to emit `message.in_progress()` → `content.in_progress()` → delta updates with full accumulated state → `content.completed()` → `message.completed()`. Remove `json.loads` on arguments to keep them as JSON strings (required by `Builder.js` string concatenation logic)
- **Tool result lifecycle**: Rewrite `TOOL_RESULT_START/TEXT_DELTA/DATA_DELTA/END` with `role=Role.TOOL` (was `ASSISTANT`), full content lifecycle, and `dict[block_id]` tracking for data blocks (was flat list append)
- **Text message finalization**: Finalize the current text message before the first `TOOL_CALL_START`, with guard conditions to prevent double-finalize or empty message emission
- **DATA_BLOCK events**: Handle `DATA_BLOCK_START/DELTA/END` for model-generated binary content (images, audio, video) that was previously silently dropped
- **Message ID prefix**: Add `msg_` prefix to all message IDs
- **Sequence numbers**: Add monotonic `sequence_number` to every yielded object
- **Response metadata**: Add `response_` prefixed ID and `created_at`/`completed_at` timestamps
- **MODEL_CALL_END**: Extract `input_tokens`/`output_tokens` into `usage` on response and message
- **EXCEED_MAX_ITERS**: Emit a warning message when the agent hits iteration limits
- **HINT_BLOCK**: Log a warning instead of silently dropping

**Security Considerations:** N/A — no auth, credential, or config changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

1. Start the QwenPaw server and open the console in a browser
2. Send a query that triggers tool calls (e.g., a web search query)
3. Open browser DevTools → Network → EventStream tab to inspect SSE events
4. **Verify interleaving**: tool call (`plugin_call`) and result (`plugin_call_output`) messages should alternate — not stack by type
5. **Verify lifecycle**: each tool message should follow `in_progress` → content deltas → `completed` sequence
6. **Verify role**: `plugin_call_output` messages should have `role: "tool"` (not `"assistant"`)
7. **Verify arguments format**: `plugin_call` content `data.arguments` should be a JSON string, not a parsed object
8. **Verify IDs**: all message IDs should start with `msg_`, response ID with `response_`
9. **Verify sequence_number**: every SSE event should carry an incrementing `sequence_number`

Verified via browser EventStream inspection — see screenshots in the linked issue discussion.

## Additional Notes

- The frontend (`Builder.js`, `v1Adapter.tsx`, `channels/base.py`) is **not modified** — all changes are in the backend envelope layer to produce the same event shape that 1.0 produced
- `extra="allow"` on pydantic models means the newly explicit fields on `AgentResponse` and `_ContentBase` are backward-compatible — previously they were set dynamically and still serialized correctly, but explicit fields are cleaner
- `DataBlock` events (model-generated images/audio) only emit content on `DATA_BLOCK_END` — partial base64 cannot be rendered, so no intermediate yield during `DATA_BLOCK_DELTA`
- `HintBlockEvent` is logged but not rendered; full rendering is deferred to a follow-up when multi-agent team scenarios are prioritized

### DIFF
--- a/src/qwenpaw/runtime/envelope.py
+++ b/src/qwenpaw/runtime/envelope.py
@@ -11,12 +11,17 @@ from __future__ import annotations
 
 import json
 import logging
+import time
 import uuid
 from typing import Any, AsyncGenerator, Dict
 
 from .message_convert import _media_type_to_block_type
 
 logger = logging.getLogger(__name__)
+
+
+def _gen_msg_id() -> str:
+    return "msg_" + uuid.uuid4().hex
 
 
 class Envelope:
@@ -39,9 +44,11 @@ class Envelope:
 
         self._response = AgentResponse(output=[], status=RunStatus.Created)
         self._response.object = "response"
+        self._response.id = "response_" + uuid.uuid4().hex
+        self._response.created_at = int(time.time())
         self._response.session_id = session_id
 
-        self._message_id = uuid.uuid4().hex
+        self._message_id = _gen_msg_id()
         self._completed_message = Message(
             id=self._message_id,
             type=MessageType.MESSAGE,
@@ -56,9 +63,24 @@ class Envelope:
         self._text_blocks: Dict[str, Dict[str, Any]] = {}
         self._reasoning_blocks: Dict[str, Dict[str, Any]] = {}
         self._tool_calls: Dict[str, Dict[str, Any]] = {}
+        self._data_blocks: Dict[str, Dict[str, Any]] = {}
+
+        self._seq_counter = 0
 
         self._error_text: str | None = None
         self._finalized = False
+
+    # ------------------------------------------------------------------
+    # Sequence number helper
+    # ------------------------------------------------------------------
+
+    def _next_seq(self) -> int:
+        self._seq_counter += 1
+        return self._seq_counter
+
+    def _tag_seq(self, obj: Any) -> Any:
+        obj.sequence_number = self._next_seq()
+        return obj
 
     # ------------------------------------------------------------------
     # Response lifecycle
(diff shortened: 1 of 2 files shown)


src/qwenpaw/runtime/envelope.py
```diff
@@ -39,9 +44,11 @@ def __init__(self, session_id: str = "") -> None:
 
         self._response = AgentResponse(output=[], status=RunStatus.Created)
         self._response.object = "response"
+        self._response.id = "response_" + uuid.uuid4().hex
+        self._response.created_at = int(time.time())
```
zhijianma: @XiuShenAl   utc timezone

zhijianma (approved):

PR #5499, opened 2026-06-24
fix: move Current date from env context to per-user-message dynamic prefix (#5455)

### LINKED ISSUE #5455: [Question]: Why not make the current time a per-user-message prefix instead of putting it in environment/system context?
QwenPaw 2.0.0b1 / git `a9ee2e83`.

Original behavior on main:

- `build_env_context()` adds only date-level context: `- Current date: YYYY-MM-DD TZ (Weekday)`.
- `_request_input_to_msgs()` passes user text through without a fresh per-turn timestamp.

I am testing a local approach on branch `local`:

- `ef05e1ad` moves current time out of `env_context`.
- `f7978a05` prepends the user timezone timestamp to user text messages, including Chinese weekday.

Example model-facing user text:

```text
Current date: 2026-06-24 星期三 Asia/Shanghai 08:09:10
hello
```

The goal is to make time turn-specific, explicit, timezone-aware, and less likely to become stale in long sessions.

Concrete case I hit in the QQ channel:

1. In the morning, I asked what time it was.
2. The agent called the time tool and got the correct morning time.
3. In the afternoon, the same session still reused that old time from history when reasoning about the current time.

This makes the agent answer with stale time context even though the real current time has changed.

## Tried so far

I tested this locally by changing request message conversion so only user messages receive the timestamp prefix. `env_context` no longer includes `Current date:`.

Original main code effectively had:

```python
blocks.append(TextBlock(type="text", text=text))
```

Local code changes that to prepend a fresh timestamp once per user message:

```python
text = _prepend_user_timestamp(text, timestamp_line)
blocks.append(TextBlock(type="text", text=text))
```

Benefits I see:

- Each user turn carries the actual time of that turn.
- Long sessions do not rely on an old context timestamp.
- Request dumps are easier to audit.
- The timestamp includes timezone, date, time, and weekday.

Main question: are there reasons QwenPaw should keep time in env/system context instead, such as token cost, provider behavior, localization, multimodal message structure, or conflict with the `get_current_time` tool?

lecheng2018:
## Description

Move `Current date` from the static `build_env_context()` (system/environment context) to a per-user-message dynamic timestamp prefix. This fixes stale time information in long sessions and improves prompt caching stability by avoiding per-request context changes.

Changes:
1. **`chats/utils.py`**: Remove `Current date` from `build_env_context()` — no more static timestamp in the system/environment context.
2. **`runtime/runtime.py`**: Add `_inject_current_time()` static method and call it in `_normalize()` — injects `Current date: YYYY-MM-DD Weekday TZ HH:MM:SS` as a prefix to the last user message's first text block.

## Related Issue

Closes #5455

## Checklist

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [ ] Implementation follows the established code style
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added unit tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/qwenpaw/app/chats/utils.py
+++ b/src/qwenpaw/app/chats/utils.py
@@ -91,14 +91,6 @@ def build_env_context(
         Formatted environment context string
     """
     parts = []
-    user_tz = load_config().user_timezone or "UTC"
-    try:
-        now = datetime.now(ZoneInfo(user_tz))
-    except (ZoneInfoNotFoundError, KeyError):
-        logger.warning("Invalid timezone %r, falling back to UTC", user_tz)
-        now = datetime.now(timezone.utc)
-        user_tz = "UTC"
-
     if session_id is not None:
         parts.append(f"- Session ID: {session_id}")
     if user_id is not None:
@@ -128,11 +120,6 @@ def build_env_context(
             )
     elif working_dir is not None:
         parts.append(f"- Working directory: {working_dir}")
-    parts.append(
-        f"- Current date: {now.strftime('%Y-%m-%d')} "
-        f"{user_tz} ({now.strftime('%A')})",
-    )
-
     if add_hint:
         parts.append(
             "- Important:\n"
--- a/src/qwenpaw/runtime/runtime.py
+++ b/src/qwenpaw/runtime/runtime.py
@@ -17,7 +17,9 @@ from __future__ import annotations
 import asyncio
 import logging
 import uuid
+from datetime import datetime, timezone
 from typing import Any, AsyncGenerator
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 from .builder import AgentBuilder
 from .envelope import Envelope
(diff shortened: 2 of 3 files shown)


lecheng2018:
Thanks for the thorough review! I've addressed all your feedback:

**1. Timing** — Moved `_inject_current_time()` from `_normalize()` to right before `executor.run(ctx.input_msgs)`. Slash commands, lifecycle hooks, and short-circuits now see raw text.

**2. Format** — Changed to `Current time: 2026-06-25 11:24:17 Asia/Shanghai (Wednesday)` using `strftime('%A')` instead of hard-coded Chinese weekday names.

**3. Media-only** — When the last user message has no text block, a `{'type': 'text', 'text': prefix}` block is inserted.

**4. Cleanup** — Removed the `_WEEKDAY_CN` constant.

rayrayraykk:
Closed as been cherry picked in #5923

PR #5510, opened 2026-06-25
fix(tool-calls): cap tool responses before context insertion

### LINKED ISSUE #5342: [Feature]: hard cap on tool result size at execution layer (defense-in-depth for context explosion)
## Problem

The `post_acting` hook in `light_context_manager.py` is the only mechanism that prunes oversized tool results. When LLM calls fail (e.g. 502 errors), the hook is skipped entirely, and tool results accumulate unpruned in context. This creates a cascading failure pattern:

1. LLM returns 502 → `post_acting` hook does not run
2. Tool results from previous calls remain unpruned in context
3. Retry sends the bloated context → more latency → more 502s
4. Context compaction triggers on a massively inflated context

I observed this in production: context ballooned from ~44K tokens (post-compaction) to **307K tokens** in ~90 seconds during a 502 outage, with only 26 messages added.

## Root Cause

The tool result pruning pipeline has two gaps:

1. **`post_acting` hook is bypassed on errors** — The hook only executes after successful agent steps. When the LLM fails, the hook is skipped.
2. **No hard limit at the tool execution layer** — Tool results (from `grep_search`, `read_file`, `memory_search`, etc.) are injected into context without any size cap. The only defense is the post-hoc pruning hook.

## Current Behavior

```
Tool executes → result (potentially 50KB+) added to context → post_acting hook prunes (if it runs)
```

When 502 occurs:
```
Tool executes → result added to context → LLM fails → post_acting skipped → result stays unpruned → retry with bloated context
```

## Proposed Solution

Add a hard cap on tool result size at the tool execution layer, before results enter the context:

```python
# In the tool execution pipeline (e.g. react_agent.py or tool dispatcher)
MAX_TOOL_RESULT_CHARS = 50_000  # configurable

def truncate_tool_result(result):
    if len(result) > MAX_TOOL_RESULT_CHARS:
        return result[:MAX_TOOL_RESULT_CHARS] + TRUNCATION_NOTICE
    return result
```

This would be a defense-in-depth measure — the `post_acting` hook would still run for fine-grained pruning, but the hard cap prevents worst-case scenarios.

## Additional Context

- The `exempt_file_extensions: [".md"]` default also contributed — 879KB of .md files in the workspace were completely exempt from pruning. I removed this exemption as a local workaround.
- The `pruning_recent_msg_max_bytes` (default 50KB, mine was 10KB) only limits per-message size, but doesn't prevent accumulation across multiple messages.

## Environment
- QwenPaw version: 1.1.12
- Config: `tool_result_pruning_config.enabled = true`, `pruning_recent_msg_max_bytes = 10000`

wananing:
## Description

This PR adds an execution-layer hard cap for final tool-response text before it
is inserted into the agent context.

The current historical pruning middleware runs after an acting step, so it can
be skipped when the next LLM call fails. This change adds a defense-in-depth
boundary inside the tool-call coordinator:

- cap final `ToolResponse` text after tool execution and after hooks complete;
- use one aggregate UTF-8 byte budget across all top-level text blocks,
  including the truncation notice;
- preserve response ID, state, metadata, and non-text blocks;
- save the full textual output to the existing tool-results cache on a
  best-effort basis;
- fail closed if limiting itself encounters an unexpected error;
- apply the same finalized response path to foreground and background tools.

Scope note: this limiter is intentionally a text-content cap, not a cap on the
full serialized `ToolResponse` payload. Non-text blocks such as media/data or
provider-specific structured blocks are preserved unchanged so their existing
formatting paths keep working.

`ToolResultPruningMiddleware` remains unchanged as the second-stage historical
pruning mechanism. The execution-layer cap reuses
`tool_result_pruning_config.enabled` and
`tool_result_pruning_config.execution_layer_max_bytes`.

Fixes #5342.

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core
- [ ] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation
- [x] Tests
- [ ] CI/CD
- [ ] Scripts

## Checklist

- [x] Run and pass relevant tests with `pytest`
- [x] Update documentation if needed: not needed; behavior reuses existing
      tool-result pruning configuration

## Testing

Relevant tests cover:

- small results remain unchanged;
- oversized ASCII and multibyte text are capped at valid UTF-8 boundaries;
- multiple text blocks share one aggregate budget;
- truncation notice counts toward the configured byte cap;
- non-text blocks, response ID, state, and metadata are preserved;
- cache-write failure still caps the text result;
- foreground finalization waits for after hooks before applying the cap;
- caller cancellation still propagates while waiting for tool completion;
- middleware callers only observe capped final text responses;
- background completion hints use the finalized capped text response.

```text
uv pip install -e \".[dev,full]\"
Installed 24 packages
```

```text
uv run --python 3.11 --extra dev pre-commit run --all-files
```

```text
uv run --python 3.11 --extra dev pytest tests/unit/tool_calls -q
............                                                             [100%]
12 passed in 0.24s
```

```text
uv run --python 3.11 --extra dev pytest tests/unit/agents/tools/test_file_io.py -q
....................................                                     [100%]
36 passed in 0.57s
```

Full pytest was also attempted locally:

```text
uv run --python 3.11 --extra dev pytest
8 failed, 582 passed, 11 skipped, 2 warnings in 777.83s (0:12:57)
```

This full run did not complete/pass in my local environment. It was stopped
after unrelated integration failures had already occurred at about 15% progress;
the first failures were existing `/api/agent/*` 404 integration assertions and
upload/provider-switching integration checks, not the tool-call limiter path.

### DIFF
--- a/src/qwenpaw/runtime/builder.py
+++ b/src/qwenpaw/runtime/builder.py
@@ -510,10 +510,44 @@ class AgentBuilder:
                 None,
             )
             if tool_coordinator is not None:
-                from ..tool_calls import ToolCoordinatorMiddleware
+                from ..tool_calls import (
+                    ToolCoordinatorMiddleware,
+                    ToolResultLimiter,
+                )
+
+                result_limiter = None
+                try:
+                    import os
+
+                    lcc = agent_config.running.light_context_config
+                    trc = lcc.tool_result_pruning_config
+                    workspace = getattr(ctx, "workspace", None)
+                    workspace_dir = (
+                        str(getattr(workspace, "workspace_dir", ""))
+                        if workspace is not None
+                        else ""
+                    )
+                    tool_results_dir = (
+                        os.path.join(workspace_dir, trc.tool_results_cache)
+                        if workspace_dir
+                        else None
+                    )
+                    result_limiter = ToolResultLimiter(
+                        enabled=trc.enabled,
+                        max_text_bytes=trc.pruning_recent_msg_max_bytes,
+                        cache_dir=tool_results_dir,
+                    )
+                except Exception:
+                    _logger.debug(
+                        "ToolResultLimiter not created",
+                        exc_info=True,
+                    )
 
                 mws.append(
-                    ToolCoordinatorMiddleware(coordinator=tool_coordinator),
+                    ToolCoordinatorMiddleware(
+                        coordinator=tool_coordinator,
+                        result_limiter=result_limiter,
+                    ),
                 )
 
         memory_manager = AgentBuilder._get_memory_manager(ctx)
--- a/src/qwenpaw/runtime/builder.py
+++ b/src/qwenpaw/runtime/builder.py
@@ -534,7 +534,7 @@ class AgentBuilder:
                     )
                     result_limiter = ToolResultLimiter(
                         enabled=trc.enabled,
-                        max_text_bytes=trc.pruning_recent_msg_max_bytes,
+                        max_text_bytes=trc.execution_layer_max_bytes,
                         cache_dir=tool_results_dir,
                     )
                 except Exception:
--- a/src/qwenpaw/tool_calls/_hint.py
+++ b/src/qwenpaw/tool_calls/_hint.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """Construct hint messages for completed background tool calls."""
+
 from __future__ import annotations
 
 from typing import Any
@@ -9,7 +10,7 @@ def make_offload_hint_msg(entry: Any) -> Any:
     """Construct a hint Msg for a completed offloaded tool call.
 
     The hint contains a system-notification TextBlock and a
-    ToolResultBlock with the full (untruncated) final_response content.
+    ToolResultBlock with the finalized response content.
     """
     from agentscope.message import Msg, TextBlock, ToolResultBlock
 
@@ -29,9 +30,10 @@ def make_offload_hint_msg(entry: Any) -> Any:
         id=entry.ctx.tool_call_id,
         name=entry.ctx.tool_name,
         output=list(entry.final_response.content or []),
+        state=entry.final_response.state,
     )
     return Msg(
         name="system",
-        role="system",
+        role="assistant",
         content=[notification, tool_result],
     )
--- a/src/qwenpaw/tool_calls/_hint.py
+++ b/src/qwenpaw/tool_calls/_hint.py
@@ -32,6 +32,9 @@ def make_offload_hint_msg(entry: Any) -> Any:
         output=list(entry.final_response.content or []),
         state=entry.final_response.state,
     )
+    # AgentScope 2.0 validates that system messages contain only text
+    # blocks.  This hint must carry a ToolResultBlock so provider formatters
+    # keep it on the tool-result path; use assistant role intentionally.
     return Msg(
         name="system",
         role="assistant",
--- a/src/qwenpaw/tool_calls/_result_limiter.py
+++ b/src/qwenpaw/tool_calls/_result_limiter.py
@@ -1,0 +1,331 @@
+# -*- coding: utf-8 -*-
+"""Execution-layer hard cap for final tool responses."""
+
+from __future__ import annotations
+
+import logging
+import re
+import uuid
+from pathlib import Path
+from typing import Any
+
+from agentscope.message import TextBlock
+from agentscope.tool import ToolResponse
+
+from ..constant import TRUNCATION_NOTICE_MARKER
+from ._context import ToolCallContext
+
+logger = logging.getLogger(__name__)
+
+
+class ToolResultLimiter:
+    """Apply a strict aggregate text-byte cap to a final ToolResponse."""
+
+    def __init__(
+        self,
+        *,
+        enabled: bool,
+        max_text_bytes: int,
+        cache_dir: Path | str | None,
+    ) -> None:
+        self._enabled = enabled
+        self._max_text_bytes = max_text_bytes
+        self._cache_dir = Path(cache_dir) if cache_dir else None
+
+    def limit(
+        self,
+        response: ToolResponse,
+        context: ToolCallContext,
+    ) -> ToolResponse:
+        """Return a response whose aggregate text content is byte-bounded."""
+        if not self._enabled:
+            return response
+
+        try:
+            return self._limit(response, context)
+        except Exception:
+            logger.exception(
+                "Failed to limit tool result",
+                extra={
+                    "tool_name": context.tool_name,
+                    "tool_call_id": context.tool_call_id,
+                },
+            )
+            return self._fail_closed(response)
+
+    def _limit(
+        self,
+        response: ToolResponse,
+        context: ToolCallContext,
+    ) -> ToolResponse:
+        content = list(response.content or [])
+        original_bytes = _total_text_bytes(content)
+        if original_bytes == 0:
+            return response
+        if self._max_text_bytes <= 0:
+            return self._fail_closed(response)
+        if original_bytes <= self._max_text_bytes:
+            return response
+
+        saved_path = self._save_original_text(content, context)
+        limited_content, retained_bytes = self._limit_content(
+            content,
+            original_bytes=original_bytes,
+            saved_path=saved_path,
+        )
+        result = _copy_response_with_content(response, limited_content)
+
+        if _total_text_bytes(result.content or []) > self._max_text_bytes:
+            return self._fail_closed(response)
+
+        logger.info(
+            "Capped tool result",
+            extra={
+                "tool_name": context.tool_name,
+                "tool_call_id": context.tool_call_id,
+                "original_bytes": original_bytes,
+                "retained_bytes": retained_bytes,
+            },
+        )
+        return result
+
+    def _limit_content(
+        self,
+        content: list[Any],
+        *,
+        original_bytes: int,
+        saved_path: str | None,
+    ) -> tuple[list[Any], int]:
+        notice = _build_notice(
+            original_bytes=original_bytes,
+            retained_bytes=0,
+            saved_path=saved_path,
+        )
+        retained_content: list[Any] = []
+        retained_bytes = 0
+
+        for _ in range(8):
+            bounded_notice = _utf8_prefix(notice, self._max_text_bytes)
+            payload_budget = max(
+                0,
+                self._max_text_bytes - _byte_len(bounded_notice),
+            )
+            retained_content, retained_bytes = _retain_text_content(
+                content,
+                payload_budget,
+            )
+            next_notice = _build_notice(
+                original_bytes=original_bytes,
+                retained_bytes=retained_bytes,
+                saved_path=saved_path,
+            )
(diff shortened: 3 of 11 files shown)


wananing:
Thanks for the guidance. I updated the PR description with the requested template sections and local verification evidence.

jinliyl:
Thanks for the feedback. We are on it
@wananing

src/qwenpaw/tool_calls/_result_limiter.py
```diff
@@ -0,0 +1,331 @@
...
+        context: ToolCallContext,
+    ) -> str | None:
+        if self._cache_dir is None:
+            return None
+
+        try:
+            self._cache_dir.mkdir(parents=True, exist_ok=True)
+            safe_id = _safe_filename_part(context.tool_call_id)
+            path = self._cache_dir / f"{safe_id}-{uuid.uuid4().hex}.txt"
+            path.write_text(_join_text_blocks(content), encoding="utf-8")
```
qbc2016, line 253: **[Blocking I/O in event loop]** `_save_original_text` calls `path.write_text(...)` synchronously. 
`ResultFinalizer` is a plain sync callable invoked inside `_finalize_completed()`, which runs on the asyncio event loop. 
For large tool results (exactly the case this PR targets), this will block the loop.

`ToolResultPruningMiddleware._truncate_tool_result` has the same pattern, 
but it runs in the tail of `on_acting` where the timing impact is softer. 
Here the finalizer is called inside the coordinator's hot path.

Suggested fix — offload the write or do it fire-and-forget:
```python
# option A: accept that cache write may be skipped under load
try:
    await asyncio.get_event_loop().run_in_executor(
        None, path.write_text, text, "utf-8"
    )
except OSError:
    ...
```
If keeping ResultFinalizer as a sync interface is a constraint,
at minimum document the blocking behaviour in the docstring so callers are aware.

src/qwenpaw/tool_calls/_hint.py
```diff
@@ -29,9 +30,10 @@ def make_offload_hint_msg(entry: Any) -> Any:
         id=entry.ctx.tool_call_id,
         name=entry.ctx.tool_name,
         output=list(entry.final_response.content or []),
+        state=entry.final_response.state,
     )
     return Msg(
         name="system",
-        role="system",
+        role="assistant",
```
qbc2016, line 40: **[Unrelated behaviour change]** Changing `role="system"` → `role="assistant"` 
in `make_offload_hint_msg` is unrelated to the byte-cap feature but 
could affect how downstream LLM formatters serialise the hint message.

`ToolResultBlock` content inside a `role="assistant"` message 
may be formatted differently depending on the provider formatter 
(Anthropic, OpenAI, Qwen, etc.). The original `role="system"` 
was presumably intentional to signal a system-level notification 
rather than an assistant turn.

Please either:
1. Revert this change and open a separate PR with the motivation explained, or
2. Add a comment in this PR explaining why the role change is necessary 
   and confirming it has been validated across all active formatters.

src/qwenpaw/tool_calls/_result_limiter.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def _build_notice(
+    *,
+    original_bytes: int,
+    retained_bytes: int,
+    saved_path: str | None,
+) -> str:
+    lines = [
+        TRUNCATION_NOTICE_MARKER,
```
qbc2016: **[Notice format incompatible with `truncate_text_output`]** `_build_notice` puts
`TRUNCATION_NOTICE_MARKER` as the *first* line (L218). The resulting text looks like:

    <<<TRUNCATED>>>
    Tool output truncated before entering agent context.
    Original size: N bytes.

The existing `ToolResultPruningMiddleware._truncate_tool_result`
(middlewares.py L576-L581) detects the marker and re-invokes `truncate_text_output`,
which splits on the marker and treats everything to the *left* as the original content:

    original = content.split(TRUNCATION_NOTICE_MARKER)[0]

When the execution-layer notice starts with the marker, `split(...)[0]` returns `""`,
so downstream re-truncation receives an empty body and produces malformed output.

This path fires when a capped message ages past `pruning_recent_n` and the threshold
drops to `pruning_old_msg_max_bytes` (default 3000 bytes). It won't crash immediately,
but the notice text will silently degrade.

Consider putting at least one line of retained content *before* the marker line,
or using a distinct sentinel for execution-layer notices to keep the two formats
cleanly separated.

src/qwenpaw/tool_calls/_result_limiter.py
```diff
@@ -0,0 +1,331 @@
...
+    return copied
+
+
+def _copy_response_with_content(
+    response: ToolResponse,
+    content: list[Any],
+) -> ToolResponse:
+    if hasattr(response, "model_copy"):
+        return response.model_copy(update={"content": content})
+    response.content = content
```
qbc2016: **[Fallback path mutates the original object]** When `ToolResponse` does not expose
`model_copy`, the else-branch at L312 writes directly to `response.content`:

    response.content = content   # in-place mutation
    return response

`entry.final_response` is referenced from `_supervise`, completion handlers, and
`make_offload_hint_msg`. A silent mutation here could cause hard-to-trace bugs if
any of those paths still hold a reference to the original content list.

Safer fallback:

    import copy
    r = copy.copy(response)
    r.content = content
    return r

src/qwenpaw/runtime/builder.py
```diff
@@ -510,10 +510,44 @@ def _build_middlewares(
...
+                        else ""
+                    )
+                    tool_results_dir = (
+                        os.path.join(workspace_dir, trc.tool_results_cache)
+                        if workspace_dir
+                        else None
+                    )
+                    result_limiter = ToolResultLimiter(
+                        enabled=trc.enabled,
+                        max_text_bytes=trc.pruning_recent_msg_max_bytes,
```
qbc2016: **[Config field reuse changes user-visible semantics]**
`pruning_recent_msg_max_bytes` is documented as:
    "Byte threshold for *recent messages* in tool result pruning"

Reusing it here as `max_text_bytes` gives it a second incompatible meaning:
a per-response ceiling applied unconditionally at execution time, with no
concept of message recency.

A user who raises this value to allow larger recent context windows will
unknowingly also raise the execution-layer hard cap, potentially defeating
the defense-in-depth purpose of this PR.

It would be cleaner to add a dedicated field to `ToolResultPruningConfig`, e.g.:

    execution_layer_max_bytes: int = Field(
        default=50_000,
        ge=1_000,
        description=(
            "Hard byte cap applied at execution time before the tool response "
            "is inserted into the agent context. Independent of the tiered "
            "historical pruning thresholds."
        ),
    )

qbc2016 (commented):

src/qwenpaw/tool_calls/_result_limiter.py
wananing: Addressed in 94036e35. ResultFinalizer now supports awaitable finalizers, ToolResultLimiter has limit_async(), and the middleware uses that async path. Cache persistence is offloaded with asyncio.to_thread so the coordinator hot path no longer performs the large write on the event loop; sync limit() remains for non-event-loop callers/tests.

src/qwenpaw/tool_calls/_hint.py
wananing: Addressed with option 2 in 94036e35. I kept role="assistant" because AgentScope 2.0 validates that system messages contain only text blocks, while this hint must carry a ToolResultBlock. I added an inline comment explaining the formatter/validation reason and a regression test for the background hint role plus finalized tool_result output.

src/qwenpaw/tool_calls/_result_limiter.py
wananing: Addressed in 94036e35. Execution-layer notices now use a distinct EXECUTION_TRUNCATION_NOTICE_MARKER instead of TRUNCATION_NOTICE_MARKER, so ToolResultPruningMiddleware will not re-interpret them as the legacy truncate_text_output format. Added assertions that capped execution-layer output does not contain the legacy marker.

src/qwenpaw/tool_calls/_result_limiter.py
wananing: Addressed in 94036e35. The no-model_copy fallback now shallow-copies the response before replacing content, instead of mutating the original object. Added a regression test with a legacy response object to verify the original content list remains unchanged.

src/qwenpaw/runtime/builder.py
wananing: Addressed in 94036e35. Added ToolResultPruningConfig.execution_layer_max_bytes and wired AgentBuilder to use it for ToolResultLimiter, leaving pruning_recent_msg_max_bytes for the tiered historical recent-message pruning path. Added a builder wiring test to keep the two limits independent.

wananing:
Follow-up pushed in 94036e35 and I replied to each inline thread.

Local verification:
- `.venv/bin/pytest tests/unit/tool_calls -q` -> 16 passed
- `.venv/bin/pytest tests/unit/agents/tools/test_file_io.py -q` -> 36 passed
- `env PRE_COMMIT_HOME=/tmp/pre-commit-cache .venv/bin/pre-commit run --all-files` -> passed

qbc2016 (commented):
LGTM. cc @jinliyl

jinliyl:
Thanks for the updates. I found one more cancellation edge case in `_await_background_task` that is worth fixing before merge.

The current implementation awaits the background task directly:

```python
await bg
```

If the caller task is cancelled while it is waiting here, asyncio will also cancel `bg`. That is not what we want for the background tool path: caller cancellation and background-task cancellation should remain distinguishable.

There is also a second issue with the detection logic. Inside `except asyncio.CancelledError`, `current_task().cancelled()` is usually still `False` because the current task has not finished cancellation yet. As a result, caller cancellation can be swallowed instead of being propagated.

Suggested change:

```python
try:
    await asyncio.shield(bg)
    # Distinguish: bg task cancelled vs caller task cancelled.
    current = asyncio.current_task()
    if current is not None and current.cancelling():
        raise
```

This keeps external cancellation from implicitly cancelling the background tool task, while still propagating cancellation when the caller itself is being cancelled. If `bg` was explicitly cancelled, the `CancelledError` can still be handled as the background task's terminal state.

wananing:
Thanks for catching this cancellation edge case.

Follow-up pushed in eacbd3e5. `_await_background_task` now awaits the background task through `asyncio.shield(bg)` and uses `current_task().cancelling()` to distinguish caller cancellation from explicit background-task cancellation.

I also added a regression test that cancels the waiter while the background task is still pending, then verifies the background task is not cancelled and can still finish normally.

Local verification:
- `.venv/bin/pytest tests/unit/tool_calls/test_coordinator_result_finalizer.py -q` -> 5 passed
- `.venv/bin/pytest tests/unit/tool_calls -q` -> 17 passed
- `env PRE_COMMIT_HOME=/tmp/pre-commit-cache .venv/bin/pre-commit run --all-files` -> passed

qbc2016 (approved):

PR #5524, opened 2026-06-25
fix(tools): register spawn_subagent in runtime

### LINKED ISSUE #5523: spawn_subagent is missing from the Runtime 2.0 tool registry
## Description

spawn_subagent was present in configuration and documentation but unusable on Runtime 2.0. Real end-to-end testing identified four migration regressions:

1. Missing runtime tool_descriptor, so automatic discovery skipped the tool.
2. Missing governance ToolRegistry entry, so SpawnSubagent was denied as unknown.
3. Foreground result collection returned the trailing turn_usage SSE event instead of the completed response, producing No text content in response.
4. Background mode called the removed /api/console/chat/task endpoint and received HTTP 405.

## Expected behavior

- Foreground subagents return their final response to the parent agent.
- Background subagents return a task id immediately.
- The task status endpoint reaches finished/completed and includes the child output.
- Governance permits subagent orchestration while continuing to govern each child file, shell, and network operation.

## Verification

A real QwenPaw FastAPI subprocess is started with an OpenAI-compatible mock model. The parent agent calls spawn_subagent through the normal console HTTP/SSE route. The test verifies an independent child model request, foreground result delivery, background task submission, status polling, and final child output.

hellozhouuu:
## Summary

- register spawn_subagent with Runtime 2.0 tool discovery
- register SpawnSubagent as an internal governance tool
- select the completed response envelope instead of trailing turn_usage SSE
- restore QwenPaw background chat submit/status endpoints
- add a real-process end-to-end test for foreground and background subagents

## Root causes found by end-to-end testing

The initial registration fix was insufficient. Running a real QwenPaw server exposed two additional Runtime 2.0 migration failures: foreground results were lost because turn_usage was treated as the final response, and background submission still targeted a removed endpoint that returned HTTP 405.

Closes #5523

## Tests

- Real QwenPaw subprocess E2E: 1 passed in 17.66s
  - parent invokes spawn_subagent through console HTTP/SSE
  - governance decision is allow
  - foreground child executes independently and returns output
  - background call returns task_id
  - status polling reaches finished/completed with child output
- Combined unit/integration verification: 44 passed
- Governance suite: 39 passed
- git diff --check

### DIFF
--- a/src/qwenpaw/agents/tools/agent_management.py
+++ b/src/qwenpaw/agents/tools/agent_management.py
@@ -291,8 +291,14 @@ def collect_final_agent_chat_response(
     to_agent: str,
     timeout: int,
 ) -> Optional[Dict[str, Any]]:
-    """Collect the last SSE payload from inter-agent chat."""
+    """Collect the final response envelope from inter-agent chat.
+
+    Console streams may append non-response events such as ``turn_usage``
+    after the completed response.  Keep the latest response-shaped payload
+    instead of blindly returning the final SSE event.
+    """
     response_data: Optional[Dict[str, Any]] = None
+    fallback_data: Optional[Dict[str, Any]] = None
     with create_agent_api_client(base_url) as client:
         with client.stream(
             "POST",
@@ -306,8 +312,13 @@ def collect_final_agent_chat_response(
                 if line:
                     parsed = parse_agent_sse_line(line)
                     if parsed:
-                        response_data = parsed
-    return response_data
+                        fallback_data = parsed
+                        if (
+                            parsed.get("object") == "response"
+                            or "output" in parsed
+                        ):
+                            response_data = parsed
+    return response_data or fallback_data
 
 
 def submit_agent_chat_task(
--- a/src/qwenpaw/agents/tools/agent_management.py
+++ b/src/qwenpaw/agents/tools/agent_management.py
@@ -653,6 +653,7 @@ def _generate_subagent_session_id() -> str:
     return f"sub-{str(uuid4())[:8]}"
 
 
+@tool_descriptor(async_execution=True)
 async def spawn_subagent(
     task: str,
     fork: bool = False,
--- a/src/qwenpaw/app/routers/console.py
+++ b/src/qwenpaw/app/routers/console.py
@@ -7,6 +7,7 @@ import json
 import logging
 import re
 import uuid
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import AsyncGenerator, Union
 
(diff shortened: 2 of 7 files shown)


maximilize (commented):
Reproduced the suite green in an isolated container (`--network none`): 56 passed, including the real-process E2E `test_spawn_subagent_foreground_and_background_end_to_end`. Traced all four fixes against Runtime 2.0 conventions on `030e6fc` and they hold up:

1. `spawn_subagent` was the only `async` inter-agent tool without `@tool_descriptor` — `list_agents` / `chat_with_agent` / `submit_to_agent` / `check_agent_task` all carry it, so discovery now treats it the same.
2. `register("SpawnSubagent", "internal", "task")` matches the sibling governance pattern, with `target_param="task"` being the tool's primary parameter (cf. `agent_id` for `ChatWithAgent`).
3. Keeping the last `object=="response" or "output" in parsed` payload is consistent with `extract_agent_text_content` reading `response_data.get("output")`; `turn_usage` stays fallback-only.
4. Client paths `/console/chat/task` (POST) + `/console/chat/task/{id}` (GET) match the restored routes, including the per-agent ownership check.

One thing worth a look before merge: `_CHAT_TASKS` (console.py:33) only grows. `_CHAT_RUNTIME_TASKS` is popped on finish (line 120), but the `_CHAT_TASKS` record — including the full `result` payload — is never removed, so every background `spawn_subagent` call leaves a permanent entry in a long-lived server. A TTL / size cap, or dropping the record once status is read post-`finished`, would bound it.

hellozhouuu:
@maximilize  Fixed in dbf478d9. Finished task records are dropped after status read, with a capped fallback for unread finished records. Added unit tests.

maximilize (commented):
Re-checked `dbf478d9` — the `_CHAT_TASKS` growth I flagged is now bounded and holds up:

- **Drop-on-read**: `get_console_chat_task_status` builds the response dict before popping the record once `finished`, so polled tasks leave nothing behind and `result` stays intact.
- **Capped fallback**: `_prune_finished_chat_tasks()` (worker `finally`) evicts oldest finished records to `_MAX_FINISHED_CHAT_TASKS=100`; the `not in _CHAT_RUNTIME_TASKS` guard is safe (runtime task popped just above).
- **No race**: `status`, `result`, and the prune run synchronously before the first `await`, so a concurrent GET can't see `finished` without `result`.

Verified offline (`--network none`): 3 `test_console_tasks.py` cases + `spawn_subagent` E2E pass.

Optional: drop-on-read makes the result single-consumption (a lost response + retry 404s); the cap covers never-read, so cap-only stays idempotent. Fine as-is. LGTM.

hellozhouuu:
@maximilize  Thanks for re-checking. This PR is now blocking my next follow-up work around subagent, so I’d appreciate a maintainer review/merge when possible.

rayrayraykk:
Had been fixed in #5665

PR #5531, opened 2026-06-25
test(integration): parallelize with pytest-xdist and harden coverage pipeline

yutai78786:
## Summary

Sprint 4.0: cut local integration runtime from **~24 min** down to
**~3.5 min** (~7x) via ``pytest-xdist`` parallel execution, while
keeping coverage data accurate across all matrix entries and
stabilising five test files that turned flaky under parallel
scheduling + agent reload races.

## Changes (3 commits)

1. **`build: split [test] deps from [dev] and add pytest-xdist`** —
   new ``[test]`` optional-dep group containing the runtime test
   stack plus ``pytest-xdist>=3.5.0``; ``[dev]`` keeps only
   developer tooling (pre-commit). Lets CI install only what it
   needs.

2. **`ci(integration): parallelize with pytest-xdist and harden coverage pipeline`** —
   - Run integration tier with ``pytest -n auto --dist=loadscope``
     (preserves class-level fixture affinity).
   - Make subprocess coverage tracing xdist-aware in
     ``tests/integration/conftest.py`` (workers stage shards;
     controller combines once at session end).
   - Run coverage collection single-process (``-n 0``) on the opt-in
     matrix entry — tracer overhead serializes ACP stdio I/O on
     2-core runners.
   - Align coverage gate with current matrix py3.11 (commit
     ``342c75e1`` swapped 3.10→3.11; gate was still pinned to 3.10
     so coverage never ran).
   - Multi-platform integration coverage is opt-in: nightly schedule
     collects on linux+macos+windows; ad-hoc dispatch defaults
     linux-only via a new ``coverage_platforms`` input. Per-OS/py
     artifact namespacing + path remap in the report job.
   - Document new runtime / coverage expectations in
     ``tests/integration/README.md``.

3. **`test(integration): stabilise ACP / MCP / Channel / Plugin under parallel + reload race`** —
   Widen cron-history poll deadlines, route inline polls through
   the shared ``wait_cron_executed`` helper (introduced in #5507),
   relax a few "must have history record" assertions to the real
   invariant (cron not deadlocked) where appropriate, and handle
   the background reload race in channel-config / MCP delete /
   plugin-provider tests.

## Test plan

- [x] Local serial: ``pytest tests/integration`` — 372 passed
- [x] Local parallel: ``pytest tests/integration -n auto --dist=loadscope`` —
  372 passed in 3:23
- [x] Local subprocess coverage: ``QWENPAW_INTEGRATION_COVERAGE=1 pytest tests/integration -n auto --dist=loadscope --no-cov`` —
  372 passed in 3:56, coverage combine succeeds, router-layer 64%, overall 35% (same as serial)
- [x] Fork CI dispatch (p0 marker): **full green** including Coverage Report
- [x] Fork CI on PR push (default p0+p1 marker): **full green** across
  unit (py3.11 + py3.13 × 3 OS), contract (same), integration
  (same), Coverage Report, prettier — see runs/28158230423 and
  28160668860

## Pre-existing dependencies

- Builds on top of #5507 which introduced ``wait_cron_executed``
  in ``tests/integration/helpers.py``. This PR routes more inline
  poll loops through that helper.
- The coverage-gate alignment fixes a regression introduced in
  ``342c75e1`` (v1→v2 merge fixup) where the matrix Python pin
  was bumped from 3.10 to 3.11 but the coverage gate was missed.

### DIFF
--- a/.github/workflows/full-tests-nightly.yml
+++ b/.github/workflows/full-tests-nightly.yml
@@ -24,6 +24,17 @@ on:
     # Off-the-hour to avoid the global GitHub Actions cron rush at :00 / :30.
     - cron: '17 18 * * *'
   workflow_dispatch:
+    inputs:
+      coverage_platforms:
+        description: 'Which platforms collect integration coverage (schedule always uses all)'
+        required: false
+        default: 'linux'
+        type: choice
+        options:
+          - 'linux'
+          - 'macos'
+          - 'windows'
+          - 'all'
 
 jobs:
   unit-tests:
@@ -74,7 +85,7 @@ jobs:
         shell: bash
         run: |
           python -m pip install --upgrade pip
-          pip install -e ".[dev,full]"
+          pip install -e ".[dev,test,full]"
 
       - name: Run unit tests
         shell: bash
@@ -155,7 +166,7 @@ jobs:
         shell: bash
         run: |
           python -m pip install --upgrade pip
-          pip install -e ".[dev,full]"
+          pip install -e ".[dev,test,full]"
 
       - name: Run contract tests
         shell: bash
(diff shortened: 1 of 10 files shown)


lalaliat (approved):
LGTM

PR #4846, opened 2026-06-01
refactor: migrate agentscope from version 1.x to 2.0.0

qbc2016:
## Description

[Describe what this PR does and why]

**Related Issue:** Fixes #(issue_number) or Relates to #(issue_number)

**Security Considerations:** [If applicable, e.g. channel auth, env/config handling]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [ ] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Feishu, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

[How to test these changes]

## Local Verification Evidence

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]

PR #5542, opened 2026-06-26
test(e2e): adapt for agentscope 2.0 — drop Plan Mode, fix selectors and fixtures

yutai78786:
## Context

After upstream merged the agentscope 1.0 → 2.0 migration (PRs #4846 / a9ee2e83 et al.), a full regression of the e2e suite on a fresh CI working_dir surfaced 6 P0 failures. This PR ships all the e2e-side adaptations in one go.

Local fresh-state P0+integration result after the changes: **71/71 PASS**.
Fork CI (E2E Integration Tests, full suite): **70 passed, 1 skipped, 125 deselected, 2 xpassed, 17m45s**, backend coverage 29.35%.

### 1. Drop Plan Mode coverage (4 cases)
Upstream commit 6500376f removed `src/qwenpaw/plan/` and `src/qwenpaw/app/routers/plan.py`. `/plan` is now a stub returning "Plan mode is being migrated to the new task system" and `/api/agents/<id>/plan/config` returns 404. Removes `test_plan.py` (PLAN-001~004), `plan_page.py`, the conftest fixture, and the `plan` marker.

### 2. ChatPage.open() — replace networkidle with textarea wait
The previous implementation waited on `wait_for_load_state("networkidle")`. The chat page keeps SSE connections open, so networkidle never fires and the call always hits the 30s timeout. The try/except around `goto()` never caught it because `goto()` uses `wait_until="commit"` and returns immediately. Replaced with a direct wait for the chat textarea to become visible. `test_slash_unknown_does_not_crash` was passing only when the SSE buffer happened to drain on time.

### 3. test_channel_distribution_display
- Frontend i18n was reorganised into two charts: "Sessions by Channel" + "Messages by Channel" (zh: 按渠道会话统计 / 按渠道消息统计). Extend `has_channel_section` to recognise these.
- The old selector `main, [class*="content"], [class*="pageContent"]` now has 41 matches in the redesigned layout, with `.first` resolving to a sidebar menu chip ("Inbox") whose text never contains the page body. Read `body.inner_text()` directly.

### 4. setup_default_model fixture — substring match
After agentscope 2.0 the backend "built-in model not found" response changed from `{"detail":"...MODEL_NOT_FOUND..."}` to `{"detail":"Model 'dashscope/qwen3.6-plus' not found"}`. The existing branch keyed on the literal `"MODEL_NOT_FOUND"` substring and never entered the add-extra-model fallback, so the global active model stayed unset on CI. Every chat test that sends a message then hit the "LLM Model Required" modal and timed out on subsequent clicks (`test_multi_turn_context_awareness`, `test_session_rename_pin_delete_switch`). Relaxed the detection to a case-insensitive `"not found"` substring match.

### 5. test_model_list_display — selector overhaul
The redesigned Models page renders Provider tiles inside CSS-module containers (`index-module__providerCards__*`). The old `.qwenpaw-card / [class*="modelList"]` triple often resolved to 0 elements during the assert (race + class rename). Rewrote Step 4 to: wait up to 15s for the "Providers" heading, accept any `[class*="providerCard"]` / `[class*="providerCards"]` tile, and fall back to body text containing known provider names (DashScope / Aliyun / OpenCode / Kilo Code / Add Provider / Available Providers).

### 6. test_toggle_agent_status — replace inline enable with page object
Step 5 hand-rolled the toggle flow but checked `popconfirm_btn.is_visible()` without first waiting for the popconfirm container to render. On fresh-state runs the container hadn't mounted yet, so the confirm click was silently skipped, the enable never landed, and Step 6 saw the Disabled tag still present. Replaced the duplicated inline logic with `agents_page.enable_agent()`, which already waits for the popconfirm before clicking confirm.

## Local validation matrix

| Scope | Before | After |
|---|---|---|
| P0+integration on fresh working_dir | 67/71 | **71/71** |
| Full non-LLM | 175/177 | 176/177 |
| Fork CI full E2E Integration Tests | 4 failed | **70 passed, 1 skipped, 2 xpassed** |

## Known frontend bug (not in this PR)
`console/src/pages/Settings/TokenUsage/components/EmptyState.tsx` references `styles.emptyState` / `styles.emptyIcon`, but neither class is defined in `index.module.less`. The wrapper div therefore renders as `class="undefined "` and any e2e selector targeting `[class*=empty]` / `[class*=emptyState]` misses it. `test_token_usage_overview` passes when there is data; on a truly empty state it fails because the test cannot identify the empty container. Leaving this for a frontend fix; the e2e test will be updated once that lands.

## Test plan
- [x] Local fresh-state P0+integration: 71/71 PASS
- [x] Fork CI full E2E Integration Tests: green (70 passed, 1 skipped, 2 xpassed)
- [ ] Upstream CI smoke + integration green


lalaliat (approved):
LGTM

PR #5549, opened 2026-06-26
fix(providers): sanitize nullable tool schemas

### LINKED ISSUE #5543: [Bug]:functionDeclaration `***.cwd` schema didn't specify the schema type field
## QwenPaw Version
Version 1.1.12.post2

## Description

请求中的函数类型 `"type":"null"` 会导致第三方中转的部分模型无法处理请求

## Steps to Reproduce

1.

Image

我把`"type":"null"`替换成`"type":"object"`就能正常响应了

Image

wananing:
## Description

Fixes #5543.

Some OpenAI-compatible relays route tool schemas into Gemini-style `functionDeclaration` validators. QwenPaw's built-in tools can emit nullable JSON Schema branches such as:

```json
{
  "anyOf": [
    { "format": "path", "type": "string" },
    { "type": "null" }
  ],
  "default": null
}
```

Those nullable branches are valid JSON Schema, but the relay rejects them with errors like `schema didn't specify the schema type field`.

This PR extends the OpenAI-compatible tool schema sanitizer so final tool payloads no longer include `type: "null"` branches:

- collapses `anyOf` / `oneOf` nullable branches to the non-null schema;
- removes `"null"` from JSON Schema `type` arrays;
- converts null-only or permissive nullable fallbacks to an explicit `type: "object"`;
- preserves useful metadata such as `description`, `default: null`, and `required`;
- keeps the existing `$ref` expansion and boolean-schema sanitization behavior intact;
- reuses one schema walker for boolean-schema and nullable-schema sanitization.

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core
- [ ] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation
- [x] Tests
- [ ] CI/CD
- [ ] Scripts

## Checklist

- [x] Update documentation if needed (not needed; provider compatibility bug fix)

## Testing

- `env BLACK_CACHE_DIR=/tmp/qwenpaw-black-cache .venv/bin/pre-commit run --all-files`
- `.venv/bin/pytest tests/unit/providers`
- `.venv/bin/pytest tests/unit/providers/test_openai_tool_schema_compat.py tests/unit/providers/test_openai_stream_toolcall_compat.py`
- `uv run --python 3.11 --extra dev pytest` attempted earlier; one unrelated OneBot watchdog test fails locally and also fails when run alone.

### pre-commit run --all-files

Command:

```bash
env BLACK_CACHE_DIR=/tmp/qwenpaw-black-cache .venv/bin/pre-commit run --all-files
```

Summary:

```text
```

### Relevant pytest

Command:

```bash
.venv/bin/pytest tests/unit/providers
```

Summary:

```text
223 passed in 2.06s
```

Command:

```bash
.venv/bin/pytest tests/unit/providers/test_openai_tool_schema_compat.py tests/unit/providers/test_openai_stream_toolcall_compat.py
```

Summary:

```text
5 passed in 0.31s
```

### Full pytest attempt

Command:

```bash
uv run --python 3.11 --extra dev pytest
```

Summary from the earlier full-suite attempt:

```text
1 failed, 3759 passed, 11 skipped, 55 warnings in 971.12s
```

The failure is unrelated to this provider schema change:

```text
FAILED tests/unit/channels/test_onebot_channel.py::TestWatchdog::test_watchdog_restarts_when_port_unreachable
AssertionError: watchdog should have created a new site
```

After rebasing this branch, I also reran that single failing test and it still fails independently:

```bash
uv run --python 3.11 --extra dev pytest tests/unit/channels/test_onebot_channel.py::TestWatchdog::test_watchdog_restarts_when_port_unreachable
```

```text
1 failed in 0.91s
```

### DIFF
--- a/src/qwenpaw/providers/openai_chat_model_compat.py
+++ b/src/qwenpaw/providers/openai_chat_model_compat.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 import json
 from datetime import datetime
 from types import SimpleNamespace
-from typing import Any, AsyncGenerator
+from typing import Any, AsyncGenerator, Callable
 
 from agentscope.model import OpenAIChatModel
 from agentscope.model._model_response import ChatResponse
@@ -235,6 +235,61 @@ _MAP_SCHEMA_KEYWORDS = frozenset(
 )
 
 
+def _walk_schema(
+    schema: dict[str, Any],
+    transform: Callable[[Any], Any],
+    *,
+    dependency_schema_types: tuple[type, ...] = (dict,),
+) -> dict[str, Any]:
+    """Walk known JSON Schema child positions using *transform*."""
+    result: dict[str, Any] = {}
+    for key, value in schema.items():
+        if key in _SINGLE_SCHEMA_KEYWORDS:
+            if key == "items" and isinstance(value, list):
+                result[key] = [transform(item) for item in value]
+            else:
+                result[key] = transform(value)
+        elif key in _ARRAY_SCHEMA_KEYWORDS:
+            if isinstance(value, list):
+                result[key] = [transform(item) for item in value]
+            else:
+                result[key] = value
+        elif key in _MAP_SCHEMA_KEYWORDS:
+            if isinstance(value, dict):
+                result[key] = {
+                    item_key: transform(item_value)
+                    for item_key, item_value in value.items()
+                }
+            else:
+                result[key] = value
+        elif key == "dependencies" and isinstance(value, dict):
+            # draft-07: value per key may be a schema or a string array.
+            result[key] = {
+                item_key: (
+                    transform(item_value)
+                    if isinstance(item_value, dependency_schema_types)
+                    else item_value
+                )
+                for item_key, item_value in value.items()
+            }
+        else:
+            result[key] = value
+    return result
+
+
+def _strip_boolean_schema_special_cases(schema: Any) -> Any:
+    if not isinstance(schema, dict):
+        return schema
+    return {
+        key: value
+        for key, value in schema.items()
+        if not (
+            (key == "additionalProperties" and value is True)
+            or (key == "required" and isinstance(value, bool))
+        )
+    }
+
+
 # pylint: disable=too-many-branches
 def _sanitize_boolean_schemas(schema: Any) -> Any:
     """Position-aware sanitizer for boolean JSON Schema values.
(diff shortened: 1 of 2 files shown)


src/qwenpaw/providers/openai_chat_model_compat.py
```diff
@@ -314,6 +314,134 @@ def _sanitize_boolean_schemas(schema: Any) -> Any:
...
+                _normalize_nullable_type(merged),
+            )
+
+        if non_null_variants:
+            node[key] = non_null_variants
+        else:
+            node.pop(key, None)
+            node = _ensure_object_type_for_untyped_schema(node)
+
+    result: dict[str, Any] = {}
```
qbc2016: The recursive keyword walker here (_SINGLE_SCHEMA_KEYWORDS, _ARRAY_SCHEMA_KEYWORDS,
_MAP_SCHEMA_KEYWORDS, dependencies) is almost identical to the one in `_sanitize_boolean_schemas`
(lines 284–313). If a new JSON Schema keyword is added or a walker bug is fixed, both copies must
be updated in lockstep.
Consider extracting a shared `_walk_schema(schema, transform_fn)` helper that both sanitizers can
delegate to. Fine to leave as a follow-up since this is a bugfix PR.

tests/unit/providers/test_openai_tool_schema_compat.py
```diff
@@ -0,0 +1,118 @@
...
+            read_file,
+            governor=None,
+            request_context={},
+        ),
+        PolicyGuardedTool(
+            execute_shell_command,
+            governor=None,
+            request_context={},
+        ),
+    ]
```
qbc2016, line 133: This test imports the real `read_file` and `execute_shell_command` tools and calls `Toolkit(tools=...).get_tool_schemas()` to get live schemas. While this is a valuable integration check, it will break if those tools' parameter signatures are refactored (e.g. parameter renamed or removed), even though the test is verifying the *sanitizer* — not the tools themselves.

Consider adding an explicit inline schema fixture alongside this test (similar to the second test `test_sanitize_tool_schemas_removes_null_from_type_arrays`) so the sanitizer's correctness is validated independently of external tool definitions.

tests/unit/providers/test_openai_tool_schema_compat.py
```diff
@@ -0,0 +1,118 @@
...
+    assert cwd["type"] == "string"
+    assert cwd["format"] == "path"
+    assert cwd["default"] is None
+    assert "anyOf" not in cwd
+
+    sandbox_config = shell_params["properties"]["sandbox_config"]
+    assert sandbox_config["type"] == "object"
+    assert sandbox_config["default"] is None
+    assert "anyOf" not in sandbox_config
+
```
qbc2016, line 174: The current assertions check individual fields one at a time (`assert cwd["type"] == "string"`, `assert cwd["format"] == "path"`, `assert "anyOf" not in cwd`, etc.). This style makes it harder to see the full expected output at a glance — a reader has to mentally reconstruct the target shape from scattered field checks.

Prefer asserting the full expected dict/list directly, e.g.:

```python
assert cwd == {"type": "string", "format": "path", "default": None}
assert start_line == {"type": "integer", "default": None}
assert sandbox_config == {"type": "object", "default": None}
```
or for the second test:
```python
assert properties == {
    "single": {"type": "string", "default": None},
    "multiple": {"type": ["integer", "number"]},
    "null_only": {"type": "object", "default": None},
}
```
Full dict equality makes the expected schema shape immediately obvious and also catches unexpected extra keys that individual field assertions would miss silently.

qbc2016 (commented):

src/qwenpaw/providers/openai_chat_model_compat.py
wananing: Addressed in d41c4e9e: extracted a shared `_walk_schema(...)` helper and made both `_sanitize_boolean_schemas` and `_sanitize_nullable_schemas` delegate schema-position recursion to it, while keeping their own sanitizer-specific normalization logic.

tests/unit/providers/test_openai_tool_schema_compat.py
wananing: Addressed in d41c4e9e: added an explicit inline schema fixture that validates nullable `anyOf` / `oneOf` branches independently of the live built-in tool definitions. I kept the built-in-tool test as an integration coverage check.

tests/unit/providers/test_openai_tool_schema_compat.py
wananing: Addressed in d41c4e9e: changed the sanitizer tests to assert full expected dict shapes for the sanitized fields/properties, so unexpected extra keys are caught directly.

qbc2016 (approved):
LGTM.

Issue #5554, opened 2026-06-26
[Bug] 企业微信发送文件后 bot 无回复，channel 重启导致文件消息处理中断


mynameyi:
通过企业微信 AI Bot 发送文件后，QwenPaw 无任何回复。文件实际已被下载到本地 media 目录，但 agent 未收到处理结果。

环境：

QwenPaw 版本：1.1.12.post2
频道：企业微信（wecom）
OS：Windows 10
模型：opencode / mimo-v2.5-free（plm-assistant agent）
复现步骤：

通过企业微信向 bot 发送一个 .docx 文件
等待处理
期望行为： bot 能识别文件、给出回复（如确认收到、读取内容等）

实际行为： bot 无任何回复

日志关键证据：

18:15:26 | wecom recv: msgtype=file text_len=0     # 文件被接收
18:15:26 | Consumer started                         # 消费者启动
                                                    # ← 缺少 _consume_with_tracker，说明消费未开始
18:15:39 | wecom channel started (aib7fxGBeUd2)     # 另一个 channel 启动
18:15:39 | Consumer stopped                         # 消费者被终止，消息丢失
文件已下载到本地：media/wecom_d5350fd9_八爪鱼项目技术方案.docx，但 agent 没有处理该消息。

疑似原因： 文件消息进入队列后、开始消费前，另一个 wecom channel 的启动导致队列被清理（Consumer stopped），消息被丢弃。

wananing:
Hi, I would like to work on this issue.

I traced the current WeCom receive path and found that the downloaded file itself is not the failing part. The file message is downloaded and converted into `FileContent`, then enqueued normally. The likely immediate cause is that WeCom does not currently override the base no-text debounce behavior.

Current path:

1. WeCom receives `msgtype=file` and downloads it into the local media directory.
2. The message is converted into a native payload containing `FileContent`.
3. `ChannelManager` consumes the payload and calls `BaseChannel._consume_one_request()`.
4. `BaseChannel._debounce_payload()` calls `_apply_no_text_debounce()`.
5. The base implementation buffers non-text content unless it is audio, so a pure file/image/video message can be stored in `_pending_content_by_session` and never reach `_consume_with_tracker()` until a later text message arrives.

That matches the reported log: `wecom recv: msgtype=file text_len=0` and `Consumer started`, but no `_consume_with_tracker` log.

[...]

PR #5560, opened 2026-06-26
fix(channels): process WeCom media-only messages

wananing:
## Description

Refs #5554.

This PR fixes WeCom media-only messages getting buffered by the base no-text debounce path and never reaching the agent until a later text message arrives.

WeCom already downloads incoming `file`, `image`, and `video` messages and converts them into content blocks. The missing piece was matching the behavior already used by Telegram and OneBot: media-only payloads should be treated as complete user input instead of waiting for text.

The change adds a WeCom-specific `_apply_no_text_debounce()` override that:

- processes media-only content immediately;
- prepends any previously buffered content for the same session;
- keeps normal text-message debounce behavior unchanged.

I am keeping the broader reload/shutdown pending-work issue out of this PR. That affects all manager-queue channels and should be handled separately in a follow-up PR for `ChannelManager` / `UnifiedQueueManager` lifecycle handling.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend
- [ ] Console
- [x] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran pre-commit checks locally through the commit hook
- [x] I ran relevant tests locally
- [ ] Documentation updated
- [x] Ready for review

## Testing

Added unit coverage for WeCom media-only debounce behavior:

- pure `FileContent` is processed immediately;
- native WeCom file payloads pass the debounce boundary instead of being buffered;
- pending media is prepended before new media;
- text-only content still follows the base behavior.

```bash
git commit --amend --no-edit
check yaml...........................................(no files to check)Skipped
check xml............................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check json...........................................(no files to check)Skipped
prettier.............................................(no files to check)Skipped
```

```bash
.venv/bin/pytest tests/unit/channels/test_wecom.py tests/unit/channels/test_base_core.py -q
139 passed, 1 skipped in 1.48s
```

### DIFF
--- a/src/qwenpaw/app/channels/wecom/channel.py
+++ b/src/qwenpaw/app/channels/wecom/channel.py
@@ -1271,6 +1271,22 @@ class WecomChannel(BaseChannel):
     # Session processing state management
     # ------------------------------------------------------------------
 
+    def _apply_no_text_debounce(
+        self,
+        session_id: str,
+        content_parts: list[Any],
+    ) -> tuple[bool, list[Any]]:
+        """Process media-only WeCom messages without waiting for text."""
+        has_media = any(
+            getattr(part, "type", None)
+            not in (ContentType.TEXT, ContentType.REFUSAL)
+            for part in content_parts
+        )
+        if has_media:
+            pending = self._pending_content_by_session.pop(session_id, [])
+            return True, pending + list(content_parts)
+        return super()._apply_no_text_debounce(session_id, content_parts)
+
     async def _consume_with_tracker(
         self,
         request: "AgentRequest",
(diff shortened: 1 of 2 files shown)


hongxicheng:
Thanks for the contribution @wananing!

However, forcefully bypassing the no-text debounce for a specific channel is not the direction we'd like to take. Different users have different needs — some prefer media-only messages to be processed immediately, while others rely on the debounce behavior to batch media with follow-up text.

I'll be closing this PR. In the near future, we plan to introduce a configurable toggle across all channels that lets users decide whether to enable or disable the no-text debounce mechanism, rather than hardcoding the behavior per channel.

Appreciate your effort on this! 🙏

PR #5575, opened 2026-06-26
feat(channels): make no-text debounce configurable

wananing:
## Description

Refs #5554.

This PR adds a common `no_text_debounce_enabled` channel config toggle so users can decide whether media-only messages should wait for follow-up text or be processed immediately.

The implementation keeps existing channel defaults when the field is unset (`null`): base channels continue to debounce media-only content, while Telegram and OneBot keep their existing media-only immediate processing defaults. When the field is explicitly set, `ChannelManager.from_config()` applies it to every channel instance through the shared BaseChannel setter.

For WeCom or any other base-channel path, setting `no_text_debounce_enabled: false` processes media/file-only messages immediately without hardcoding a WeCom-specific bypass. Setting it to `true` makes Telegram/OneBot use the shared base debounce path.

Docs for the common channel fields were updated in both English and Chinese.

## Type of Change

- Feature

## Component(s) Affected

- Channels
- Config
- Documentation
- Tests

## Checklist

- [x] Run and pass relevant tests (`pytest`)
- [x] Update documentation if needed

## Testing

- Verified the shared BaseChannel behavior when no-text debounce is disabled.
- Verified ChannelManager applies the config toggle to constructed channels.
- Verified WeCom channel tests still pass.
- Verified Telegram and OneBot preserve their current default media-only behavior, and use the base debounce behavior when the toggle is explicitly enabled.
- Verified updated docs with Prettier 3.0.0.

```text
env BLACK_CACHE_DIR=/tmp/qwenpaw-black-cache .venv/bin/pre-commit run --all-files
# Passed
```

```text
.venv/bin/pytest tests/unit/channels/test_base_core.py::TestNoTextDebounceBuffering tests/unit/channels/test_manager.py tests/unit/channels/test_wecom.py tests/unit/channels/test_telegram.py::TestTelegramNoTextDebounce tests/unit/channels/test_onebot_channel.py::TestOneBotNoTextDebounce
# 81 passed
```

```text
npx --yes prettier@3.0.0 --check public/docs/channels.en.md public/docs/channels.zh.md
# All matched files use Prettier code style
```

### DIFF
--- a/src/qwenpaw/app/channels/base.py
+++ b/src/qwenpaw/app/channels/base.py
@@ -4,6 +4,7 @@
 """
 Base Channel: bound to AgentRequest/AgentResponse, unified by process.
 """
+
 from __future__ import annotations
 
 import asyncio
@@ -133,6 +134,7 @@ class BaseChannel(ABC):
         streaming_enabled: bool = False,
         access_control_dm: bool = False,
         access_control_group: bool = False,
+        no_text_debounce_enabled: bool = True,
     ):
         self._process = process
         self._on_reply_sent = on_reply_sent
@@ -149,6 +151,7 @@ class BaseChannel(ABC):
         self.require_mention = require_mention
         self.access_control_dm = access_control_dm
         self.access_control_group = access_control_group
+        self._no_text_debounce_enabled = bool(no_text_debounce_enabled)
         self._language = "zh"
         self._enqueue: EnqueueCallback = None
         self._workspace = None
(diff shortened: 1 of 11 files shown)


hongxicheng:
Thanks for the contribution! We've submitted a more comprehensive PR (#5617 ) that addresses the same feature. A few shortcomings in this approach:

1. **Duplicate overrides retained** — telegram/onebot still carry `_apply_no_text_debounce` overrides with conditional `super()` delegation instead of unifying the logic in BaseChannel.

2. **Inconsistent config wiring** — Uses a post-construction `set_no_text_debounce_enabled()` setter rather than passing through `from_config` → `__init__` like `filter_tool_messages`/`filter_thinking`.

3. **Unnecessary nullable field** — `Optional[bool] = None` requires extra `is not None` checks throughout; a simple `bool = True` suffices.

4. **No frontend UI** — Users can only configure via config file; no ChannelDrawer toggle or i18n support.

Closing this in favor of #5617 .

PR #5586, opened 2026-06-28
fix(context): prioritize runtime model over static config for compaction threshold

zorrofox1121:
### Description

When a user switches the model **inside a conversation session** (via the UI dropdown, not via Settings → Set as Default), the `light_context_config` compaction logic still reads `max_input_length` from the default model defined in `agent.json`. This causes the compact threshold to be calculated against the wrong model's context length, potentially leading to `MODEL_CONTEXT_LENGTH_EXCEEDED` errors or premature/overdue compaction.

### Root Cause

QwenPaw has two model-switching mechanisms, but the compaction code only reads from one:

| Mechanism | Scope | Persistence |
|-----------|-------|-------------|
| Settings → Models → Set as Default | Global / Agent-wide | `agent.json.active_model` + `active_model.json` |
| In-session dropdown selection | Current conversation only | Agent model instance only — **no** config file is updated |

The compaction threshold (`compact_threshold = max_input_length × compact_threshold_ratio`) is calculated via `get_model_max_input_length(agent_config)`, which queries `agent.json` / `active_model.json` — both of which always point to the **default** model, not the session-level override.

### Model Object Chain (for reference when fixing)

The session-level model information is available in the agent's runtime model chain:

```
agent.model
  └─ RetryChatModel
       └─ ._inner → TokenRecordingModelWrapper
            ├─ ._provider_id      → actual provider in use
            ├─ .model_name        → actual model in use
            └─ .model             → OpenAIChatModelCompat
```

`TokenRecordingModelWrapper` is created in `model_factory.py::create_model_and_formatter()` and always carries the correct `_provider_id` and `model_name`, regardless of how the model was selected.

### Expected Behavior

When a user switches models in-session, the compaction threshold should be calculated against the **currently active model**'s `max_input_length`, not the default model's.

### Actual Behavior

The compaction threshold is always calculated against the default model, ignoring in-session overrides.

### Steps to Reproduce

1. Set the default model to one with a large `max_input_length` (e.g., 131072)
2. Configure another provider/model with a smaller `max_input_length` (e.g., 65536) and set `compact_threshold_ratio = 0.5`
3. In the conversation UI, switch to the smaller model (via dropdown — **do not** set it as default)
4. Accumulate context by sending messages
5. **Expected**: compaction triggers at ~32K (`65536 × 0.5`)
6. **Actual**: compaction triggers at ~65K (`131072 × 0.5`), which may already exceed the backend's context window limit

### Environment

- QwenPaw version: main branch (as of 2026-06-28)
- Backend: llama.cpp / any OpenAI-compatible API
- Custom provider with `ModelInfo.max_input_length` set via `update_model_config()` API

### Additional Context

This issue was discovered and fixed locally against a later build that includes a `light_context_manager.py` module not yet present in the public `main` branch. The fix involves two parts:

#### Part 1: Fix `config.py` (can be applied now)

**File:** `src/qwenpaw/config/config.py`

**Function:** `get_model_max_input_length(agent_config)` (~line 2354)

Change the priority from `agent_config.active_model → runtime` to `runtime → agent_config.active_model`:

```diff
 # Before: try agent.json first, fall back to runtime
-model_slot = agent_config.active_model
+model_slot = ProviderManager.get_active_model()
 if not model_slot or not model_slot.provider_id:
-    model_slot = ProviderManager.get_active_model()
+    model_slot = agent_config.active_model
```

This fixes the path for `/model` command switches (which write to `agent.json`).

#### Part 2: Fix the session-level model path (needs the new code)

In the upcoming version that includes `agents/context/light_context_manager.py`, the `pre_reasoning()` method should read `max_input_length` from `agent.model` instead of `agent_config`. The approach is:

1. Add a new function `get_model_max_input_length_from_agent()` that traverses `agent.model → _inner → TokenRecordingModelWrapper` to extract `_provider_id` and `model_name`, then looks up the provider's `ModelInfo.max_input_length`, falling back to `get_model_max_input_length(agent_config)` on failure
2. In `pre_reasoning()`, replace `get_model_max_input_length(agent_config)` with this new function call

This ensures the compaction threshold always follows the conversation-level model, no matter which selection mechanism was used.

### DIFF
--- a/src/qwenpaw/config/config.py
+++ b/src/qwenpaw/config/config.py
@@ -2354,14 +2354,16 @@ def get_model_max_input_length(
     """
     from ..providers import ProviderManager
 
-    model_slot = agent_config.active_model
+    # First try the runtime active model (set by zorrofox)
+    try:
+        manager = ProviderManager.get_instance()
+        model_slot = manager.get_active_model()
+    except Exception:
+        model_slot = None
+        
     # Fallback: if agent.json doesn't have active_model, try ProviderManager
     if not model_slot or not model_slot.provider_id:
-        try:
-            manager = ProviderManager.get_instance()
-            model_slot = manager.get_active_model()
-        except Exception:
-            pass
+        model_slot = agent_config.active_model
 
     if model_slot and model_slot.provider_id and model_slot.model:
         try:
--- a/src/qwenpaw/config/config.py
+++ b/src/qwenpaw/config/config.py
@@ -2354,7 +2354,7 @@ def get_model_max_input_length(
     """
     from ..providers import ProviderManager
 
-    # First try the runtime active model (set by zorrofox)
+    # First try the runtime active model (set by zorrofox1121)
     try:
         manager = ProviderManager.get_instance()
         model_slot = manager.get_active_model()


src/qwenpaw/config/config.py
```diff
@@ -2354,14 +2354,16 @@ def get_model_max_input_length(
     """
     from ..providers import ProviderManager
 
-    model_slot = agent_config.active_model
+    # First try the runtime active model (set by zorrofox1121)
```
qbc2016, line 2357: Please remove the GitHub username from the code comment — it is not meaningful to future readers.

src/qwenpaw/config/config.py
```diff
@@ -2354,14 +2354,16 @@ def get_model_max_input_length(
...
+    try:
+        manager = ProviderManager.get_instance()
+        model_slot = manager.get_active_model()
+    except Exception:
+        model_slot = None
+        
     # Fallback: if agent.json doesn't have active_model, try ProviderManager
```
qbc2016, line 2364: The comment is stale after the priority swap — the fallback is now `agent_config.active_model`, not ProviderManager.

qbc2016 (commented):

RerankerGuo (changes requested):
I found a blocking model-scope regression at `7bcb14f48b34f35734763feba85bc26f3cb273c7`.

`ProviderManager.get_active_model()` is the process-wide global default, not the model selected for the current agent/session. The changed priority at [config.py lines 2357-2366](https://github.com/agentscope-ai/QwenPaw/blob/7bcb14f48b34f35734763feba85bc26f3cb273c7/src/qwenpaw/config/config.py#L2357-L2366) therefore replaces every valid `agent_config.active_model` with the global model whenever one is configured.

I exercised the production helper with:

```text
agent_config.active_model = agent-provider:agent-small   (65,536)
ProviderManager.active_model = global-provider:global-large (131,072)
```

The two-sided contract produced:

```text
current main:
  agent model precedes global model       PASS (65,536)
  missing agent model falls back globally PASS (131,072)

exact head and clean current-main trial:
  agent model precedes global model       FAIL (actual 131,072)
  missing agent model falls back globally PASS (131,072)
```

On current main, the real per-request/session override is passed separately as `model_slot_override` to `create_model_and_formatter()` and applied after the persisted agent slot ([model_factory.py lines 1393-1460](https://github.com/agentscope-ai/QwenPaw/blob/bacac7410cf2ac122dd892152beff8b03f38729f/src/qwenpaw/agents/model_factory.py#L1393-L1460)). It is never represented by `ProviderManager.active_model`, and this helper only receives `agent_config`, so the priority swap cannot implement the PR's stated session-override behavior. Instead, it makes `/history`, daemon status, and fallback usage snapshots report the global context window for agents with independent models.

Please keep `agent_config.active_model` ahead of the global fallback. If context reporting must follow a per-request override, pass the resolved model slot/context size from the request model path into those consumers rather than reading unrelated process-global state.

Validation:

```text
exact related suites: 255 passed
current-main trial related suites: 445 passed
current-main focused baseline: 62 passed
external scope contract: main 2 passed; exact/trial 1 failed, 1 passed
```

The branch also fails `git diff --check` and the repository's `trailing-whitespace` hook at `config.py:2363`; the remaining applicable Python hooks pass.

lalaliat:
Thank you for your contribution! This PR aims to address an important scenario: the compaction threshold does not update when the model is switched within a conversation.

However, `ProviderManager.get_active_model()` returns the global default model, not the model selected for the current conversation. Prioritizing it would cause agents with their own model configurations to use the wrong context window limit.

The current main branch already calculates the compaction threshold directly from `agent.model.context_size`, so we’re closing this PR for now. We truly appreciate your effort and hope to see future contributions from you!

PR #5629, opened 2026-06-29
fix(memory): fix memory-related system prompt to avoid over-eager memory writing

niceIrene:
## Expose context strategy in the UI & curb over-eager memory writing

This PR makes two related context/memory improvements that were previously invisible or counterproductive to users.

### 1. Console: expose the native/scroll context strategy selector

The light context manager already supports two strategies — `native` (AgentScope in-place compression) and `scroll` (durable `history.db` + recall REPL, the default) — but the choice lived only in `light_context_config.strategy` and could not be set from the UI. The `Context Manager Backend` dropdown only ever shows the single `light` option, so the meaningful choice was buried.

This adds a top-level **Context Strategy** dropdown next to *Context Manager Backend*, with options **Scroll** (default) and **Native**, bound to `light_context_config.strategy`.

image

### 2. Memory prompt: stop the agent from writing memory documents on every turn

The memory guidance prompt pushed the agent to record aggressively — the *"Write It Down – No Mental Notes!"* and *"Proactive Recording"* sections, especially the **"record first, answer second / write it down immediately"** mandate. This duplicated and fought the existing dream promotion process, which already consolidates recalled daily-note content into `MEMORY.md`, and caused the agent to create memory files constantly.

- **Slimmed** the memory guidance to: file roles (`MEMORY.md` / daily notes) + one restrained "when to edit `MEMORY.md`" line + the Retrieval Tool section. Removed the aggressive proactive-write sections.
- **Scoped `memory_search`** to curated long-term memory (durable preferences, profile/personal facts, settled decisions, open to-dos), removing the raw-history terms ("past work, dates, people") that belong to scroll's `recall_history_python`.
- **Aligned the scroll prompt's** vocabulary and made the boundary explicit: recall is the source of truth for the raw conversation record; memory is the curated subset.

Files: `src/qwenpaw/agents/memory/prompts.py`, `src/qwenpaw/agents/context/scroll/prompt.py`


jinliyl (approved):
LGTM

PR #5637, opened 2026-06-30
feat(subagent): add event-driven background lifecycle and parent wakeup

### LINKED ISSUE #4873: [Bug]: 同时开两个subagent会导致主agent无限快速轮询直至任务结束，并且飞书侧无法打断轮询
## QwenPaw Version
1.1.10b

## Description

  Starting two subagent tasks as background tasks at the same time causes very frequent check_agent_task tool call and LLM requests. The running task can be interrupted from the Dashboard, but cannot interrupt from the Feishu side.

  Because foreground subagent execution cannot run tasks in parallel, I started both subagent tasks in the background. During execution, QwenPaw keeps querying the background task
  status very aggressively(using check_agent_task tool). This looks like a busy polling loop and can make the Feishu conversation flooded with messages..

**Related PR(s):** #(optional)
  ## Related Issues
  - #4031
  - #4044

## Component(s) Affected

  - [x] Core / Backend
  - [x] Channels: Feishu
  - [x] Multi-agent / Subagent management
  - [x] Dashboard / task control
  - [ ] Console frontend
  - [ ] Skills
  - [ ] CLI
  - [ ] Documentation
  - [ ] Tests

## Environment

  - QwenPaw version: 1.1.10b1
  - Install method: curl -fsSL https://qwenpaw.agentscope.io/install.sh | bash
  - OS: Ubuntu 22.04.5 LTS
  - Channel: Feishu & Dashboard
  - Model/provider: deepseek-v4-flash

## Steps to Reproduce

  1. Enable the Feishu channel or on dashboard.
  2. From Feishu, ask the agent to start two subagent tasks (ask each subagent to make a ppt).
  3. Both subagent tasks will be started in background, because foreground execution cannot run them in parallel.
  4. While the tasks are running, observe the backend logs or Dashboard.

## Actual vs Expected

- **Actual:**
-   - Both subagent tasks run in the background.
  - The parent/main session keeps polling their status very frequently.
  - The polling looks excessive and may continue like a busy loop during task execution.
  - Feishu cannot effectively interrupt/cancel the running background subagent tasks.

- **Expected:**

  - When each background subagent finishes, it should emit a completion event back to the parent/main agent.
  - The parent/main agent should be notified when both subagents have completed.
  - The parent/main agent should wait for subagent completion events instead of blindly polling task status.

## Logs / Screenshots

```
check_agent_task tool call: 
INPUT
{"task_id": "779c898c-afb9-44e7-9ed1-8dfcfd2a6bdb"}

OUTPUT
[
  {
    "type": "text",
    "text": "[TASK_ID: 241607c3-6f4e-4da4-bb77-2cca71fca386]\n[STATUS: running]\n\nTask is still running...\nStarted at: 1780315630.1314218"
  }
]

keep repeating... until subagent finished
```

splash-li:
## Summary

This PR changes `spawn_subagent(background=True)` from a polling-based task
into an event-driven child lifecycle. A background child returns immediately,
runs independently, and wakes its parent with the result when it finishes.

It also adds heartbeat detection, parent-to-child cancellation, per-parent
concurrency limits, and reliable event delivery. Healthy long-running children
have no total runtime timeout, and nested subagents are intentionally disabled.

Fixes https://github.com/agentscope-ai/QwenPaw/issues/4873.

## What Problem This Solves

Background subagents currently rely on parent-side polling to discover completion. That wastes turns, cannot reliably wake an idle parent, and leaves cancellation, heartbeat loss, ownership, and callback delivery spread across unrelated runtime paths. This change adds a workspace-scoped lifecycle manager so a background child returns immediately, runs independently, and delivers a terminal result back to its owning parent session without a polling loop.

## Evidence

- Post-rebase focused lifecycle, agent-management, Console, and integration suites pass: 64 passed in 21.19s.
- `git diff --check` passes against official `origin/main` at `af13614c`.
- The manual runtime captures below show that background children start without `check_agent_task` polling and that completion wakes the parent and delivers the result automatically.
- The lifecycle tests cover callback claiming/acknowledgement, retry after save failure, heartbeat expiry, cancellation propagation, parent isolation, concurrency limits, stale-task handling, and nested-subagent rejection.

## Upstream basis

This PR is rebased on the latest official `main`, including the official
Runtime 2.0 `spawn_subagent` and generic background-chat implementation merged
in https://github.com/agentscope-ai/QwenPaw/pull/5660. It does not replace or
duplicate those routes; it adds subagent-specific parent/child lifecycle
behavior on top.

### 1. Build on the official Runtime 2.0 background-task path

- Reuse the official `spawn_subagent` tool descriptor and governance target.
- Reuse the official `/console/chat/task` routes and `_bg_tasks` store for
  generic background agent chat (`submit_to_agent`).
- Keep foreground response selection and generic background behavior from the
  upstream implementation.
- Add only the `CancelSubagent` governance entry and subagent-specific
  lifecycle integration required by the sections below.

### 2. Add a workspace-scoped background subagent manager

`SubagentTaskManager` owns only background subagents; it is not a new generic
task framework. Each Workspace has one manager and one watchdog.

It tracks:

- parent/root/child agent and session identities;
- the real `asyncio.Task` handle;
- lifecycle state, result/error, heartbeat lease, cancellation reason, and
  optional fork branch;
- terminal/stale events waiting for the parent session.

Lifecycle states:

```text
SUBMITTED -> RUNNING -> COMPLETED
                     -> FAILED
                     -> CANCELLING -> CANCELLED
                                   -> STALE -> eventual real terminal state
```

`STALE` is deliberately non-terminal. It means cooperative cancellation did
not finish within the grace period; the task handle and concurrency slot are
retained instead of pretending the child was killed.

### 3. Replace polling with callback wakeup

For `spawn_subagent(background=True)`:

- run the child through the existing `workspace.stream_query` Runtime path;
- return `[TASK_ID]` and `[SESSION]` immediately;
- tell the model not to poll or sleep;
- publish a terminal/stale event into the parent-session inbox;
- wait for an active parent run to become idle, then wake an idle parent via
  `ChannelManager.enqueue`;
- inject child results before parent reasoning as AgentScope 2.0
  `HintBlock`/`HintBlockEvent` values;
- acknowledge events only after parent session persistence succeeds;
- release a claim on failure so a later run can retry delivery.

`check_agent_task` remains available as a manual/debug fallback. It is no
longer the normal coordination mechanism for background subagents.

### 4. Add heartbeat and watchdog semantics

- Reuse QwenPaw's existing Runtime heartbeat envelope; do not create a
  per-child heartbeat task.
- Runtime heartbeat interval: 25 seconds.
- Missing-heartbeat threshold: four intervals (about 100 seconds).
- Cancellation grace period: 10 seconds.
- No total runtime limit for background subagents. The `timeout` argument is
  ignored in background mode; model/tool/network operations retain their own
  timeouts.
- If the watchdog itself is delayed by event-loop suspension, refresh leases
  instead of mass-cancelling healthy children.

### 5. Add ownership, cancellation, and isolation

- Add `cancel_subagent(task_id)`; only the creating parent session can cancel
  its child.
- Console stop, `/stop`, Runtime cancellation, and Workspace shutdown cancel
  linked children.
- Parent termination discards pending child callbacks and suppresses late
  wakeups.
- Limit each parent session to eight non-terminal background children.
- Managers are Workspace-scoped, so different agents cannot mix records or
  events even if session IDs happen to match.
- Background subagents are leaf nodes: nested `spawn_subagent` calls are
  rejected and delegation tools are hidden from the child toolkit.
- Channel metadata is filtered before reuse so tokens, secrets, callbacks,
  and runtime-only objects are not copied into child routing metadata.

### 6. Keep foreground and generic background behavior compatible

- `spawn_subagent(background=False)` keeps the existing synchronous path and
  foreground timeout semantics from the official implementation.
- Forked foreground behavior remains unchanged.
- Forked background children use the same lifecycle manager and report the
  worktree branch in the parent notification.
- `submit_to_agent` continues to use the generic Console background task plus
  `check_agent_task`; this PR changes only background subagent coordination.

## AgentScope 2.0 integration

The implementation reuses AgentScope/QwenPaw extension points rather than
patching AgentScope internals:

- `MiddlewareBase.on_reasoning` for inbox delivery;
- `HintBlock` and `HintBlockEvent` for lifecycle notifications;
- existing Runtime `stream_query` and heartbeat envelopes;
- QwenPaw lifecycle hooks for request-local context and post-save ack;
- Workspace `ServiceManager` for manager startup/shutdown;
- existing `ChannelManager` for parent wakeup.

QwenPaw's normal Runtime does not run the full `agentscope.app` service stack,
so AgentScope's distributed `BackgroundTaskManager`/wakeup dispatcher cannot
be used directly here. The new manager is a small in-process adapter for
subagent-specific parent/child semantics, not a replacement for that stack.

## Expected behavior

| Scenario | Expected result |
| --- | --- |
| Foreground child | Parent blocks and receives the final child result as before. |
| Healthy long background task | Continues indefinitely while Runtime heartbeats arrive. |
| Background completion/failure | Parent receives a HintBlock and is automatically woken if idle. |
| Parent is already running | Wakeup waits; the active run may consume the event on its next reasoning step. |
| Delivery/session save fails | Event claim is released and remains available for retry. |
| Parent stop/cancel | Direct children are cancelled and no late callback re-wakes the parent. |
| Heartbeat disappears | Child is cancelled; parent receives CANCELLED or one STALE event if cancellation does not finish. |
| Event loop pauses | Leases are refreshed; children are not falsely killed in bulk. |
| Ninth child from one parent | Spawn is rejected until one of the eight non-terminal children exits. |
| Different parent/Workspace | Tasks and notifications remain isolated. |
| Child attempts nested spawn | Tool is hidden and runtime validation rejects the call. |

### New runtime package

- `src/qwenpaw/app/subagents/__init__.py`
- `src/qwenpaw/app/subagents/context.py`
- `src/qwenpaw/app/subagents/hooks.py`
- `src/qwenpaw/app/subagents/manager.py`
- `src/qwenpaw/app/subagents/middleware.py`

### Modified production files

- `src/qwenpaw/agents/tools/agent_management.py`
- `src/qwenpaw/agents/tools/__init__.py`
- `src/qwenpaw/app/_app.py`
- `src/qwenpaw/app/workspace/workspace.py`
- `src/qwenpaw/app/workspace/local_workspace.py`
- `src/qwenpaw/runtime/builder.py`
- `src/qwenpaw/runtime/commands/control/stop_handler.py`
- `src/qwenpaw/app/routers/console.py`
- `src/qwenpaw/app/channels/console/channel.py`
- `src/qwenpaw/hooks/error/error_hook.py`
- `src/qwenpaw/hooks/session/session_hook.py`
- `src/qwenpaw/config/config.py`
- `src/qwenpaw/governance/tool_registry.py`

### Tests

- `tests/unit/app/test_subagent_lifecycle.py`
- `tests/integration/test_spawn_subagent.py`
- `tests/unit/agents/tools/test_agent_management.py`
- `tests/unit/channels/test_console.py`
- `tests/unit/governance/test_policy.py`

## Verification

Against official `origin/main` commit
`af13614c`:

```text
Focused subagent lifecycle, agent-management, Console, and integration suites:
64 passed in 21.19s

git diff --check:
passed
```

The full repository test suite and `pre-commit run --all-files` have not been
run in this verification pass.

### Manual runtime verification

**1. Background subagents start without a `check_agent_task` polling loop**

WDjMvcgAQ3

**2. Child completion wakes the parent and delivers the result automatically**

XDGghYqM3l

## Scope and known limitations

- Lifecycle records and inbox events are in memory. Process restart does not
  resume children or recover undelivered callbacks.
- This targets QwenPaw's current single-process Workspace model; it is not a
  distributed queue or persistent task scheduler.
- `asyncio.Task.cancel()` is cooperative. A child stuck in non-interruptible
  synchronous code may remain `STALE`; process/tool-level cleanup remains the
  tool runner's responsibility.
- Only one subagent level is supported intentionally.
- Heartbeats are liveness signals, not semantic progress reports.
- Terminal records are retained until delivery is acknowledged (or the parent
  is stopped/Workspace shuts down), prioritizing reliable delivery over TTL
  eviction. A future persistent/distributed implementation should add durable
  retention and cleanup policy.

## Rebase and conflict-resolution note

This branch was rebased after the official Runtime 2.0 fix landed in #5660.
The duplicated #5524 commits, legacy `_CHAT_TASKS` store, duplicate background
routes, and obsolete tests for that store were removed. The remaining changes
are the parent/child manager, heartbeat watchdog, HintBlock delivery, callback
wakeup, cancellation propagation, one-level restriction, and no-polling
behavior.

### DIFF
--- a/src/qwenpaw/app/subagents/manager.py
+++ b/src/qwenpaw/app/subagents/manager.py
@@ -24,6 +24,8 @@ logger = logging.getLogger(__name__)
 DEFAULT_HEARTBEAT_TIMEOUT_SECONDS = HEARTBEAT_INTERVAL_SECONDS * 4
 DEFAULT_WATCHDOG_INTERVAL_SECONDS = HEARTBEAT_INTERVAL_SECONDS
 DEFAULT_CANCEL_GRACE_SECONDS = 10.0
+DEFAULT_PARENT_WAKEUP_WAIT_TIMEOUT_SECONDS = 300.0
+DEFAULT_PARENT_WAKEUP_RETRY_INTERVAL_SECONDS = 5.0
 
 
 class SubagentStatus(str, Enum):
@@ -114,6 +116,12 @@ class SubagentTaskManager:
         heartbeat_timeout_seconds: float = DEFAULT_HEARTBEAT_TIMEOUT_SECONDS,
         watchdog_interval_seconds: float = DEFAULT_WATCHDOG_INTERVAL_SECONDS,
         cancel_grace_seconds: float = DEFAULT_CANCEL_GRACE_SECONDS,
+        parent_wakeup_wait_timeout_seconds: float = (
+            DEFAULT_PARENT_WAKEUP_WAIT_TIMEOUT_SECONDS
+        ),
+        parent_wakeup_retry_interval_seconds: float = (
+            DEFAULT_PARENT_WAKEUP_RETRY_INTERVAL_SECONDS
+        ),
     ) -> None:
         self.workspace = workspace
         self.max_running_per_parent = max_running_per_parent
@@ -126,11 +134,20 @@ class SubagentTaskManager:
             0.01,
         )
         self.cancel_grace_seconds = max(float(cancel_grace_seconds), 0.0)
+        self.parent_wakeup_wait_timeout_seconds = max(
+            float(parent_wakeup_wait_timeout_seconds),
+            0.0,
+        )
+        self.parent_wakeup_retry_interval_seconds = max(
+            float(parent_wakeup_retry_interval_seconds),
+            0.01,
+        )
         self._records: dict[str, SubagentTaskRecord] = {}
         self._events: list[SubagentLifecycleEvent] = []
         self._lock = asyncio.Lock()
         self._watchdog_task: asyncio.Task | None = None
         self._notification_tasks: set[asyncio.Task] = set()
+        self._wakeup_retry_parents: set[str] = set()
         self._stopping = False
 
     async def start(self) -> None:
@@ -442,8 +459,8 @@ class SubagentTaskManager:
     async def _watchdog_loop(self) -> None:
         """Cancel children whose Runtime heartbeat stream has stopped."""
         previous_tick = time.monotonic()
-        try:
-            while True:
+        while True:
+            try:
                 await asyncio.sleep(self.watchdog_interval_seconds)
                 now = time.monotonic()
                 tick_delay = now - previous_tick
@@ -459,10 +476,10 @@ class SubagentTaskManager:
                 stale_records = await self._watchdog_tick(now)
                 for record in stale_records:
                     self._schedule_parent_wakeup(record)
-        except Exception:  # pragma: no cover - defensive service boundary
-            logger.exception(
-                "Subagent heartbeat watchdog stopped unexpectedly",
-            )
+            except Exception:  # pragma: no cover - defensive service boundary
+                logger.exception(
+                    "Subagent watchdog tick failed; will retry next interval",
+                )
 
     async def _reset_leases_after_loop_pause(self, now: float) -> None:
         async with self._lock:
@@ -536,6 +553,51 @@ class SubagentTaskManager:
         )
         self._notification_tasks.add(task)
         task.add_done_callback(self._notification_tasks.discard)
+
+    def _schedule_parent_wakeup_retry(
+        self,
+        record: SubagentTaskRecord,
+    ) -> None:
+        """Retry an automatic parent wakeup without a tight idle poll loop."""
+        if self._stopping:
+            return
+        parent_session_id = record.parent_session_id
+        if parent_session_id in self._wakeup_retry_parents:
+            return
+        self._wakeup_retry_parents.add(parent_session_id)
+        task = asyncio.create_task(
+            self._retry_parent_wakeup_when_idle(record),
+            name=f"qwenpaw-subagent-wakeup-retry:{parent_session_id}",
+        )
+        self._notification_tasks.add(task)
+
+        def _done(done: asyncio.Task) -> None:
+            self._notification_tasks.discard(done)
+            self._wakeup_retry_parents.discard(parent_session_id)
+
+        task.add_done_callback(_done)
+
+    async def _retry_parent_wakeup_when_idle(
+        self,
+        record: SubagentTaskRecord,
+    ) -> None:
+        """Retry wakeup delivery at low frequency until the parent is idle."""
+        while not self._stopping:
+            await asyncio.sleep(self.parent_wakeup_retry_interval_seconds)
+            if not await self.has_pending_events(record.parent_session_id):
+                return
+            if await self._parent_session_is_running(record):
+                continue
+            manager = getattr(self.workspace, "channel_manager", None)
+            if manager is None or not record.channel:
+                logger.warning(
+                    "Cannot wake parent for subagent %s: channel unavailable",
+                    record.task_id,
+                )
+                return
+            if await self.has_pending_events(record.parent_session_id):
+                self._enqueue_parent_request(record, manager)
+            return
 
     async def get(self, task_id: str) -> SubagentTaskRecord | None:
         async with self._lock:
@@ -658,7 +720,18 @@ class SubagentTaskManager:
                     record.parent_session_id,
                     record.channel,
                 )
+            deadline = (
+                time.monotonic() + self.parent_wakeup_wait_timeout_seconds
+            )
             while chat_id and tracker is not None:
+                if time.monotonic() > deadline:
+                    logger.warning(
+                        "Timed out waiting for parent session %s to become "
+                        "idle; retrying wakeup when the parent becomes idle",
+                        record.parent_session_id,
+                    )
+                    self._schedule_parent_wakeup_retry(record)
+                    return
                 if await tracker.get_status(chat_id) != "running":
                     break
                 # The active parent may consume the HintBlock on its next
@@ -671,28 +744,50 @@ class SubagentTaskManager:
 
             if not await self.has_pending_events(record.parent_session_id):
                 return
-
-            from ...schemas import AgentRequest
-
-            request = AgentRequest(
-                input=[],
-                session_id=record.parent_session_id,
-                user_id=record.user_id or record.parent_session_id,
-                channel=record.channel,
-                root_session_id=record.root_session_id,
-                root_agent_id=record.parent_agent_id,
-                request_context={
-                    "subagent_wakeup": True,
-                    "subagent_event_task_id": record.task_id,
-                },
-            )
-            request.channel_meta = dict(record.channel_meta)
-            manager.enqueue(record.channel, request)
+            self._enqueue_parent_request(record, manager)
         except Exception:
             logger.exception(
                 "Failed to enqueue parent wakeup for %s",
                 record.task_id,
             )
+
+    @staticmethod
+    def _enqueue_parent_request(
+        record: SubagentTaskRecord,
+        manager: Any,
+    ) -> None:
+        from ...schemas import AgentRequest
+
+        request = AgentRequest(
+            input=[],
+            session_id=record.parent_session_id,
+            user_id=record.user_id or record.parent_session_id,
+            channel=record.channel,
+            root_session_id=record.root_session_id,
+            root_agent_id=record.parent_agent_id,
+            request_context={
+                "subagent_wakeup": True,
+                "subagent_event_task_id": record.task_id,
+            },
+        )
+        request.channel_meta = dict(record.channel_meta)
+        manager.enqueue(record.channel, request)
+
+    async def _parent_session_is_running(
+        self,
+        record: SubagentTaskRecord,
+    ) -> bool:
+        chat_manager = getattr(self.workspace, "chat_manager", None)
+        tracker = getattr(self.workspace, "task_tracker", None)
+        if chat_manager is None or tracker is None:
+            return False
+        chat_id = await chat_manager.get_chat_id_by_session(
+            record.parent_session_id,
+            record.channel,
+        )
+        if not chat_id:
+            return False
+        return await tracker.get_status(chat_id) == "running"
 
     @staticmethod
     def _build_child_request(
--- a/src/qwenpaw/app/subagents/manager.py
+++ b/src/qwenpaw/app/subagents/manager.py
@@ -1,0 +1,779 @@
+# -*- coding: utf-8 -*-
+"""Workspace-scoped background subagent task manager.
+
+The design follows AgentScope 2.0's background-tool lifecycle: keep the
+asyncio task handle locally, publish lifecycle events into a parent-session
+inbox, then enqueue an idle-session wakeup.  QwenPaw supplies its own inbox and
+wakeup bridge because it does not run the ``agentscope.app`` service stack.
+"""
+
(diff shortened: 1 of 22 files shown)


splash-li:
Rebased this PR onto the latest official `main`.

The Runtime 2.0 `spawn_subagent` registration and generic `/console/chat/task` implementation now come directly from the official #5660 implementation. I removed the duplicated #5524 commits, the legacy `_CHAT_TASKS` store, the duplicate background routes, and the obsolete tests for that store.

This PR now contains only the subagent-specific lifecycle layer on top: parent/child tracking, heartbeat watchdog, callback wakeup, cancellation propagation, and lifecycle tests.

Verification after the rebase:
- 122 passed, 1 skipped
- real-process E2E passed
- pre-commit passed

src/qwenpaw/app/subagents/manager.py
```diff
@@ -0,0 +1,779 @@
...
+            if chat_manager is not None:
+                chat_id = await chat_manager.get_chat_id_by_session(
+                    record.parent_session_id,
+                    record.channel,
+                )
+            while chat_id and tracker is not None:
+                if await tracker.get_status(chat_id) != "running":
+                    break
+                # The active parent may consume the HintBlock on its next
+                # reasoning step.  If so, no follow-up wake is necessary.
```
XiuShenAl, line 738: **Bug: unbounded spin loop — can leak notification tasks indefinitely**

This `while` loop has no upper bound. If the parent session stays in `"running"` state indefinitely (e.g. blocked on a long tool call, waiting for user approval, or encountering a tracker state anomaly), this coroutine spins at 10 Hz forever.

Each notification task (one per completed child) occupies a slot in `_notification_tasks` and holds a reference to `record`. When multiple children complete while the parent is stuck, you accumulate multiple unbounded loops — 864,000 iterations/task/day of pure overhead, eventually leading to resource exhaustion.

**Suggested fix:** Add a deadline. When it expires, log a warning and exit. The event remains in the inbox and will be delivered on the next user-initiated interaction or child-completion wakeup that finds the parent idle.

```python
deadline = time.monotonic() + 300  # 5 minutes
while chat_id and tracker is not None:
    if time.monotonic() > deadline:
        logger.warning(
            "Timed out waiting for parent session %s to become idle; "
            "event remains in inbox for later delivery",
            record.parent_session_id,
        )
        break
    if await tracker.get_status(chat_id) != "running":
        break
    if not await self.has_pending_events(record.parent_session_id):
        return
    await asyncio.sleep(0.1)
```

The timeout value could be made configurable (similar to `cancel_grace_seconds`), but it must exist.

src/qwenpaw/app/subagents/middleware.py
```diff
@@ -0,0 +1,125 @@
...
+            f"{self._parent_session_id}:"
+            f"{agent.state.reply_id}:subagent-events"
+        )
+        events = await self._manager.claim_events(
+            self._parent_session_id,
+            claim_id,
+        )
+        if events:
+            setattr(agent, "_qwenpaw_subagent_event_claim_id", claim_id)
+            hints = [self._to_hint(event) for event in events]
```
XiuShenAl, line 49: **Bug: claim_id overwrite causes permanently orphaned events in inbox**

In a ReAct loop, `on_reasoning` is invoked multiple times within a single agent execution (reasoning → acting → reasoning → ...). If events are claimed on iteration N and new events arrive and are claimed on iteration N+1 (with a different `claim_id`), this `setattr` overwrites the previous value.

The downstream hooks (`SubagentEventAckHook` at POST_RESPONSE and `SubagentContextCleanupHook` at FINALLY) only see the *latest* `claim_id` stored on the agent. Events claimed under a prior `claim_id` are then neither acknowledged nor released — they remain permanently locked in the inbox with a stale `claim_id` that nothing will ever reference again.

**Suggested fix:** Before overwriting, release (or ack) the previously held claim:

```python
if events:
    existing_claim = getattr(agent, "_qwenpaw_subagent_event_claim_id", None)
    if existing_claim:
        # Prior claim was not yet ack'd by session_save; release it so
        # those events become available again on this or a future run.
        await self._manager.release_events(existing_claim)
    setattr(agent, "_qwenpaw_subagent_event_claim_id", claim_id)
```

Alternatively, accumulate all claim IDs in a list and have both hooks iterate over the full list. The single-attribute approach is simpler but requires this guard.

src/qwenpaw/app/subagents/manager.py
```diff
@@ -0,0 +1,779 @@
...
+                now = time.monotonic()
+                tick_delay = now - previous_tick
+                previous_tick = now
+
+                # If this watchdog was delayed too, the whole event loop was
+                # paused. Give every child a fresh lease instead of killing
+                # healthy work after suspend, debugger pause, or starvation.
+                if tick_delay > self.watchdog_interval_seconds * 2:
+                    await self._reset_leases_after_loop_pause(now)
+                    continue
```
XiuShenAl, line 474: **Bug: entire watchdog terminates permanently on a single non-CancelledError exception**

The `except Exception` wraps the entire `while True` loop. Any unexpected error in a single tick (e.g. a transient `AttributeError` if a record is in an inconsistent state, or a future regression in `_watchdog_tick`) kills the watchdog permanently for the lifetime of the Workspace.

After that, no heartbeat monitoring occurs — stuck or dead children will never be detected, their concurrency slots will never be reclaimed, and their parents will never be notified. The failure is silent from the user's perspective (only a log line), and the system appears to function normally until orphaned children accumulate.

**Suggested fix:** Move the exception boundary inside the loop so individual tick failures are isolated and the watchdog self-heals:

```python
async def _watchdog_loop(self) -> None:
    previous_tick = time.monotonic()
    while True:
        try:
            await asyncio.sleep(self.watchdog_interval_seconds)
            now = time.monotonic()
            tick_delay = now - previous_tick
            previous_tick = now

            if tick_delay > self.watchdog_interval_seconds * 2:
                await self._reset_leases_after_loop_pause(now)
                continue

            stale_records = await self._watchdog_tick(now)
            for record in stale_records:
                self._schedule_parent_wakeup(record)
            raise
        except Exception:
            logger.exception(
                "Subagent watchdog tick failed; will retry next interval"
            )
```

This ensures the watchdog survives transient failures while still allowing clean shutdown via `CancelledError`.

XiuShenAl (changes requested):
Overall this is a well-designed PR with clear architectural layering (context/hooks/middleware/manager) and solid test coverage. The event-driven wakeup model is a significant improvement over the polling approach.

However, there are three correctness/reliability issues that should be addressed before merge. Each can manifest in normal production scenarios (long-running parent sessions, multi-step ReAct loops, and transient internal errors). Detailed inline comments follow.

src/qwenpaw/app/subagents/manager.py
splash-li: Thanks, fixed. The watchdog now catches exceptions per tick, so a transient tick failure is logged and the watchdog continues running.

src/qwenpaw/app/subagents/middleware.py
splash-li: Fixed. I used the alternative approach you suggested: the middleware now tracks all claim IDs in the current agent run, and the POST_RESPONSE / FINALLY hooks ack or release the full set.

src/qwenpaw/app/subagents/manager.py
splash-li: Thank you for your detailed review. Fixed in the latest commit. I bounded the initial wait and added a deduplicated low-frequency retry per parent session. The wakeup is not enqueued while the parent is still running, but it will be enqueued automatically once the parent becomes idle.

PR #5652, opened 2026-06-30
feat(crons): honor per-request model override

### LINKED ISSUE #5638: [Feature]: Support per-cron-job model override (`request.model` honored in cron executor)
## Problem

Currently, `jobs.json` supports a `request.model` field in the cron job spec (it's part of `AgentRequest`), but the cron executor ignores it entirely. The executor passes `job.request.model_dump()` to `stream_query()`, which creates an `AgentRequest(**req)`, but the runner's `query_handler` only ever reads `agent.model` (the agent's active_model), never `request.model`.

This means all cron jobs always use the agent's active_model, with no way to specify a different model per job or per schedule.

## Use Case

1. **Cost optimization**: Cron jobs run frequently (e.g., 170 times/day for monitoring tasks). These can use faster/cheaper models, while the main conversation uses a stronger model.
2. **Multi-model strategies**: Different cron tasks have different needs. Stock monitoring might need a code-capable model, while a simple ping check could use a tiny model.
3. **Queue separation**: Heavy cron traffic on one model can cause queuing delays for real-time conversations. Per-cron model selection allows isolating cron traffic to a different model/provider.

## Expected Behavior

When `request.model` is set in a cron job spec:

```json
{
  "request": {
    "input": [...],
    "model": "provider:model-name"
  }
}
```

The cron executor should honor this field, overriding the agent's active_model for that specific job execution.

## Implementation Hint

The code path is:
```
cron executor → job.request.model_dump() → stream_query(req) → AgentRequest(**req)
  → runner.query_handler(request) → agent.model  ← only reads active_model
```

The fix would be in `runner.query_handler()` to check `request.model` before falling back to `agent.model`:
```python
if request.model:
    # Use per-request model override (e.g., from cron)
    ...
else:
    # Use agent's active_model
    ...
```

## Priority

Medium. The workaround (changing agent active_model to a shared model) works but prevents model-level traffic isolation.

WujunhaoGPU:
## Summary
- Add explicit `model` fields to `AgentRequest` and cron job requests.
- Pass `request.model` from runtime agent builds into the shared model factory.
- Resolve request-level `provider:model` before agent/global active models, with clear errors for invalid providers or models.

Fixes #5638

## Why
Cron jobs could carry `request.model` in `jobs.json`, but the runtime still selected the agent active model or global active model when building the agent.

## User impact
A cron job can now set `request.model: "provider:model-name"` for that run only. Jobs without `request.model` keep the existing agent active model, then global active model fallback behavior.

## Verification
- `uv sync --python 3.11 --extra test`
- `uv run --python 3.11 pytest tests/unit/agents/test_model_factory_request_override.py tests/unit/agents/test_model_factory_message_normalization.py tests/unit/agents/test_cross_provider_normalization.py tests/unit/app/crons/test_models.py tests/unit/app/crons/test_manager.py tests/integration/test_cron_execution.py -q`
- `uv run --python 3.11 python -m compileall -q src/qwenpaw/schemas.py src/qwenpaw/app/crons/models.py src/qwenpaw/runtime/builder.py src/qwenpaw/agents/model_factory.py tests/unit/agents/test_model_factory_request_override.py tests/unit/app/crons/test_models.py`
- `git diff --check`

### DIFF
--- a/src/qwenpaw/agents/model_factory.py
+++ b/src/qwenpaw/agents/model_factory.py
@@ -38,7 +38,11 @@ except ImportError:
 from .utils.message_request_normalizer import (
     normalize_messages_for_model_request,
 )
-from ..exceptions import ProviderError, ModelFormatterError
+from ..exceptions import (
+    ModelNotFoundException,
+    ProviderError,
+    ModelFormatterError,
+)
 from ..providers import ProviderManager
 from ..providers.capping_formatter import MAX_INLINE_MEDIA_BYTES
 from ..providers.retry_chat_model import (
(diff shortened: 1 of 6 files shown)


xieyxclack:
Supported in https://github.com/agentscope-ai/QwenPaw/pull/5731

PR #5672, opened 2026-06-30
fix(scroll): strip ⟦…⟧ headline in HTTP history path, not just SSE, add mechanism to 

niceIrene:
## Summary

Two scroll/context fixes that surfaced as noisy logs and a UI glitch.

### 1. Stop-during-write no longer logs a false durability error

**Symptom:** Stopping a chat sometimes logged a `ERROR` with `sqlite3.ProgrammingError: Cannot operate on a closed database`, and flipped the history store to "durability degraded".

**Cause:** A teardown race. On stop, the agent closes the history store's SQLite connection, but a final in-flight `on_save` write-through can land *after* that close. It was caught — so nothing crashed — but it was misclassified as a disk outage rather than the expected close race.

**Fix:** `HistoryStore` now exposes a `closed` flag; the manager skips a write-through quietly when the store is already closed. Genuine disk failures still degrade durability and block eviction exactly as before.

### 2. `⟦…⟧` headline no longer reappears after switching tabs

**Symptom:** The hidden milestone headline showed up in the chat after switching to the inbox and back.

**Cause:** Two display paths, only one stripped the headline. The live **SSE stream** stripped it; the **HTTP `GET /chats/{id}`** history path (used when re-loading a chat) did not. Navigating away and back re-fetches over the HTTP path, so the headline leaked back into view.

**Fix:** Strip the headline (and the injected `<skill>` block) in the shared `agentscope_msg_to_message` converter via a new `clean_display_text` helper, so every display path is consistent.

## Testing

- `tests/unit/agents/context/test_scroll_manager.py` — write-through after close is a quiet no-op (not degraded).
- `tests/unit/app/chats/test_utils.py` — headline (incl. Qwen's `〚 〛` lookalike brackets) stripped on the history path; regression test through `agentscope_msg_to_message`.

### DIFF
--- a/src/qwenpaw/agents/context/scroll/history.py
+++ b/src/qwenpaw/agents/context/scroll/history.py
@@ -71,6 +71,9 @@ class HistoryStore:
         # degraded; callers/monitoring can read this.
         self.degraded = False
         self.write_failures = 0
+        # Flipped True by ``close()`` so callers can tell an intentional
+        # teardown race from a real disk outage (see ``closed``).
+        self._closed = False
         try:
             self._open_and_init()
         except sqlite3.DatabaseError as exc:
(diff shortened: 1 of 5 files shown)


cuiyuebing (approved):
LGTM

PR #5673, opened 2026-06-30
feat(tui): add live context-usage bar to the status bar

ekzhu:
## Description

Adds a live **context-usage bar** to the TUI status bar so it's clear at a glance how full the current model context is — and how close it is to auto-compaction.

Previously the status bar only showed *cumulative* session token counts (`tok ↑/↓`), which sum every API call's input and routinely run into the millions on a long session. That number is easy to mistake for context-window usage, and there was no indication of how full the *current* context was or when compaction would fire. This adds a separate, compact occupancy bar:

```
ctx [█░░░░░░░╵░] 123k / 1M   <- current tokens-in-context / window; the tick marks the compaction threshold
```

- Fixed 10-cell bar; fill colour ramps green -> amber -> red with raw occupancy.
- A single-cell tick marks the configured auto-compaction threshold (`compact_threshold_ratio`), so you can watch the fill approach it. Because that ratio is user-configurable, it is sourced from the agent config rather than hardcoded; the tick is hidden when compaction is disabled or the window/threshold is unknown.
- When auto-compaction fires and evicts history, the bar visibly shrinks.

**Plumbing:** reuses the existing per-call usage side channel. `TokenRecordingModelWrapper` now also carries the model's `context_size` and the agent's `compact_threshold_ratio`; these flow through the ACP server into the native `usage_update` notification (the threshold via `_meta`, since the schema has no field for it), then `normalize` -> `Usage` event -> status bar. The `usage_update` consumer path already existed in the TUI but was never emitted by the server; this completes that wiring.

**Related Issue:** N/A (no tracked issue)

**Security Considerations:** None — UI/telemetry display only; no auth, network, or config-write changes.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [x] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit` locally and it passes (on the changed files)
- [ ] Documentation updated (if needed) — N/A
- [x] Ready for review

## Testing

1. Run the TUI against any model/provider and send a few messages.
2. The status bar shows ` ctx [..bar..] <used> / <window>` reflecting the latest request's prompt size vs. the model context window.
3. With auto-compaction enabled, a tick appears at the configured `compact_threshold_ratio`; as context fills, the bar approaches it, and when compaction fires the bar shrinks.
4. Set `context_compact_config.enabled = false` (or use a model whose window is unknown) -> the tick is omitted.

Automated: new unit tests cover the render math (fill always 10 cells, colour ramp, clamp, min-fill), the marker (position per threshold, visibility through the lead-in zone, absorbed at/above threshold, hidden when unknown), and the threshold plumbing at each hop (wrapper -> usage meta -> `usage_update` `_meta` -> `Usage`).

```bash
pre-commit run --files <changed files>
# check python ast..Passed  add-trailing-comma..Passed  mypy..Passed
# black..Passed  flake8..Passed  pylint..Passed  (10.00/10)

pytest tests/unit/cli/tui tests/unit/token_usage tests/unit/agents/test_acp_available_commands.py
# 152 passed, 1 warning
```

Note: the repo-wide `pytest` has 4 failures in `tests/unit/agents/context/` that are pre-existing and unrelated to this change (they fail identically on `main`).

## Additional Notes

The `usage_update` ACP path and the `Usage` TUI event already existed but were never emitted by the server; this PR completes that intended wiring and adds the compaction-threshold marker on top of the occupancy bar.

### DIFF
--- a/src/qwenpaw/agents/acp/server.py
+++ b/src/qwenpaw/agents/acp/server.py
@@ -61,6 +61,7 @@ from acp.schema import (
     SseMcpServer,
     TextContentBlock,
     ToolCallUpdate,
+    UsageUpdate,
 )
 from qwenpaw.schemas import (
     AgentRequest,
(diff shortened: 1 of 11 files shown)


cuiyuebing (approved):
LGTM

PR #5686, opened 2026-07-01
fix(browser): dispose Playwright driver on cleanup

xiaoming-qxm:
## Description

This PR:
- Fixes Playwright driver process leaks in `browser_use` cleanup paths.
- Adds a unified async browser-state disposer that closes context/browser handles, stops the Playwright driver, applies bounded cleanup timeouts, and resets state only after disposal has been attempted.
- Replaces disconnect-time direct `_reset_browser_state()` calls with `_dispose_browser_state()` so CDP disconnects do not drop the last reference to a live Playwright driver.
- Makes `action="stop"` resilient when `playwright.stop()` hangs or raises, while still stopping QwenPaw-owned Chrome processes.
- Cleans up Playwright instances created during `action="start"` when later launch/context creation fails before the instance is stored in state.
- Keeps `_reset_browser_state()` as a pure state-reset helper instead of mixing async resource lifecycle into a synchronous function.

**Related Issue:** N/A

**Security Considerations:** No new user input surface, permission model, or persisted configuration is introduced. The fallback driver termination uses best-effort introspection of Playwright's private driver subprocess only during cleanup after graceful stop fails or times out.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Documentation updated (not needed)
- [x] Ready for review

## Testing

Local verification:

```bash
uv run python -m pytest tests/local/unit/agents/test_browser_control_lifecycle.py -v
# 3 passed

uv run black --check src/qwenpaw/agents/tools/browser_control.py tests/local/unit/agents/test_browser_control_lifecycle.py
# 2 files would be left unchanged

uv run flake8 src/qwenpaw/agents/tools/browser_control.py
# Passed

uv run python -m compileall -q src/qwenpaw/agents/tools/browser_control.py
# Passed

uv run python -c "from qwenpaw.agents.tools.browser_control import browser_use; print('OK')"
# OK

git diff --check
# Passed
```

Manual smoke test:

```bash
uv run python - <<'PY'
import asyncio
from qwenpaw.agents.tools.browser_control import browser_use

def text(chunk):
    return chunk.content[0].text

async def main():
    try:
        start = await browser_use(action="start")
        print("START", text(start))
    finally:
        stop = await browser_use(action="stop")
        print("STOP", text(stop))

asyncio.run(main())
PY
```

Result:

- `start` returned `ok: true`, `launch_mode: managed_cdp`, `owned_browser_process: true`, and a Chrome PID.
- `stop` returned `ok: true` and `browser_stopped: true`.
- A follow-up PID check found the Chrome PID no longer existed.

## Before/After Leak Experiment Report

Experiment setup:

- Before commit: `c49b247b`
- After commit: `157d2040` for the behavioral fix. Later commits only add review polish and pre-commit-required formatting/pylint fixes.
- Same experiment script was run against exported source trees for both commits with `PYTHONPATH` pointing at each tree.
- The real-driver scenario starts a real Playwright driver process, injects a disconnected CDP browser object, calls `_ensure_browser()`, and checks whether the driver process remains alive.

| Scenario | Before | After | Result |
| --- | --- | --- | --- |
| Real Playwright driver after lost CDP connection | Node driver still alive after `_ensure_browser()`, PID `50447`, RSS `138512 KiB` | Driver exited after `_ensure_browser()`, PID `50759`, returncode `0` | Fix removes the residual driver process; about 135 MiB of driver RSS no longer remains per lost connection |
| Fake CDP disconnect cleanup | `context_closed=false`, `playwright_stopped=false`, state reset | `context_closed=true`, `playwright_stopped=true`, state reset | Resources are disposed before state references are cleared |
| Hung `playwright.stop()` during owned CDP stop | `_action_stop()` timed out after `0.501s`; owned browser process was not signalled | `_action_stop()` returned in `0.108s`; owned browser process was signalled; response includes cleanup warning | Hung driver stop no longer blocks owned browser cleanup |
| Start failure after `async_playwright().start()` | Start returns error, but `playwright_stopped=false` | Start returns error and `playwright_stopped=true` | Driver is cleaned up when launch fails after driver creation |

Final residual check:

```bash
uv run python - <<'PY'
import psutil
for pid in (50447, 50759):
    print(pid, psutil.pid_exists(pid))
PY
# 50447 False
# 50759 False
```

```bash
pre-commit run --all-files
# Passed

pytest
# Full pytest not run. Targeted lifecycle regression suite passed:
# tests/local/unit/agents/test_browser_control_lifecycle.py: 3 passed

GitHub checks on PR head `fe9eafe9`:
- `Pre-commit Checks / run`: pass
- `NPM Format (website & console) / console format check`: pass
- `NPM Format (website & console) / website format check`: pass
- `PR Spam Gate / Check Open PR Count`: pass
- `Maintainer Approval`: pending human approval gate
```

## Additional Notes

- `tests/local` is intentionally ignored per repository policy for non-integration local regression tests, so the lifecycle test file is not included in this PR.
- While exploring broader checks, `tests/contract/runner/test_control_command_contract.py` failed during collection because `qwenpaw.app.runner.control_commands.base` is missing on the current branch. That appears unrelated to this browser cleanup change and was not used as validation evidence.

### DIFF
--- a/src/qwenpaw/agents/tools/browser_control.py
+++ b/src/qwenpaw/agents/tools/browser_control.py
@@ -22,11 +22,13 @@ import signal
 import socket
 import subprocess
 import sys
+import threading
 import time
 from typing import Any, Optional
 from urllib.parse import urljoin
 from urllib import request as urllib_request
 
+import psutil
 from agentscope.message import TextBlock
 from agentscope.tool import ToolChunk
 from agentscope.message import ToolResultState
@@ -47,6 +49,7 @@ logger = logging.getLogger(__name__)
 
 _MAX_DIRECT_URL_DOWNLOAD_BYTES = 10 * 1024 * 1024
 _CDP_CONNECT_TIMEOUT_SECONDS = 30.0
+_BROWSER_CLEANUP_TIMEOUT_SECONDS = 5.0
 _HEADLESS_VERIFICATION_WARNING = (
     "Headless browser launches are more likely to trigger verification. "
     "If verification appears, call browser_use with action='stop' to stop "
@@ -312,8 +315,21 @@ def _is_browser_running(state: dict) -> bool:
         return (
             state.get("_sync_context") is not None
             or state.get("_sync_browser") is not None
+            or state.get("_sync_playwright") is not None
         )
-    return state.get("browser") is not None or state.get("context") is not None
+    if (
+        state.get("browser") is not None
+        or state.get("context") is not None
+        or state.get("playwright") is not None
+    ):
+        return True
+    proc = state.get("browser_process")
+    if proc is None:
+        return False
+    try:
+        return proc.poll() is None
+    except Exception:
+        return True
 
 
 def _reset_browser_state(state: dict) -> None:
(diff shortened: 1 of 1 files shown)


xiaoming-qxm:
## Local Fault-Injection Validation: Playwright Driver Stop Timeout

I also validated the fix locally with a deterministic fault-injection scenario that simulates `playwright.stop()` getting stuck while waiting for the Playwright driver process to exit.

For this local validation only, I temporarily changed the browser idle timeout from the production default of `600s` to `60s` so the idle watchdog path could be exercised quickly. The actual default timeout in the PR remains unchanged.

Test setup:

- Branch: `fix/playwright-memory-issue`
- Playwright: `1.60.0`
- Browser cleanup path: idle watchdog -> `_action_stop()` -> `_dispose_browser_state()` -> `_stop_playwright_instance()`
- Fault injection: after starting a managed CDP browser, I paused the Playwright driver process with `kill -STOP <driver_pid>`.

Observed timeline:

```text
T+60s
idle watchdog detects that the browser has been idle for 60 seconds
  -> calls _action_stop()

T+60s ~ T+60s+?
_action_stop()
  -> _dispose_browser_state()
    -> closes browser context/browser resources
    -> stops/kills the owned Chromium process
    -> calls _stop_playwright_instance(playwright)

T+60s+5s
_stop_playwright_instance()
  -> await asyncio.wait_for(playwright.stop(), timeout=5s)
  -> because the Playwright driver was paused with SIGSTOP, Playwright stop blocks at:
       await self._transport.wait_until_stopped()
  -> asyncio.wait_for reaches the cleanup timeout and cancels the stop coroutine
  -> the log shows CancelledError inside Playwright, followed by TimeoutError from wait_for
  -> QwenPaw records:
       Playwright driver stop timed out after 5.0s

At this moment
  -> only the graceful Playwright stop has timed out
  -> fallback process cleanup has not necessarily completed yet
  -> seeing the driver process still present in `ps` immediately after this warning is expected

Then
  -> _force_stop_playwright_driver()
  -> resolves the driver subprocess through pw._impl_obj._connection._transport._proc
  -> _terminate_process()
       first sends terminate / SIGTERM
       waits up to the cleanup timeout
       if the process is still alive, sends kill / SIGKILL
  -> the Playwright driver process disappears

Finally
  -> _reset_browser_state()
  -> clears state["playwright"] and the rest of the browser state
```

This confirms the intended behavior of the fix: when graceful `playwright.stop()` hangs, cleanup does not wait indefinitely and does not simply drop the Python reference. The code logs the timeout, falls back to process-level termination for the Playwright driver, and only resets the browser state after the cleanup path has run.

One important detail from the observation: the driver can still be visible via `ps` at the exact moment the timeout warning is emitted. That is expected because the warning is logged at the transition point from graceful stop to forced cleanup. The success condition is that the driver process disappears shortly afterward, once the fallback terminate/kill path completes.

x1n95c:
Hello, I tested the changes manually: when I stop the Playwright driver process (with kill -STOP)  first and then call `browser_use(action="stop")`, the new code still shuts down the browser successfully, whereas the old code would hang and eventually time out. I this this is one of the potential causes of the Playwright process leak we've been seeing. The only question is `_sync_browser_close` is now dead code since _dispose_browser_state has fully replaced it, we should remove it.

xiaoming-qxm:
> Hello, I tested the changes manually: when I stop the Playwright driver process (with kill -STOP) first and then call `browser_use(action="stop")`, the new code still shuts down the browser successfully, whereas the old code would hang and eventually time out. I this this is one of the potential causes of the Playwright process leak we've been seeing. The only question is `_sync_browser_close` is now dead code since _dispose_browser_state has fully replaced it, we should remove it.

Thanks for checking this.  `_sync_browser_close` has already been removed in the latest version of this branch, and all cleanup paths now go through `_dispose_browser_state()`.

cuiyuebing (approved):
LGTM

PR #5669, opened 2026-06-30
feat(memory): add qwen3-rerank to memory search

iluv7:
## Summary

Wrap ReMe's hybrid search results with DashScope `qwen3-rerank` for precision re-ranking in `ReMeLightMemoryManager.memory_search()`. No changes to ReMe itself — just an outer rerank stage controlled by a `rerank_enabled` toggle (default off).

**Related Issue:** Closes #5588

### Backend: Rerank Integration (`src/qwenpaw/agents/memory/reme_light_memory_manager.py`)

- **`_build_search_answer(candidates)`** — formats re-ranked candidates back into ReMe-standard answer text, appending `rerank=` scores
- **`_rerank_dashscope(query, candidates, top_n)`** — calls DashScope `TextReRank.call()` with full error handling; falls back to original ordering on any failure
- **`memory_search()`** — rerank off (default): behavior unchanged. rerank on: fetches `max_results * 3` candidates from ReMe, reranks, returns top `max_results`
- **`auto_memory_search()`** — same rerank pipeline applied

### Backend: Configuration (`src/qwenpaw/config/config.py`)

New fields in `ReMeLightMemoryConfig`:

| Field | Type | Default |
|-------|------|---------|
| `rerank_enabled` | `bool` | `False` |
| `rerank_model` | `str` | `"qwen3-rerank"` |
| `dashscope_api_key` | `str` | `""` |

### Flow

```
ReMe hybrid search (limit = max_results * 3)
  → response.metadata["results"] (structured candidates)
  → dashscope.TextReRank.call(qwen3-rerank)
  → sort by relevance_score desc
  → _build_search_answer (ReMe format + rerank score)
  → _tool_chunk → LLM
```

### Degradation

Every error path falls back to original ordering: dashscope not installed, API key not configured, API call failure, empty results.

## Usage

```json
{
  "running": {
    "reme_light_memory_config": {
      "rerank_enabled": true,
      "rerank_model": "qwen3-rerank",
      "dashscope_api_key": "sk-xxx"
    }
  }
}
```

## Tests

`tests/unit/agents/memory/test_rerank.py` — 10 core tests, all passing:

### Formatting

| Test | What it verifies |
|------|------------------|
| `test_formats_candidates_in_reme_style` | `_build_search_answer` output matches ReMe format |

### Config Reading

| Test | What it verifies |
|------|------------------|
| `test_load_agent_config_end_to_end` | Real files → real `load_agent_config` → all 3 fields deserialize correctly |
[...]

Issue #5588, opened 2026-06-28
[Feature]: 记忆搜索支持专用 Reranker 模型实现两阶段检索


cjc0045:
## 背景

QwenPaw 当前的 `memory_search` 仅使用 embedding 模型做**单阶段检索**：
```
用户查询 → embedding 向量化 → Chroma 相似度检索 → 返回 Top-N
```

这种方式在记忆库较小时工作良好，但随着记忆积累，仅靠 embedding 相似度的召回精度会下降。

## 问题

1. **无精排环节**：embedding 粗筛后直接返回，缺少对候选结果的二次精细排序
2. **reME 的 LLM-based rerank 未启用**：reME service 模式有 `enable_llm_rerank`（调用 LLM 文本重排），但 QwenPaw 用的 `remelight` 模式不支持
3. **无法利用本地专用 Reranker**：Ollama 0.30.x 已原生支持 `/v1/rerank` 端点，本地有 `dengcao/Qwen3-Reranker-4B` 等专用模型可用但无法接入

### 两阶段检索流程

```
用户查询 → embedding 模型向量化 → Chroma 粗筛 Top-20
                                               ↓
                               Reranker 模型精排 → 返回 Top-5
```

### agent.json 配置示例

```json
{
  "reme_light_memory_config": {
    "reranker_model_config": {
      "enabled": true,
      "backend": "openai",
      "api_key": "ollama",
      "base_url": "http://localhost:11434/v1",
      "model_name": "dengcao/Qwen3-Reranker-4B",
      "top_n": 5,
      "candidate_multiplier": 4
    }
  }
}
```

- `backend`：兼容 OpenAI-compatible 的 `/v1/rerank` 端点（Ollama、vLLM 等均支持）
- `candidate_multiplier`：embedding 阶段取 `top_n * candidate_multiplier` 条候选，再交给 reranker 精排

### 技术要点

- Ollama 的 `/v1/rerank` API 格式：
```bash
POST /v1/rerank
{
  "model": "dengcao/Qwen3-Reranker-4B",
  "query": "用户查询",
  "documents": ["候选文档1", "候选文档2", ...],
  "top_n": 5
}
```
- Reranker 模型通常很小（0.6B~4B），显存开销远小于 embedding 模型
- 应与 embedding 模型共用 `OLLAMA_KEEP_ALIVE` 配置

## 收益

- **检索精度提升**：embedding 粗筛 + reranker 精排是业界 RAG 标准实践
[...]

PR #5692, opened 2026-07-01
feat(memory): add reranker for search results on reme0.4

lecheng2018:
## Description

Add standard reranker support for memory search on top of reme0.4.

reme0.4 provides the base `memory_search` pipeline with hybrid retrieval (BM25 + vector). This PR adds a post-retrieval reranking stage: after the initial top-K results are fetched, a dedicated reranker API re-orders them by relevance using the standard `/rerank` endpoint (SiliconFlow-compatible, `relevance_score` response). This improves result precision at the cost of one extra API call per search.

**Related Issue:** (none)

**Security Considerations:** API key stored in agent config via existing `RerankerConfig` model. No new external endpoints or auth surface introduced.

## Type of Change

- [x] New feature

## Component(s) Affected

## Checklist

- [x] Ready for review

## Testing

1. Enable in Web UI: ReMeLight config → Search Result Reranker → Enable
2. Fill model_name (e.g. `bge-reranker-v2-m3`), base_url (e.g. `https://api.siliconflow.cn/v1`), api_key
3. Save, restart, perform a memory search — results will be reordered by the reranker
4. Check logs for `[rerank] reordered N results with model=...`

Without config, behavior is unchanged (reranker disabled by default). reme0.4 base search pipeline is untouched.

bash

pre-commit run --all-files

black --line-length=79 → OK

add-trailing-comma → OK

flake8 --extend-ignore=E203 --max-line-length=79 → OK

2 files changed, 212 insertions(+), 4 deletions(-)

python3 -c "

import ast

ast.parse(open('src/qwenpaw/config/config.py').read())

ast.parse(open('src/qwenpaw/agents/memory/reme_light_memory_manager.py').read())

print('syntax OK')

## Additional Notes

- Built on reme0.4's `memory_search` pipeline — reranker is a pure post-processing step
- Compatible with any provider supporting standard `/rerank` endpoint with `[{"index": N, "relevance_score": 0.99}, ...]` response
- Graceful fallback: if reranker call fails, original reme0.4 result order is preserved
- Config schema: `RerankerConfig(enabled, api_key, base_url, model_name, temperature)`


jinliyl:
Thanks for working on this. I like the provider-agnostic direction here: a generic reranker config with `base_url`, `api_key`, and `model_name` is a better long-term fit than binding memory search to a single vendor.

Before this is ready to merge, I think a few issues need to be addressed:

1. Please move the rerank implementation into a separate helper/module instead of embedding the HTTP/API details directly in `reme_light_memory_manager.py`. Ideally the ReMe memory code should only need a few simple lines such as: load config, call the rerank helper with `query` + candidates, then format/return the result. This keeps the ReMe integration small and makes the reranker easier to test and extend.
2. Please add tests. This changes memory search behavior and currently has no unit coverage for config parsing, successful rerank, fallback behavior, or response formatting.
3. `auto_memory_search()` should use the same rerank path as `memory_search()`, otherwise automatic memory injection and explicit memory search can return differently ordered results.
4. The search should fetch a larger candidate pool when rerank is enabled, for example `max_results * 3`, then rerank and return the final `max_results`. Re-ranking only the already-truncated top K limits the benefit.
5. The `/rerank` request should include `top_n` if the target API supports it, and the code should not require the response result count to exactly match the input document count. Many rerank APIs return only the requested top N results.
6. It would be useful to preserve or expose the rerank score in the rebuilt answer, similar to `rerank=...`, so we can debug and understand the final order.
7. `temperature` is currently part of `RerankerConfig` but is not used in the request. Please either wire it if the API expects it or remove it from this config.
8. `_refresh_config_cache()` is added but I do not see it being called. Either integrate the cache invalidation path or avoid caching the config for now.

You may also want to compare with #5669. That PR is DashScope-specific, but it has useful pieces we should probably keep: larger candidate fetch, `auto_memory_search()` integration, answer formatting, and tests.

lecheng2018:
> 感谢您为此付出的努力。我喜欢这种与提供商无关的方向：一个通用的重排序器配置（包含 `<name>` `base_url`、` `api_key`<name>` 和 `<name>`）`model_name`比将内存搜索绑定到单一供应商更符合长远发展趋势。
> 
> 在合并之前，我认为需要解决几个问题：
> 
> 1. 请将重排序的实现移到一个单独的辅助函数/模块中，而不是直接将 HTTP/API 细节嵌入到代码中`reme_light_memory_manager.py`。理想情况下，ReMe 内存代码应该只需要几行简单的代码，例如：加载配置、调用带有候选值的重排序辅助函数`query`，然后格式化并返回结果。这样可以保持 ReMe 集成的简洁性，并使重排序器更易于测试和扩展。
> 2. 请添加测试用例。此更改会改变内存搜索行为，目前尚无针对配置解析、成功重排序、回退行为或响应格式化的单元测试。
> 3. `auto_memory_search()`应该使用相同的重排序路径`memory_search()`，否则自动内存注入和显式内存搜索可能会返回不同排序的结果。
> 4. 启用重排序后，搜索应该获取更大的候选池，例如`max_results * 3`，然后重新排序并返回最终结果`max_results`。仅对已截断的前 K 个候选结果进行重新排序会限制其优势。
> 5. 请求`/rerank`中应包含`top_n`目标 API 是否支持该功能的信息，并且代码不应要求响应结果的数量与输入文档的数量完全匹配。许多重新排序 API 只返回请求的前 N ​​个结果。
> 6. 如果能像 一样在重建的答案中保留或显示重排序分数，那就很有用了`rerank=...`，这样我们就可以调试并了解最终的排序结果。
> 7. `temperature`目前该参数已包含在配置中`RerankerConfig`，但未在请求中使用。请根据 API 的需要添加该参数，或者将其从配置中移除。
> 8. `_refresh_config_cache()`已添加，但我没看到它被调用。要么集成缓存失效路径，要么暂时不要缓存配置。
> 
> 您可能还想与#5669进行比较。该 PR 是专门针对 DashScope 的，但其中包含一些我们应该保留的有用部分：更大的候选对象获取范围、`auto_memory_search()`集成、答案格式化和测试。

Hi @jinliyl, thank you for the thorough review. I've addressed all your feedback in the latest push:

1. Extracted rerank into a separate module — src/qwenpaw/agents/memory/reranker.py now contains rerank() and build_search_answer(). The memory manager only loads config, calls the helper, and formats results.

2. Added tests — tests/unit/agents/memory/test_rerank.py covers: answer formatting, reordering logic, top_n in request, partial response tolerance, timeout/error fallback, and integration with both memory_search() and auto_memory_search().

3. auto_memory_search() uses the same rerank path — both methods now share _load_reranker_config() + rerank() + build_search_answer().

4. Larger candidate pool — when rerank is enabled, fetch max_results * 3 from ReMe, rerank, then return the top max_results.

5. top_n in /rerank request + flexible response handling — top_n is sent to the API. The code tolerates responses with fewer items than input docs.

6. Rerank score preserved in answer — exposed as rerank=... in the formatted output header, alongside existing score/vector/keyword.

7. Removed temperature — dropped from RerankerConfig since no current API uses it.

8. Removed _refresh_config_cache() — replaced with on-demand config loading to avoid stale cache issues.

I also took inspiration from #5669 for the answer format (with rerank=) and the auto_memory_search() integration pattern. Appreciate the suggestion to compare with that PR.

Let me know if there's anything else you'd like adjusted.

iluv7:
@lecheng2018 
Nice approach — one thing to watch out for: different providers use different rerank paths (DashScope is `/reranks`, not `/rerank`), so hardcoding `f"{base_url}/rerank"` won't work across all providers.

I've handled this in #5588 by letting users configure the full URL, and all tests pass with both DashScope and SiliconFlow keys.

lecheng2018:
> @lecheng2018 Nice approach — one thing to watch out for: different providers use different rerank paths (DashScope is `/reranks`, not `/rerank`), so hardcoding `f"{base_url}/rerank"` won't work across all providers.
> 
> I've handled this in #5588 by letting users configure the full URL, and all tests pass with both DashScope and SiliconFlow keys.

Thanks for pointing out this compatibility issue and providing the complete solution in #5588! I’ve followed your approach to refactor the 4 related files: rename base_url to endpoint_url, remove hardcoded /rerank path splicing, and let users configure the full request URL directly. All unit tests for both DashScope and SiliconFlow have been updated and passed successfully.

iluv7:
@lecheng2018 
Thanks for the kind words! Since #5588 already covers the same ground with tests across two providers, might make sense to review that one directly to avoid duplicate effort. Happy to incorporate any feedback you have there.

lecheng2018:
> @lecheng2018 Thanks for the kind words! Since #5588 already covers the same ground with tests across two providers, might make sense to review that one directly to avoid duplicate effort. Happy to incorporate any feedback you have there.

Got it, thanks for the heads up! My implementation includes matching front-end and back-end supporting logic, which differs from the scope of #5588, so this PR addresses a separate set of requirements. I’d appreciate your review here.

jinliyl:
Thanks a lot for this PR — the overall approach is solid, and we appreciate that reranking is opt-in and the legacy path is preserved byte-for-byte when disabled. The nested `RerankerConfig` model is also nicely aligned with the existing `EmbeddingModelConfig` / `AutoMemorySearchConfig` style in the codebase, which we like.

Before we can merge, here are the points we'd like to see addressed:

### 1. A unit test is currently failing (blocks CI)
`tests/unit/agents/memory/test_rerank.py::TestManagerRerankIntegration::test_auto_memory_search_uses_rerank` fails with `assert None is not None`.

Root cause is in the test itself: it creates a local `auto_cfg = MagicMock()` with `auto_cfg.enabled = True`, but that mock is never wired into the config returned by the patched `load_agent_config`. Meanwhile the fixture sets `reme_cfg.auto_memory_search_config.enabled = False`, so `auto_memory_search()` returns early at the `if not memory_cfg.auto_memory_search_config.enabled: return None` guard. Please set `auto_memory_search_config.enabled = True` on the mocked config for that test (and drop the unused `auto_cfg`).

### 2. `build_search_answer` uses hard key access and runs outside the try/except
In `reranker.py`, `build_search_answer` accesses `c['path']`, `c['start_line']`, `c['end_line']`, `c['text']` directly. In `memory_search()` / `auto_memory_search()` the `try/except` only wraps the `rerank()` call — `build_search_answer(...)` is called outside it. If ReMe's `metadata["results"]` items are missing any of those keys, this raises `KeyError` and propagates out of the search, breaking the request entirely (instead of gracefully degrading). Could you either use `.get()` with sensible fallbacks in `build_search_answer`, or bring the formatting step inside the fallback-protected block?

### 3. Please confirm the ReMe `metadata["results"]` contract
The formatting relies on each result having `path` / `start_line` / `end_line` / `text` / `scores`. Since this is the load-bearing assumption of the whole feature, it would be great to confirm against reme0.4's actual `memory_search` output (ideally with a note or a small integration test), so we don't hit surprises once rerank is switched on.

### 4. Minor: redundant config load on the auto path
`_load_reranker_config()` calls `load_agent_config()` again, but `auto_memory_search()` already loads `memory_cfg`. Reusing the already-loaded config there would avoid a second read on every auto-search.

Once the failing test and the `KeyError` guard (items 1 & 2) are resolved, this should be in good shape. Thanks again!

lecheng2018:
> Thanks a lot for this PR — the overall approach is solid, and we appreciate that reranking is opt-in and the legacy path is preserved byte-for-byte when disabled. The nested `RerankerConfig` model is also nicely aligned with the existing `EmbeddingModelConfig` / `AutoMemorySearchConfig` style in the codebase, which we like.
> 
> Before we can merge, here are the points we'd like to see addressed:
> 
> ### 1. A unit test is currently failing (blocks CI)
> `tests/unit/agents/memory/test_rerank.py::TestManagerRerankIntegration::test_auto_memory_search_uses_rerank` fails with `assert None is not None`.
> 
> Root cause is in the test itself: it creates a local `auto_cfg = MagicMock()` with `auto_cfg.enabled = True`, but that mock is never wired into the config returned by the patched `load_agent_config`. Meanwhile the fixture sets `reme_cfg.auto_memory_search_config.enabled = False`, so `auto_memory_search()` returns early at the `if not memory_cfg.auto_memory_search_config.enabled: return None` guard. Please set `auto_memory_search_config.enabled = True` on the mocked config for that test (and drop the unused `auto_cfg`).
> 
> ### 2. `build_search_answer` uses hard key access and runs outside the try/except
> In `reranker.py`, `build_search_answer` accesses `c['path']`, `c['start_line']`, `c['end_line']`, `c['text']` directly. In `memory_search()` / `auto_memory_search()` the `try/except` only wraps the `rerank()` call — `build_search_answer(...)` is called outside it. If ReMe's `metadata["results"]` items are missing any of those keys, this raises `KeyError` and propagates out of the search, breaking the request entirely (instead of gracefully degrading). Could you either use `.get()` with sensible fallbacks in `build_search_answer`, or bring the formatting step inside the fallback-protected block?
> 
> ### 3. Please confirm the ReMe `metadata["results"]` contract
> The formatting relies on each result having `path` / `start_line` / `end_line` / `text` / `scores`. Since this is the load-bearing assumption of the whole feature, it would be great to confirm against reme0.4's actual `memory_search` output (ideally with a note or a small integration test), so we don't hit surprises once rerank is switched on.
> 
> ### 4. Minor: redundant config load on the auto path
> `_load_reranker_config()` calls `load_agent_config()` again, but `auto_memory_search()` already loads `memory_cfg`. Reusing the already-loaded config there would avoid a second read on every auto-search.
> 
> Once the failing test and the `KeyError` guard (items 1 & 2) are resolved, this should be in good shape. Thanks again!

Thanks for the thorough review! Here's what we changed in response:

**1. Failing test fixed** — The fixture was setting `reme_cfg.auto_memory_search_config.enabled = False` which, through MagicMock's attribute caching, caused `auto_memory_search()` to return early. Changed to properly wire a dedicated `auto_cfg` object via `reme_cfg.auto_memory_search_config = auto_cfg`. Also updated the rerank mock's `return_value` to match the test's overridden `metadata["results"]`, so the content assertion also passes.

**2. `build_search_answer` hardened** — Switched from hard dict access (`c['path']`, `c['start_line']`, `c['end_line']`, `c['text']`) to `.get()` with sensible fallbacks (`"?"` for path/line numbers, `""` for text). Missing keys produce a degraded entry rather than a `KeyError` crash.

**3. ReMe contract documented** — Added a note in `build_search_answer`'s docstring specifying the expected fields: `path`, `start_line`, `end_line`, `text`, and optional `scores`.

**4. Redundant config load removed** — `auto_memory_search()` now reuses the already-loaded `memory_cfg.reranker_config` instead of calling `_load_reranker_config()` again. This also fixed a latent bug where `limit=max_results` was passed to `_run_reme_job` instead of `limit=fetch_limit`.

Additionally, we added the missing `RerankerConfig` import to resolve a pre-commit `F821` error.

jinliyl:
Thanks for the update. CI is still failing because `test_auto_memory_search_uses_rerank` patches `reme_light_memory_manager.uuid`, but that module no longer has a `uuid` attribute. Please remove this obsolete patch or update it to patch the symbol currently used, then rerun the tests. Thanks!

lecheng2018:
> Thanks for the update. CI is still failing because `test_auto_memory_search_uses_rerank` patches `reme_light_memory_manager.uuid`, but that module no longer has a `uuid` attribute. Please remove this obsolete patch or update it to patch the symbol currently used, then rerun the tests. Thanks!

This PR has been reimplemented via [https://github.com/agentscope-ai/QwenPaw/pull/6398](https://link.wtturl.cn/?target=https%3A%2F%2Fgithub.com%2Fagentscope-ai%2FQwenPaw%2Fpull%2F6398&scene=im&aid=582478&lang=zh)

lecheng2018:
> Thanks for the update. CI is still failing because `test_auto_memory_search_uses_rerank` patches `reme_light_memory_manager.uuid`, but that module no longer has a `uuid` attribute. Please remove this obsolete patch or update it to patch the symbol currently used, then rerun the tests. Thanks!

Hi @jinliyl, just a quick note — PR #6398 is the redesigned backend that supersedes #5692 (which addressed all your previous review feedback but needed a clean restart). The old #5692 can be closed once #6398 is reviewed. Happy to close it now if that helps.

PR #5617, opened 2026-06-29
feat(channels): add per-channel no_text_debounce toggle

hongxicheng:
## Summary

Add a per-channel toggle (`no_text_debounce`, default `true`) to control
the no-text debounce feature. When enabled, media-only messages are
buffered until a text message arrives; when disabled, all messages are
processed immediately.

## Changes

- **Config**: Added `no_text_debounce: bool = True` to `BaseChannelConfig`
- **BaseChannel**: Added parameter to `__init__` and `from_config`;
  short-circuit in `_apply_no_text_debounce` when disabled
- **Manager**: Centrally read from config (dict/pydantic) and pass via
  `from_config_kwargs`, following the `filter_tool_messages` pattern
- **18 channels**: Parameter declaration + passthrough in `__init__`
  and `from_config`
- **Telegram/OneBot**: Removed `_apply_no_text_debounce` overrides,
  unified to base class behavior
- **Frontend**: Switch toggle in channel settings; type definition;
  i18n for 6 locales (zh/en/ja/vi/ru/pt-BR)
- **Tests**: Added 2 tests for `no_text_debounce=False`; removed
  obsolete `TestTelegramNoTextDebounce` class

## Behavior Change

Removing Telegram/OneBot overrides means their media-only messages are
now buffered by default (base class behavior) instead of being processed
immediately. Users can disable debounce per-channel via the new toggle.

**Related Issue:** Fixes #5554 

**Security Considerations:** [If applicable, e.g. channel auth, env/config handling]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [ ] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Feishu, QQ, Console, etc.)

[...]

PR #5693, opened 2026-07-01
fix: file-only messages bypass no-text debounce

mynameyi:
## Summary

File-only messages (e.g. Excel files sent via WeChat Work) were being silently dropped because they were caught by the no-text debounce logic and never processed when no text message followed.

## Problem

`_apply_no_text_debounce()` in `base.py` only had a bypass for audio-only messages. File-only messages (no text, no audio) were buffered in `_pending_content_by_session` and waited indefinitely for a text message that never came.

This means: **sending a file via WeChat Work (or any channel) with no accompanying text would produce zero response**.

## Root Cause

```python
def _apply_no_text_debounce(self, session_id, content_parts):
    if not self._content_has_text(content_parts):
        if self._content_has_audio(content_parts):
            # Audio bypasses debounce ✓
            return (True, merged)
        # File-only: buffered forever ✗
        self._pending_content_by_session[...].extend(content_parts)
        return (False, [])
```

## Fix

Add `_content_has_file()` check alongside the existing audio check, so file-only messages are processed immediately:

```python
if self._content_has_file(content_parts):
    pending = self._pending_content_by_session.pop(session_id, [])
    merged = pending + list(content_parts)
    return (True, merged)
```

## Reproduction

1. Send a file (e.g. Excel) via WeChat Work to any agent
2. Do not send any text message after
3. Agent receives the file but produces no response

## Impact

- Fixes file uploads via WeChat Work, DingTalk, Telegram, and all channels using the no-text debounce
- No change for text messages or mixed content (text + file)
- Images are handled separately via time-debounce

### DIFF
--- a/src/qwenpaw/app/channels/base.py
+++ b/src/qwenpaw/app/channels/base.py
@@ -297,6 +297,13 @@ class BaseChannel(ABC):
             for c in (contents or [])
         )
 
+    def _content_has_file(self, contents: List[Any]) -> bool:
+        """True if contents has at least one FILE block."""
+        return any(
+            getattr(c, "type", None) == ContentType.FILE
+            for c in (contents or [])
+        )
+
     def _apply_no_text_debounce(
         self,
         session_id: str,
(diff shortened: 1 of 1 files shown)


hongxicheng:
Thanks for the PR, @mynameyi ! 

This functionality has already been implemented in #5617 . Closing this as a duplicate.

PR #5716, opened 2026-07-02
feat(channels): add generic webhook channel

### LINKED ISSUE #338: [Feature]: 建议添加webhook功能
可以通过我的软件发信息给copaw ，copaw 回答后调用我接口
或者我自己去论查 最好。发过去后返回一个key 我用这个key 去查 有没有信息

AnandSundar:
## Summary

Adds a built-in **Webhook** channel so QwenPaw can receive arbitrary JSON over HTTP and POST agent replies back to a configurable URL, with HMAC-SHA256 signing on both directions. Closes #338.

This is a generic, low-friction integration point for any system that can POST JSON (internal services, home automation, custom UIs, CI bots) — without writing a custom channel plugin.

## What ships

- **Inbound**: `POST /webhooks/<channel_id>` on `bind_address:port` (default `127.0.0.1:9070`)
  - Optional `X-QwenPaw-Signature: sha256=<hex>` verification (HMAC-SHA256 of the raw body)
  - Rejects bodies > 1 MiB (`413`), bad JSON (`400`), wrong channel id (`404`), bad signature (`401`)
- **Outbound**: agent replies POSTed to `outbound_url` (or a per-reply `to_handle`)
  - Same `X-QwenPaw-Signature` header on outbound (verified by the receiving service)
  - 2xx → success; 4xx → no retry; 5xx / network error → retry with exp backoff (1s/2s/4s)
- **Lifecycle**: runs its own uvicorn server in a background task; respects `enabled` flag
- **Session id**: defaults to `webhook:<channel_id>`; can be overridden per request via `meta.session_id`

## Files

```
src/qwenpaw/app/channels/webhook/__init__.py   # public exports
src/qwenpaw/app/channels/webhook/channel.py    # WebhookChannel + config + GenericWebhookEvent
src/qwenpaw/app/channels/webhook/server.py     # FastAPI app + uvicorn lifecycle
src/qwenpaw/app/channels/webhook/sender.py     # signed POST + retry
src/qwenpaw/app/channels/webhook/signature.py  # HMAC-SHA256 verify
src/qwenpaw/app/channels/registry.py           # register webhook in _BUILTIN_SPECS
src/qwenpaw/config/config.py                   # WebhookConfig schema
src/qwenpaw/cli/channels_cmd.py                # configurator + display name
tests/unit/channels/test_webhook.py            # 43 unit tests (signature, sender, channel, server)
tests/contract/channels/test_webhook_contract.py  # 23 contract tests (inherits ChannelContractTest)
website/public/docs/channels.en.md              # English docs
website/public/docs/channels.zh.md              # 中文文档
```

## Verification

- `pre-commit run --files <changed files>` — **all hooks pass** (mypy, black, flake8, pylint, trailing-comma)
- `pytest tests/contract/channels -q` — **266 passed, 2 skipped** (all channels, including webhook's 23 contract tests)
- `pytest tests/unit/channels/test_webhook.py -v` — **43 passed**
- `bash scripts/check-channels.sh webhook` — **all checks passed**

## Deferred (not in this PR)

- Streaming responses back to the inbound caller — the current behavior is fire-and-forget; the inbound HTTP request returns 200 immediately and replies go to `outbound_url`. A follow-up could add SSE or "wait-for-reply" semantics if requested.
- Multi-channel multiplexing on a single port — currently each `WebhookChannel` instance binds its own port. If users want multiple channels on one port, they can run a reverse proxy (nginx/caddy) in front.
- Inbound body schema validation — the channel accepts arbitrary JSON; if a stricter contract is needed, callers can validate `event.parsed` in a downstream channel plugin.

## Issue

Closes #338


xieyxclack:

hongxicheng:
Thanks a lot for this contribution, @AnandSundar — a generic webhook entry point is genuinely useful.

Before we move forward, I'd like to raise one design concern about shipping this as a **built-in** channel.

**Security surface of an exposed port.** To be useful for external senders, this channel has to listen on a reachable address/port. Once exposed, the inbound JSON is turned directly into an `AgentRequest` and drives the agent (which can call tools), so the endpoint is effectively a remote trigger into the agent. The current auth is fairly permissive — e.g. even with a `secret` configured, requests without a signature header are still accepted, and there's no rate limiting / early body-size rejection — which makes an exposed instance risky. Making every deployment carry this listening surface by default (built-in) isn't ideal.

We actually hit the same trade-off recently with the **Azure Bot** channel (#5849): because it also exposes a public webhook endpoint, we deliberately shipped it as an **opt-in plugin** rather than a built-in, so that the surface isn't forced onto deployments that don't need it.

Would you be open to reworking this the same way — as a standalone **plugin channel** (register via `api.register_channel(...)`, config via `config_fields`, deps via `requirements.txt`) and publishing it directly to the [Plugin Marketplace](https://platform.agentscope.io/plugins) so users can install it on demand? #5849 is a good reference implementation for the plugin-channel shape (icon / doc_url / localized config fields included).

If you go that route, it'd also be great to tighten the auth while you're at it: reject unsigned requests when a `secret` is set, and add basic rate limiting. Thanks again!

AnandSundar:
Reposting here for context (closing #5716 after this).

@hongxicheng — your three-point review landed as #6126 (https://github.com/agentscope-ai/QwenPaw/pull/6126) on a rebase of `feat/webhook-channel-338` onto the post-#5849 plugin framework:

1. **Plugin shape (match #5849)** — `plugins/channel/webhook/{plugin.json, plugin.py, __init__.py, requirements.txt}` plus `WebhookChannelPlugin.register(api)` calling `api.register_channel(channel_class=WebhookChannel, icon=..., doc_url={zh,en}, config_fields=[...])`. Built-in entries (`_BUILTIN_SPECS["webhook"]`, `WebhookConfig`, `configure_webhook`) removed. Existing `channels.webhook.*` config blocks are still recognized via the plugin's `config_fields`.
2. **Auth tightening** — `verify_signature()` now returns `False` when a `secret` is configured but `X-QwenPaw-Signature` is missing or empty (was: warn + accept). Server maps that to `401` with `{"error": "signature_mismatch"}`. Unsigned + no-secret path is unchanged.
3. **Rate limiting** — per-client-IP token-bucket, default `5 RPS` / `burst 10`, exposed as `rate_limit_rps` / `rate_limit_burst` config fields and `WEBHOOK_RATE_LIMIT_RPS` / `WEBHOOK_RATE_LIMIT_BURST` env vars. Empty bucket ⇒ `429` with `Retry-After: 1`.

I also updated `scripts/check-channels.sh` to fall back to `plugins/channel/<name>/tests/{contract,unit}/` and run tests via `python -m pytest` (so the namespace package import works) — without that, plugin channels would fail the pre-commit gate by construction. Verified: `bash scripts/check-channels.sh webhook` ⇒ green; full `pytest` sweep ⇒ 305 passed, 2 skipped.

Docs: `website/public/docs/channels.{en,zh}.md` lead the Webhook section with a "plugin channel, not built-in" callout and document the new fields / 429 status / secret-mandatory rule. Deferred items (streaming reply, multi-channel multiplexing, inbound schema validation) carry over.

PR #5716 → superseded by #6126. Happy to iterate on any of the three points.

PR #5722, opened 2026-07-02
fix(feishu): retain per-message sender in shared group sessions

### LINKED ISSUE #5721: Feishu group chats (share_session_in_group) lose per-message sender in history
### Summary

In Feishu group chats with `share_session_in_group=True`, all members'
messages are routed into one shared session, but each user message is
persisted as only `role=user` + `content`. The real speaker is not stored
per message, so once the context grows the model can no longer tell who
said what in the history.

### Environment

- QwenPaw 2.0.0b1
- Channel: Feishu, `share_session_in_group=True` (also affects topic
  threads)

### Current behavior / evidence

The Feishu channel already parses the sender (nickname, open_id,
`sender_display`) and puts it in the inbound `meta` (`user_name`,
`feishu_sender_id`). But:

- `BaseChannel.build_agent_request_from_user_content(...)` builds the
  `Message` with only `type/role/content` — `Message.metadata` is never
  populated, even though the schema supports it.
- `_request_input_to_msgs(...)` converts to an agentscope `Msg` keeping
  only `role/content` (name is set to the role, not the sender).
- With `share_session_in_group=True`, `meta["feishu_sender_id"]` is
  overwritten to `"group"` for routing — further proof the real sender
  must be stored separately or it is lost.

Net effect: persisted history is `role=user` + `content` with `name="user"`
for everyone.

### Impact

Breaks any group capability that needs to distinguish historical speakers:

- "What did <name> just say?"
- "What did A and B each ask for?"
- "Continue answering what <name> asked earlier"
- Per-user profiles / preferences / long-term memory in group chats

### Expected

Each user message in a shared group session should persist the real
sender, and the model should be able to attribute each historical user
message to its speaker. Internal metadata must not leak into user-visible
content.

### Proposed fix

PR incoming: persist a compact, safe sender metadata subset per message,
separate routing identity from the real sender, and project the sender
into the model's view (and summarize/memory) as a `<msg sender="Name">`
envelope on a throwaway copy — without persisting the envelope or adding
an outbound stripper. Feishu-scoped with a generic base hook so other
group channels can opt in.

hellozhouuu:
## What

Fixes loss of the per-message sender in Feishu group chats with
`share_session_in_group=True` (and topic threads). All members share one
session, but history previously stored only `role=user` + `content`, so
the model could not tell who said what once the context grew.

Closes #5721

## How

- **Persist sender metadata.** New `BaseChannel.build_message_metadata`
  hook lets a channel opt in a small, stable metadata subset per user
  `Message`; Feishu supplies `is_group` + `sender_label` (+ ids). The
  metadata is carried into the agentscope `Msg` (`_request_input_to_msgs`).
- **Separate routing identity from the real sender.** Group/thread
  routing now uses `feishu_session_sender_id` (`"group"` / `"thread:*"`),
  while `feishu_sender_id` keeps the real open_id for history. `user_id`
  resolution prefers the routing id, so session routing is unchanged.
- **Project the sender for the model** as `<msg sender="Name">…</msg>`,
  applied on a throwaway clone at request time only — never persisted.
  The label and user content are stripped of the tag tokens so a crafted
  message cannot forge the envelope.
- **Retain sender in long-term memory.** The same projection is applied
  at the `summarize` / auto-memory choke point, so compacted history and
  ReMe memory keep the speaker.
- **Group-only system-prompt hint** teaches the envelope and marks
  message bodies as untrusted input.

## Design note

Rather than persisting the envelope into the transcript and stripping it
on every outbound path (which would break `/`-command parsing, mention
handling, debounce keys, title derivation, and risk leaking to users),
this stores clean content + structured metadata and **projects** the
envelope only where it is needed (model request + summarize/memory), on
throwaway clones. The projection is a deterministic function of
`(content, metadata)`, so eval/replay can reconstruct exactly what the
model saw from persisted state. Storage stays clean; there is no outbound
stripper to forget.

## Scope

Feishu-scoped. Other group-capable channels can opt in via the same base
hook by supplying `is_group` + `sender_label`; the projection layer is
channel-agnostic.

## Testing

- New/extended unit tests: envelope format, forged-tag stripping,
  nickname sanitization, `project_group_sender_labels` clone-safety,
  the group-only prompt hint, metadata carry-through, and a Feishu
  card-approval round-trip that stays in the shared session.
- Full unit suite passes (one unrelated pre-existing failure on `main`
  in `test_coding_project_override.py`).

### DIFF
--- a/src/qwenpaw/agents/memory/reme_light_memory_manager.py
+++ b/src/qwenpaw/agents/memory/reme_light_memory_manager.py
@@ -381,10 +381,20 @@ class ReMeLightMemoryManager(BaseMemoryManager):
         if not messages:
             return ""
 
+        # Project group-sender envelopes so long-term memory retains who
+        # spoke (summarize is the single choke point for both /compact
+        # summaries and auto-memory). Operates on clones; stored context
+        # stays clean.
+        from ..utils.message_request_normalizer import (
+            project_group_sender_labels,
+        )
+
+        projected = project_group_sender_labels(messages)
+
         response = await self._run_reme_job(
             "auto_memory",
             needs_llm=True,
-            messages=[msg.model_dump(mode="json") for msg in messages],
+            messages=[msg.model_dump(mode="json") for msg in projected],
             session_id=str(kwargs.get("session_id") or ""),
             memory_hint=str(kwargs.get("memory_hint") or ""),
         )
--- a/src/qwenpaw/agents/utils/__init__.py
+++ b/src/qwenpaw/agents/utils/__init__.py
@@ -38,7 +38,10 @@ from .message_processing import (
 )
 
 # Message request normalizer
-from .message_request_normalizer import normalize_messages_for_model_request
+from .message_request_normalizer import (
+    normalize_messages_for_model_request,
+    project_group_sender_labels,
+)
 
 # Registry
 from .registry import Registry
@@ -87,6 +90,7 @@ __all__ = [
     "prepend_to_message_content",
     # Message request normalizer
     "normalize_messages_for_model_request",
+    "project_group_sender_labels",
     # Registry
     "Registry",
     # Setup utilities
--- a/src/qwenpaw/agents/utils/message_request_normalizer.py
+++ b/src/qwenpaw/agents/utils/message_request_normalizer.py
@@ -9,6 +9,7 @@ stored conversation state.
 
 from __future__ import annotations
 
+import re
 from copy import deepcopy
 from typing import Any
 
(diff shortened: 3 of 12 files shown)


RerankerGuo (changes requested):
The structured-metadata direction is useful, but the sender projection currently covers only text blocks, so it does not satisfy the per-message attribution contract for Feishu media.

At exact head `41e71a09e8a807d9e9aa78d91b92e76ebf09ae45`, I passed an image-only group message and both mixed block orders through the production `normalize_messages_for_model_request(..., supports_multimodal=True)` path:

```text
image_only=['data:<media>']
image_then_text=['data:<media>', 'text:<msg sender="Alice">caption</msg>']
text_then_image=['text:<msg sender="Alice">caption</msg>', 'data:<media>']
```

`_inject_group_sender_envelope()` returns early when `text_idx` is empty, so an image/file-only message has no sender at all. With mixed content, it wraps only the text block; media before or after that text remains outside `<msg sender=...>...</msg>`. Consequently the model still cannot answer who sent an attachment, despite the Issue/PR promise that each historical user message retains its real author.

Please cover the entire content sequence, including zero-text messages. One possible representation is synthetic opening/closing text blocks around all blocks, but the exact shape should be validated through the actual provider formatting path. Please add image-only plus media-before-text and media-after-text regressions.

This is not exposed by the current tests: all five changed exact-head test scopes pass (`254 passed, 1 skipped`). The same external contract fails after auto-merging the conflict-free normalizer into current main, while its 33 normalization tests pass, confirming a semantic coverage gap rather than a stale dependency artifact.

The branch also needs a semantic rebase. It currently conflicts in three files. In particular:

- `reme_light_memory_manager.py` must retain main's empty-session guard and `_to_reme_session_id(...)` while projecting the messages.
- `message_convert.py` must retain main's `_request_message_metadata(...)`, which already copies incoming metadata and adds the external-user query tag. Taking the PR side would drop that current-main invariant.
- The add/add `test_message_convert.py` files need to be combined.

Finally, the exact remote pre-commit run is red: Black rewrites three test files, and Pylint reports `R1711`, `R0915`, and two unescaped zero-width-space findings. I reproduced the Black failures locally. Please fix the behavior, rebase conflicts, and repository gates before re-review.

PR #5741, opened 2026-07-02
feat(config,agents): env var resolution for config + dialog log sanitization

wjt0321:
PR for #5705. Implements two of the three security improvements proposed in the issue.

## Env var placeholders in config

Any string field in `agent.json` can now use `${ENV_VAR}` syntax:

```
"app_secret": "${QWENPAW_FEISHU_APP_SECRET}"
```

This means channel tokens (Discord, Feishu, Telegram) no longer need to be stored as plaintext.

Missing env vars keep the placeholder and log a warning, so existing configs work as-is. Only values are resolved, never dict keys.

## Dialog log sanitization

Before persisting to `dialog/*.jsonl` and `tool_results/*.txt`, seven API-key patterns are redacted:

- `sk-*`
- `ghp_*`
- `github_pat_*`
- `tvly-*`
- `mkt_*`
- `agent-world-*`
- `dashscope-*`

This prevents permanent leakage when an agent reads a sensitive file (e.g. `grep_search` on `agent.json`) and the raw tool output lands in the dialog archive.

Only the on-disk copy is sanitized. In-memory values stay untouched so tools continue to work.

## Not included

ReMe log sanitization (Plan C from #5705). The ReMe module lives in a separate `reme-ai` package — will follow up there.

## Tests

- 15 new unit tests (8 for env var resolution, 7 for sanitization)
- 659 existing tests, zero regressions
- pre-commit clean (mypy / black / flake8 / pylint)

### DIFF
--- a/src/qwenpaw/agents/offloader.py
+++ b/src/qwenpaw/agents/offloader.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import re
 import sys
 import uuid
 from datetime import datetime, timedelta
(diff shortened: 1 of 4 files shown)


wjt0321:
Although this isn't my first PR, I'm still not very experienced. I made this PR based on my own experience using the system. If anything is not quite right, please let me know — I'll do my best to address it!

PR #5765, opened 2026-07-03
fix(scroll): protect the active turn, add graduated pressure relief, and make recall failures unmistakable

### LINKED ISSUE #5746: [Bug]: 2.0 beta：scroll 上下文压缩可能错误折叠当前任务，导致模型回复旧消息
# QwenPaw 2.0 beta：scroll 上下文压缩可能错误折叠当前任务，导致模型回复旧消息

## 问题概述

在 QwenPaw `v2.0.0b2` 中，使用默认的 `scroll` 上下文策略时，我遇到一次比较严重的上下文错乱：

用户发送 `/heartbeat` 后，QwenPaw 已经开始执行心跳任务，并且实际读取了 `HEARTBEAT.md`、读取了当天心跳日报、执行了多轮 Tavily 搜索。但任务进行到后半段时，最终回复突然变成了对很久之前一条「你好呀」消息的闲聊回复，表现得像是“失忆”或“任务中途丢失上下文”。

从本地日志和 `history.db` 看，这不是超时、不是 channel 断连、也不是模型主动拒绝，而更像是 `scroll` 上下文压缩把当前正在执行的 `/heartbeat` 任务折叠进了 `conversation_history`，只在 live context 中保留了旧的 pinned head，导致最后一次模型生成时看见的是旧消息而不是当前任务。

## 环境信息

- QwenPaw 版本：`v2.0.0b2`
- 模型：`minimax-cn / MiniMax-M3`
- 运行方式：macOS LaunchAgent
- 运行参数：`qwenpaw app --host 0.0.0.0 --port 8088`
- 触发渠道：元宝频道对应的控制台会话
- 心跳配置：
  - `enabled: true`
  - `every: 2h`
  - `target: inbox`
  - `timeout_seconds: 300`

## 复现过程

1. 打开 QwenPaw Console 中的一个已有元宝会话。
2. 该会话此前曾被 `/clear` 清理过，live context 中只剩较少消息。
3. 在该会话中发送 `/heartbeat`。
4. QwenPaw 开始执行心跳任务：
   - 获取当前时间
   - 检查并读取 `HEARTBEAT.md`
   - 读取当天心跳日报
   - 执行多轮 Tavily 搜索
5. 任务尚未正常完成汇总时，最终回复变成对旧消息「你好呀」的普通闲聊回复。

实际 UI 中可以看到，`/heartbeat` 后面已经出现了心跳任务过程：

- “收到～开始执行心跳任务”
- “读取 HEARTBEAT.md”
- “读取 心跳日报_07-02.md”
- “执行 Tavily__tavily_search”

但最后的文本回复却是：

```text
你好呀，欧尼酱！

今天周四，想起今天要做点啥不？是选题、要产出，还是想让我帮忙整点啥资料/搞点啥自动化？

随便打个招呼也行，我都在这儿～
```

这明显不是 `/heartbeat` 的正确收尾。

### 1. 任务不是超时，也不是取消

状态接口显示本轮任务已经结束，且没有正在运行的任务：

```json
{
  "status": "idle",
  "running_task_count": 0,
  "last_run_at": "2026-07-02T14:57:32.511684Z",
  "last_finish_at": "2026-07-02T14:58:44.626257Z"
}
```

也就是说，本轮 `/heartbeat` 大约运行了 72 秒后结束，并没有触发 `timeout_seconds: 300`。

日志中这一轮也没有看到 `CancelledError`、`timeout`、新的 `ERROR` 或任务取消。

### 2. `/heartbeat` 和工具结果实际写入了 `conversation_history`

本地 `history.db` 中可以看到，当前 `/heartbeat` 本身和工具执行结果都已经被写入：

```text
3000 | context_msg | user      | /heartbeat
3001 | model_turn  | assistant | 收到～开始执行心跳任务...试试抖音/微博侧搜...
3002 | tool_result | assistant | get_current_time
3003 | tool_result | assistant | execute_shell_command
3004 | tool_result | assistant | read_file HEARTBEAT.md
3005 | tool_result | assistant | execute_shell_command
3006 | tool_result | assistant | execute_shell_command
3007 | tool_result | assistant | read_file 心跳日报_07-02.md
3008 | tool_result | assistant | Tavily__tavily_search
3009 | tool_result | tool      | Tavily__tavily_search
3010 | tool_result | assistant | Tavily__tavily_search
3011 | tool_result | tool      | Tavily__tavily_search
```

这说明 `/heartbeat` 任务并不是没有进入系统，也不是工具没有执行。

### 3. 但最终 live context 只剩旧消息和压缩索引

保存后的会话文件中，`agent.state.context` 只剩 3 条：

```text
message 0: user
内容：你好呀
时间：2026-07-02T21:05:39

message 1: user
内容：
<system-info>
[context compressed] The turns below were evicted from the live window...
  [seq 2880–3011]
</system-info>

message 2: assistant
内容：
你好呀，欧尼酱！...
```

也就是说，当前 `/heartbeat` 请求和中间工具结果虽然进入了 `conversation_history`，但没有留在 live context 中供最后一次模型生成使用。

### 4. scroll index 把当前任务也折叠进去了

会话文件中的 scroll 状态显示：

```text
seq_by_id:
  old user "你好呀"      -> [2879, 2879]
  old assistant reply    -> [2880, 2880]
  current "/heartbeat"   -> [3000, 3000]
  current assistant turn -> [3001, 3011]

index:
  seq_lo: 2880
  seq_hi: 3011
```

这意味着 `scroll` 压缩索引覆盖了 `2880–3011`，其中包括当前 `/heartbeat` 任务的 `3000–3011`。

最终 live context 中保留下来的 pinned head 却是更早的用户消息「你好呀」，于是最后模型生成时自然会对“你好呀”作答，而不是继续完成心跳任务。

## 预期行为

在一次用户请求或命令仍处于当前 active turn 时，scroll 压缩不应该把当前 user turn 和当前 assistant working turn 整体折叠出 live context。

至少应该保证：

1. 当前用户请求必须保留在 live context 中。
2. 当前 assistant 已完成的中间 reasoning / tool call / tool result 摘要必须保留在 live context 中，或者生成一个明确、可靠的 active-turn summary。
3. 如果当前 turn 被压缩进 `conversation_history`，模型必须能自动、无歧义地 recall 当前 turn，而不是依赖模型自己读 `[context compressed]` 索引。
4. pinned head 不应优先保留一个很久以前的旧用户消息，从而覆盖当前任务语义。

## 实际行为

当前行为似乎是：

1. 当前 `/heartbeat` turn 执行过程中触发 scroll 压缩。
2. 压缩逻辑保留 pinned head。
3. pinned head 在这个会话中是旧消息「你好呀」。
4. 当前 `/heartbeat` 和工具结果被折叠进 `conversation_history`。
5. 最终模型只看到旧消息 + 压缩索引，没有主动 recall 当前任务。
6. 最终回复变成对旧消息「你好呀」的闲聊回复。

## 影响

这个问题对命令型任务影响很大：

- `/heartbeat` 这类长任务可能中途“丢任务”。
- 用户会看到工具确实执行了，但最终结果完全不对应当前请求。
- 表现形式很像模型幻觉或失忆，但根因可能是上下文压缩策略错误保留了旧 live context。
- 对自动化任务尤其危险，因为用户可能误以为任务完成了，实际最后没有正确汇总、推送或写日报。

类似风险可能不只影响 `/heartbeat`，也可能影响所有长链路命令：

- 多轮搜索
- 多工具调用
- 文件读取 + 写入任务
- inbox / cron 触发的后台任务
- 需要最后汇总的自动化工作流

### 方案 A：scroll 压缩时强制保留 active turn

压缩时不要只保留 `pinned head + index + tail`，而要显式识别当前 active turn：

- 当前最新 user message
- 当前 assistant message
- 当前 assistant message 中尚未完成最终回复所需的 tool call / tool result 摘要

这些内容必须在 live context 中保留，直到本轮请求完成。

优点：

- 最符合用户预期。
- 不需要模型主动 recall。
- 对命令型任务、长工具链任务更安全。

缺点：

- 需要在 scroll manager 或 AgentScope split 结果之外额外维护 active-turn 边界。

### 方案 B：压缩后自动注入 active-turn summary

如果当前 turn 因 token 压力必须被压缩，可以生成一个不可折叠的 active-turn summary，例如：

```text
Current active request: /heartbeat
Progress so far:
- read HEARTBEAT.md
- read HeartBeatLog/2026-07/心跳日报_07-02.md
- ran Tavily searches for ...
Next step:
- continue heartbeat task and produce final heartbeat result
```

优点：

- 对超长工具结果更节省上下文。
- 可以兼容现有 scroll index。

缺点：

- summary 质量会影响后续行为。
- 如果 summary 生成不完整，仍可能遗漏关键步骤。

### 方案 C：命令型任务禁用或延后 scroll 压缩

对 `/heartbeat`、`/topic`、`/idea` 等命令型任务，在当前 turn 完成前暂时不要触发 scroll 压缩。

优点：

- 实现可能相对直接。
- 可以快速降低命令中途丢上下文的概率。

缺点：

- 只覆盖命令入口，普通长对话仍可能遇到同类问题。
- 不是根治。

## 临时规避建议

在修复前，建议用户可以选择：

1. 将 `running.light_context_config.strategy` 临时改回 `native`，避免 scroll 压缩。
2. 对 `/heartbeat` 这类任务使用新的独立会话，减少旧 pinned head 干扰。
3. 避免在长任务会话中频繁 `/clear` 后继续复用同一个 channel session。

但这些只是规避，不是根因修复。

## 补充说明

这次问题容易被误判为：

- 模型幻觉
- 心跳超时
- 元宝频道断连
- 工具调用失败

但本地证据显示：

- 工具确实执行了。
- `/heartbeat` 和工具结果确实写入了 `history.db`。
- 任务没有超时，也没有取消。
- 最终 live context 中缺失当前 `/heartbeat`，只保留了旧消息「你好呀」。

因此更可能是 scroll context compression 的 active turn 保留策略存在问题。

### LINKED ISSUE #5776: [Bug]: stale pinned first user message treated as active task in long-lived IM session
## QwenPaw Version

  2.0.0b3, git commit `2fb13111`.

  ## Description

  In long-lived QQ/IM sessions, the AI can treat an old, stale user message as the current task.

  A message from June 28, 2026 saying “create a test task in DeskQuill” was still pinned in context on July 3, 2026. The user later asked an unrelated WeChat
  group summary question, but the model still treated the old DeskQuill context as current and acted on it.

  Initial analysis suggests this is caused by the `scroll` context strategy assuming that the first user message is always the active task.

  **Related PR(s):** N/A

  **Security considerations:**
  This can cause the agent to perform local file operations based on stale context. No direct credential exposure was observed, but it increases the risk of
  incorrect filesystem actions.

  ## Component(s) Affected

  - [x] Tests

  ## Environment

  - **QwenPaw version:** 2.0.0b3, git `2fb13111`
  - **OS:** Linux WSL2, `6.6.114.1-microsoft-standard-WSL2`
  - **Install method:** from source / local workspace checkout
  - **Python version (if applicable):** Python 3.12.3

  ## Steps to Reproduce

  1. Use QQ channel with default `running.light_context_config.strategy = "scroll"` and default `scroll_config.pinned = 1`.
  2. In a QQ session, send a task-like message, for example: `Create a test task in DeskQuill`.
  3. Let the session continue long enough for scroll compression and session persistence to occur.
  4. Days later, send an unrelated request in the same QQ sender session.
  5. Inspect the generated `llm_requests/*.json` or observe the model behavior.

  ## Actual vs Expected

  - **Actual:** The old first user message remains pinned in the model context and is treated as current task context.
  - **Expected:** In long-lived IM sessions, old first-turn user messages should not be treated as the active task indefinitely.

  ## Logs / Screenshots

  Captured request:

  `/home/chiye/worklab/QwenPawSpace/QwenPaw/workspaces/chenlingyue/llm_requests/20260703192209.json`

  ```text
  meta.session_id = qq:ACFFF9507059FC38800D1AAEAE2A1E52

  conversation_history[0].role = user
  conversation_history[0].text starts with:
  Current time: 2026-06-28 01:46:55 Asia/Shanghai (Sunday)
  在deskquill里创建一个测试任务
  <system-info>
  [context compressed] ...
  ```

  Persisted session state also contains:

  ```text
  context[0]: role=user, name=user
  Current time: 2026-06-28 01:46:55 ...
  在deskquill里创建一个测试任务

  context[1]: role=user, name=memory
  <system-info>
  [context compressed] ...
  the pinned task is above and the recent live turns follow.
  ```

  Relevant code locations:

  ```text
  src/qwenpaw/app/channels/base.py
  - resolve_session_id() defaults to f"{channel}:{sender_id}"

  src/qwenpaw/config/config.py
  - ScrollContextConfig.pinned default = 1

  src/qwenpaw/agents/context/scroll/manager.py
  - keeps pinned head via to_compress[:self._pinned]
  - injects compressed index via UserMsg(name="memory", ...)
  ```

  ## Additional Notes

  Suggested fixes:

  - For long-lived IM/channel sessions, default `scroll_config.pinned` should be `0`, or pinning should only apply to explicitly task-scoped sessions.
  - Avoid injecting compressed history as an ordinary user-role message.
  - Change scroll wording from “the pinned task is above” to wording that does not imply the pinned message is the current task.
  - Add regression tests for long-lived channel sessions where an old first user message must not dominate later unrelated requests.

### LINKED ISSUE #5778: scroll 压缩后上下文丢失，后续回复完全跑偏 / scroll compression loses context, causes off-track responses
### 问题描述 / Problem Description

QwenPaw 2.0 默认的 scroll 上下文压缩策略在触发压缩后，会出现严重的上下文丢失问题：**之前讨论的关键信息、决策、上下文被压缩成几句模糊标题，后续回复完全跑偏，牛头不对马嘴**。模型忘记了原本在做什么任务，像换了一个人。

低版本（native 策略）没有这个问题——所有对话都在上下文里，不会丢。

此外，thinking 模式模型还有额外兼容性问题：压缩丢弃 `reasoning_content`，配合 `auto_memory_search` 会直接导致 API 400 错误。

In QwenPaw 2.0, after the default scroll compression triggers, critical context is lost: key discussion points, decisions, and context are compressed into vague headlines, and subsequent responses go completely off-track — the model forgets what it was doing. Lower versions (native strategy) did not have this problem.

Additionally, thinking-mode models have a separate compatibility issue: compression discards `reasoning_content`, which combined with `auto_memory_search` causes API 400 errors.

---

### 复现步骤 / Steps to Reproduce

1. 使用 QwenPaw 2.0（scroll 默认策略）进行较长会话
2. 会话触发上下文压缩（出现 `[context compressed]` 块）
3. 继续对话——模型回复与原始任务方向完全不相关

1. Use QwenPaw 2.0 (default scroll strategy) for a long session
2. Compression triggers (`[context compressed]` block appears)
3. Continue the conversation — model responses become completely irrelevant to the original task

---

### 具体痛点 / Pain Points

- **核心问题：压缩后上下文丢失，回复跑偏**
  多轮对话积累的信息和决策在压缩后只剩几句标题，模型无法从中恢复完整上下文，后续回复方向完全错误。这对需要长会话的任务（排障、开发、深度分析）是致命打击。
  After compression, accumulated context is reduced to a few headlines. The model cannot recover the full context and goes off-track. This is fatal for long-session tasks like debugging, development, and deep analysis.

- **标题不可靠**
  压缩索引完全依赖模型自写的单句标题。标题质量取决于模型当时的状态——状态差时标题毫无意义。压缩越多次，信息丢失越严重。
  The compression index relies entirely on model-generated single-line headlines. Poor headlines lead to compounding information loss across compression rounds.

- **thinking 模式兼容性（附带问题）**
  压缩丢弃 `reasoning_content`，`auto_memory_search` 回注残缺消息 → API 400。
  Compression discards reasoning_content; auto_memory_search re-injects broken messages → API 400.

- **Windows 下 recall 频繁弹审批窗（附带问题）**
  `recall_history_python` 注册为 shell 类型，Windows 无 sandbox → 每次回忆都弹确认。
  On Windows without sandbox, recalling history triggers approval prompts every time.

---

### 根因分析 / Root Cause Analysis

压缩策略的核心缺陷：**标题索引无法替代完整上下文**。当模型需要依赖之前的多轮推理链、决策过程时，几句标题完全不够。这和低版本 native 策略的表现差距巨大——native 保留一切，scroll 只保留标题。

The core flaw: headline-based indexing cannot substitute for full context. When the model needs to follow a multi-turn reasoning chain, a few headlines are nowhere near enough. This is a significant regression from the native strategy.

---

### 建议修复方向 / Suggested Fix

- 关键：压缩后的上下文摘要质量需要大幅提升，不能只靠单句标题
- 可考虑用压缩前的完整上下文调用模型生成结构化摘要
- 短期缓解：允许用户方便地切换回 native 策略，或在文档中明确告知 scroll 的局限性

- Critical: compression summaries need to be much richer, not just single-line headlines
- Consider using the pre-compression context to generate structured summaries
- Short-term mitigation: make it easier for users to switch back to native strategy; document scroll's limitations clearly

---

### 环境信息 / Environment

- QwenPaw 版本: 2.0.0b2
- 操作系统: Windows 11
- 策略: scroll（2.0 默认）
- 对比: 低版本 native 策略无此问题

- QwenPaw version: 2.0.0b2
- OS: Windows 11
- Strategy: scroll (2.0 default)
- Comparison: lower versions with native strategy do not have this problem

niceIrene:
Fixes #5746, #5778, and #5776

Supersedes #5747 — lands its phase-1 patch as-is and implements the planned phases 2–3.

## Background

Under the scroll context strategy, the token-based split could push the **current user request** and its running tool chain into the evictable half of the window. Once folded into the eviction index, the model saw only stale context (an old pinned message + the index) and answered the wrong thing. This PR protects the active turn end-to-end, replaces the pinned-head heuristic with a correct-by-construction anchor, adds a graduated pressure pipeline, makes recall failures impossible to misread, and fixes per-model context-window resolution so compaction triggers at the right threshold.

## Key changes

**1. Active-turn protection**
The latest *real* user message and everything after it are excluded from the evictable middle and merged back into the reserved tail (using the full live `Msg`, not a partial boundary copy) — mid-task compression can no longer lose the current request.

**2. Anchor on the real request, not synthetic stubs**
Loop gates / stop handlers inject a `role='user'` "Continue working on the task." stub (tagged `loop_continuation`). The anchor and the recall SQL floor now skip these tagged stubs (shared `SYNTHETIC_USER_MESSAGE_TAGS` constant; `qwenpaw_tag` persisted into the row metadata), so an extended turn stays anchored on the request that started it.

**3. Pinned-head mechanism removed**
The assumption that the first message is the task brief breaks in long-lived channel sessions (the first message is arbitrary chat, and was the exact bait for the stale-answer bug). Active-turn protection supersedes it. Old configs carrying `pinned:` still load (`extra="ignore"`); the value is ignored.

**4. Unified, trigger-gated pressure pipeline**
`compress()` is one pipeline: evict middle → compact the index while over reserve → **fold the active turn as a last resort**. The fold is *not* a synthetic summary: each completed `tool_result` is replaced **in place** with a one-line recall stub pointing at its durable `seq` (`ms.expand`), keeping the `Msg` id stable and avoiding write-through hazards. The fold (and the still-overflowing warning) are gated on the compression **trigger**, so an in-progress turn is only folded when the active turn itself is the pressure — not on every round of an ordinary long chat.

**5. Structured `recall_history` tool + sandbox-gated REPL**
New in-process `recall_history` tool (`op=expand/search/recall_tool`) runs bound read-only queries over `MemorySpace` — no model-authored code, registered as an `internal` governance type, so **no sandbox and no approval prompt on any platform** (including Windows without WSL2). The Python REPL (`recall_history_python`) is now registered only when a sandbox is actually available or is explicitly opted in, and is repositioned as the advanced escape hatch. Fold stubs, the eviction-index placeholder, the cap pointer, and the en/zh system prompt all point at `recall_history` first.

**6. Recall failures are unmistakable + stale-context guard**
A non-zero recall exit now leads with a **RECALL FAILED — the history was NOT read** banner; a cell that printed real hits *then* crashed leads with **RECALL INCOMPLETE** ("the output below is real — use it") instead of telling the model to discard valid rows; an exit-0 cell that printed nothing says silence is not evidence of absence. The eviction-index placeholder and system prompt instruct the model to recall the current request if it isn't visible, and never to answer an older visible message as if it were the current one.

**7. Honest `/compact` reporting + single steady-state count**
A fold-only run changes no message count, so the old before/after `len()` heuristic reported "Nothing to compact." `compress()` now records what it did in `last_compress` (evicted/compacted/folded) and `/compact` reports the folded count explicitly. In the sustained-overflow steady state the trigger count is reused while nothing rebuilds the context (one `count_tokens` per compress instead of two), and the over-trigger warning fires once per overflow episode instead of every step.

**8. Per-model context windows — single resolver**
Display/usage (`/history`, usage %, daemon status) and the compaction trigger now resolve through one entry point — `Provider.get_context_size` → `resolve_context_window` (explicit config > catalog > 128k default) — so they can no longer diverge (previously a 200k Claude session could show 114% usage while compaction never fired). A static catalog of known windows is added; **OpenRouter self-reports** its authoritative `context_length` (wins outright, no hand-maintained entries); **Ollama opts out** of the catalog (local `num_ctx` truncation made the cloud value worse than the default). Matching is longest-pattern-first so specific entries beat family catch-alls.

**9. Search echo prevention**
`ms.search` now excludes the **active turn** (all current-session rows at/after the latest real user message). Because AgentScope accumulates the reply in one assistant `Msg`, a second recall round would otherwise top-k-match the agent's own current turn (which quotes the recalled findings) instead of real history. Earlier evicted turns and other sessions remain searchable; `ms.expand` / `recall_tool` / `session` stay unfiltered.

**10. Performance**
- Compress-path whole-window write-through is offloaded off the event loop via `asyncio.to_thread`; `HistoryStore` is made thread-safe (`check_same_thread=False` + a lock). The incremental `on_save` hook stays synchronous and shares the lock.
- The `MemorySpace` active-turn floor is memoized (a single `search()` consulted it twice, each running a `MAX(seq)` scan); cached at instance granularity since `MemorySpace` is built fresh per recall call.

## Compatibility

- `scroll_config.pinned` removal is non-breaking: `ScrollContextConfig` ignores unknown fields.
- Checkpoints: `seq_by_tcid` is additive; old snapshots load with an empty map (stubs then fall back to `ms.recall_tool(...)`).
- Context-window catalog: an explicitly configured `max_input_length` always wins; models not in the catalog keep the 128k default, so nothing changes for them. (Edge: explicitly configuring a cataloged model to exactly 128k is indistinguishable from "unset" — use any other value, e.g. 130000, to force it.)

## Testing

- New unit tests: active turn survives a split that reserves nothing; empty middle still compacts an existing index under pressure; fold stubs older results, keeps the newest verbatim, preserves `Msg` identity, never rewrites durable rows, and is idempotent; `seq_by_tcid` checkpoint round-trip; recall failure/silent/success observation shapes; `ms.search` active-turn exclusion (FTS + LIKE); context-window catalog lookups across provider id formats, boundary matching, and the three-level precedence.
- Existing fixtures updated from single-user-message task shapes to multi-turn channel shapes (the single-message shape is now itself covered by the fold tests).
- Full `tests/unit`: **4051 passed, 4 skipped**.

## Related

- Fixes #5746 (active turn evicted mid-task; model answers a stale message)
- Fixes #5778 (compaction trigger identical for all models; large-context windows unused)
- Fixes #5776 ( stale pinned first user message treated as active task in long-lived IM session)
- Supersedes #5747 (phase-1 patch landed as-is; phases 2–3 implemented here)


src/qwenpaw/agents/context/scroll/manager.py
rayrayraykk, line 150: Please make this function async (and use await where appropriate). Running this synchronously will block the main event loop.

rayrayraykk, line 127: Same as below.

src/qwenpaw/agents/context/scroll/memoryspace.py
```diff
@@ -441,6 +497,10 @@ def _search_like(self, query, targets, kind, k) -> list[dict]:
             f"(name IS NULL OR name NOT IN ({_RECALL_EXCL_PLACEHOLDERS}))",
         ]
         params: list = [f"%{query}%", *_RECALL_TOOL_NAMES]
+        excl = self._active_turn_exclusion()
```
rayrayraykk, line 537: `MemorySpace.search()` runs in a subprocess (no event-loop blocking) but querying `SELECT MAX(seq)` twice per call accumulates sub-process latency with large histories; we should cache this value at the session level and refresh it only after compression.

vanwaals:
## Edge case: `recall_history_python` still triggers approval popup in OFF mode when sandbox IS available

### Scenario

1. Platform has a working sandbox (e.g. macOS Seatbelt, Linux bubblewrap)
2. User sets `approval_level = "OFF"` (via Agent Config or per-session toggle)
3. Both `recall_history` and `recall_history_python` are registered (since sandbox probe passes)
4. Model decides to call `recall_history_python` (e.g. for complex queries, custom SQL, cross-session lookups)

### What happens

In `_policy_tool_check_permissions` (tool_adapter.py), the OFF check short-circuits to ALLOW **without compiling a sandbox_config**:

```python
if effective_level is not None and effective_level.is_disabled():
    return PermissionDecision(behavior=ALLOW, ...)
    # ← sandbox_config is never compiled; _qp_sandbox_config stays None
```

Then in `_policy_tool_call`:

- `sandbox_mode = False` (was never set) → `sandbox_config` not injected into kwargs
- `recall_history_python` receives `sandbox_config=None` → fail-closed returns `ToolChunk(state=DENIED)`
- `_policy_tool_call` misreads the DENIED as a sandbox violation → triggers approval popup
- Even after user approves, the retry still has no sandbox_config → recurring popup

### Suggested fix

In `_scroll_repl_runnable` (or wherever the REPL registration gate lives), add an OFF-mode check: when `approval_level = "off"`, do **not** register `recall_history_python` — only expose `recall_history`.

The logic already gates on "sandbox available"; the additional condition is:

```python
# Do not register the REPL when approval_level is OFF:
# OFF mode skips sandbox_config compilation, so the REPL will
# fail-closed (sandbox_config=None) and trigger an unresolvable
# approval loop.
if effective_level is not None and effective_level.is_disabled():
    repl_runnable = False
```

This way, in OFF mode the model only sees `recall_history` (internal type, no sandbox, no approval) and can never hit the sandbox-violation → approval-popup loop.

**Why this is better than patching `_policy_tool_call`**: it prevents the problem at the source (tool not registered = model can't call it) rather than papering over it after the fact. It's also consistent with the existing pattern of not registering the REPL when sandbox is unavailable.

### Edge case: mid-session toggle

Tool registration happens at `AgentBuilder.build()` time (once per session). If a user switches from AUTO → OFF **mid-session**, the already-registered `recall_history_python` remains in the toolkit until the next session.

This is acceptable (consistent with how all tools behave — toolkit is session-scoped), but please add a code comment in `_scroll_repl_runnable` (or equivalent) documenting this boundary:

```python
def _scroll_repl_runnable(...):
    """Whether the sandboxed REPL should be registered for this build.

    The REPL is omitted when:
      - governor is absent (no sandbox_config source)
      - sandbox probe failed (platform cannot sandbox)
      - approval_level is OFF (OFF skips sandbox_config compilation,
        so the REPL would fail-closed and trigger an unresolvable
        approval loop)

    NOTE: this check runs at session-build time. If the user toggles
    approval_level mid-session, the change only takes effect on the
    NEXT session (toolkit is immutable within a session). This is
    consistent with all other tool registration decisions.
    """
```

This helps future maintainers understand why the OFF gate exists and what its timing semantics are.

PR #5766, opened 2026-07-03
fix(utils): recover JSON objects after leading whitespace

VectorPeak:
﻿## Description

This PR fixes a narrow JSON recovery edge case in `safe_json_loads`.

### What Problem This Solves

`safe_json_loads` already recovered a valid JSON object followed by trailing garbage, such as:

```text
{"k":"v"} trailing
```

However, the same recoverable object previously failed when the content also started with JSON whitespace:

```text
\n  {"k":"v"} trailing
```

The failure happens only after the primary `json.loads()` path rejects the trailing bytes. At that point, the fallback called `JSONDecoder.raw_decode(content)` from index `0`. Unlike `json.loads()`, this fallback expects parsing to begin at the JSON document start index, so leading whitespace caused the recovery path to treat an otherwise recoverable object as completely corrupted.

This can affect chat session or mission state recovery when a JSON state file is pretty-printed, hand-edited, or partially overwritten with trailing bytes. In that case, data that could otherwise be recovered is treated as unrecoverable and returns `{}`.

### Why This Change Was Made

The fix belongs in `qwenpaw.utils.json_utils.safe_json_loads` because that helper is the shared recovery boundary used by chat session loading and mission state loading. Fixing it at the utility layer keeps the behavior consistent for all callers instead of patching one session path.

The implementation intentionally stays small: it advances past leading JSON whitespace with `json.decoder.WHITESPACE.match(content, 0).end()` before calling `raw_decode`. This keeps the fallback aligned with JSON whitespace rules without allocating a trimmed copy or accepting broader non-JSON leading text.

This PR does not broaden recovery beyond the existing contract. It still leaves these behaviors unchanged:

- valid JSON continues through the normal `json.loads()` path
- valid dicts with only leading/trailing whitespace still parse normally
- completely invalid JSON still returns `{}`
- recovered non-dict values still return `{}`
- BOMs, arbitrary leading non-whitespace garbage, multiple concatenated objects, and non-dict JSON values are not newly accepted

### Evidence

The regression test covers the exact previously missed case:

```text
\n  {"k": "v"}garbage
```

Before this change, this input returned `{}` because the fallback `raw_decode` started at index `0`.

After this change, the fallback skips JSON whitespace first and recovers:

```python
{"k": "v"}
```

### Possible Call Chain / Impact

```text
chat or mission state file
  -> qwenpaw.app.chats.session._safe_json_loads
  -> qwenpaw.utils.json_utils.safe_json_loads
  -> json.loads(...) fails on trailing garbage
  -> raw_decode fallback previously starts at index 0
  -> leading whitespace causes recovery to fail
  -> caller receives {}
```

**Related Issue:** N/A

**Security Considerations:** No auth, env/config boundary, or external input trust policy changes; this narrows a local JSON recovery edge case.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

I ran the focused chat session recovery tests and the adjacent raw-decode utility tests to confirm the broader raw JSON recovery behavior still passes.

```bash
D:\ZXY\Dev\Miniforge3\condabin\mamba.bat run -n prw-qwenpaw-py3.11 pytest tests/unit/app/chats/test_session.py tests/unit/agents/utils/test_tool_message_utils.py -q
# 74 passed

git diff --check
# passed

D:\ZXY\Dev\Miniforge3\condabin\mamba.bat run -n prw-qwenpaw-py3.11 pre-commit run --files src/qwenpaw/utils/json_utils.py tests/unit/app/chats/test_session.py
# passed
```

`pre-commit run --all-files` has not been run locally for this PR.

Channel-specific checklist items are not applicable because this PR does not change any channel implementation or contract surface.

## Additional Notes

The code change is limited to the `raw_decode` fallback start index and one regression test. It does not change the normal JSON parsing path or the `{}` fallback contract for unrecoverable content.

### DIFF
--- a/src/qwenpaw/utils/json_utils.py
+++ b/src/qwenpaw/utils/json_utils.py
@@ -39,9 +39,11 @@ def safe_json_loads(content: str, filepath: str = "") -> dict:
     except json.JSONDecodeError:
         pass
 
-    # Try to extract the first valid JSON object.
+    # Try to extract the first valid JSON object. ``raw_decode`` does not
+    # skip leading whitespace after ``json.loads`` has already failed.
     try:
-        result, _ = json.JSONDecoder().raw_decode(content)
+        start = json.decoder.WHITESPACE.match(content, 0).end()
+        result, _ = json.JSONDecoder().raw_decode(content, start)
         if not isinstance(result, dict):
             logger.warning(
                 _NOT_DICT_RAW_MSG,
--- a/tests/unit/app/chats/test_session.py
+++ b/tests/unit/app/chats/test_session.py
@@ -57,6 +57,14 @@ def test_safe_json_loads_recovers_trailing_garbage():
     assert _safe_json_loads(content) == {"k": "v"}
 
 
+def test_safe_json_loads_recovers_leading_whitespace_and_trailing_garbage():
+    # Pretty-printed or hand-edited files may include whitespace before the
+    # recoverable object while still retaining unrelated trailing bytes.
+    content = '\n  {"k": "v"}garbage'
+
+    assert _safe_json_loads(content) == {"k": "v"}
+
+
 def test_safe_json_loads_completely_corrupted_returns_empty_dict():
     # Contract: unparseable content returns {} rather than raising. The
     # function also logs a warning, but the warning message is an


rayrayraykk (approved):
LGTM

rayrayraykk:
The test fails due to ACP breaking changes, which has been fixed in #5798

PR #5792, opened 2026-07-05
fix(agents): stop dropping self-paired tool messages during sanitation

Osamaali313:
## Description

When a model request contains an unpaired tool block, `_sanitize_tool_messages` runs `_reorder_tool_results` + `_remove_unpaired_tool_messages` to repair ordering/pairing. In that path, a **valid** AgentScope 2.0 *self-paired* assistant message — one that carries its own `tool_call` and matching `tool_result` in the same `Msg` (`[text, tool_call, tool_result]`) — was **silently dropped**, losing the text and the whole turn from the request sent to the model.

Root cause (two functions):
- `_reorder_tool_results` treated every result-bearing message as a movable result-carrier, skipping it to re-insert after a `tool_call` in *another* message; for a self-paired message the call is in the *same* (skipped) message, so it was never re-inserted.
- `_remove_unpaired_tool_messages` only scanned *following* messages for a `tool_call`'s results, so a self-paired message's own `tool_result` never satisfied its own use and the message was judged unpaired.

Fix: keep messages that carry their own `tool_call` in place during reordering; satisfy a message's `tool_call`s from its own `tool_result`s first (and don't consume a following self-paired message as a result provider); and let a message's own `tool_call`s pair its own `tool_result`s in the final sweep.

## Type of Change

## Component(s) Affected

- [x] Core (`src/qwenpaw/agents/utils/tool_message_utils.py`)
- [x] Tests

## Checklist

- [x] Run and pass `pre-commit run --all-files` (on the changed files — see evidence)
- [x] Run and pass relevant tests (`pytest`)
- [x] No documentation change needed (internal sanitation helper)

## Testing

Added `TestSanitizeToolMessages::test_self_paired_message_kept_when_another_block_unpaired`: a batch with an unrelated unpaired `tool_call` plus a valid self-paired `[text, tool_call, tool_result]` message. Before the fix the self-paired message is dropped; after, it is preserved (and the unpaired orphan is still removed). Classic separate `tool_call`/`tool_result` pairs, result-before-call reordering, and orphan removal remain unchanged.

```
$ pre-commit run --files src/qwenpaw/agents/utils/tool_message_utils.py tests/unit/agents/utils/test_tool_message_utils.py
(yaml/xml/toml/json/prettier: no matching files -> Skipped)
```

```
$ pytest tests/unit/agents/utils/test_tool_message_utils.py tests/unit/agents/test_message_request_normalizer.py -q
71 passed
```

### DIFF
--- a/src/qwenpaw/agents/utils/tool_message_utils.py
+++ b/src/qwenpaw/agents/utils/tool_message_utils.py
@@ -87,11 +87,11 @@ def _reorder_tool_results(msgs: list) -> list:
     result_msg_ids: set[int] = set()
     for msg in msgs:
         if isinstance(msg.content, list):
-            # A message that carries its own tool_call (e.g. an AgentScope 2.0
-            # self-paired assistant msg [text, tool_call, tool_result]) must stay
-            # in place; only standalone tool_result messages are movable. Moving
-            # a self-carrying msg would skip it here and never re-insert it,
-            # silently dropping the whole turn.
+            # Keep a message that carries its own tool_call in place (e.g. an
+            # AgentScope 2.0 self-paired assistant msg
+            # [text, tool_call, tool_result]); only standalone tool_result
+            # messages are movable. Skipping a self-carrying msg here would
+            # drop it, since its own tool_call can never re-insert it.
             has_own_call = any(
                 _is_tool_call(block) and _block_attr(block, "id")
                 for block in msg.content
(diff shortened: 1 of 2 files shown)


qbc2016 (approved):
LGTM.

Osamaali313:
Thanks so much, and glad it landed! 🎉 Happy to be shouted out — here are the handles I'm comfortable sharing:

x: @OsamaAli_72 (https://x.com/OsamaAli_72)
github: Osamaali313 (https://github.com/Osamaali313)
linkedin: https://www.linkedin.com/in/syedosamaali72
instagram: https://www.instagram.com/osamaali_72

(No Discord/Xiaohongshu.) Thanks again — looking forward to contributing more!

PR #5809, opened 2026-07-06
test(unit): inbox module unit tests — PR-A1

hanson-hex:
## Summary

Add unit tests for the `inbox` module (`inbox_store.py` + `inbox_trace_store.py`) — 64 cases, all green. Also fixes a source bug in `_load_events` where corrupted JSON would crash every subsequent read.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Problem / Motivation

Backend line coverage was 42% at end of June. The `inbox` module had zero unit tests despite being on the heartbeat → inbox write path (heavy traffic, real users). During test writing, a real bug was found: `_load_events` called `json.loads()` without exception handling, so a single corrupted bytes write would crash every subsequent `list_events` / `mark_read` / `mark_all_read` / `delete_event` — the inbox becomes permanently unreadable until manual cleanup.

## Proposed Solution

Two test files, 64 cases:
- `tests/unit/app/inbox/test_inbox_store.py` (22 cases) — append_event, list/pagination, status filter, mark_read/mark_all_read, delete, atomic save, large payload write
- `tests/unit/app/inbox/test_inbox_trace_store.py` (42 cases) — `_to_jsonable` (primitives, list/dict recursion, key stringification, pydantic `model_dump`, repr fallback), `create_trace`, `append_trace_events`, `flatten_session_messages`, `parse_session_timestamp` (both formats + invalid), `read_session_messages`, `append_trace_from_session_delta` (baseline→delta, negative baseline clamped), `finalize_trace`, `get_trace`, `delete_trace`

**Source fix**: `src/qwenpaw/app/inbox_store.py` `_load_events` now wraps `json.loads()` in `try/except (json.JSONDecodeError, OSError)` returning `[]`. The next `append_event` atomically replaces the file via tmp-rename.

Style: real file IO via `tmp_path` + `monkeypatch.setattr` on module-level `_INBOX_PATH` / `_TRACE_DIR`. No `importorskip`, no `type:ignore`. Matches `tests/unit/app/chats` and `tests/unit/app/crons` patterns.

## Alternatives Considered

- Mock the file IO (rejected — real atomic-rename path is where bugs hide)
- Split into `test_models.py` (rejected — no dedicated model classes; dicts only, so model-ish logic is tested in the store test files)

## Additional Context

Part of the July backend test plan (PR-A1). Follows the chats/crons template. Tests verified green with `/Users/remy/work/QwenPaw/.venv/bin/python -m pytest tests/unit/app/inbox/ -q`.

## Willing to Contribute

## Verification

- **Tests**: 64 cases, all green locally (`pytest tests/unit/app/inbox/ -q`)
- **CI**: pre-commit + Unit/Contract/Integrated Tests passing
- **Source changes**: 1 bug fix — `inbox_store.py::_load_events` now catches JSONDecodeError/OSError (corrupt file no longer crashes every subsequent read)
- **Commits**: 2 commits (tests + source fix + pre-commit formatting) — see commit history for the fix trail

### DIFF
--- a/src/qwenpaw/app/inbox_store.py
+++ b/src/qwenpaw/app/inbox_store.py
@@ -17,7 +17,13 @@ _MAX_EVENTS = 5000
 def _load_events() -> list[dict[str, Any]]:
     if not _INBOX_PATH.exists():
         return []
-    data = json.loads(_INBOX_PATH.read_text(encoding="utf-8"))
+    try:
+        data = json.loads(_INBOX_PATH.read_text(encoding="utf-8"))
+    except (json.JSONDecodeError, OSError):
+        # Corrupted or unreadable inbox file — treat as empty rather than
+        # crashing every subsequent read. The next append_event will
+        # atomically replace the file with a valid payload.
+        return []
     if not isinstance(data, list):
         return []
     return [item for item in data if isinstance(item, dict)]
--- a/src/qwenpaw/app/inbox_store.py
+++ b/src/qwenpaw/app/inbox_store.py
@@ -3,11 +3,14 @@ from __future__ import annotations
 
 import asyncio
 import json
+import logging
 import time
 import uuid
 from typing import Any
 
 from ..constant import WORKING_DIR
+
+logger = logging.getLogger(__name__)
 
 _INBOX_PATH = WORKING_DIR / "inbox_events.json"
 _LOCK = asyncio.Lock()
@@ -19,10 +22,16 @@ def _load_events() -> list[dict[str, Any]]:
         return []
     try:
         data = json.loads(_INBOX_PATH.read_text(encoding="utf-8"))
-    except (json.JSONDecodeError, OSError):
+    except (json.JSONDecodeError, OSError) as exc:
         # Corrupted or unreadable inbox file — treat as empty rather than
         # crashing every subsequent read. The next append_event will
-        # atomically replace the file with a valid payload.
+        # atomically replace the file with a valid payload. Log the
+        # error so permission/disk-full issues are not silently lost.
+        logger.warning(
+            "Failed to load inbox events from %s: %s",
+            _INBOX_PATH,
+            exc,
+        )
         return []
     if not isinstance(data, list):
         return []
--- a/tests/unit/app/inbox/test_inbox_store.py
+++ b/tests/unit/app/inbox/test_inbox_store.py
@@ -2,12 +2,11 @@
 # pylint: disable=redefined-outer-name,protected-access,unused-argument
 """Unit tests for qwenpaw.app.inbox_store.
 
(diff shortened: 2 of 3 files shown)


yutai78786 (commented):
Coverage is comprehensive and the `_load_events` hardening is the right direction. A few things worth addressing before merge:

1. `except (json.JSONDecodeError, OSError): return []` also silently swallows `PermissionError`, disk-full, etc., which would make the inbox appear empty with no diagnostic trail. Consider at least a `logger.warning`.
2. Inconsistent handling within this same PR: `test_get_trace_on_invalid_json_raises` expects the trace store to actively `raise ValueError` on a corrupted file ("callers must not silently swallow"), while the inbox path silently returns `[]`. Worth unifying, or documenting why they differ.
3. `test_list_events_handles_invalid_json_file` writes `"not a json list"` (valid JSON that isn't a list), so it exercises the `isinstance(data, list)` branch rather than the new `except`. Add a case with genuinely malformed JSON (e.g. `"{bad"`) to actually cover the fix.

Also, these tests are missing priority markers (p0/p1/p2); CI gates by tier, so please add them.

hanson-hex:
Thanks @yutai78786! All 3 items addressed in the latest push:

**1. Silent swallowing of PermissionError/disk-full** — ✅ Added `logger.warning("Failed to load inbox events from %s: %s", _INBOX_PATH, exc)` in the except branch. Permission and disk errors now leave a diagnostic trail.

**2. Inconsistent handling (inbox returns [] vs trace raises ValueError)** — ✅ The inbox path now logs a warning before returning [], which is the safer default (inbox is non-critical: users can still chat without it). The trace store raises because trace data is critical for debugging — callers must know it failed. This difference is intentional but now documented via the log.

**3. Test only exercises isinstance branch, not the except** — ✅ Split into two tests:
- `test_list_events_handles_non_list_json` — valid JSON that isn't a list (exercises `isinstance` guard)
- `test_list_events_handles_malformed_json` — genuinely broken JSON `{bad` (exercises the `except` branch + asserts the log warning)

**Priority markers** — ✅ Added `pytestmark = [pytest.mark.unit, pytest.mark.p0]` at file level. 23 tests, all passing.

yutai78786 (approved):
Approve. All three points have been addressed:
1. The `except` branch in `_load_events` now emits `logger.warning` (with the path and exception), so permission / disk-full errors are no longer silently swallowed.
2. Added `test_list_events_handles_malformed_json` (writes `"{bad"`) that genuinely hits the `except` branch, and split the original non-list case out into its own test.
3. Added `unit` / `p0` markers.

Thanks!

lalaliat (approved):

PR #5815, opened 2026-07-06
refactor(memory): simplify auto-memory search state and context handling

jinliyl:
## Summary

This PR simplifies auto-memory search by removing the `persist_to_context`
configuration, keeping injected search results in the current live context for
the active conversation window, and moving turn-level auto-memory state onto the
memory manager so it survives middleware rebuilds.

### auto-memory search

- Removed `persist_to_context` from auto-memory search config for both ReMe Light and ADBPG memory.
- Consolidated the simulated memory search interaction into a single assistant message containing:
  - search hint text
  - thinking block
  - completed `memory_search` tool call
  - successful tool result
- Injected search results are appended to the current live context, so follow-up model calls in the same tool loop can still see them until normal context management evicts them.
- Added metadata-based block tracking via auto-memory search block ids, so injected search blocks can be identified and stripped later.
- Changed the synthetic assistant message sender to fixed `memory_search`, avoiding accidental merging into the real agent reply.
- Removed unused `agent_name` handling from auto-search message construction.

### auto-memory turn state

- Added per-session auto-memory turn state to `BaseMemoryManager`.
- Tracks pending and seen turn markers on the manager instead of middleware-local fields.
- Tracks the already-searched user turn on the manager, so rebuilt middleware instances do not repeat auto-memory search for the same turn.
- Added TTL cleanup for stale session state to avoid unbounded state growth.
- Ensures auto-memory interval tracking survives middleware reconstruction.

### context sanitization

- Added helpers to remove auto-memory search blocks before persisting conversation messages into memory.
- Drops synthetic auto-search-only messages entirely when no real content remains.
- Preserves normal assistant reply blocks when they share a message with auto-search blocks.
- Applies sanitization in ADBPG and ReMe Light auto-memory persistence paths.

### middleware

- Refactored memory middleware to read and update auto-memory turn state through the memory manager.
- Keeps automation-request guards while clearing pending markers consistently.
- Preserves existing flush behavior for normal user requests and context compression.

### console config UI

- Removed `persist_to_context` from the frontend `AutoMemorySearchConfig` type.
- Removed the “Persist Search Result to Context” switch from:
  - ADBPG memory config card
  - ReMe Light memory config card
- Removed related locale strings across supported console locales.

### configuration and docs

- Removed `persist_to_context` from backend memory config schema/default expectations.
- Updated English and Chinese memory docs so `auto_memory_search_config` only documents:
  - `enabled`
  - `max_results`
- Documented that auto-memory search results remain in the current live context until normal context management evicts them.
- Updated ADBPG config examples to match the new shape.

### tests

- Added unit coverage for memory-manager auto-memory turn state:
  - same-session state reuse
  - session isolation
  - stale state TTL cleanup
- Added middleware coverage proving rebuilt middleware instances do not repeat auto-memory search for the same user turn.
- Added sanitization tests for stripping auto-memory search blocks.
- Updated middleware tests to assert manager-backed state behavior.
- Updated ADBPG and base memory manager tests for the single-message `memory_search` format.
- Removed config test assertions for the deleted `persist_to_context` field.


src/qwenpaw/agents/memory/base_memory_manager.py
```diff
@@ -1,19 +1,31 @@
 # -*- coding: utf-8 -*-
```
rayrayraykk, line 1: Should we modify dummy manager accordingly?

jinliyl: 看过 PR 5815 的 memory 相关改动了，/Users/yuli/workspace/QwenPaw/src/qwenpaw/agents/memory/dummy.py 不需要修改。

  原因是 PR 新增的 get_auto_memory_turn_state()、auto-search 消息构造和清理逻辑都在 BaseMemoryManager，NoopMemoryManager 会继承这些默认实现；同时它现在显式 build_middlewares() -> []，仍然符合 no-op 后端“完全禁用 memory 功能”的语义。

rayrayraykk (approved):

PR #5838, opened 2026-07-07
fix(runtime): persist interrupted turn state and suppress empty cancel messages

qbc2016:
## Description

This PR updates QwenPaw for the AgentScope 2.0.4 and ReMe 0.4.0.8 integration path, and adds runtime/session handling needed for safer interruption behavior. It also refreshes the ReMeLight memory documentation so the docs match the current implementation instead of the earlier high-level design.

### 1. Dependency Upgrades

- Bumps `agentscope` from `2.0.2` to `2.0.4`.
- Bumps `reme-ai` from `0.4.0.7` to `0.4.0.9`.
- Removes the stale inline note about the standalone `agentscope-runtime` package from `pyproject.toml`.

### 2. Runtime Cancel-Save Support

- Adds `Envelope.collect_partial_blocks()` so the runtime can extract in-progress streaming content before a cancellation fully tears down the run.
- Adds `Envelope.collect_tool_output()` so partial tool output accumulated before interruption can be preserved.
- Stores the active `Envelope` on the runtime hook context while a run is active.
- On `asyncio.CancelledError`, attempts a best-effort session save before emitting the cancellation envelope.
- Uses `asyncio.shield()` around `session.save_session_state()` so the save can continue even if the outer task is re-cancelled.
- Snapshots the agent via `state_dict()` into a `StateProxy` before the async save, avoiding mutation/corruption from later cleanup.
- Injects unfinished assistant text and thinking blocks into the agent context before saving, with a deduplication guard to avoid double-saving content already present in the final assistant message.
- Closes dangling tool calls by appending interrupted `ToolResultBlock`s for any `ToolCallBlock` that does not yet have a matching result.
- Preserves any partial tool output and appends an interruption reminder so reloaded sessions have a valid tool-call/result structure.

### 3. ReMeLight Memory Configuration

- Adds `mem_session_dir` to QwenPaw's ReMeLight config and passes it into the embedded ReMe application config.
- Clarifies that `session_dir` stores source conversation logs used by auto-memory.
- Introduces `mem_agent` as the default ReMe internal memory-agent session directory.
- Updates embedded ReMe embedding config to pass `dimensions` as a top-level `as_embedding` field, matching the ReMe 0.4.0.8 component API.
- Keeps vector search enablement behavior unchanged: vector indexing is enabled only when the embedding backend has the required model and credentials.

### 4. Gemini Provider Compatibility Notes

- Updates TODO comments around Gemini tool-schema sanitization to reflect that the local compatibility override is still needed until a later AgentScope version.
- Moves the removal target from `agentscope >= 2.0.3` to `agentscope >= 2.0.5`.

### 5. Configuration Documentation

- Removes obsolete memory/retrieval environment variables from the config docs.
- Documents the current ReMeLight directory fields:
  - `metadata_dir`
  - `session_dir`
  - `mem_session_dir`
  - `resource_dir`
  - `daily_dir`
  - `digest_dir`
- Updates `auto_memory_interval` documentation to the current default of `5`, with `None` or `<= 0` disabling periodic auto-memory.
- Updates `auto_memory_search_config.max_results` documentation to the current default of `2`.
- Clarifies that some runtime config values apply after saving, while embedded ReMe component settings such as directories and embedding config require restarting the agent process.
- Applies these config documentation updates in both English and Chinese docs.

### 6. Memory, Auto Dream, Proactive, and Search Documentation

- Rewrites the memory-evolving/proactive documentation around the actual current code paths rather than the earlier conceptual flow.
- Documents the file layout used by embedded ReMe:
  - source conversation logs under `mem_session/dialog/*.jsonl`
  - internal memory-agent sessions under `mem_agent`
  - daily notes and day indexes under `memory`
  - digest memory under `digest/personal`, `digest/procedure`, and `digest/wiki`
- Explains the Auto Memory trigger path through `MemoryMiddleware`, including turn-count flushing and pre-compaction flushing.
- Documents Auto Memory output behavior, including source-log persistence, daily-note updates, day-index refresh, and inbox event behavior.
- Documents the Auto Resource background watcher and supported resource suffixes.
- Documents Auto Dream entry points (`/dream`, `dream_cron`, and `ReMeLightMemoryManager.dream()`), default job parameters, four ReMe steps, digest buckets, and traceability through wikilinks.
- Clarifies the boundary between ReMe's low-level `proactive` job and QwenPaw's current `/proactive` command.
- Documents that QwenPaw `/proactive` currently uses recent chat/session context and optional screen context instead of directly consuming `memory/<date>/interests.yaml`.
- Documents ReMe search/indexing behavior, including watched directories, raw-log search behavior, hybrid search, BM25, vector recall, RRF fusion, and wikilink expansion.
- Applies the memory/proactive documentation updates in both English and Chinese docs.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [x] Refactoring / compatibility

## Component(s) Affected

- [x] Core / Backend (runtime, config, providers)
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [x] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Security Considerations

- Cancel-save preserves interrupted assistant/tool state but does not introduce new external network access or new credential handling.
- The runtime now persists partial tool output on cancellation, including an explicit interruption marker, so restored context remains structurally valid.
- The proactive documentation explicitly calls out that QwenPaw proactive mode may read historical session memory and may analyze screenshots when multimodal screen analysis is available.

## Testing

- Not run as part of this PR-description update.

## Additional Notes

- The PR includes both implementation changes and substantial documentation updates because the ReMeLight docs were out of sync with the current embedded ReMe behavior.
- One known follow-up is to revisit the legacy `use_dimensions` user-facing option: ReMe 0.4.0.8 expects `dimensions` as a top-level embedding component field, while some UI/docs wording may still imply that `use_dimensions=false` skips sending dimensions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "playwright>=1.49.0",
     "questionary>=2.1.1",
     "mss>=9.0.0",
-    "reme-ai==0.4.0.8",
+    "reme-ai==0.4.0.9",
     "transformers>=4.30.0",
     "python-dotenv>=1.0.0",
     "python-socks>=2.5.3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "playwright>=1.49.0",
     "questionary>=2.1.1",
     "mss>=9.0.0",
-    "reme-ai==0.4.0.7",
+    "reme-ai==0.4.0.8",
     "transformers>=4.30.0",
     "python-dotenv>=1.0.0",
     "python-socks>=2.5.3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,10 +5,7 @@ description = "QwenPaw is a **personal assistant** that runs in your own environ
 readme = "README.md"
 requires-python = ">=3.11,<3.14"
 dependencies = [
-    "agentscope==2.0.2",
-    # agentscope-runtime is no longer needed: agentscope 2.0 ships the
-    # runtime (`agentscope.app`, `agentscope.session`, `agentscope.app._schema`)
-    # in-tree, replacing the standalone agentscope-runtime package.
+    "agentscope==2.0.4",
     "httpx>=0.27.0",
     "packaging>=24.0",
     "discord-py>=2.3",
--- a/src/qwenpaw/_compat/message.py
+++ b/src/qwenpaw/_compat/message.py
@@ -180,4 +180,8 @@ def msg_from_dict(data: Mapping[str, Any]) -> Any:
     if "name" not in payload:
         payload["name"] = payload.get("role") or "assistant"
 
+    # ``metadata`` must be a dict; legacy sessions may store ``None``.
+    if payload.get("metadata") is None:
+        payload["metadata"] = {}
+
     return Msg.model_validate(payload)
--- a/src/qwenpaw/agents/memory/reme_config.py
+++ b/src/qwenpaw/agents/memory/reme_config.py
@@ -38,6 +38,7 @@ def build_reme_app_config(
             "workspace_dir": working_dir,
             "metadata_dir": reme_config.metadata_dir,
             "session_dir": reme_config.session_dir,
+            "mem_session_dir": reme_config.mem_session_dir,
             "resource_dir": reme_config.resource_dir,
             "daily_dir": reme_config.daily_dir,
             "digest_dir": reme_config.digest_dir,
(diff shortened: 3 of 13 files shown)


rayrayraykk:
The cancel-save functionality is correct and the intent (preserving interrupted context) is valuable, but the implementation bypasses the project's established hook architecture. We already have SessionSaveHook (POST_RESPONSE) for normal saves and CancelCleanupHook (ON_ERROR) for cancel cleanup — the cancel-save logic should follow the same pattern as a CancelSessionSaveHook rather than being hardcoded into runtime.py and exposing Envelope internals via new public methods. This would keep the runtime orchestrator thin (per its own docstring: “All insertable features live in LifecycleHook / AgentMode instances”) and make the behavior pluggable, testable, and consistent with the rest of the codebase.

PR #5843, opened 2026-07-07
fix(tool_guard): detect find -delete as dangerous shell command

### LINKED ISSUE #5842: file_guard: find -delete bypasses out-of-workspace file deletion check
## Steps to reproduce

1. `allow_preview_outside_workspace: true`
2. Create a file outside workspace: `touch ~/test.txt`
3. `rm ~/test.txt` → blocked by file_guard (expected)
4. `find ~ -name "test.txt" -delete` → **NOT blocked** (bug)

## Expected behavior

`find ... -delete` should be intercepted by file_guard, same as `rm`.

## Environment

- QwenPaw latest
- Ubuntu VPS
- file_guard enabled, allow_preview_outside_workspace: true

manjieqi:
## Problem

`find ... -delete` bypasses the file guard. When `rm` is used to delete files outside the workspace, the tool guard shows a confirmation dialog. But `find -delete` achieves the same result without triggering any check.

Fixes #5842

## Root Cause

The `dangerous_shell_commands.yaml` only matches `rm`/`del`/`Remove-Item` as file deletion commands. `find -delete` is a completely different pattern that is not recognized.

## Fix

Add a new YAML rule `TOOL_CMD_DANGEROUS_FIND_DELETE` that matches the pattern `find ... -delete` with HIGH severity, same as `TOOL_CMD_DANGEROUS_RM`.

## Testing

- [x] All 503 security-related tests pass (`pytest tests/unit/security/ tests/contract/security/`)
- [x] Verified locally: `find ~ -name "test.txt" -delete` now triggers a HIGH severity finding and shows confirmation dialog
- [x] Regex tested against edge cases:
  - `find . -name "*.tmp" -delete` ✅ intercepted
  - `find /tmp -type f -mtime +30 -delete` ✅ intercepted
  - `find . -name foo` ✅ not intercepted (no -delete)
  - `find-delete foo` ✅ not intercepted (not find command)
  - `echo find && rm -rf` ✅ not intercepted (false positive)


xieyxclack (approved):
LGTM

PR #5861, opened 2026-07-08
fix(desktop): resolve login-shell PATH for packaged macOS backend

alvinlee518:
## What Problem This Solves

GUI apps launched from Finder/Dock on macOS inherit `launchd`'s minimal PATH and do not load the user's login shell environment. As a result, packaged backend processes cannot discover user-installed tools managed through shell configuration (`~/.zshrc` / `~/.bashrc`), such as Homebrew, nvm, mise, pyenv, or asdf managed runtimes.

Commands like `node`, `git`, `ffmpeg`, etc. may fail to resolve even though they work correctly from a terminal session.

This only affects packaged builds. Development builds (`cargo tauri dev`) inherit the terminal environment and already have the expected PATH.

The initial implementation also contained a marker parsing bug: it returned the text after the END marker instead of the content between the markers. This caused valid shell output to be rejected (`None` fallback to the launchd PATH), while any rc-file output after the marker could incorrectly become part of PATH.

## Fix

On macOS, `path_with_backend_dir` (`console/src-tauri/src/backend/command.rs`) now performs a one-shot login shell probe during packaged backend startup and prepends the resolved PATH:

```
[backend_dir] → login-shell PATH → inherited env PATH
```

Using the user's shell is required because PATH customization is arbitrary and may depend on user-specific setup, including custom Homebrew prefixes and non-standard tool locations.

Implementation details:

* Probe command:

  ```
  $SHELL -l -i -c 'printf "__QWENPAW_LOGIN_PATH_BEGIN__%s__QWENPAW_LOGIN_PATH_END__" "$PATH"'
  ```

  Markers isolate the PATH value from arbitrary stdout produced by shell rc files.

* Fixed `parse_marked_login_path` to return only the content between markers (`|(path, _)|` instead of `|(_, path)|`).

* Added regression tests covering:

  * rc-file noise before and after markers
  * clean marker output
  * missing markers

* Increased login shell probe timeout from 3s to 5s because real-world shell initialization (for example gvm + conda + nvm setups) can exceed the previous timeout.

* Any failure case falls back safely to inherited PATH:

  * timeout
  * non-zero shell exit
  * invalid UTF-8 output
  * empty PATH payload

## Evidence

* Controlled zsh reproduction with stdout noise before/after markers:

  * old parser returned `None` or captured trailing rc output
  * fixed parser returns exactly the marked PATH payload

* arm64 validation:

  * under the buggy parser, a fake Homebrew `ffmpeg` installation was not discoverable through `shutil.which()` or subprocess execution
  * after the fix, both resolve successfully using the login-shell PATH

* Local validation with a shell setup containing gvm + conda + nvm:

  * login-shell probe succeeded consistently within the 5s timeout budget
  * resolved PATH contained expected version-manager entries
  * tools unavailable under launchd-minimal PATH (`node`, `go`, `conda`, `brew`, `ffmpeg`, `uv`) became discoverable after applying the resolved PATH

* Validation completed:

  * `cargo test --lib` (16 passed)
  * `cargo check` (dev + release, warning-free)
  * `rustfmt`
  * `git diff --check`

## Checklist

* [x] Conventional Commit title
* [x] Format/lint and tests pass locally
* [x] No behavior change on non-macOS targets


jinglinpeng:
Hi @alvinlee518 , Thanks for the PR. Please consider the following small robustness pass before merge:

1. `resolve_login_shell_path()` currently runs the interactive login shell synchronously during backend startup, before the backend sidecar is spawned, and there is no timeout. If a user’s `.zshrc` / `.bashrc` runs a slow or hanging command, the packaged desktop app can block during startup. Could we add a short timeout and fall back to the inherited PATH on timeout/failure?

2. The code treats the shell’s entire stdout as PATH. Because `-i` sources interactive rc files, any stdout from those files, such as banners, update notices, or user `echo`s, can pollute the parsed PATH. Could we print PATH between unique markers and parse only the marked value?

alvinlee518:
> Hi @alvinlee518 , Thanks for the PR. Please consider the following small robustness pass before merge:
> 
> 1. `resolve_login_shell_path()` currently runs the interactive login shell synchronously during backend startup, before the backend sidecar is spawned, and there is no timeout. If a user’s `.zshrc` / `.bashrc` runs a slow or hanging command, the packaged desktop app can block during startup. Could we add a short timeout and fall back to the inherited PATH on timeout/failure?
> 2. The code treats the shell’s entire stdout as PATH. Because `-i` sources interactive rc files, any stdout from those files, such as banners, update notices, or user `echo`s, can pollute the parsed PATH. Could we print PATH between unique markers and parse only the marked value?

@jinglinpeng 

  Both points addressed:

  1. Added a 3s timeout with kill-on-expiry; caller falls back to the inherited PATH on timeout/failure.
  2. PATH is now printed between unique markers and only the marked value is parsed.

  Ready for re-review.

RerankerGuo:
I reproduced a blocker in the marker parser at exact head `6942eb208da79521c5e16b8b5073e90988b21e58`.

After:

```rust
.and_then(|(_, rest)| rest.split_once(END))
```

the tuple is `(marked_path, suffix_after_END)`, but the next line:

```rust
.map(|(_, path)| path.trim())
```

selects the suffix, not the PATH between the markers.

I exercised the exact shell command on macOS with a controlled interactive zsh profile:

```text
initial PATH: /usr/bin:/bin
.zshrc:       export PATH="/custom/version-manager/bin:$PATH"
stdout:       profile-banner-before
              __QWENPAW_LOGIN_PATH_BEGIN__/custom/version-manager/bin:...__QWENPAW_LOGIN_PATH_END__
```

Contract result:

```text
marked PATH starts with /custom/version-manager/bin: true
exact PR parser result: None
exact PR parser with text after END: "profile-banner-after"
```

Therefore a clean shell returns `None` and `path_with_backend_dir()` falls back to the inherited launchd PATH, so the packaged-app bug remains. If a profile prints after the marker, that banner is instead treated as a PATH entry.

The clean current-main trial merge tree `5c7b56f0` retains the same parser behavior. `git diff --check origin/main...6942eb20` passes, and the remote executable test/format checks are green; this production behavior is not covered by a committed Rust regression.

The tuple selection should be:

```rust
.map(|(path, _)| path.trim())
```

Please also add a parser test with noise both before and after the markers, asserting that only the marked value is returned.

Separately, the required `Real behavior proof` check still fails because the PR body does not contain the exact `## What Problem This Solves` and `## Evidence` headings.

RerankerGuo (changes requested):
作者意图：在 packaged macOS app 启动 backend 前，从 login shell 恢复 Homebrew/version-manager PATH，同时用 marker 和 timeout 抵御 profile 输出与挂起。

当前 marker parser 取错 `split_once(END)` tuple 元素：此时 tuple 是 `(marked_path, suffix_after_END)`，但 `.map(|(_, path)| path.trim())` 选择了 suffix。

```mermaid
flowchart LR
 A[Shell prints BEGIN + PATH + END] --> B[split_once BEGIN]
 B --> C[split_once END gives path, suffix]
 C --> D[Current code selects suffix]
 D --> E[Clean profile returns None]
 E --> F[Fallback to launchd PATH]
```

受控 macOS interactive zsh：marker 内 PATH 以 `/custom/version-manager/bin` 开头，但 exact parser 返回 `None`；若 END 后有 banner，则错误返回 banner。进一步 arm64 probe 证明 fake Homebrew `ffmpeg` 无法被 `shutil.which()` 或 bare subprocess 发现；选择 `(path, _)` 后两者恢复。

请改为选择第一个元素，并提交含 BEGIN 前/END 后噪声的 parser regression，断言只返回 marked PATH。另请补齐 required RBP 的精确标题 `## What Problem This Solves` 与 `## Evidence`。

alvinlee518:
> 作者意图：在 packaged macOS app 启动 backend 前，从 login shell 恢复 Homebrew/version-manager PATH，同时用 marker 和 timeout 抵御 profile 输出与挂起。
> 
> 当前 marker parser 取错 `split_once(END)` tuple 元素：此时 tuple 是 `(marked_path, suffix_after_END)`，但 `.map(|(_, path)| path.trim())` 选择了 suffix。
> 
> ```mermaid
> flowchart LR
>  A[Shell prints BEGIN + PATH + END] --> B[split_once BEGIN]
>  B --> C[split_once END gives path, suffix]
>  C --> D[Current code selects suffix]
>  D --> E[Clean profile returns None]
>  E --> F[Fallback to launchd PATH]
>  style D fill:#fff3e0,color:#e65100
>  style F fill:#ffcdd2,color:#b71c1c
> ```
> 受控 macOS interactive zsh：marker 内 PATH 以 `/custom/version-manager/bin` 开头，但 exact parser 返回 `None`；若 END 后有 banner，则错误返回 banner。进一步 arm64 probe 证明 fake Homebrew `ffmpeg` 无法被 `shutil.which()` 或 bare subprocess 发现；选择 `(path, _)` 后两者恢复。
> 
> 请改为选择第一个元素，并提交含 BEGIN 前/END 后噪声的 parser regression，断言只返回 marked PATH。另请补齐 required RBP 的精确标题 `## What Problem This Solves` 与 `## Evidence`。

@RerankerGuo 

已根据反馈完成修改，麻烦帮忙再 review 一下。

jinglinpeng (changes requested):
## AI-assisted code review

> This review was generated by Codex (AI) and submitted at a maintainer request.

The earlier marker-parser tuple bug has been fixed, and the new parser regression tests look appropriate. However, the current head (`c024019b`) still has three startup reliability and compatibility issues that should be addressed before merge.

### [P1] Apply the timeout to stdout collection as well

After `wait_child_with_timeout()` returns, the code still calls `wait_with_output()` without any deadline ([`command.rs` L243-L249](https://github.com/agentscope-ai/QwenPaw/blob/c024019b178718c4bc1588f88f24689f186c19f5/console/src-tauri/src/backend/command.rs#L243-L249)). The five-second timeout therefore bounds only the exit of the login shell, not the wait for stdout EOF.

If an rc file starts a background process that inherits the stdout pipe, the shell can exit successfully while `wait_with_output()` continues waiting for every writer to close the pipe. In a controlled macOS zsh probe, the shell exited in 0.035 seconds, but `sleep 2 &` kept stdout open for another 2.006 seconds. A long-lived process would block backend startup indefinitely.

Please collect output under the same deadline and stop once the END marker has been received, without waiting for unrelated descendant writers to close the pipe.

### [P2] Terminate the complete probe process tree on timeout

The timeout path calls only `child.kill()` ([`command.rs` L187-L191](https://github.com/agentscope-ai/QwenPaw/blob/c024019b178718c4bc1588f88f24689f186c19f5/console/src-tauri/src/backend/command.rs#L187-L191)), which terminates only the login-shell process. Commands started by shell rc files can remain alive after the parent shell receives SIGKILL. A controlled zsh test confirmed that the child process was still running after the shell was killed. Repeated backend restarts can therefore leak orphaned processes.

Please run the probe in its own process group or session and terminate and reap the complete group on timeout, or provide equivalent process-tree cleanup.

### [P2] Handle macOS login shells with different command-line syntax

The code passes `-l -i -c` to every value of `$SHELL` ([`command.rs` L232-L236](https://github.com/agentscope-ai/QwenPaw/blob/c024019b178718c4bc1588f88f24689f186c19f5/console/src-tauri/src/backend/command.rs#L232-L236)). macOS `/etc/shells` includes `/bin/csh` and `/bin/tcsh`, but `/bin/tcsh -l -i -c ...` immediately fails with `Unknown option: -l`. The implementation then falls back to the minimal launchd PATH, so the original Homebrew and ffmpeg problem remains for those users.

Please dispatch the arguments and command syntax according to the shell type, or provide a reliable Homebrew/common-path fallback when the configured shell is unsupported.

### Suggested regression coverage

- A background writer holds stdout open, but the complete probe remains bounded by the deadline.
- A timeout cleans up the complete process tree.
- csh/tcsh or another unsupported shell follows a reliable fallback path.
- The updated change is tested in a macOS Rust release build against the latest `main`.

jinglinpeng:
@alvinlee518, please take a look at the [AI-assisted review](https://github.com/agentscope-ai/QwenPaw/pull/5861#pullrequestreview-4978998043). The earlier marker-parser issue is fixed, but three startup reliability and shell-compatibility issues still need to be addressed before merge. Thank you.

PR #5866, opened 2026-07-08
fix(security): split rm detection/extraction to prevent ${HOME} bypass (#5090)

hanson-hex:
## Summary

Fix the `rm -rf ${HOME}` bypass reported in #5090. The `_RM_ESCAPE_PATTERNS` list applied a `${VAR}` blanking substitution to the rm-part before shlex tokenisation, which erased real targets like `${HOME}` — so `rm -rf ${HOME}` was silently treated as having no target and was NOT flagged as outside-workspace.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Problem / Motivation

Issue #5090: `rm -rf ${HOME}` was not being blocked by the tool guard. The root cause was identified during PR #5813 (runtime/security/install regression tests). The old code used a single `_RM_ESCAPE_PATTERNS` list for both detection ("is this an rm command?") and extraction ("what are the targets?"). The `${VAR}` → `""` substitution in that list blanked `${HOME}` before shlex parsing, so the target disappeared and the guard saw nothing to flag.

## Proposed Solution

Split into two pattern sets:

- **`_RM_ESCAPE_PATTERNS`** — applied to BOTH detection and extraction passes (e.g. `$(cmd)`, backticks, backslash-escaped `rm`)
- **`_RM_DETECTION_ONLY_PATTERNS`** — applied ONLY to the detection pass to recognise `${RM}` as `rm`, but NOT to the extraction pass so `${HOME}` survives to be flagged

The extraction pass now re-applies only `_RM_ESCAPE_PATTERNS` (not detection-only), preserving real arguments for shlex and path expansion.

### Test coverage (11 cases)

- `rm -rf ${HOME}` → blocked ✅ (the core bypass — was passing before the fix)
- `rm -rf $HOME` → blocked ✅
- `rm -rf ~` → blocked ✅
- `rm -rf /`, `/etc`, `/tmp`, `/usr`, `/var`, `/*` → blocked ✅
- `rm file.txt` → allowed ✅ (in-workspace)
- `${RM} -rf /` → skipped (requires env expansion in shell, tracked as follow-up)

## Alternatives Considered

- **Block all `${VAR}` in rm commands** — too aggressive, breaks legitimate uses like `rm ${BACKUP_DIR}/old.log`
- **Use a full shell parser instead of regex** — correct but heavy refactor, out of scope for this fix

## Additional Context

Extracted from test-only PR #5813 per reviewer (@yutai78786) request for focused security review. The test PR keeps the test suite (with `${HOME}` case marked `xfail`); this PR carries the source fix and its own focused tests.

## Evidence

```
PYTHONPATH=src python -m pytest tests/unit/security/test_rm_bypass_fix_5090.py -v
...
10 passed, 1 skipped in 0.14s
```

Pre-commit clean: mypy + black + flake8 + pylint all pass.

## Willing to Contribute


yutai78786 (commented):
The fix itself is correct: splitting into a detection pass and an extraction pass, preserving `${HOME}` for shlex and path expansion, precisely closes the #5090 bypass, and the tests cover the main vectors.

One issue: this PR includes an unrelated file, `docs/plans/qwenpaw-july-quality-plan.md` (+212 lines, a quality-plan doc), which has nothing to do with this rm-bypass security fix. Since the whole point of splitting this out was a focused security review, please drop that doc to keep the PR single-purpose. Happy to approve once it's removed.

yutai78786 (approved):
Approve. The unrelated quality-plan doc has been removed, so the PR is now single-purpose: the security fix plus its test. The detection-pass / extraction-pass split is correct and closes the #5090 `${HOME}` bypass. Thanks!

lalaliat (approved):

PR #5687, opened 2026-07-01
feat: add per-model preserve_thinking toggle to control reasoning_content relay

qbc2016:
## Description

Introduces a `preserve_thinking` flag on `ModelInfo` (default `true`) that controls whether thinking traces (`reasoning_content`) are relayed back to the model in subsequent turns. When disabled, the formatter omits `reasoning_content` from assistant wire messages.

- Backend: adds `preserve_thinking` field to `ModelInfo`, `ModelConfigRequest`, and `update_model_config`; each provider (DashScope, OpenAI, Gemini) reads the flag and passes it to the formatter as `relay_reasoning_content`; `ModelFormatter.format()` skips `reasoning_content` injection when the flag is off

- Frontend: adds a Preserve Thinking switch in the model config editor with i18n support; fixes stale-state bug where reopening the editor showed the old toggle value by propagating the configure_model API response directly via a new onProviderUpdated callback instead of relying on an async re-fetch cycle

**Related Issue:** Fixes #(issue_number) or Relates to #(issue_number)

**Security Considerations:** [If applicable, e.g. channel auth, env/config handling]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [ ] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
[...]

PR #5870, opened 2026-07-08
fix(model): default preserve_thinking to false to prevent reasoning repetition

niceIrene:
### Summary

Flips the default of the per-model `preserve_thinking` flag from `true` to `false`. When enabled, prior turns' `reasoning_content` (chain-of-thought) is relayed back to the model on every subsequent request. For reasoning models this is off-distribution — they are trained to drop prior-turn CoT from history — and it drives **reflux loops**: the model re-reads its own stale reasoning and re-issues repeated / near-identical tool calls, spinning until the turn is cancelled (cron `wait_for` timeout or a client stream drop, surfaced in the UI as "Answers have stopped").

### Root cause

- `preserve_thinking` defaulted to `true` (it was the pre-existing unconditional relay behavior; the per-model toggle in #5687 only added the ability to turn it off, keeping the old default).
- With it on, `format()` re-injects each prior assistant turn's `thinking` as `reasoning_content` on the wire message.
- Fed its own prior CoT across a multi-step tool loop, the reasoning model regurgitates earlier reasoning and repeats tool calls (observed: the identical thinking block reappearing verbatim across different tool-call iterations).

### Why a global default flip is safe / targeted

The relay is already a **no-op for any model that emits no `thinking` blocks** — `reasoning_contents` is empty and the whole relay branch short-circuits. So flipping the global default only changes behavior for thinking-emitting (reasoning) models, which is exactly the target. No model-type detection needed. Models that genuinely want relay can still opt in explicitly per-model; an explicit config value always overrides the default.

### Changes

**Backend (effective behavior)**
- `providers/provider.py` — `ModelInfo.preserve_thinking` field `default=False` (+ docstring explaining the rationale).
- `providers/provider.py` — `_get_preserve_thinking()` fallback returns `False` when no model info is present.
- `providers/capping_formatter.py` — formatter field `relay_reasoning_content` `default=False`.
- `agents/model_factory.py` — `getattr(self, "relay_reasoning_content", …)` fallback default `False` (consistency safety net).

**Frontend (UI default display)**
- `RemoteModelManageModal.tsx` — the model-edit toggle now defaults to off (`?? false`) in both the initial `useState` and the `useEffect` reset, matching the new backend default.

**Tests**
- `tests/unit/local_models/test_llamacpp_backend.py` — updated the two expected serialized `ModelInfo` dicts to `"preserve_thinking": False`.

### Impact

- **Non-reasoning models:** no behavioral change (relay was a no-op either way).
- **Reasoning models without an explicit setting:** prior-turn CoT is no longer relayed — across all providers (DashScope, Gemini, Ollama, OpenAI, OpenRouter), not just DashScope. This is the intended fix and aligns with how these models are trained (drop prior CoT between turns). The model still reasons normally each turn; it just no longer sees its *previous* turns' thinking, only their conclusions and tool results.
- **Models with an explicit `preserve_thinking`:** unaffected.

### Testing

- `pytest tests/unit/local_models/test_llamacpp_backend.py` → 30 passed.
- Verified no remaining `True` / `?? true` defaults for this flag and no other tests assert the old default.
- Manual: with `preserve_thinking=false`, the verbatim-repeated thinking + repeated tool calls no longer reproduce.

### Notes

Behavior change: anyone who implicitly relied on CoT relay for a reasoning model will lose it and can re-enable it per-model via the model settings toggle (or `preserve_thinking: true` in config).


rayrayraykk:
Has been fixed.

PR #5885, opened 2026-07-09
Reset stop gates on conversation reset

### LINKED ISSUE #5884: Doom loop gate state persists after conversation reset
## Bug

After a conversation hits the `DoomLoopGate` stop stage, sending `/new` or `/clear` clears conversation history but does not reset the registered stop gates. The next turns can still inherit the previous `DoomLoopGate` session state and keep returning:

```text
Doom loop: agent stuck after 6 consecutive repetitions
```

## Observed behavior

Logs showed `DoomLoopGate` continuing to increment `consecutive_hits` across reset commands, for example moving from 6 up through later values instead of starting fresh after `/new`.

## Expected behavior

Commands that reset conversation context should also reset stateful loop gates and pending gate state so a new conversation starts with a clean stop-handler state.

## Proposed fix

Add a reset path on `StopHandler` that calls `reset()` on stateful gates, then invoke it from conversation-reset commands such as `/new` and `/clear`.

FIELA:
## Summary

- Add `StopHandler.reset()` to reset stateful gates without unregistering them.
- Reset stop gates and pending gate state when `/new` or `/clear` resets conversation context.
- Add a regression test for clearing stop gate state from the command handler.

Closes #5884

## Tests / Verification

- `uv run --extra test pytest tests/unit/agents/test_command_handler.py -q`
- `python3 -m compileall -q src/qwenpaw/agents/command_handler.py src/qwenpaw/loop/gates/handler.py tests/unit/agents/test_command_handler.py`

## Risks / Rollback

Low risk. The change only calls `reset()` on gates that explicitly expose it and leaves registered gates in place. Rollback by reverting this commit.

### DIFF
--- a/src/qwenpaw/agents/command_handler.py
+++ b/src/qwenpaw/agents/command_handler.py
@@ -206,6 +206,21 @@ class CommandHandler(ConversationCommandHandlerMixin):
         """Write the rolling compaction summary."""
         self._state.summary = value or ""
 
+    def _reset_stop_gates(self) -> None:
+        """Clear loop-gate state for commands that reset conversation state."""
+        workspace = getattr(self._prompt_context, "workspace", None)
+        handler = getattr(workspace, "_stop_handler", None)
+        reset = getattr(handler, "reset", None)
+        if callable(reset):
+            reset()
+
+        agent = self._agent or getattr(self._prompt_context, "agent", None)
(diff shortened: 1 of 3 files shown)


rayrayraykk:
Since this PR is a draft, I am going to cherry pick your commits and pull a merge request.

PR #5923, opened 2026-07-10
fix: per-message current time injection, drop get_current_time tool (#5455)

### LINKED ISSUE #5455: [Question]: Why not make the current time a per-user-message prefix instead of putting it in environment/system context?

wangfei010313:
## Description

Builds on #5499 by @lecheng2018: cherry-picks its per-user-message time injection (the `Current date` line is removed from the static `build_env_context()` and a fresh timestamp is injected before `executor.run()`), then **removes the now-redundant `get_current_time` tool**.

In long sessions the agent tended to reuse the stale return value of an earlier `get_current_time` call. With every user turn now carrying a fresh timestamp, the tool is both redundant and the root cause of the stale-time reuse, so it is removed. The injected prefix is only visible to the model and is NOT persisted into session history.

**Related Issue:** Fixes #5455
**Supersedes:** #5499 (its commits are included here, preserving authorship)

**Security Considerations:** None. `set_user_timezone` is kept; only the read-only `get_current_time` tool is removed.

## Type of Change
- [x] Bug fix

## Component(s) Affected
- [x] Core / Backend (runtime, agents/tools, config, governance, app/chats)
- [x] Console (frontend) — removed the obsolete GetCurrentTimeCard
- [x] Documentation (website) — mcp/config docs
- [x] Tests

## Checklist
- [x] I ran `pre-commit` on the changed files and it passes
- [x] Documentation updated
- [x] Ready for review

## Testing
Ask the agent "what time is it now?": before it called `get_current_time` (extra round-trip, stale-prone in long sessions); after, it reads the injected `Current time: ...` prefix and answers directly, fresh every turn.

## Local Test
{4DB3B67C-461D-47CE-B7BC-151442728B22}


website/public/docs/config.en.md
```diff
@@ -494,8 +494,7 @@ application is constructed with the new configuration.
 
 This timezone is used for:
 
-- Displaying the current time in the agent's system prompt
-- The `get_current_time` tool
+- Displaying the current time prepended to each user message
```
rayrayraykk, line 497: No change in `config.zh`?

rayrayraykk:
Link to #5499

xieyxclack:
@wangfei010313 Please pass the unit test

rayrayraykk (approved):

PR #5949, opened 2026-07-10
fix(mcp): apply access policy updates immediately

### LINKED ISSUE #5947: V2.0.0版本 MCP中禁用了某些子工具的访问,但是agent还是可以调用. 允许和拒绝失效
Image

V2.0.0版本 中的MCP工具,拒绝和允许设置,无效. 子工具中及时设置了拒绝,agent还是可以访问.

wananing:
## Description

Fixes #5947.

MCP access-policy updates were persisted and then applied through a
fire-and-forget Driver reload. Until that reload completed, the active handler
continued evaluating the previous policy; if reconnecting failed, the stale
policy could remain active indefinitely.

This change adds a reconnect-independent policy path:

- `DriverManager.sync_driver_policy()` serializes DriverCard persistence and
  the live handler update under the manager lock.
- `DriverConfigWatcher` uses `refresh_driver()` for on-disk changes. Policy and
  config-only edits update the active handler in place, while name, protocol,
  endpoint, credential, and enabled-state changes retain the full reload path.

**Related Issue:** Fixes #5947

**Security Considerations:** Policy enforcement still occurs before credential
resolution and tool execution. After a successful API update, subsequent
invocations use the new policy. Manual policy edits take effect on the next
watcher poll without depending on transport reconnection.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Documentation updated (not needed; documented immediate-effect behavior is restored)
- [x] Ready for review

Not applicable. This change is in the shared Driver policy lifecycle and does
not modify channel contracts.

## Testing

Regression coverage verifies:

- Console/API policy updates immediately affect the active MCP handler without
  attempting a transport reload.
- Concurrent policy updates serialize persistence and runtime mutation, leaving
  the on-disk and active policies consistent.
- A policy-only manual YAML edit is applied by the watcher without reconnecting
  or creating a second MCP client.
- A manual endpoint edit still closes the old client and establishes a new
  transport connection.
- Existing MCP stdio, HTTP, and OAuth Driver flows remain green.

## Evidence

The focused Driver/MCP suite and every repository pre-commit hook pass locally.

```bash
.venv/bin/pytest tests/integration/test_driver_mcp_*.py --no-cov -q
# 16 passed

.venv/bin/pre-commit run --all-files
# all hooks passed
```

## Additional Notes

**Chosen layer:** shared Driver lifecycle synchronization, rather than a
Console-, channel-, transport-, or individual-tool workaround.

**Verification matrix:**

- MCP stdio: covered by regression and flow tests
- MCP HTTP/OAuth: existing flow tests pass
- Console/API policy updates: covered
- Manual DriverCard policy edits: covered
- Concurrent policy writes: covered
- Endpoint changes: covered; retain full reload behavior
- Credential/enabled changes: retain full reload behavior
- Channels and stream modes: not applicable

### DIFF
--- a/src/qwenpaw/app/driver_config_service.py
+++ b/src/qwenpaw/app/driver_config_service.py
@@ -115,6 +115,14 @@ class DriverConfigService:
             await self.reload_driver_best_effort(card.name)
         return path
 
+    async def save_policy(self, card: DriverCard) -> Path:
+        """Persist policy changes and update the active Driver immediately."""
+        path = await self.card_store.save(card)
+        manager = getattr(self._workspace, "driver_manager", None)
+        if manager is not None:
+            await manager.sync_driver_policy(card)
+        return path
+
     async def reload_driver_best_effort(self, name: str) -> None:
         manager = getattr(self._workspace, "driver_manager", None)
         if manager is None:
--- a/src/qwenpaw/app/driver_config_service.py
+++ b/src/qwenpaw/app/driver_config_service.py
@@ -117,11 +117,12 @@ class DriverConfigService:
 
     async def save_policy(self, card: DriverCard) -> Path:
         """Persist policy changes and update the active Driver immediately."""
-        path = await self.card_store.save(card)
         manager = getattr(self._workspace, "driver_manager", None)
         if manager is not None:
             await manager.sync_driver_policy(card)
-        return path
+        else:
+            await self.card_store.save(card)
+        return self.card_path(card.name, protocol=card.protocol)
 
     async def reload_driver_best_effort(self, name: str) -> None:
         manager = getattr(self._workspace, "driver_manager", None)
--- a/src/qwenpaw/app/driver_config_watcher.py
+++ b/src/qwenpaw/app/driver_config_watcher.py
@@ -2,8 +2,8 @@
 """Watch DriverCard files and hot-reload active Drivers.
 
 This is the protocol-neutral successor to the old MCP-only config watcher.
-Console/API saves already trigger reloads directly; this watcher preserves the
-manual-edit path for ``drivers/<protocol>/<name>.yaml`` files.
+Console/API saves already apply runtime updates directly. This watcher
+preserves the manual-edit path for ``drivers/<protocol>/<name>.yaml`` files.
 """
 
 from __future__ import annotations
@@ -24,7 +24,7 @@ DEFAULT_POLL_INTERVAL = 2.0
 
 
 class DriverConfigWatcher:
-    """Poll DriverCard storage and reload changed external capabilities."""
+    """Poll DriverCard storage and refresh changed external capabilities."""
 
     def __init__(
         self,
@@ -114,17 +114,17 @@ class DriverConfigWatcher:
 
         for name in sorted(changed_names):
             try:
-                await self._driver_manager.reload_driver(name)
-                logger.info("Driver '%s' reloaded after card change", name)
+                await self._driver_manager.refresh_driver(name)
+                logger.info("Driver '%s' refreshed after card change", name)
             except Exception:
                 logger.warning(
-                    "Failed to reload Driver '%s' after card change",
+                    "Failed to refresh Driver '%s' after card change",
                     name,
                     exc_info=True,
                 )
 
         # Use the snapshot that triggered this pass as the baseline. If a
-        # manual edit lands while reload_driver() is running, the next poll
-        # will compare against this observed state and reload it instead of
+        # manual edit lands while refresh_driver() is running, the next poll
+        # will compare against this observed state and apply it instead of
         # accidentally swallowing the newer write.
         self._last_snapshot = current
--- a/src/qwenpaw/app/mcp/config_service.py
+++ b/src/qwenpaw/app/mcp/config_service.py
@@ -252,7 +252,7 @@ class MCPConfigService:
             card.policy,
             access,
         )
-        await self._driver_config.save_card(card)
+        await self._driver_config.save_policy(card)
         return mcp_access_policy_from_card(card)
 
     async def create_client(
--- a/src/qwenpaw/drivers/manager.py
+++ b/src/qwenpaw/drivers/manager.py
@@ -173,6 +173,19 @@ class DriverManager:
             await self._shutdown_handler(old)
         return self._runtime_info_from_card(card)
 
+    async def sync_driver_policy(self, card: DriverCard) -> None:
+        """Apply a persisted policy to an active Driver without reconnecting.
+
+        This is a no-op when the Driver is not active; the persisted policy
+        will be loaded the next time the Driver starts.
+        """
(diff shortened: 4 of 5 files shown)


xiaoming-qxm:
Nice fix — the approach is exactly right. Routing policy updates through a dedicated `save_policy` → `sync_driver_policy` → `handler.set_policy` path decouples a policy change from the MCP transport rebuild, so a slow or unreachable endpoint can no longer strand a stale (allow) policy on the live handler. Reusing the existing `set_policy` under the manager lock is clean, and the regression test asserting *no transport reload is attempted* nails the core of #5947. 👍

Two things before I'd call this complete.

**1. Make persist + in-place update atomic.** `save_policy` currently persists the card *outside* the lock and then calls `sync_driver_policy`, which takes the lock:
```python
path = await self.card_store.save(card)      # outside lock
await manager.sync_driver_policy(card)        # lock + set_policy
```
There's an `await` boundary between the two, so two concurrent policy updates can interleave and briefly leave the on-disk card and the live handler out of sync. Folding the save into the manager lock closes that window:
```python
# manager.py
async def sync_driver_policy(self, card: DriverCard) -> None:
    card = self._validate_card_for_registered_protocol(card)
    async with self._lock:
        await self._card_store.save(card)                 # persist inside the lock
        handler = self._handlers.get(card.name)
        if handler is not None and handler.card.protocol == card.protocol:
            handler.set_policy(card.policy)

# driver_config_service.py
async def save_policy(self, card: DriverCard) -> Path:
    manager = getattr(self._workspace, "driver_manager", None)
    if manager is not None:
        await manager.sync_driver_policy(card)            # persist + sync atomically
    else:
        await self.card_store.save(card)                  # no manager yet: persist only
    return self.card_path(card.name, protocol=card.protocol)
```

**2. The manual-edit path still has the old failure mode.** `DriverConfigWatcher` calls `reload_driver(name)` for *any* card change, including policy-only edits to `drivers/<protocol>/<name>.yaml`. So editing a policy on disk still rebuilds the transport and can strand a stale policy when the reconnect fails — the exact bug this PR fixes for the Console/API path. There's also a second-order effect: the new `save_policy` writes the card, which the watcher then observes and turns into a redundant full reload on its next poll.

Both disappear if the watcher applies the lightest sufficient action — sync in place when only policy/metadata changed, and reload only when `endpoint`/`credentials`/`enabled` actually changed:
```python
# manager.py
def _requires_reconnect(self, old: DriverCard, new: DriverCard) -> bool:
    # endpoint/credential refs need a transport rebuild; enabled is a lifecycle
    # transition. policy/config-only changes do not need a reconnect.
    return (
        old.endpoint != new.endpoint
        or old.credentials != new.credentials
        or old.enabled != new.enabled
    )

async def refresh_driver(self, name: str) -> DriverRuntimeInfo | None:
    """Apply an on-disk card change with the lightest sufficient action:
    policy/metadata-only edits are applied in place; endpoint/credential/
    enabled changes fall back to a full reload."""
    path = await self._card_store.stored_path(name)
    if path is not None:
        new_card = self._validate_card_for_registered_protocol(
            await self._card_store.load_path(path)
        )
        async with self._lock:
            handler = self._handlers.get(name)
            if handler is not None and not self._requires_reconnect(
                handler.card, new_card
            ):
                handler.sync_runtime_metadata(new_card)   # in place, no reconnect
                return self._runtime_info_from_card(new_card)
    return await self.reload_driver(name)                 # needs reconnect / (re)start
```
```python
# driver_config_watcher.py  (in the changed-names loop)
-            await self._driver_manager.reload_driver(name)
+            await self._driver_manager.refresh_driver(name)
```
This reuses the already-present-but-unused `sync_runtime_metadata`, makes manual policy edits behave like the Console/API path, and incidentally drops the redundant reload triggered by `save_policy`'s own write. A small test — edit only the policy in a card file, assert the tool becomes denied and that `_setup`/connect is **not** called — would lock the behavior in.

With both, policy updates become robust and reconnect-independent across *both* the Console/API and manual-edit paths.

wananing:
Thanks for the detailed review. I addressed both points in `b80af7f7` and
`9defba7c`:

1. `sync_driver_policy()` now validates the card, persists it, and updates the
   active handler under the same manager lock. `save_policy()` only persists
   directly when no manager is available. A deterministic concurrent-update
   test verifies that saves serialize and the on-disk and live policies finish
   in the same state.
2. `DriverConfigWatcher` now calls `refresh_driver()`. Policy/config-only edits
   use `sync_runtime_metadata()` without reconnecting, while name, protocol,
   endpoint, credential, and enabled-state changes retain the full reload path.
   The regression coverage verifies both a manual policy edit with no second
   MCP connection and a manual endpoint edit that replaces the connection.

Verification:

```text
.venv/bin/pytest tests/integration/test_driver_mcp_*.py --no-cov -q
16 passed

.venv/bin/pre-commit run --all-files
all hooks passed
```

The current PR checks are green. Please take another look when convenient.

xiaoming-qxm:
> Thanks for the detailed review. I addressed both points in `b80af7f7` and `9defba7c`:
> 
> 1. `sync_driver_policy()` now validates the card, persists it, and updates the
>    active handler under the same manager lock. `save_policy()` only persists
>    directly when no manager is available. A deterministic concurrent-update
>    test verifies that saves serialize and the on-disk and live policies finish
>    in the same state.
> 2. `DriverConfigWatcher` now calls `refresh_driver()`. Policy/config-only edits
>    use `sync_runtime_metadata()` without reconnecting, while name, protocol,
>    endpoint, credential, and enabled-state changes retain the full reload path.
>    The regression coverage verifies both a manual policy edit with no second
>    MCP connection and a manual endpoint edit that replaces the connection.
> 
> Verification:
> 
> ```
> .venv/bin/pytest tests/integration/test_driver_mcp_*.py --no-cov -q
> 16 passed
> 
> .venv/bin/pre-commit run --all-files
> all hooks passed
> ```
> 
> The current PR checks are green. Please take another look when convenient.

Thanks for the quick turnaround — this addresses both points and looks complete now.

- **Atomicity (Part A):** folding `card_store.save` into the `sync_driver_policy` lock makes persistence + the live-handler update a single critical section, so concurrent policy writes can no longer interleave the on-disk card and the running policy.
- **Manual-edit path (Part B):** `refresh_driver` + `_requires_reconnect` is exactly the right shape — policy/metadata-only edits are applied in place via `sync_runtime_metadata`, while endpoint/credential/enabled changes still fall back to a full reload, and the watcher now routes through it. This also drops the redundant reload that `save_policy`'s own write used to trigger.

LGTM 👍

lalaliat (approved):

Issue #5929, opened 2026-07-10
[Bug]:`ToolResultLimiter` not usable in scroll mode


rayrayraykk:
## QwenPaw Version

[Provide the version of QwenPaw you are using, e.g. 0.x.x or git commit hash.]
[Using `qwenpaw --version` in your command line or checking the version in the console UI can help.]

## Description

[Describe the bug clearly and concisely. What happened vs what you expected?]

ToolResultLimiter doesn't take effect in scroll mode because ToolResultCapMiddleware runs first and already compresses the output below 50KB. By the time ToolResultLimiter sees the content, it's already small enough and doesn't need to be truncated.

And recall won't take any effect, because the raw tool output was never saved.

**Related PR(s):** #(optional)

**Security considerations:** [If applicable, e.g. auth, env/config exposure]

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** [e.g. 0.x.x or git commit]
- **OS:** [e.g. macOS 14, Ubuntu 22.04, Windows 11]
- **Install method:** [pip / one-line install / Docker / from source]
- **Python version (if applicable):** [e.g. 3.10]

## Steps to Reproduce

1.
2.
3.

## Actual vs Expected

- **Actual:**
- **Expected:**

## Logs / Screenshots

[Paste relevant log output or attach screenshots. Use code blocks for logs.]

```
(paste logs here)
```

## Additional Notes

[Optional: workarounds, similar issues, etc.]

<img width="1976" height="1154" alt="Image" src="https://github.com/user-attachments/assets/aa62cfa4-e16f-49e2-8fdf-a1198288c340" />

niceIrene:
Resolved by #5953

PR #5953, opened 2026-07-10
fix: use standard truncation hint for scroll-capped tool results

### LINKED ISSUE #5946: [Bug] [Full output preserved durably] 提示导致 agent 对仍在上下文中的内容发起无效 recall_history 调用
## 现象

工具输出（如 `execute_shell_command`）较长时，界面会显示截断内容并附带 system-info 提示：

```
<<<TRUNCATED ~18948 tokens>>
<system-info>Full output preserved durably. Recall it with recall_history(op="recall_tool", tool_call_id="call_00_...").</system-info>
```

agent 看到这条消息后，会认为内容已被逐出上下文窗口，于是调用 `recall_history` 去数据库拉取——**但此时该工具的完整输出其实还在 live context 中**，没有被 evict。

## 复现步骤

1. 对话中执行一个输出较长的工具（如 `curl` 大段文本）
2. 工具输出被截断显示 + `[Full output preserved durably]` 提示
3. agent 在后续轮次中需要该内容时，不直接引用上下文中的完整输出，而是调用 `recall_history`
4. 实际上 `recall_history` 拉回来的内容就在同一个会话的可见轮次中

## 根因分析

`[Full output preserved durably]` 的语言暗示「内容已持久化（需要 recall 才能取回）」，但实际上在 Scroll Context 架构中，工具输出在被 eviction 之前，完整内容仍然在 live context 的滚动窗口内。

换句话说，这条 system-info 是为 eviction 后的场景设计的，但它在 eviction 之前就已经出现了，导致 agent 产生误判。

## 建议

方案 A：system-info 中区分两种情况：
- 「仍在上下文，但被截断显示」→ 提示 agent 直接引用即可
- 「已被逐出上下文」→ 保持现在的 recall 提示

方案 B：调整提示文案，让 agent 理解截断 ≠ 丢失，先尝试在上下文中查找再 recall。

### LINKED ISSUE #6009: [Bug]: scroll context 压缩触发不准 + .md 豁免 + 无硬上限导致大 session 上游拒绝
## QwenPaw Version

v2.0.0

## Description

scroll 模式下累积数轮对话后（涉及 make-skill 流程，多次读取 .md 文件），后续对话全部失败：
openai.BadRequestError: Error code: 400 Error from provider (Console Go): Upstream request failed

重试切换同供应商（opencode）的其他模型同样被拒，只能开新 session 恢复。

## Root Cause

从 dump 的 `agent_state.context` 分析，整个 session 历史被完整带入请求，压缩未触发，工具结果裁剪也失效：

1. **压缩触发不准**：`token_count_estimate_divisor: 4.0` 低估实际 token 数（中文+代码场景），系统认为未达到 `compact_threshold_ratio: 0.8`，压缩跳过。
2. **`.md` 豁免**：make-skill 等操作读取大量 .md 文件，这些结果被 `exempt_file_extensions` 豁免裁剪，原样堆入上下文。
3. **硬上限在 scroll 下不生效**：参见 #5929，`ToolResultLimiter` 在 scroll 模式下被绕过，缺少兜底截断。

## Environment

- QwenPaw: v2.0.0
- context_manager_backend: scroll
- compact_threshold_ratio: 0.8
- token_count_estimate_divisor: 4.0
- tool_result_pruning: enabled, exempt_file_extensions: [".md"]
- 模型: deepseek-v4-pro（通过 Console Go 供应商）

## Suggestions

1. 增加硬性单次请求 token 上限，超过时强制裁剪（不依赖压缩触发条件）
2. .md 豁免应设最大字节上限，或仅在非 tool_result 场景生效
3. 压缩触发条件应使用更准确的 token 计数方式

## Related

- #5342 / #5510：加了硬上限但仅限 light 模式
- #5929：scroll 下 ToolResultLimiter 不生效
- #5950：类似「按字符数而非 token 数截断」问题

niceIrene:
## Summary

Fixes #5946, #5929, #6009.

This PR aligns scroll-mode tool result handling with the unified `ToolResultPruningMiddleware` introduced by #5935.

`ToolResultPruningMiddleware` is now the single owner of initial tool result pruning for current, background, and historical responses. This PR removes scroll's separate token-based `ToolResultCapMiddleware`, avoiding a second truncation and persistence path.

Large tool outputs are persisted as artifacts under `tool_results/`, while live context and `history.db` retain bounded previews with structured, block-scoped truncation metadata and a standard `read_file` continuation hint. When scroll compaction runs, retained live previews can be reduced further without losing the original artifact path or continuation state.

The PR also improves scroll recall, headline handling, and manual compaction UX so large-output turns remain compact and recallable without exposing internal index markers in user-facing output.

### Tool result handling

- Remove `ToolResultCapMiddleware` from the scroll pipeline.
- Use the unified `ToolResultPruningMiddleware` for initial tool result pruning.
- Use `pruning_recent_msg_max_bytes` for execution-time and recent previews under scroll.
- Persist oversized raw output under `tool_results/`.
- Store only bounded previews and recovery metadata in live context and `history.db`.
- Re-truncate retained live previews to `pruning_old_msg_max_bytes` during scroll compaction.
- Preserve block-scoped `qwenpaw_truncation` metadata when previews are re-truncated.
- Keep the deprecated `scroll_config.tool_output_token_cap` field for configuration compatibility, but no longer use it for scroll result capping.

### Recall and history

- Allow `recall_history` and `recall_history_python` searches to surface matches from saved tool-output artifacts when matching content falls outside the stored DB preview.
- Include saved artifact information when recalling a truncated tool result by `tool_call_id`.
- Avoid duplicating large raw tool output in SQLite.

### Scroll UX

- Improve headline instructions to reduce missing headlines and copied examples.
- Hide scroll headline comments from console and SSE output, including streamed deltas.
- Avoid evicting only the user half of a completed exchange, allowing compaction to reuse the assistant headline without an extra summarization call.
- Simplify manual `/compact` output to show user-facing archived-turn headlines instead of raw tier and sequence details.
- Add compaction-stage timing logs for scroll diagnostics.

### Configuration and documentation

- Clarify the tool result pruning settings:
  - `pruning_recent_msg_max_bytes`: preview size before entering context and while recent.
  - `pruning_old_msg_max_bytes`: smaller preview size retained after scroll compaction.
- Update console labels and English/Chinese documentation to describe the unified pruning pipeline.

## Testing

```bash
uv run pytest -q \
  tests/unit/agents/test_tool_result_pruning_middleware.py \
  tests/unit/agents/tools/test_file_io.py \
  tests/unit/tool_calls/test_coordinator_completion.py \
  tests/unit/agents/context/test_scroll_manager.py

### CODE REVIEW

**Comment by jinliyl** [comment]
Thanks for the comprehensive fix. I checked the current PR head (`30daa17b`) and agree with the main direction: removing the scroll-only `ToolResultCapMiddleware` eliminates two competing execution-time truncation paths. There are, however, several correctness, compatibility, and scalability issues worth addressing before merge.

## 1. Scroll and `ToolResultCapMiddleware`: what is actually overlapping

Deleting `ToolResultCapMiddleware` is conceptually correct. Its execution-time responsibilities are now covered by `ToolResultPruningMiddleware`:

- intercept the completed `ToolResponse` before it enters context;
- keep a bounded preview;
- persist the full output for recovery;
- attach a continuation hint and structured truncation metadata.

The scroll manager still contains tool-result reduction logic, but it runs at a different lifecycle stage:

- `ToolResultPruningMiddleware` caps a fresh/recent result at `pruning_recent_msg_max_bytes` (default 50,000 bytes);
- `ScrollContextManager._compact_live_tool_results()` runs only after scroll compaction is triggered and reduces retained live previews to `pruning_old_msg_max_bytes` (default 3,000 bytes);
- `_fold_active_turn_results()` remains the last-resort pressure fallback when eviction and ordinary compaction are insufficient.

This means the PR removes the duplicate *initial cap*, but it does not—and should not simply—remove all scroll-side result handling. In the builder, scroll deliberately receives `old_max_bytes=pruning_recent_msg_max_bytes`, so without `_compact_live_tool_results()` a retained preview would remain at the recent limit and would never be reduced to the compact limit.

I suggest preserving scroll's responsibility for deciding **when** compaction happens, while moving the duplicated truncation mechanics out of `ScrollContextManager`. In particular, `_compact_tool_output()`, `_compact_tool_text()`, and `_merge_truncation_metadata()` duplicate behavior already owned by the pruning path. A shared `ToolResultPruner` (or a public reusable pruning API) could be called by both:

```text
fresh ToolResponse
  -> ToolResultPruningMiddleware
  -> shared pruner with recent_max_bytes

scroll compaction
  -> ScrollContextManager decides when to compact
  -> shared pruner with old_max_bytes
```

That keeps the lifecycle boundary clear and prevents the two implementations from drifting in notice format, block indexing, artifact handling, or failure behavior.

## 2. Blocking correctness issue: artifact write failure can lose the full result

The old scroll cap was fail-open: if durable persistence failed, it yielded the full result rather than truncating data that could no longer be recovered.

The unified middleware currently catches `OSError` while creating/writing the artifact, leaves `saved_path=None`, and still calls `truncate_text_output()`. The live context and `history.db` can therefore retain only a preview with no usable recovery path. A read-only workspace, permission error, disk-full condition, or transient filesystem failure is enough to make the original output irrecoverable.

Suggested behavior:

- only replace the raw result after the artifact has been saved successfully; or
- persist through another durable fallback first;
- otherwise fail open and return the original content, with a warning/health signal.

Please add a regression test that forces `Path.write_text()`/artifact saving to raise `OSError` and verifies that the original `ToolResponse` is unchanged and no misleading truncation/recovery notice is emitted.

## 3. Configuration compatibility needs an explicit migration path

`scroll_config.tool_output_token_cap` remains accepted but is now silently ignored. This changes behavior for existing deployments:

- a user-specified low cap no longer protects the live context;
- the default changes from 3,000 model-estimated tokens to 50,000 bytes, which can be materially larger depending on the content/model;
- when `tool_result_pruning_config.enabled=false`, scroll now has no execution-time per-result cap at all, whereas the old scroll cap still applied.

It is reasonable to unify on the byte-based pruning configuration, but this should not be a silent compatibility change. At minimum:

- detect an explicitly configured `tool_output_token_cap` and emit a clear deprecation/migration warning;
- document the corresponding `pruning_recent_msg_max_bytes` setting and the unit change;
- explicitly document that disabling tool-result pruning also disables scroll's execution-time result bound;
- add tests for an explicit legacy value and for `enabled=false` under scroll.

If preserving old behavior for one transition release is desired, the legacy field could be treated as a compatibility upper bound. If not, failing validation or producing a prominent startup warning is safer than accepting and ignoring it.

## 4. Saved artifact search has bounded-recall and memory-scaling problems

Saved tool-output search only checks the 200 most recent truncated results. Older artifacts can never be matched even when the full saved file contains the query. This creates a non-obvious recall cliff: the database row is durable, but content outside its preview effectively becomes unsearchable once it falls outside the candidate window.

Artifact search also uses `read_text().splitlines()`, loading each full tool-output file into memory. Since artifacts may be very large and up to 200 candidates are scanned, one recall call can cause a large memory spike or OOM, and the synchronous scan can create substantial latency.

Suggested improvements:

- stream files line by line instead of reading/splitting the whole file;
- enforce a total scan byte/time budget and report partial/degraded search explicitly;
- page through candidates rather than imposing an invisible permanent 200-row cutoff; or
- maintain a lightweight artifact index when the file is created.

Tests should include a match beyond the first candidate page and a large-file search that demonstrates bounded memory behavior.

## 5. Console headline streaming can suppress ordinary comments and subsequent output

The console headline stream filter treats the last ordinary `<!-- ...` comment as a potential headline start whenever it is not already a headline prefix. It stores that suffix in `pending` indefinitely. If the comment is not a scroll headline, subsequent streamed response content can remain withheld until the stream state is reset.

The filter should buffer only prefixes that can still become the exact scroll-headline grammar (for example `<!--`, followed by optional whitespace and `⟦`/`〚`). Once buffered text can no longer match that grammar, it should be flushed immediately as ordinary output. Please add chunk-boundary tests for:

- an ordinary HTML comment split across deltas;
- `<!-- ordinary comment -->` followed by normal response text;
- a real scroll headline split at every relevant prefix boundary;
- a literal `<`, `<!`, or `<!--` at end of stream.

## 6. Test gaps around the unified ownership model

The related backend tests pass locally (`182 passed, 2 skipped`), but the edge cases above do not appear to be covered. In addition, `test_background_completion_is_pruned_before_hint` is currently skipped, while the PR description explicitly claims unified handling for background results.

I recommend covering the ownership model as an end-to-end matrix:

1. a fresh result is capped once at the recent limit and produces one recoverable artifact;
2. scroll persists only the bounded preview to `history.db`;
3. scroll compaction re-truncates that preview to the old limit without changing the original artifact path or block-scoped continuation metadata;
4. multi-block results retain independent metadata and continuation offsets;
5. artifact write failure is fail-open;
6. background completion passes through the same pruning path before a hint is stored/emitted;
7. pruning-disabled and legacy-config scroll behavior is explicit and tested.

## Recommendation

The architectural direction is sound: keep `ToolResultCapMiddleware` deleted and make `ToolResultPruningMiddleware` the sole owner of execution-time pruning. Before merge, I would treat the artifact-write failure as a correctness blocker and require an explicit legacy-config migration story. The shared-pruner extraction, bounded/streaming artifact search, and headline-filter fix would prevent the PR from replacing one duplicate path with new long-term divergence and scalability problems.

niceIrene:
Thanks for the detailed review. I’ve addressed all six areas in the latest PR head (`58738124`).

1. **Shared pruning mechanics**

   I extracted a reusable `ToolResultPruner` and now use it from both `ToolResultPruningMiddleware` and Scroll compaction. Scroll remains responsible for deciding *when* an older live preview should be compacted, while artifact handling, truncation notices, continuation offsets, and metadata merging are owned by the shared implementation.

2. **Fail-open artifact persistence**

   Fresh tool results are now truncated only after the complete output has been persisted successfully. If the artifact directory is unavailable or saving raises `OSError`, the original `ToolResponse` is returned unchanged and no misleading recovery metadata or truncation notice is emitted. Regression tests cover both failure cases.

3. **Legacy configuration migration**

   An explicitly configured `scroll_config.tool_output_token_cap` now produces a clear migration warning directing users to `tool_result_pruning_config.pruning_recent_msg_max_bytes`, including the token-to-byte unit change.

   The deprecated field is excluded from future config serialization, and the warning is emitted at most once per process to avoid request-time log spam. The English and Chinese documentation also clarify that disabling tool-result pruning disables Scroll’s execution-time per-result bound. Tests cover the legacy field, serialization behavior, and `enabled=false`.

4. **Bounded artifact search**

   Artifact search now uses keyset pagination instead of permanently limiting recall to the latest 200 rows. Files are scanned incrementally in binary mode rather than loaded with `read_text().splitlines()`.

   A total 32 MiB / 2-second scan budget bounds memory and latency. When the budget is exhausted, the result explicitly reports that the search was partial. Tests cover matches beyond the first page, streaming behavior, and budget exhaustion.

5. **Console headline streaming**

   The stream filter is now a small state machine that buffers only prefixes which can still become the exact Scroll headline grammar. Ordinary HTML comments and literal `<`, `<!`, or `<!--` content are flushed as soon as they can no longer match, and pending literal content is flushed when the stream completes.

   The frontend tests cover ordinary comments, content following comments, split real headlines, split closing markers, and incomplete prefixes at end-of-stream.

6. **Ownership and background-result coverage**

   The previously skipped pruning-specific background test has been replaced with an active test verifying that a configured background result processor applies the same pruner before storing/emitting the completion hint.

   The existing coordinator offload integration test remains skipped because offload is still globally disabled pending its separate fix; the PR documentation now scopes the background-path claim accordingly.

   Coverage also verifies fail-open persistence, pruning-disabled behavior, multi-block metadata, continuation offsets, Scroll re-truncation, artifact preservation, and bounded history persistence.

During manual verification I also fixed two diagnostic issues:

- deprecated configuration warnings no longer repeat on every config reconstruction;
- manual `/compact` no longer reports a false `context still over ... > 0` warning when it has already reached the configured reserve target. Genuine automatic-compaction overflow warnings remain unchanged.

Current verification:

- Backend reviewer-related suite: **186 passed, 1 skipped**
  - The single skip is the pre-existing `offload globally disabled pending fix` integration test.
- Frontend headline-filter tests: **7 passed**
- TypeScript type checking: passed
- Pre-commit hooks for the changed files: passed

Thanks again for identifying these correctness and scalability gaps. Please take another look when convenient.

jinliyl:
I found one remaining P2 issue in the saved tool-output search path: multi-block tool results only search the first artifact.

`ToolResultPruner` can persist each oversized text block to a separate file, and `flatten_output()` then joins all block notices into one `tool_result` history row. However, `MemorySpace._saved_tool_path()` uses a single regex `search()`, and both `_search_saved_tool_files()` and `_attach_saved_tool_file_matches()` process only that one returned path. As a result, content outside the preview of the second or any later text block cannot be found through `recall_history(op="search", ...)`.

I reproduced this with one history row containing notices for two saved artifacts: the first file contained only decoy text and the search keyword existed only in the second file. `MemorySpace.search()` returned an empty list.

Suggested fix: extract and validate all saved artifact paths from the row, then scan/annotate each path within the existing shared byte/time budget. Please also add a regression test where the match exists only in the second block's artifact.

rayrayraykk (approved):
LGTM

PR #6021, opened 2026-07-13
perf(scroll): batch session history migration writes

niceIrene:
## Summary

Speed up the one-time session history migration during startup.

Previously, every history row was committed in a separate SQLite transaction, which could make large migrations take several minutes. This PR batches all rows from each session file into one transaction.

The existing behavior is preserved:

- Messages older than the configured retention period are still skipped.
- Duplicate rows are still ignored.
- FTS indexing and migration manifests continue to work.

## Testing

- All context tests pass: `176 passed`
- Added tests for batch insertion, deduplication, and FTS indexing

### DIFF
--- a/src/qwenpaw/agents/context/scroll/history.py
+++ b/src/qwenpaw/agents/context/scroll/history.py
@@ -8,6 +8,7 @@ import logging
 import sqlite3
 import sys
 import threading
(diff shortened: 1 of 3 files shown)


rayrayraykk (approved):
LGTM

Issue #5906, opened 2026-07-09
[Bug]: 防重复功能异常触发


MCQSJ:
## QwenPaw Version

2.0.0b4

## Description

开启防止重复后，正常对话和读取历史上下文错误触发Doom loop: agent stuck after 6 consecutive repetitions，但是实际对话未重复

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **OS:** [Ubuntu 26.04]
- **Install method:** [one-line install]

## Steps to Reproduce

1.正常聊天
2.搜索历史记忆或者正常聊天出现

## Actual vs Expected

- **Actual:**正常回答提示循环中断
- **Expected:**正常回答

## Logs / Screenshots

[Paste relevant log output or attach screenshots. Use code blocks for logs.]

```
2026-07-09 14:01:02 | INFO | root/.qwenpaw/venv/lib/python3.12/site-packages/qwenpaw/loop/gates/runner.py:166 | Gate pending stop applied: Doom loop: agent stuck after 6 consecutive repetitions
2026-07-09 14:01:02 | INFO | root/.qwenpaw/venv/lib/python3.12/site-packages/qwenpaw/loop/gates/runner.py:147 | Gate wants stop (deferred): Doom loop: agent stuck after 6 consecutive repetitions
2026-07-09 14:01:02 | INFO | root/.qwenpaw/venv/lib/python3.12/site-packages/qwenpaw/loop/gates/doom_loop.py:141 | DoomLoopGate: STOP after 17 hits
2026-07-09 14:01:02 | WARNING | root/.qwenpaw/venv/lib/python3.12/site-packages/qwenpaw/loop/gates/doom_loop.py:242 | Doom loop: sim=1.00 thr=1.00
2026-07-09 14:00:17 | INFO | root/.qwenpaw/venv/lib/python3.12/site-packages/qwenpaw/loop/gates/doom_loop.py:141 | DoomLoopGate: STOP after 16 hits
[...]

PR #6041, opened 2026-07-13
fix(loop): exempt read-only tools from doom loop detection (#5906)

JiJiaF1020:
## Problem

`DoomLoopGate` with `similarity_threshold=1.0` flags legitimate repeated read-only tool calls as doom loops. Users reading memory 3+ times get a warning; 6+ times gets terminated — making normal memory recall trigger doom loop false positives.

**Issue:** #5906

## Root Cause

`DoomLoopGate.record()` and `_auto_record_from_ctx()` count ALL tool calls equally, including read-only tools that are legitimately called repeatedly (e.g. recalling conversation history, reading files).

## Fix

Add `exempt_tools` config to `DoomLoopConfig` (default: `["recall_history", "recall_history_python", "read_file"]`). `DoomLoopGate` skips recording any tool in the exempt set, so repeated read-only calls never enter the sliding window.

Key insight: `recall_history_python` is the sandboxed REPL that is the **default** recall tool in scroll context. `recall_history` (structured) is only the fallback for environments without a sandbox (e.g. Windows without WSL2). Both must be exempt — omitting `recall_history_python` leaves the main code path unfixed.

## P1 Fix (control flow bug found during review)

`_auto_record_from_ctx` previously advanced `last_recorded_iter` before checking exempt — an exempt tool as the last call in context would swallow non-exempt calls earlier in the same iteration. Now: `last_recorded_iter` advances only after a record is appended or the iteration is fully scanned; exempt tools use `continue` (not `return`) so earlier non-exempt calls are still found.

## Trade-off

Fully exempting read-only tools means a model stuck in an infinite read-only loop is invisible to this gate and will burn tokens unchecked. This is an accepted hotfix cost — the alternative (higher tolerance instead of full exempt) is tracked as a follow-up.

## Changes

- `config.py`: Add `exempt_tools` field to `DoomLoopConfig` (includes `recall_history_python` for default scroll context path)
- `doom_loop.py`: Add `exempt_tools` param; skip in `record()` and `_auto_record_from_ctx()` (with `continue` + deferred iter advance); document trade-off in `__init__` comment
- `react_gates.py`: Pass `exempt_tools` from config to `DoomLoopGate`
- `goal/helpers.py`: Same wiring for `/goal` mode
- `test_doom_loop_gate.py`: 12 new tests covering exempt behavior, P1 swallow scenario, `recall_history_python` main path, and all-exempt iter advancement

## Test Results

All 34 tests in `tests/unit/loop/` pass (12 new + 22 existing), zero regression.

### DIFF
--- a/src/qwenpaw/config/config.py
+++ b/src/qwenpaw/config/config.py
@@ -1104,6 +1104,25 @@ class DoomLoopConfig(BaseModel):
         default=False,
         description=("Also run during /goal and " "/mission loop modes"),
     )
+    exempt_tools: List[str] = Field(
+        default_factory=lambda: [
+            "recall_history",
+            "recall_history_python",
+            "read_file",
+        ],
+        description=(
+            "Tool names exempt from doom loop "
+            "detection. Read-only tools that are "
+            "legitimately called repeatedly "
+            "(e.g. memory recall, file reading) "
+            "should be listed here to avoid "
+            "false positives (#5906)."
+            " recall_history_python is the "
+            "sandboxed REPL used by default in "
+            "scroll context; recall_history is "
+            "the structured fallback."
+        ),
+    )
 
 
 class IterationGateConfig(BaseModel):
--- a/src/qwenpaw/loop/gates/doom_loop.py
+++ b/src/qwenpaw/loop/gates/doom_loop.py
@@ -65,6 +65,7 @@ class DoomLoopGate(LoopGate):
         window_size: int = 3,
         similarity_threshold: float = 1.0,
         stages: list | None = None,
+        exempt_tools: set[str] | None = None,
     ) -> None:
         super().__init__()
         self._window_size = max(2, window_size)
@@ -73,6 +74,14 @@ class DoomLoopGate(LoopGate):
             stages or [],
             key=lambda s: s.after,
         )
+        self._exempt_tools = set(exempt_tools) if exempt_tools else set()
+
+        # Trade-off (#5906): fully exempting read-only tools means a
+        # model stuck in an infinite read-only loop (e.g. same
+        # recall_history call forever) is invisible to this gate and
+        # will burn tokens unchecked.  This is an accepted hotfix cost
+        # — the alternative (higher tolerance instead of full exempt)
+        # is tracked as a follow-up.
 
     def _ensure_state(self) -> _DoomState:
         """Get or create per-session state."""
@@ -93,6 +102,8 @@ class DoomLoopGate(LoopGate):
     ) -> None:
         """Record a completed tool call."""
         state = self._ensure_state()
+        if tool_name in self._exempt_tools:
+            return
         state.history.append(
             _ToolCallRecord(
                 tool_name=tool_name,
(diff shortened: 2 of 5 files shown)


RerankerGuo:
I found a false-positive regression in the exemption behavior at exact head `47beb330a5450c983f2ed9c070a600c2ecd32d2f`.

`record()` / `_auto_record_from_ctx()` drop exempt calls entirely. That compresses non-consecutive calls into a consecutive repetition window. For example, this productive sequence:

```text
search_web({"query":"same"})
read_file({"path":"file-a"})
search_web({"query":"same"})
read_file({"path":"file-b"})
search_web({"query":"same"})
```

becomes three identical `search_web` records, so the fifth iteration incorrectly returns `INTERRUPT_AND_CONTINUE`. The two different file reads should break the repetition sequence.

I ran the same external contract through `DoomLoopGate.check()` (one real `_auto_record_from_ctx()` call per iteration) on macOS / Python 3.11.15:

| Revision | Recorded history | Final action |
|---|---|---|
| current main `bacac741` | search, read A, search, read B, search | `BYPASS` |
| exact PR head `47beb330` | search, search, search | `INTERRUPT_AND_CONTINUE` |
| clean trial merge tree `424051b1` | search, search, search | `INTERRUPT_AND_CONTINUE` |

Existing tests remain green, so they do not cover this ordering contract:

```text
exact head:  pytest -q tests/unit/loop  -> 45 passed
trial merge: pytest -q tests/unit/loop  -> 75 passed
```

Could exempt calls act as repetition boundaries (or otherwise remain represented in chronological order) instead of being removed from the window? A regression test should assert that the interleaved sequence above stays `BYPASS`, while preserving the intended behavior for repeated recall calls.

Separately, the required `Real behavior proof` check currently fails because the PR body lacks the exact `## What Problem This Solves` and `## Evidence` headings.

RerankerGuo (changes requested):
作者意图：避免重复调用只读 recall/file 工具时触发 doom-loop 误报，同时保留对真正重复操作的检测。

当前实现引入了另一类确定性误报：`record()` 和 `_auto_record_from_ctx()` 将 exempt calls 从历史中完全删除，导致原本不连续的普通工具调用在滑动窗口中被压缩成连续重复。

```mermaid
flowchart LR
    A[search_web same query] --> B[read_file A exempt]
    B --> C[search_web same query]
    C --> D[read_file B exempt]
    D --> E[search_web same query]
    E --> F{Chronological history}
    F -->|current main| G[search read search read search: BYPASS]
    F -->|PR drops reads| H[search search search: INTERRUPT]
    style B fill:#bbdefb,color:#0d47a1
    style D fill:#bbdefb,color:#0d47a1
    style H fill:#ffcdd2,color:#b71c1c
```

独立生产路径复现每轮调用一次 `_auto_record_from_ctx()`：

```text
sequence:
  search_web(q)
  read_file(file-a)
  search_web(q)
  read_file(file-b)
  search_web(q)

current main bacac741:
  history = search, read A, search, read B, search
  result  = BYPASS

exact head 47beb330:
  history = search, search, search
  result  = INTERRUPT_AND_CONTINUE

current-main trial tree 424051b1:
  history = search, search, search
  result  = INTERRUPT_AND_CONTINUE
```

现有测试未覆盖该顺序边界，尽管都通过：

```text
exact loop suite: 45 passed
current-main trial loop suite: 75 passed
```

请让 exempt calls 作为重复序列的边界保留，或采用等价的顺序感知表示，而不是直接从历史删除。新增回归测试必须同时证明：

1. `search(q) -> read(A) -> search(q) -> read(B) -> search(q)` 保持 `BYPASS`；
2. 连续重复的 recall/read 仍按 #5906 的目标不被误判；
3. 真正连续的非豁免相同调用仍会触发既有 stage。

此外，当前 required `Real behavior proof` 因 PR 正文缺少精确的 `## What Problem This Solves` 和 `## Evidence` 标题而失败；修复行为后请同时更新正文门禁。

PR #6080, opened 2026-07-14
feat(crons): add silent agent delivery

### LINKED ISSUE #6075: [Feature]: Allow cron agent jobs to run without channel delivery
## Problem

Agent cron jobs always forward every generated event to their target channel. That makes background work such as token refreshes, health checks, indexing, and log maintenance add messages to an interactive chat even when only the task's side effects matter.

`save_result_to_inbox` controls Inbox records, but it does not suppress channel delivery.

## Proposed change

Add an optional `dispatch.silent` flag for agent cron jobs. When it is enabled, the executor should still consume the full agent stream and keep the normal session, trace, history, timeout, and optional Inbox behavior, but skip `channel_manager.send_event()`.

The flag should default to `false`, be rejected for text-only jobs, and be available through the Console and cron CLI. Existing jobs would keep their current behavior.

## Validation

Regression tests should cover schema compatibility, text-job validation, silent execution without delivery, and unchanged delivery for the default path.

AaronZ345:
Closes #6075.

## What Problem This Solves

Agent cron jobs currently forward every generated event to their target channel. That makes background maintenance jobs such as index refreshes, health checks, and token renewal post unwanted replies even when their useful work is entirely in tools or persisted state.

This adds an optional `dispatch.silent` flag for agent jobs that should run normally without channel delivery.

## What changed

- consumes the full agent stream, so tool side effects and normal execution still happen
- preserves the session, trace, timeout handling, and optional Inbox record
- skips only `send_event` delivery and reports the result as `suppressed`
- keeps delivery enabled by default and rejects silent mode for fixed-text jobs
- exposes the option through the CLI and Console, with English and Chinese docs and cron skill guidance

## Evidence

On current `main`, the new regression suite failed in six places: `DispatchSpec` had no `silent` field, text jobs accepted the unsupported option, the CLI dropped it, and a silent agent job still awaited `send_event` twice.

After this change, the executor test consumes both generated events, awaits `send_event` zero times, finalizes the trace with `success`, and returns `delivery_status="suppressed"`. The matching default-delivery test still sends both events.

Validation run locally:

- `uv run pytest tests/unit/app/crons/test_models.py tests/unit/app/crons/test_executor.py tests/unit/cli/test_cron_cmd.py -q` — 22 passed
- `uv run pytest tests/unit -q` — 4650 passed, 7 skipped
- `uv run pre-commit run --all-files` — passed
- `npm run format:check` — passed
- `npm run test:run` — 1118 passed
- `npm run build` — passed

## Type and scope

Feature; affects cron core, CLI, Console, built-in cron skill, documentation, and tests. Existing jobs remain unchanged because silent delivery defaults to false.


AaronZ345:
@lalaliat You have worked on the recent cron execution and delivery paths, so I would appreciate your review here. This keeps agent execution, session history, traces, and optional Inbox records intact while suppressing only channel delivery. If that boundary looks right, could you help move it toward merge? Thanks.

lalaliat (approved):
Great work! Thank you for your contribution!

Issue #6082, opened 2026-07-14
[Bug]: /goal completions block subsequent chat — stale session triggers TERMINATE on every message


cmradix:
## QwenPaw Version

v2.0.0.post1

## Description

After a `/goal` task completes, every subsequent normal chat message is blocked by `TERMINATE`, making conversation impossible.

Root cause: PR #5727 introduced `GoalTurnGate` detecting `active=False` → `TERMINATE`, but `update_goal("complete")` only sets `active=False` without removing the session from `_sessions`. Every subsequent message, `session_by_ctx_var()` returns the stale session, triggering an infinite TERMINATE loop.

**Related PR(s):** #5727

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)

## Environment

- **QwenPaw version:** v2.0.0.post1
- **OS:** macOS 15 (Darwin 25.5.0, arm64)
- **Install method:** uv pip
- **Python version:** 3.12

## Steps to Reproduce

1. `/goal write a hello world script`
2. Wait for agent to finish and call `update_goal(status="complete")`
3. Send a new message: `"hey"`
4. Agent returns `TERMINATE` instead of chatting — repeats on every subsequent message

## Actual vs Expected

- **Actual:** After goal completion, normal chat is impossible — every message gets `TERMINATE`
- **Expected:** After goal completion, session is cleaned up automatically and the next message chats normally

## Additional Notes

Fix is straightforward — `pop` the session from `_sessions` in three exit paths: `update_goal` complete/blocked, `GoalTurnGate` detecting `active=False`, and max_iterations exceeded. 2 files,

rayrayraykk:
Great thanks, I will look into this issue.

rayrayraykk:
**Root cause:** `update_goal("complete")` sets `session.active = False` but never removes the session from `GoalMode._sessions`. On subsequent messages, `GoalTurnGate.check()` finds the stale session via `session_by_ctx_var()` and returns `TERMINATE("Goal completed: satisfied")` every time.

**Fix (PR #6086):**
- Add `GoalMode.deactivate()` — pops the current session from `_sessions`
- Call it from `update_goal("complete")` after setting `active=False`
- Also renamed `first_active_session()` → `active_session()` for clarity

2 files changed, ~30 lines added, ~20 removed.

rayrayraykk:
**Why deactivating mid-loop is safe:**

[...]

PR #6094, opened 2026-07-14
fix(goal): reset stop gates on /new and /clear commands

rayrayraykk:
## Description

`/new` and `/clear` reset conversation history and summary but did not reset loop gate state (iteration counters, doom loop history, pending stop/continue). This caused stale gate state to bleed into new conversations — e.g. a completed goal's `TERMINATE` persisting after `/new`.

Cherry-picked from #5885 by @FIELA, with a follow-up commit for lint compliance.

**Changes:**

1. `StopHandler.reset()` — iterates all registered gates and calls `gate.reset()`
2. `CommandHandler._reset_stop_gates()` — resets handler gates + clears `agent._gate_pending_stop` / `_gate_pending_continue`
3. Called from all three exit paths of `_process_new` and `_process_clear`

**Commits:**

1. `2ec492c7` — Cherry-pick of PR #5885 (author: @FIELA)
2. `0adc5cbc` — Lint fixes (pylint/flake8 compliance)

**Related Issue:** Relates to #6082

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

```bash
$ pytest tests/unit/agents/test_command_handler.py -v
14 passed in 0.34s
```

## Evidence

```
pre-commit run --files <all changed files>
All checks passed
```

### DIFF
--- a/src/qwenpaw/agents/command_handler.py
+++ b/src/qwenpaw/agents/command_handler.py
@@ -207,19 +207,35 @@ class CommandHandler(ConversationCommandHandlerMixin):
         """Write the rolling compaction summary."""
         self._state.summary = value or ""
 
-    def _reset_stop_gates(self) -> None:
-        """Clear loop-gate state for commands that reset conversation state."""
-        workspace = getattr(self._prompt_context, "workspace", None)
-        handler = getattr(workspace, "_stop_handler", None)
-        reset = getattr(handler, "reset", None)
-        if callable(reset):
-            reset()
+    def _reset_stop_gates(  # pylint: disable=protected-access
+        self,
+    ) -> None:
+        """Clear loop-gate state on conversation reset."""
+        workspace = getattr(
+            self._prompt_context,
+            "workspace",
+            None,
+        )
+        handler = getattr(
+            workspace,
+            "_stop_handler",
+            None,
+        )
+        if handler is not None:
+            handler.reset()
 
-        agent = self._agent or getattr(self._prompt_context, "agent", None)
+        agent = self._agent or getattr(
+            self._prompt_context,
+            "agent",
+            None,
+        )
         if agent is not None:
             if hasattr(agent, "_gate_pending_stop"):
                 agent._gate_pending_stop = None
-            if hasattr(agent, "_gate_pending_continue"):
+            if hasattr(
+                agent,
+                "_gate_pending_continue",
+            ):
                 agent._gate_pending_continue = None
 
     def is_command(self, query: str | None) -> bool:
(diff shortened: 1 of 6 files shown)


jinliyl:
Thank you for working on this fix. I took a closer look at the gate lifecycle and found one case that appears to remain unresolved by the current implementation.

`StopHandler.reset()` delegates to `reset()` on each registered gate. However, the goal-mode gates (`GoalTurnGate`, `GoalBudgetGate`, and `RubricGate`) do not override `reset()`, so they inherit the no-op implementation from `StopGate`. Their effective state is owned by `GoalMode._sessions`, and that state is not cleared by either `StopHandler.reset()` or `CommandHandler._reset_stop_gates()`.

As a result, a terminal goal session can survive `/new` or `/clear`. For example, sessions made inactive by the iteration limit, token budget, or blocked status remain in `GoalMode._sessions`. On the next request, `GoalTurnGate.check()` retrieves that inactive session and returns `TERMINATE` again:

```python
set_current_session_id("s1")

mode = GoalMode()
mode.sessions["s1"] = GoalSession(
    goal="done",
    active=False,
    last_verdict="satisfied",
)

handler = StopHandler()
gate = GoalTurnGate(mode)
handler.register(gate)

handler.reset()
result = await gate.check({})

assert mode.session_by_ctx_var() is not None
assert result.action == StopAction.TERMINATE
```

I reproduced this against the current PR head (`134c5a2d`); the observed result after `handler.reset()` was:

```text
session_still_present= True
action_after_reset= terminate
reason= Goal completed: satisfied
```

The newly added tests pass, but they use a mocked stop handler and therefore verify only that `reset()` is invoked, not that the real goal-owned session state is removed.

One implementation detail worth considering is that the existing gate `reset()` operation is also used for ordinary new-user-turn/sub-turn resets. Therefore, changing the default `LoopGate.reset()` to always deactivate the current session could unintentionally cancel a live goal on every turn. A separate conversation/session reset operation, or an explicit goal-mode cleanup hook that removes the current entry from `GoalMode._sessions`, may keep these two lifecycle semantics distinct.

Would you please add a regression test using the real `GoalMode`/`GoalTurnGate` and ensure that `/new` and `/clear` remove the current terminal goal state? This seems important because stale goal termination is one of the primary behaviors described in the PR motivation.

xieyxclack (approved):

PR #6096, opened 2026-07-14
feat(channels): split tool call and output rendering controls

### LINKED ISSUE #5976: [Feature]: 可以分开设置channel的工具调用参数信息和调用结果信息的发送
每次channel的工具调用结果信息太长了，希望可以单独控制结果是否发送

另外希望工具调用结果可以截断，显示前几行和后几行发送给chennel，用户预览大致信息即可

qiuC123:
## Description

This PR adds channel-level rendering controls for tool call arguments and tool outputs. It retires the legacy `filter_tool_messages` config surface in favor of independent `filter_tool_calls` and `filter_tool_outputs` fields, while migrating existing configs on load.

**Related Issue:** Fixes #5976

**Security Considerations:** This change only affects channel rendering visibility and preview length. It does not add new external calls, permissions, credential handling, or tool execution paths.

## What Problem This Solves

Channel rendering previously had a single `filter_tool_messages` switch. When enabled, it hid both the tool call arguments and the tool output together. That made it hard for a channel to show what tool the agent chose while suppressing long or sensitive outputs, or to hide tool calls while still showing useful results.

Tool outputs were also truncated with a fixed renderer limit, so channel owners could not tune previews for different channel UX constraints. This PR separates those controls, exposes them in the Console UI, and keeps legacy configs compatible through a one-shot migration.

## Evidence

I verified the renderer and migration behavior with focused unit tests covering:

- `filter_tool_calls=True` hides tool call arguments while keeping tool outputs visible.
- `filter_tool_outputs=True` hides tool outputs while keeping tool calls visible.
- `tool_call_args_limit` controls the visible tool call argument preview length.
- `tool_output_head_chars` and `tool_output_tail_chars` produce configurable head/tail output previews.
- List-structured tool output text remains untruncated by default, preserving previous behavior.
- Legacy `filter_tool_messages` config is removed and migrated to the split filters when active.

Local verification passed with `43 passed` for the focused Python tests, and the Console TypeScript/Vite build completed successfully.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Changes

- Add separate `filter_tool_calls` and `filter_tool_outputs` channel config fields.
- Add configurable `tool_call_args_limit`, `tool_output_head_chars`, and `tool_output_tail_chars`.
- Add one-shot config migration from legacy `filter_tool_messages`.
- Remove `filter_tool_messages` from the new config and Console UI surface.
- Preserve default behavior for list-structured tool output text/thinking blocks.
- Wire the new controls into Console channel types, drawer fields, and i18n strings.
- Add focused renderer and config migration tests.

## Checklist

- [x] Ready for review

## Testing

- `.\.venv\Scripts\python.exe -m py_compile src/qwenpaw/app/channels/renderer.py src/qwenpaw/app/channels/base.py src/qwenpaw/app/channels/manager.py src/qwenpaw/config/config.py tests/unit/app/channels/test_renderer.py tests/unit/config/test_channel_tool_filter_migration.py`
- `.\.venv\Scripts\python.exe -m pytest tests/unit/app/channels/test_renderer.py tests/unit/config/test_channel_tool_filter_migration.py -q`
- `cd console && npm.cmd run build -- --mode test`

Result: `43 passed` for focused Python tests. Console build completed successfully.

Notes:

- Pytest emitted a local cache permission warning for `.pytest_cache`, which does not affect the test result.
- Console build emitted existing chunk/dynamic import size warnings, but completed successfully.

```bash
.\.venv\Scripts\python.exe -m py_compile src/qwenpaw/app/channels/renderer.py src/qwenpaw/app/channels/base.py src/qwenpaw/app/channels/manager.py src/qwenpaw/config/config.py tests/unit/app/channels/test_renderer.py tests/unit/config/test_channel_tool_filter_migration.py

.\.venv\Scripts\python.exe -m pytest tests/unit/app/channels/test_renderer.py tests/unit/config/test_channel_tool_filter_migration.py -q
# 43 passed, 1 warning

cd console
npm.cmd run build -- --mode test
# built successfully
```

## Additional Notes

The legacy `filter_tool_messages` constructor path is still accepted at the channel base layer for compatibility, but it is no longer exposed as a config/UI field. Existing `agent.json` configs are migrated to the new split fields when loaded.


hongxicheng:
Thanks for the contribution. After reviewing it, though, I think the implementation still has some gaps in its logic, and I'd like to share my findings along with the design changes I think we should make.

The overall direction is reasonable — splitting the single `filter_tool_messages` switch into independent call/output controls plus configurable truncation maps directly onto both asks in #5976. That said, here are a few issues I ran into:

1. **Frontend isn't wired up.** The new fields land in `BaseChannelConfig`, so they work if you edit `agent.json` / hit the API directly, but the Console UI (`channel.ts` types, `BASE_FIELDS`, `ChannelDrawer`, `index.tsx`, i18n) still only exposes `filter_tool_messages` / `filter_thinking`. So users can't actually configure the new knobs from the UI yet.

2. **A default-behavior change slipped in.** The PR says it doesn't change default behavior, but text/thinking blocks inside *list-structured* tool outputs used to be sent in full and are now truncated to `tool_output_head_chars=500` by default. Tools that return rich/structured output (web search, fetch, etc.) will silently lose content that users saw before. Worth either preserving the old default for list blocks or calling this out explicitly.

3. Minor: `manager.py` now both passes the fields via `from_config` kwargs *and* re-applies them via `apply_render_config`, which is redundant; and docs/tests aren't updated yet.

**On the design itself:** rather than keeping `filter_tool_messages` around as a third overlapping switch, I think we should fully retire it in favor of the two independent booleans, and handle old configs with a one-shot startup migration. The repo already has this pattern (the `weixin → wechat` rename and the `dm_policy/allow_from → access_control_*` migration in `config.py`): on load, if `filter_tool_messages` is present, drop it and set `filter_tool_calls` / `filter_tool_outputs` accordingly (following the existing convention of only writing the new field when it's the active/`true` case). That keeps the config surface minimal and avoids three switches with overlapping semantics.

This is just my initial pass on the PR. If you have the bandwidth, feel free to take these ideas and iterate on it. If you're busy or would rather not, no problem at all — feel free to leave this one on hold and I'm happy to pick up the feature and carry it forward from here. Either way, thanks again for kicking this off.

qiuC123:
Thanks for the detailed review. I pushed an update addressing the points you raised:

- Wired the new controls into the Console UI/types/drawer/i18n.
- Retired `filter_tool_messages` from the config/UI surface and added a one-shot startup migration to `filter_tool_calls` / `filter_tool_outputs`.
- Preserved the old default behavior for list-structured tool output text/thinking blocks; they remain untruncated unless the new truncation settings are changed.
- Removed the redundant manager pass-through and now applies render config in one place after channel construction.
- Added focused tests for renderer behavior and legacy config migration.

Validation:

- `python -m py_compile ...`
- `pytest tests/unit/app/channels/test_renderer.py tests/unit/config/test_channel_tool_filter_migration.py -q` (`43 passed`)
- `cd console && npm.cmd run build -- --mode test`

hongxicheng:
Thanks for iterating on this based on my earlier review — the updated approach is heading in the right direction.

That said, while digging deeper I found a couple of things that make this harder to land as-is. A few channels don't render through the shared `renderer.py` path: XiaoYi implements its own part extraction (it emits a channel-specific `kind` protocol), and Slack has an independent streaming layer. Both currently re-implement the tool/thinking filtering policy locally, which duplicates the logic and can silently drift out of sync with the renderer. Beyond centralizing the shared policy, these channels will also need their own rendering logic reworked so they align with the unified approach rather than carrying parallel implementations. There are also some related rough edges in how the render config is threaded through channel construction.

Cleaning this up properly means centralizing the filtering policy and reworking how render config is applied across every channel, so all channels stay functional and the logic stays clean and non-redundant. That's a fairly broad change touching all channels and needs thorough testing to make sure each one keeps working correctly.

Given the scope, I'm going to close this PR and open a new one shortly that refactors the channel modules as a whole. This isn't a reflection on your work — the direction here was genuinely helpful in surfacing the underlying issues.

Thanks again for your support of QwenPaw, and please keep the contributions coming!

PR #6102, opened 2026-07-14
test(isolation): boundary meta-test pinning the #5813 failure modes

hanson-hex:
## Description

What problem this solves: the #5813 class of "passes in isolation, fails in the full suite" test failures. Two static patterns caused multi-hour debugging sessions across this batch:

1. A test that imports a side-effectful module (`qwenpaw.app._app`) leaks it into `sys.modules` and poisons later tests on the same `pytest -n auto --dist=loadscope` worker (e.g. `test_entry._ensure_qwenpaw_app_not_loaded` then fails).
2. An autouse isolation fixture that blanket-deletes a whole module family (`for name in sys.modules: if name.startswith("qwenpaw.app"): del`) re-imports siblings (`agent_context`) as NEW module objects, so monkeypatches in other tests land on the old object and silently break (this was #5813's second bite — broke `test_token_usage`).

These are static properties of a test file. This PR adds a meta-test (`tests/unit/test_isolation_boundary.py`) that scans `tests/unit` and fails if either pattern appears, so the next contributor copying either is caught before CI.

Also adds a bullet to CONTRIBUTING.md: `fix:` PRs must include a regression test. (Earlier version mentioned `CLAUDE.md`/`AGENTS.md`; those files are intentionally excluded from the repo by `.gitignore`, so the pointer is removed after reviewer feedback.)

No production code changed.

## Component(s) Affected
- [ ] Core / Backend
- [ ] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [x] Refactoring

## Checklist
- [x] I ran `pre-commit run` locally and it passes
- [x] Documentation updated (CONTRIBUTING.md regression-test bullet)
- [x] Ready for review

## Testing

- `pytest tests/unit/test_isolation_boundary.py -q` → 2 passed
- Crafted a violator file (fixture deleting `qwenpaw.app.*` by prefix) → the meta-test flagged it as expected; removed the violator.
- Full `pytest tests/unit -q --cov` on main → 4738 passed, the meta-test ran green inside the full suite without polluting others. (Baseline measured on main at PR creation, 2026-07-14; #6489 adds 76 drivers cases, so the pass count and TOTAL% will rise on rebase — the meta-test itself is baseline-independent.)

## Evidence

The branch was rebased onto current main (14141d73) and the meta-test was verified both in isolation and as part of the full suite using the repo `.venv`:

```bash
$ pytest tests/unit/test_isolation_boundary.py -q --no-header
..                                                                       [100%]
2 passed in 0.83s

$ pytest tests/unit/test_isolation_boundary.py -q
FAILED tests/unit/test_isolation_boundary.py::test_no_isolation_fixture_deletes_module_family_by_prefix
1 failed, 1 passed

$ pytest tests/unit -q --cov=qwenpaw
4738 passed, 6 skipped
TOTAL 46%
```

## Additional Notes

- Heavy-module isolation fixture may be declared in the test file itself or in a `conftest.py`; both forms satisfy the meta-test (the assertion message is updated to mention `conftest.py` explicitly after reviewer feedback).
- Coexists with #6489 (drivers unit tests, +76 cases; disjoint files). The meta-test scans all of `tests/unit/` including the new `tests/unit/drivers/` cases; on rebase, confirm the drivers tests don't import `qwenpaw.app._app` or do prefix `sys.modules` deletes (they target `qwenpaw.drivers.*`, so they should be clean — verify, don't assume).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -211,8 +211,7 @@ If you add or change platform support, please test on the affected OS and mentio
 - **Include a regression test for bug fixes** — a `fix:` PR should include a
   test that fails before the fix and passes after, pinning the behavior
   contract so it doesn't silently regress. Trivial fixes (typos, dep bumps,
-  docs) are exempt. AI agents and contributors: see `CLAUDE.md`/`AGENTS.md`
-  for the full test-isolation and CI-matrix rules.
+  docs) are exempt.
 - Update documentation for user-facing changes.
 - Use conventional commit messages and PR titles.
 - Be respectful and constructive (we follow a welcoming Code of Conduct).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -208,6 +208,11 @@ If you add or change platform support, please test on the affected OS and mentio
 - Start with small, focused changes.
 - Discuss large or design-sensitive changes in an issue first.
 - Write or update tests where applicable.
+- **Include a regression test for bug fixes** — a `fix:` PR should include a
+  test that fails before the fix and passes after, pinning the behavior
+  contract so it doesn't silently regress. Trivial fixes (typos, dep bumps,
+  docs) are exempt. AI agents and contributors: see `CLAUDE.md`/`AGENTS.md`
+  for the full test-isolation and CI-matrix rules.
 - Update documentation for user-facing changes.
 - Use conventional commit messages and PR titles.
 - Be respectful and constructive (we follow a welcoming Code of Conduct).
--- a/tests/unit/test_isolation_boundary.py
+++ b/tests/unit/test_isolation_boundary.py
@@ -28,8 +28,6 @@ import ast
 import re
 from pathlib import Path
 
-import pytest
-
 # Modules whose import has strong global side-effects and which later tests
 # assert the absence of (see tauri/test_entry.py::_ensure_qwenpaw_app_not_loaded).
 # A test file importing any of these MUST isolate with an autouse fixture.
@@ -39,7 +37,7 @@ HEAVY_MODULES: frozenset[str] = frozenset({"qwenpaw.app._app"})
 # specific name) is the #5813 footgun: it re-creates sibling module objects.
 # e.g. `for name in list(sys.modules): if name.startswith("qwenpaw.app"): del`
 _PREFIX_DEL_PATTERN = re.compile(
-    r"\.startswith\(\s*[\"']qwenpaw\.[^\"']+[\"']\s*\)"
+    r"\.startswith\(\s*[\"']qwenpaw\.[^\"']+[\"']\s*\)",
 )
 
 TESTS_UNIT = Path(__file__).resolve().parent
@@ -139,11 +137,12 @@ def test_files_importing_heavy_module_isolate_with_autouse() -> None:
         if _imports_heavy_module(path) and not _has_autouse_fixture(path):
             offenders.append(
                 f"{path.relative_to(TESTS_UNIT)}: imports a heavy module "
-                f"({HEAVY_MODULES}) without an autouse isolation fixture"
+                f"({HEAVY_MODULES}) without an autouse isolation fixture",
             )
     assert not offenders, (
         "Tests importing heavy/side-effectful modules must isolate with an "
-        "autouse fixture that restores sys.modules on teardown:\n  - "
+        "autouse fixture that restores sys.modules on teardown "
+        "(declared in the test file itself or in a conftest.py):\n  - "
         + "\n  - ".join(offenders)
     )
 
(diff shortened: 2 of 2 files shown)


yutai78786 (commented):
Really like the direction here — pinning the two #5813 isolation failure modes as a static meta-test turns multi-hour debugging sessions into an instant CI signal. The AST-based heavy-module check plus the self-exclusion trick (the meta-test contains the regex it scans for) show good care.

Two small things worth a look:

1. **Dead pointer in CONTRIBUTING.md**: the new bullet references `CLAUDE.md`/`AGENTS.md`, but neither file exists in the repo (they're git-ignored), so contributors following the pointer will hit a dead end. Suggest dropping that sentence or pointing at a section within CONTRIBUTING itself.

2. **Autouse fixtures in conftest.py**: `_has_autouse_fixture` only detects fixtures declared inside the test file itself. A contributor who puts a perfectly valid isolation fixture in `conftest.py` would still be flagged — the error message could mention that case so nobody is confused when their conftest-based isolation gets reported.

Neither is a big deal, but the first one especially is worth fixing since it affects every new contributor reading the guide. Thanks!

RerankerGuo:
Independent validation on macOS 26.5.1 arm64, Python 3.11.15, pytest 9.1.1.

The current meta-test has a false negative: any in-file `autouse=True` fixture satisfies `_has_autouse_fixture()`, even when that fixture does not isolate `qwenpaw.app._app` or touch `sys.modules` at all.

## Minimal bypass

I added this temporary file under `tests/unit/`:

```python
import pytest

import qwenpaw.app._app


@pytest.fixture(autouse=True)
def unrelated_environment_fixture(monkeypatch):
    monkeypatch.setenv("SYNTHETIC_FLAG", "1")
    yield


def test_placeholder():
    assert qwenpaw.app._app is not None
```

This is exactly the hazardous heavy-module import the new boundary intends to reject. Its only autouse fixture changes an unrelated environment variable and leaves the imported app module in `sys.modules`.

At exact head `de074bc5a2f9e2581adfc874dc7a1c5e82d45f30`, direct classification showed:

```text
imports_heavy_module=True
has_autouse_fixture=True
```

The meta-test then remained green:

```text
pytest tests/unit/test_isolation_boundary.py -q
2 passed
```

The same false negative reproduces after a clean merge into current main:

```text
main:         bacac7410cf2ac122dd892152beff8b03f38729f
trial commit: d570bd9fea7b18b097dacf8556193ae22806f50f
trial tree:   27082642312ec7e1615615060daeeaba5ebddb51
imports_heavy_module=True
has_autouse_fixture=True
2 passed
```

## Positive validation

After removing the synthetic file, both exact head and the current-main trial pass the intended isolation and order-sensitive suites:

```text
pytest \
  tests/unit/test_isolation_boundary.py \
  tests/unit/runtime/test_install_smoke.py \
  tests/unit/tauri/test_entry.py -q
16 passed
```

The trial also passes add-trailing-comma, mypy, Black, Flake8, Pylint, and `git diff --check`.

## Requested fix boundary

Please make the predicate prove module isolation rather than only the presence of an arbitrary autouse fixture. For example, the AST check could require the qualifying fixture to snapshot/remove/restore each imported `HEAVY_MODULES` entry, or the isolation contract could be enforced dynamically in a subprocess/postcondition instead of inferred from the decorator alone.

Please add the synthetic unrelated-autouse case as a negative-control regression. Without it, a future test can copy the original leak and silence the new gate by already having any unrelated autouse setup fixture.

RerankerGuo (changes requested):
作者意图：用静态 meta-test 阻止 unit tests 导入重型 app 模块后污染 `sys.modules`，固定 #5813 的顺序依赖故障。

当前 predicate 只检查文件内是否存在任意 `autouse=True` fixture，并不证明 fixture 隔离了重型模块。一个只设置无关环境变量的 autouse fixture 就能让危险 import 通过 gate。

```mermaid
flowchart LR
 A[Test imports qwenpaw.app._app] --> B{Any autouse fixture exists?}
 B -->|yes| C[Gate treats file as isolated]
 C --> D[Heavy module remains in sys.modules]
 D --> E[Order-dependent suite contamination]
```

最小 negative control 在 exact head 和 current-main trial 上均得到：

```text
imports_heavy_module=True
has_autouse_fixture=True
isolation meta-tests: 2 passed
```

该 fixture 仅 `monkeypatch.setenv()`，从未 snapshot/remove/restore `qwenpaw.app._app`。移除 synthetic file 后，原有 16 项 isolation/order tests 与全部 hooks 通过，因此 blocker 是 classifier false negative。

请让 predicate 证明实际 module isolation，例如 AST 要求 qualifying fixture 对每个 imported heavy module 执行 snapshot/remove/restore，或改用 subprocess/postcondition 动态验证。必须提交“heavy import + unrelated autouse fixture”这一 negative-control regression，确保 gate 会失败。

Issue #5996, opened 2026-07-12
[Bug]: 2.0.0对话时会产生MODEL_EXECUTION_ERROR


kukuegg:
QwenPaw 的 _hint.py 中 make_offload_hint_msg() 创建了这样的 hint 消息：

assistant: [系统通知 TextBlock] + [ToolResultBlock]
当 OpenAI formatter 序列化时，ToolResultBlock 被转成 role=tool 消息，但当前 assistant 消息中没有 tool_calls，导致 OpenAI API 返回 400 错误：

"Messages with role 'tool' must be a response to a preceding message with 'tool_calls'"

zhuzhuppx:
加1奔溃了，一会儿一清理会话，才行

ningblue:
+1

ltnetcase:
+1

anye3508:
+1,而且感觉不如之前的版本好用了。有时候任务总会在那循环执行。

wananing:
I reproduced the failure path and would like to work on this.

`make_offload_hint_msg()` currently creates an assistant message containing a `TextBlock` followed by a `ToolResultBlock`, but no matching `ToolCallBlock`. AgentScope formatters therefore serialize an orphan tool result; with the OpenAI formatter this becomes a `role=tool` message without preceding `tool_calls`, which the API rejects.

I plan to fix this in the shared offload-hint builder rather than adding an OpenAI-specific workaround: add a matching synthetic tool call using the same call ID/name, preserve the finalized result payload, and add formatter-level regression coverage for OpenAI-compatible, Anthropic, and Gemini message shapes.

ymofen:
+1， 照理说这个是底层的执行交互，为什么还是会出现这种问题？难道之前的1.x版本也是这样？

rayrayraykk:
We've done a full root-cause analysis using two independent error dumps and confirmed the same failure path described above.

**Root cause summary:**

`make_offload_hint_msg()` builds an assistant message with `[TextBlock, ToolResultBlock]` but no `ToolCallBlock`. The OpenAI formatter flushes the `TextBlock` as an assistant message (without `tool_calls`), then emits a `role=tool` message for the `ToolResultBlock` — violating the API contract. The `_sanitize_tool_messages` pass does not catch this because the duplicate `tool_call_id` from the original offloaded interaction satisfies the surviving-ID check.

**Proposed fix — flatten content blocks (alternative to #6004):**

[...]

PR #6108, opened 2026-07-14
fix(context): keep tool results paired with assistant calls during context compression

nguyenthanhthe:
## What Problem This Solves

When using DeepSeek V4 Pro or other OpenAI-compatible API backends in long conversations, triggering QwenPaw's context compression (inside the scroll manager) could lead to a permanent `400 Bad Request` error:
`Messages with role 'tool' must be a response to a preceding message with 'tool_calls'`

This occurs because `split_for_compression` splits the live context into `to_compress` and `to_reserve` purely based on token count, which can split right between the `assistant(tool_calls)` message and the following `tool` result messages. This leaves orphaned `tool` role messages at the beginning of the reserved context, breaking the strict format constraints of the DeepSeek API.

This PR fixes this by post-processing the split lists in `_as_internals.py`. If the reserved list starts with a `"tool"` message, it shifts messages from the end of `to_compress` to `to_reserve` until it starts with a non-tool message (the corresponding `assistant` message), maintaining the pairing integrity.

## Evidence

I added a new unit test in `test_as_internals_signatures.py` (`test_split_for_compression_keeps_tool_messages_with_assistant`) which mocks the split and asserts that the `assistant` message is correctly moved to keep the tool response paired.

All tests and pre-commit checks pass:
```bash
pytest tests/unit/agents/context/test_as_internals_signatures.py
pre-commit run --files src/qwenpaw/agents/context/scroll/_as_internals.py tests/unit/agents/context/test_as_internals_signatures.py
```

Output:
```
tests\unit\agents\context\test_as_internals_signatures.py ....           [100%]
======================== 4 passed in 0.62s =========================
```

### DIFF
--- a/src/qwenpaw/agents/context/scroll/_as_internals.py
+++ b/src/qwenpaw/agents/context/scroll/_as_internals.py
@@ -32,4 +32,22 @@ async def split_for_compression(
     """Pairing-safe split of the live context into (to_compress, to_reserve),
     reserving roughly ``reserve`` tokens of the recent tail."""
     # pylint: disable-next=protected-access
-    return await agent._split_context_for_compression(reserve, tools)
+    to_compress, to_reserve = await agent._split_context_for_compression(
+        reserve,
+        tools,
+    )
+
+    # DeepSeek/OpenAI constraint: tool messages must be preceded by
+    # assistant message with tool_calls. Shift messages from to_compress
+    # to to_reserve to prevent orphaned tool messages at start of reserve.
+    while to_reserve and to_compress:
+        first_msg = to_reserve[0]
+        role = getattr(first_msg, "role", None) or (
+            first_msg.get("role") if isinstance(first_msg, dict) else None
+        )
+        if role == "tool":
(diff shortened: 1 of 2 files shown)


niceIrene:
Thanks for digging into this!
This class of orphan tool_result errors should already be covered by #5989 and #5996. If the issue persists after those changes, that would be important to know.
One note on the implementation: the proposed role == "tool" check doesn't match AgentScope 2.0's in-memory representation. Tool calls and tool results live as content blocks inside an assistant Msg; the wire-level tool role is only produced later by the formatter. So this check would never fire against in-memory messages, even when an orphan actually exists.
To move this forward, could you:

Confirm the failure still reproduces on latest main (with #5989 and #5996 included)?
Share the actual provider error message or a sanitized dump of the offending context?

nguyenthanhthe:
Thank you for the clarification! Yes, the issue is already fully resolved by the multi-layer orphan tool_result message defense (#5989). Closing this PR.

PR #6109, opened 2026-07-14
fix(governance): honor sandbox_enabled switch in OFF-mode sandbox path

vanwaals:
fix(governance): honor sandbox_enabled switch in OFF-mode sandbox path

`_prepare_off_mode_sandbox` only checked the platform probe (`sandbox_available`), ignoring the global `security.sandbox_enabled` switch. With `approval_level=OFF` + `sandbox_enabled=false` the REPL was still forced into a sandbox, silently overriding the user's opt-out.

Use `governor.sandbox_usable` (platform support AND global switch), matching the normal policy path (`_sandbox_usable`). Document that tool registration in `_scroll_repl_runnable` is a build-time decision (not hot-reloaded).

Tests: add OFF-mode switch-off no-op case, real-governor hot-reload cases, and the missing HTTP roundtrip for `PUT`/`GET /security/sandbox`.

## Description

The global `security.sandbox_enabled` switch is meant to globally control whether sandbox execution is active. On the **normal policy path**, `assert_policy` → `_sandbox_usable()` correctly degrades `SANDBOX_FALLBACK` to unsandboxed `ALLOW` when the switch is off.

However, the **OFF-mode path** (`_prepare_off_mode_sandbox` in `tool_adapter.py`) never consulted the switch — it only checked `governor.sandbox_available` (a one-time platform probe). As a result:

- User sets `approval_level=OFF` and `sandbox_enabled=false`
- The REPL tool still gets a `sandbox_config` injected and runs inside the sandbox
- The user's explicit "sandbox off" preference is silently ignored

This PR closes that inconsistency so both code paths evaluate sandbox eligibility the same way.

**Root cause**

```python
# tool_adapter.py (before)
def _prepare_off_mode_sandbox(tool, governor):
    if governor is None or not getattr(governor, "sandbox_available", False):
        return
    # ↑ platform capability only — never reads sandbox_enabled
```

vs. the normal policy path:

```python
# resource_governor.py
def _sandbox_usable(self) -> bool:
    return self._sandbox_available and self._sandbox_globally_enabled()
    # ↑ platform capability AND the user's sandbox_enabled switch
```

**Fix**

```python
# tool_adapter.py (after)
if governor is None or not getattr(governor, "sandbox_usable", False):
    return
```

`sandbox_usable` is the existing public property that combines the platform probe with the live, mtime-cached `sandbox_enabled` switch — so an explicit `sandbox_enabled=false` is now honored on the OFF-mode path too. Using the public property (rather than the private `_sandbox_usable()`) avoids reaching into private members while keeping the `getattr` fallback for duck-typed governors.

**Secondary items from the report**

1. *Tool registration is not hot-reloadable.* `_scroll_repl_runnable` in `builder.py` already reads `sandbox_usable`, but registration is a build-time decision — toggling `sandbox_enabled` at runtime does not add/remove the REPL from a running agent. This is now documented in the method docstring; the execution path still honors the switch.
2. *No hot-reload test.* Added coverage (see Testing).

**Related Issue:** Relates to the OFF-mode sandbox switch inconsistency report.

**Security Considerations:** This is a fail-safe direction change. Previously the REPL was *more* restrictive than requested (forced sandbox); the fix makes it respect the user's explicit opt-out. Fail-open shell tools (e.g. `Bash`) are deliberately left untouched — they already run unsandboxed by design in OFF mode. A compile failure still leaves `sandbox_config` unset, so the tool's own fail-closed guard denies rather than running unsandboxed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

> N/A — no channel code changed.

## Testing

Unit tests (`tests/unit/governance/test_off_mode_sandbox.py`):

- `test_sandbox_switch_off_is_noop` — `sandbox_available=True` + `sandbox_enabled=false` ⇒ OFF-mode provisioning is skipped (no `sandbox_config`).
- `TestSandboxSwitchHotReload::test_sandbox_usable_follows_switch_without_restart` — with a **real** `ResourceGovernor`, flipping the switch changes `sandbox_usable` on the next read, no `start()`/restart needed.
- `TestSandboxSwitchHotReload::test_off_mode_provisioning_skipped_when_switch_off` — end-to-end with a real governor: OFF-mode path skips provisioning once the switch is off.
- Existing cases (REPL gets sandbox, `Bash` untouched, no-platform no-op, no-governor no-op, compile-failure fails closed) still pass.

Integration test (`tests/integration/test_security_config.py`):

- `test_global_sandbox_put_get_roundtrip` — fills the previously missing HTTP roundtrip for `GET`/`PUT /api/config/security/sandbox`, asserting the switch flips both directions and is readback-consistent (hot-reload via the mtime-cached config).

How to run:

```bash
pytest tests/unit/governance/test_off_mode_sandbox.py -q
pytest tests/integration/test_security_config.py -q -k sandbox
```

## Evidence

Verified locally on macOS (Python 3.11) against branch `fix/off-mode-sandbox-enabled`:

- **pre-commit** on all four changed files: every hook (black, flake8, pylint, mypy, ast, trailing-comma, etc.) reports **Passed** — no auto-fixes required.
- **pytest** for `tests/unit/governance/test_off_mode_sandbox.py`: **11 passed**, including the new switch-off no-op case and the two real-`ResourceGovernor` hot-reload cases. The existing OFF-mode cases (REPL sandboxed, `Bash` untouched, no-platform no-op, compile-failure fails closed) remain green, confirming no regression.

The raw transcripts are below.

```bash
$ pre-commit run --files \
    src/qwenpaw/governance/tool_adapter.py \
    src/qwenpaw/runtime/builder.py \
    tests/unit/governance/test_off_mode_sandbox.py \
    tests/integration/test_security_config.py
```

```bash
$ pytest tests/unit/governance/test_off_mode_sandbox.py -q
11 passed, 2 warnings in 0.04s
```

## Additional Notes

Files changed (4 files, +195 / -5):

| File | Change |
|------|--------|
| `src/qwenpaw/governance/tool_adapter.py` | Core fix: `sandbox_available` → `sandbox_usable` compound check + docstring |
| `src/qwenpaw/runtime/builder.py` | Document that REPL registration is a build-time decision (not hot-reloaded) |
| `tests/unit/governance/test_off_mode_sandbox.py` | Switch-off no-op + real-governor hot-reload cases |
| `tests/integration/test_security_config.py` | Missing HTTP roundtrip for `PUT`/`GET /security/sandbox` |

### DIFF
--- a/src/qwenpaw/governance/tool_adapter.py
+++ b/src/qwenpaw/governance/tool_adapter.py
@@ -194,11 +194,19 @@ def _prepare_off_mode_sandbox(tool: Any, governor: Any) -> None:
     unsandboxed by design, and forcing a sandbox on them would silently narrow
     their filesystem access.
 
-    A no-op (leaving ``sandbox_config=None``) when the platform has no sandbox:
-    such a REPL is never registered in the first place, or it tolerates
-    unsandboxed execution via ``allow_unsandboxed``.
+    A no-op (leaving ``sandbox_config=None``) when the sandbox is not usable
+    -- either the platform has no sandbox, or the operator turned the global
+    ``security.sandbox_enabled`` switch off. In both cases such a REPL is
+    never registered in the first place, or it tolerates unsandboxed
+    execution via ``allow_unsandboxed``.
+
+    Uses ``governor.sandbox_usable`` (platform support AND the global switch)
+    rather than the platform-only ``sandbox_available`` probe, so that an
+    explicit ``sandbox_enabled=false`` is honoured on the OFF-mode path just
+    like it is on the normal policy path (see
+    :meth:`ResourceGovernor._sandbox_usable`).
     """
-    if governor is None or not getattr(governor, "sandbox_available", False):
+    if governor is None or not getattr(governor, "sandbox_usable", False):
         return
     policy_name = DEFAULT_REGISTRY.python_to_policy_name(
         getattr(tool, "name", "Unknown"),
--- a/src/qwenpaw/runtime/builder.py
+++ b/src/qwenpaw/runtime/builder.py
@@ -734,6 +734,14 @@ class AgentBuilder:
         tool, which needs no sandbox. This is narrower than
         :meth:`_scroll_recall_runnable`, which gates whether scroll is wired at
         all; here scroll is already wired and structured recall is present.
+
+        Note: this is a build-time decision. Registration is evaluated once
+        when the agent is built; toggling ``security.sandbox_enabled`` at
+        runtime does NOT add or remove the REPL from an already-running
+        agent. The switch is still honoured on the execution path
+        (``_prepare_off_mode_sandbox`` / ``_sandbox_usable``) -- an already
+        registered REPL will simply skip sandbox provisioning once the switch
+        is off. Rebuild the agent to change which tools are offered.
         """
         if governor is not None:
             sandbox_usable = getattr(governor, "sandbox_usable", None)
--- a/tests/integration/test_security_config.py
+++ b/tests/integration/test_security_config.py
@@ -7,6 +7,76 @@ import pytest
 
 @pytest.mark.integration
 @pytest.mark.p0
+def test_global_sandbox_put_get_roundtrip(app_server) -> None:
+    """Test purpose:
+    - Verify the global sandbox switch supports update and readback, and that
+      the change is hot-reloaded (no restart) via the mtime-cached config.
+
+    Test flow:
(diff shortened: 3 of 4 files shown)


cuiyuebing (approved):
LGTM

PR #6118, opened 2026-07-14
feat(channels): add Zalo Bot channel

### LINKED ISSUE #5776: [Bug]: stale pinned first user message treated as active task in long-lived IM session

lamnguyen3119:
## Description

Adds a Zalo Bot Platform channel as a built-in channel under `src/qwenpaw/app/channels/zalo/`. Uses long-polling (`getUpdates`) so no public webhook URL is required.

**Related Issue:** Closes #5776

**Security Considerations:** Channel requires `bot_token` stored in config/environment. Token is never logged. Channel uses no external HTTP callbacks.

## Type of Change

- [x] New feature

## Component(s) Affected

- [x] Channels (Zalo)

## Checklist

- [x] Ready for review

### For Channel Changes

- [x] I ran `./scripts/check-channels.sh` and it passes
- [x] **Contract test** exists in `tests/contract/channels/test_zalo_contract.py`
- [x] Contract test implements `create_instance()` with proper channel initialization
- [x] All 19 contract verification points pass
- [x] **Optional**: Unit tests in `tests/unit/channels/test_zalo.py` for complex internal logic

## Testing

All 78 tests pass: 59 unit tests + 19 contract tests. Pre-commit (mypy, black, flake8, pylint) and `scripts/check-channels.sh zalo` all pass clean.

Key test coverage:
- Channel init + factory methods (from_config, from_env, health_check)
- Session routing: private (zalo:<user_id>), group shared (zalo:group:<chat_id>), group not-shared (zalo:<user_id>:<chat_id>)
- Smart outbound routing: markdown images, bare image URLs, magic tokens [IMAGE:], [STICKER:], [VOICE:], [FILE:]
- Thinking / null-token stripping: removes Thinking:/Reasoning: labels, null/None/undefined/nil lines
- File dispatch: local file upload with name in message, file-not-found fallback
- build_agent_request_from_native: group prefix injection ([Tên trong nhóm chat_id]:)
- _handle_update: private/group dispatch, empty-text skip, offset tracking, mention stripping
- Send path: text-only, group routing, no-http fallback, empty parts
- Lifecycle: start/stop, double-start guard, stop-idle safety

pre-commit run --all-files: all 12 hooks pass (python ast, trailing commas, mypy, black, flake8, pylint, etc.)

pytest tests/unit/channels/test_zalo.py tests/contract/channels/test_zalo_contract.py -v: 78 passed in 0.83s

scripts/check-channels.sh zalo: 59 unit tests passed, all checks passed

## Additional Notes

Channel architecture:
- **Long-poll** (`getUpdates`) — no webhook server needed, works behind NAT/firewall
- **Session routing** — private chats get individual sessions; group chats share a session with sender-name prefix
- **Smart outbound routing** — LLM replies can embed `[IMAGE:]`, `[STICKER:]`, `[VOICE:]`, `[FILE:]` magic tokens for rich messages
- **Thinking/null-token strip** — removes `Thinking:`/`Reasoning:` labels and `null`/`None`/etc. artifacts from streaming LLM output
- **split_text** utility for long messages exceeding Zalo's 2000-char limit
- **78 unit + 19 contract tests** — extract_actions, session routing, thinking strip, file dispatch, send path, lifecycle

### DIFF
--- a/src/qwenpaw/app/channels/zalo/__init__.py
+++ b/src/qwenpaw/app/channels/zalo/__init__.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+"""Zalo Bot channel package.
+
+Connects to the Zalo Bot Platform (https://bot.zalo.me/) via long-polling
+of the ``getUpdates`` endpoint. No public URL / webhook is required,
+making the channel suitable for personal bots running behind NAT.
+
+Group chats are routed with a per-group session id
+(``zalo:group:<chat_id>``) so the agent keeps one conversation context
+per group, while each private chat retains its own per-sender session
+(``zalo:<user_id>``).
+"""
+
+from .channel import ZaloChannel
+
+__all__ = ["ZaloChannel"]
--- a/src/qwenpaw/app/channels/zalo/channel.py
+++ b/src/qwenpaw/app/channels/zalo/channel.py
@@ -1,0 +1,932 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=too-many-instance-attributes,too-many-public-methods
+"""Zalo Bot channel: long-polling getUpdates + REST send."""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple
+
+import httpx
+
+from qwenpaw.app.channels.base import (
+    AudioContent,
+    BaseChannel,
+    ContentType,
+    FileContent,
+    ImageContent,
+    OnReplySent,
+    ProcessHandler,
+    TextContent,
+    VideoContent,
+)
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Tunables / defaults
+# ---------------------------------------------------------------------------
+
+ZALO_DEFAULT_API_BASE = "https://bot.zalo.me"
+ZALO_DEFAULT_POLL_INTERVAL = 30.0
+ZALO_DEFAULT_POLL_TIMEOUT = 25.0
+ZALO_DEFAULT_MAX_RETRIES = 3
+ZALO_DEFAULT_MAX_MESSAGE_LEN = 2000
+
+# Outbound content markers an LLM may emit inside a text reply.
+_IMAGE_TOKEN_RE = re.compile(
+    r"\[IMAGE\s*:\s*(\S+?)\s*\]",
+    flags=re.IGNORECASE,
+)
+_STICKER_TOKEN_RE = re.compile(
+    r"\[STICKER\s*:\s*(\S+?)\s*\]",
+    flags=re.IGNORECASE,
+)
+_VOICE_TOKEN_RE = re.compile(
+    r"\[VOICE\s*:\s*(\S+?)\s*\]",
+    flags=re.IGNORECASE,
+)
+_FILE_TOKEN_RE = re.compile(
+    r"\[FILE\s*:\s*(/\S+?)\s*\]",
+    flags=re.IGNORECASE,
+)
+_MARKDOWN_IMG_RE = re.compile(
+    r"!\[[^\]]*\]\(\s*(https?://\S+?)\s*\)",
+    flags=re.IGNORECASE,
+)
+_URL_RE = re.compile(
+    r"(https?://\S+?\.(?:png|jpg|jpeg|gif|webp|bmp)(?:\?[^\s)]*)?)",
+    flags=re.IGNORECASE,
+)
+_THINK_BLOCK_RE = re.compile(
+    r"\s*<think(?:ing)?\b[^>]*>.*?</think(?:ing)?\s*>",
+    re.DOTALL | re.IGNORECASE,
+)
+_LEADING_NULL_RE = re.compile(
+    r"^\s*(null|none|undefined|nil|nan)\s*\n\n",
+    flags=re.IGNORECASE,
+)
+_NULL_TOKENS = frozenset(
+    {"null", "none", "undefined", "nil", "nan", "n/a", "()"},
+)
+
+
+# ---------------------------------------------------------------------------
+# Data model
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class OutboundAction:
+    """A non-text action extracted from LLM text reply."""
+
+    kind: str  # photo | sticker | voice | local_file
+    payload: str
+    position: int = 0  # source position in original text
+
+
+# ---------------------------------------------------------------------------
+# Channel class
+# ---------------------------------------------------------------------------
+
+
+class ZaloChannel(BaseChannel):
+    """Zalo Bot channel driven by long-polling ``getUpdates``.
+
+    Long-polling avoids the need for a public webhook URL — useful when
+    QwenPaw runs behind NAT or on a developer laptop.
+    """
+
+    # Channel registry key — read by ``ChannelManager``.
+    channel = "zalo"
+    uses_manager_queue = True
+
+    def __init__(
+        self,
+        process: ProcessHandler,
+        enabled: bool,
+        bot_token: str,
+        api_base_url: str = ZALO_DEFAULT_API_BASE,
+        secret_token: str = "",
+        show_typing: bool = True,
+        poll_interval: float = ZALO_DEFAULT_POLL_INTERVAL,
+        max_retries: int = ZALO_DEFAULT_MAX_RETRIES,
+        max_message_len: int = ZALO_DEFAULT_MAX_MESSAGE_LEN,
+        share_session_in_group: bool = True,
+        on_reply_sent: OnReplySent = None,
+        show_tool_details: bool = True,
+        filter_tool_messages: bool = False,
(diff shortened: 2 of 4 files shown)


hongxicheng:
Thanks for the contribution @lamnguyen3119! I did a thorough review and I don't think this can be merged as a built-in channel yet — as it stands the channel is only the implementation class + tests, and it's missing most of the wiring needed to actually function inside QwenPaw. A few things are also runtime-breaking against the real BaseChannel contract. Details below.

## Blockers: registering Zalo as a built-in channel needs a lot more

The PR only adds `zalo/__init__.py`, `zalo/channel.py` and the two test files. A built-in channel needs to be wired into ~15 registration points; almost all are missing:

- **Not registered anywhere** → the channel can never be loaded/enabled/configured:
  - `app/channels/registry.py` `_BUILTIN_SPECS` (missing `zalo`)
  - `app/channels/schema.py` `BUILTIN_CHANNEL_TYPES` (missing `zalo`)
  - `config/config.py` — no Pydantic Config class mounted on `ChannelConfig`
- **CLI**: no `channels_cmd.py` entry / configurator
- **Frontend**: nothing in `api/types/channel.ts`, `constants/channel.ts`, `components/channelIcons.ts`, `components/constants.ts`, or `components/ChannelDrawer.tsx` (case branch, `CHANNELS_WITH_ACCESS_CONTROL`, `CHANNEL_DOC_*_URLS`, streaming判断)
- **i18n**: none of the 7 locales updated (`en/zh/id/ja/pt-BR/ru/vi`)
- **Docs**: no section/anchor in `website/public/docs/channels.{en,zh}.md`

So even though the unit/contract tests are green, the channel is a no-op in a running app.

## Contract / runtime issues (would crash against the real framework)

1. **`send()` signature mismatch.** `BaseChannel.send` is called by the framework as `send(to_handle, text: str, meta: dict)` (see base.py, and the Telegram channel). This PR defines `send(to_handle, content_parts: List)`, so the framework call passes an extra positional arg → `TypeError`. The tests only pass because they call the invented signature directly.
2. **`_enqueue` misuse.** `EnqueueCallback` is a sync `Callable[[Any], None]` (Telegram uses `self._enqueue(native)`). Here it's `await self._enqueue(native, is_group=is_group)` — awaiting a non-coroutine plus an unexpected kwarg. Masked by `AsyncMock` in tests, but will fail at runtime.
3. **`_on_reply_sent` misuse.** Base type is `Callable[[str, str, str], None]`, invoked by the base as `self._on_reply_sent(self.channel, *args)`. Here it's called with kwargs `_on_reply_sent(chat_id=..., is_group=...)` — signature mismatch.

## Access control / config wiring

4. `from_config` / `from_env` never wire `access_control_dm/group`, `allow_from`, `dm_policy/group_policy`, `require_mention`, `streaming_enabled`, `filter_thinking`. `__init__` accepts them, but user config is silently dropped, so ACL/mention/streaming settings won't take effect.
5. `build_agent_request_from_native` puts only `sender_id` in meta; per our channel guidelines the raw stable id should also be passed as `acl_sender_id` (matters for shared group sessions).

## Minor

6. Group prefix `[{name} trong nhóm {chat_id}]:` is hardcoded — should be neutral/localized.
7. `show_typing` and `secret_token` are accepted but never used (dead config).
8. `_dispatch_local_file` doesn't actually upload — it only sends a text description, whereas the PR description says "local file upload".

## Zalo only supports private chats — has this been verified against the real API?

We looked into the Zalo Bot Platform docs. The long-poll `getUpdates` approach and the 2000-char split look correct and consistent with the (Telegram-style) Zalo Bot API. **However, the docs indicate Zalo Bot API currently supports 1:1 direct messages only — groups are "coming soon" and not yet available.**

This PR implements a lot of group logic (`zalo:group:<chat_id>` routing, `share_session_in_group`, the group name-prefix injection, `supergroup`/`channel` chat types, in-group mention stripping). Against the real API today, that path is effectively dead/untestable. Combined with the unused `secret_token` (webhook secret) in a polling-only implementation, it looks like this may have been adapted from the Telegram channel and carried over capabilities Zalo doesn't have yet.

**Could you confirm whether this was actually tested end-to-end against a live Zalo bot** (inbound getUpdates + outbound sendMessage), or whether it's currently only covered by mocked unit tests? If groups aren't supported by Zalo yet, I'd suggest either removing the group code or clearly gating it as "enable once Zalo groups ship".

## Alternative: ship it as a plugin instead of a built-in channel

If you'd prefer a faster path, you can also publish Zalo as a **channel plugin** to the plugin marketplace (https://platform.agentscope.io/plugins) rather than as a built-in channel. That route has a much lighter, more self-contained spec — you don't need to touch registry/schema/config, the CLI, the frontend, the 7 i18n locales, or the core docs — and it lets you version and iterate the channel independently on your own release cadence, without waiting on a core PR review cycle. Since the Zalo Bot API is still actively evolving (e.g. groups not shipped yet), the plugin path is likely the better fit.

Happy to help once the registration wiring and the `send`/`_enqueue` contract issues are addressed. Thanks again!

lamnguyen3119:
﻿@hongxicheng Thanks for the detailed review — really appreciate you taking the time to go through it thoroughly.

You make a solid point that a built-in channel needs full registration wiring (registry, schema, config, frontend, i18n, docs) before it's functional. That's a significant amount of scaffolding, and I agree it doesn't make sense to land it half-wired.

Since the Zalo Bot API is still evolving (e.g. group support isn't live yet), I'll take your suggested route and ship this as a channel plugin instead. The plugin architecture is a better fit — lighter spec, independent release cadence, no coupling to the core review cycle. I've been iterating on it separately and it's been running in production with webhook + polling, typing indicator, content filtering, etc.

I'll close this PR for now. The code here (magic token extraction, thinking strip, text splitting) has pieces I'll port over to the plugin. The send/enqueue contract lessons are noted for any future built-in attempt.

Thanks again for the honest review — it saved me from merging something that wouldn't actually work at runtime. 🙏

lamnguyen3119:
Closing this in favor of a Zalo channel plugin on the plugin marketplace.

PR #5849, opened 2026-07-08
feat(channel): add Azure Bot channel as a plugin, with i18n/custom-icon/doc_url support for plugin channels

hongxicheng:
## Summary

Adds **Azure Bot Service (Bot Framework)** as a first-class **plugin channel**,
and, along the way, extends the plugin-channel framework with two generic,
optional capabilities — a **custom card icon** and a **documentation link** —
so any plugin channel (not just Azure) can supply them.

Azure Bot ships as a plugin under `plugins/channel/azure_bot/` rather than a
built-in channel: it exposes a public webhook endpoint, so keeping it opt-in
(installed from the Plugin Marketplace) avoids forcing that surface on every
deployment.

### New plugin: `azure_bot`
- Full Bot Framework webhook integration (Teams / Web Chat / DirectLine, etc.).
- **Inbound security** built on Microsoft JWT verification: JWKS fetched from
  Microsoft OpenID metadata (cached 1h, fail-closed), `jwt.decode` with
  `RS256` + `audience=app_id` + `verify_exp`, plus an issuer allowlist.
  Requests are **verified before the body is read** to avoid buffering
  unauthenticated payloads.
- **Async, non-blocking I/O**: disk reads/writes and JSON persistence run via
  `asyncio.to_thread`; concurrent saves are serialized with an `asyncio.Lock`;
  pending saves are drained on shutdown with a 5s timeout guard.
- **Log redaction**: MSAL token errors log only the error code at WARNING,
  descriptions at DEBUG.
- **Access control**: `access_control_dm` / `access_control_group` /
  `require_mention` declared in `config_fields`.
- Config fields fully localized across all 7 project locales
  (zh-CN / en-US / ja-JP / ru-RU / pt-BR / vi-VN / id-ID); technical fields
  (app_id / app_password / tenant_id / host / port) kept as plain strings.
- `requirements.txt` (`aiohttp`, `PyJWT`, `msal`) so the framework auto-installs
  dependencies on plugin install.

### Framework: generic plugin-channel capabilities
- `register_channel` gains optional `icon` and `doc_url` parameters
  (`registry.py`, `api.py`), surfaced through `/config/channels/schemas`.
- `doc_url` accepts a plain http(s) string **or** a localized dict
  (`{"zh": "...", "en": "..."}`), resolved by the Console per UI language.
- Both are **optional and backward-compatible** — channels that don't set them
  behave exactly as before.

### Console (frontend)
[...]

PR #6126, opened 2026-07-15
feat(channels): generic webhook channel as opt-in plugin

AnandSundar:
## Summary

Addresses the design-concern review from **@hongxicheng** on #5716 by shipping the generic webhook channel as an **opt-in plugin** rather than a built-in, matching the Azure Bot (#5849) shape. The inbound-listener attack surface is now opt-in: deployments that don't need webhooks don't carry the listening port.

Replaces #5716 (kept open during the rework; this new PR is the one to review).

cc @hongxicheng

### Plugin migration (matching #5849 shape)

- New plugin under `plugins/channel/webhook/`:
  - `plugin.json` — id / version / qwenpaw version range / dependencies
  - `plugin.py` — `WebhookChannelPlugin.register` calling `api.register_channel(channel_class=WebhookChannel, label, description, icon, doc_url={zh, en}, config_fields=[channel_id, bind_address, port, outbound_url, secret, rate_limit_rps, rate_limit_burst])` — uses the new `register_channel(icon=..., doc_url=...)` signature introduced in #5849
  - `__init__.py` — re-exports `WebhookChannel` and friends
  - `requirements.txt` — `fastapi`, `uvicorn`, `httpx`
- Removed from the built-in path:
  - `"webhook"` entry from `_BUILTIN_SPECS` in `src/qwenpaw/app/channels/registry.py`
  - `WebhookConfig` Pydantic model from `src/qwenpaw/config/config.py`
  - `configure_webhook` + `"webhook"` slot in `_ALL_CHANNEL_NAMES` / `_ALL_CHANNEL_CONFIGURATORS` in `src/qwenpaw/cli/channels_cmd.py`
- Existing `channels.webhook` config blocks are preserved through `extra="allow"` on `AgentConfig` and rendered via the plugin's `config_fields`.

### Auth tightening

- `signature.py:verify_signature()` now returns **`False`** when a `secret` is configured but the `X-QwenPaw-Signature` header is missing or empty — previously it warned and accepted. The server already maps that to **HTTP 401** with `{"error": "signature_mismatch"}`.
- The permissive unsigned-but-no-secret path is preserved (deployments that rely on network isolation alone continue to work).
- New tests: `TestVerifySignature::test_returns_false_when_signature_missing_with_secret`, `TestWebhookServer::test_post_returns_401_when_unsigned_with_secret`, plus the negative `test_post_returns_200_when_unsigned_and_no_secret`.

### Per-client-IP rate limiting

- `server.py` builds an in-memory token-bucket (`_RateLimiter`, one bucket per `request.client.host`) when the app is constructed. Empty bucket returns **HTTP 429** with `Retry-After: 1`.
- Defaults: **5 RPS sustained / 10-request burst**, both exposed via `config_fields` (`rate_limit_rps`, `rate_limit_burst`) and read in `from_env()` as `WEBHOOK_RATE_LIMIT_RPS` / `WEBHOOK_RATE_LIMIT_BURST`.
- New tests: `TestWebhookServer::test_post_returns_429_when_rate_limited`, `test_post_returns_429_includes_retry_after_header`.

### Tests

- Relocated `tests/unit/channels/test_webhook.py` and `tests/contract/channels/test_webhook_contract.py` to `plugins/channel/webhook/tests/{unit,contract}/`. The contract test is self-contained (mirrors the relevant subset of `tests.contract.channels.ChannelContractTest`) so the plugin test tree has no hard dependency on the main `tests/` tree and can be exercised against the Plugin Marketplace bundle directly.
- New coverage for the 401-when-unsigned-with-secret and 429 paths.
- All 62 plugin tests (49 unit + 13 contract) pass; surrounding 305-test sweep green; pre-commit clean on the full changed file set.

### Docs

- Both `website/public/docs/channels.en.md` and `.../zh.md` now lead the Webhook section with a "plugin channel, not built-in" callout and direct users to install via the Plugin Marketplace.
- Field tables, status codes, and the limits/behaviour section cover the new `rate_limit_rps` / `rate_limit_burst` fields, the new `429` status, and the tightened "secret ⇒ mandatory signature" rule.

### Tooling fix (companion commit)

- `scripts/check-channels.sh` previously hardcoded `tests/contract/channels/<name>.py` paths and exited 1 when a plugin channel's tests lived under `plugins/channel/<name>/tests/`. Now it falls back to the plugin path and uses `python -m pytest` (rather than bare `pytest`) so the project root is on `sys.path` and the `plugins.channel.*` namespace-package imports work. Verified with `bash scripts/check-channels.sh webhook` ⇒ "All checks passed!".

## Verification

- `pre-commit run --files <changed files>` — all hooks pass (mypy, black, flake8, pylint, trailing-comma)
- `python -m pytest plugins/channel/webhook/tests/ tests/contract/channels/ -q` — **305 passed, 2 skipped**
- `bash scripts/check-channels.sh webhook` — contract + unit tests green
- Plugin test tree (49 unit + 13 contract) green in isolation

## Deferred (carried over from #5716)

- Streaming responses back to the inbound caller — current behavior is fire-and-forget; the inbound HTTP request returns 200 immediately and replies go to `outbound_url`. A follow-up could add SSE or "wait-for-reply" semantics if requested.
- Multi-channel multiplexing on a single port — each `WebhookChannel` instance binds its own port. A reverse proxy (nginx / caddy) can fan in if needed.
- Inbound body schema validation — the channel accepts arbitrary JSON; stricter contracts belong in a downstream plugin.

Refs #5716


hongxicheng:
Thanks for the rework, @AnandSundar — really appreciate you turning this around into an opt-in plugin and tightening the signature verification. 🙏

One thing though: **publishing a plugin doesn't require a PR to this repo.** Plugins are distributed through the Plugin Marketplace rather than merged into the main codebase, so you can upload your webhook plugin directly at [Plugin Marketplace](https://platform.agentscope.io/plugins) and users will be able to install it on demand.

For that reason I'll go ahead and close this PR. If you have any questions about the marketplace publishing flow, feel free to leave a comment — happy to help. Thanks again for the contribution!

PR #6133, opened 2026-07-15
fix(mcp): make legacy MCP migration one-shot via schema watermark (#6130)

### LINKED ISSUE #6130: [Bug]: Deleted MCP clients reappear after QwenPaw restarts
## QwenPaw Version

2.0.0.post2

## Description

After an MCP client is deleted from the Console, it disappears immediately but reappears after QwenPaw restarts.

This occurs when the client exists in both the legacy `agent.json -> mcp.clients` configuration and the migrated `drivers/mcp/<client>.yaml` storage. Deletion removes the DriverCard YAML, but startup migration recreates it from the legacy entry.

**Related PR(s):** None

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.0.0.post2
- **OS:** Windows 11
- **Install method:** pip
- **Python version (if applicable):** 3.12.10

## Steps to Reproduce

1. Have an MCP client migrated from `agent.json -> mcp.clients` to `drivers/mcp/*.yaml`.
2. Delete the client from **Workspace -> MCP** in the Console.
3. Restart QwenPaw.
4. Open **Workspace -> MCP** again.

## Actual vs Expected

- **Actual:** The deleted MCP client reappears after restart.
- **Expected:** The MCP client remains deleted after restart.

## Logs / Screenshots

Not required.

## Additional Notes

The startup call to `migrate_legacy_mcp_if_needed()` recreates the missing DriverCard from the unchanged legacy `agent.json` entry.

xiaoming-qxm:
## Summary

Deleting an MCP client from the Console removes its DriverCard and credentials but leaves the legacy `agent.json` `mcp.clients` entry untouched. On the next start, `migrate_legacy_mcp_if_needed` used **"does the DriverCard file exist?"** as its *already-migrated* test — so it re-derived a card from the still-present legacy entry and the **deleted client reappeared** (#6130).

That judgement conflated two different states — *never migrated* vs *migrated then deleted* — because DriverCard existence cannot distinguish them.

## Fix

Move the *already-migrated* judgement onto an explicit, persisted schema watermark, decoupled from card existence:

- **`MCPConfig.migration_version: int = 0`** — persisted in `agent.json`.
- **`migrate_legacy_mcp_if_needed` is now version-gated**: once a workspace reaches `CURRENT_MCP_MIGRATION_VERSION` (=2) the migration is **skipped entirely**, so a deleted card is never resurrected.
  - `v0 → v1`: migrate legacy `mcp.clients` into DriverCard storage. The per-client "card exists" check is now only an in-run idempotency / crash-rerun guard, **not** the migration judgement.
  - `v1 → v2`: fold the #6029 `env:` credential-ref self-heal in as a **one-shot** step, instead of re-scanning every card on every start. This is safe because bad `${VAR}` literals are historical — post-fix migration never produces new ones.
- The watermark is persisted (`save_agent_config`) only **after all steps succeed**; on partial failure the version stays put and the next start re-runs (idempotent).

`legacy mcp.clients` is intentionally **preserved** (rollback-safe; consistent with the explicit *"DO NOT clear … mcp"* contract in `app/migration.py`).

### Why a version field (and not a marker file, or clearing `mcp.clients`)

The migration-completed state cannot be derived from any surviving runtime artifact — deleting a card also deletes its credentials — so it must be persisted explicitly. That is exactly 1 bit of information, best modeled as a schema version on the config that already owns the data, requiring no new state file. It also lets the #6029 upgrade become a proper one-shot step instead of an every-boot scan.

## Changes

- `config/config.py`: add `MCPConfig.migration_version`.
- `drivers/adapters/mcp_legacy_config.py`: add `CURRENT_MCP_MIGRATION_VERSION`, version-gate `migrate_legacy_mcp_if_needed`, add `_persist_mcp_migration_version`.
- `app/workspace/service_factories.py`: drop the standalone `upgrade_legacy_mcp_credentials` call (now folded into the v1→v2 step).
- `tests/unit/drivers/adapters/test_mcp_migration_version.py`: new tests.

## Verification

**New unit tests (5), TDD** (watched them fail on the missing symbol first):
- skip entirely when already at current version;
- migrate + persist watermark when unmigrated;
- **deleted card is not resurrected after the watermark**;
- env-ref upgrade folded in at v1→v2;
- config default `migration_version == 0`.

**Hermetic cold-boot proof** (isolated `QWENPAW_WORKING_DIR`, a *separate process* per boot):
```
[SEED]   {"clients": ["tavily_search"], "migration_version": 0}
[BOOT#1] {"card_exists": true,  "migration_version": 2, "clients": ["tavily_search"]}
[DELETE] {"deleted": true, "card_exists_after": false}
[BOOT#2] {"card_exists": false, "migration_version": 2, "clients": ["tavily_search"]}
[GREEN] deleted MCP client stays deleted across restart
```

- No regressions: `tests/unit/drivers/ tests/unit/config/` + env-ref integration → **30 passed**.
- `pre-commit` (mypy / black / flake8 / pylint) clean.

## Steady-state note

The first start after this change performs one final reconciliation for any client deleted *before* the watermark existed (such a client may reappear once), then settles at `version=2` and never re-migrates. Once the watermark is established, steady-state boots perform no migration writes.

Closes #6130

### DIFF
--- a/src/qwenpaw/app/workspace/service_factories.py
+++ b/src/qwenpaw/app/workspace/service_factories.py
@@ -25,7 +25,6 @@ async def create_driver_service(ws: "Workspace", _service):
     # pylint: disable=protected-access
     from ...drivers.adapters.mcp_legacy_config import (
         migrate_legacy_mcp_if_needed,
-        upgrade_legacy_mcp_credentials,
     )
     from ...drivers.credentials.store import AsyncCredentialStore
     from ...drivers.handlers import MCPDriverHandler
@@ -50,7 +49,6 @@ async def create_driver_service(ws: "Workspace", _service):
     # endpoint validator and tests.  This PR intentionally keeps the concrete
     # runtime surface to MCP while leaving DriverManager protocol-neutral.
     await migrate_legacy_mcp_if_needed(ws, driver_manager)
-    await upgrade_legacy_mcp_credentials(driver_manager)
     await driver_manager.start()
     ws._service_manager.services["driver_manager"] = driver_manager
     logger.debug(
--- a/src/qwenpaw/config/config.py
+++ b/src/qwenpaw/config/config.py
@@ -1696,6 +1696,12 @@ class MCPConfig(BaseModel):
             ),
         },
     )
+    # One-shot migration watermark, persisted in agent.json.  Decoupled from
+    # DriverCard existence so that deleting a migrated client no longer lets
+    # startup migration resurrect it (#6130).  0 = not migrated; steps are
+    # defined by CURRENT_MCP_MIGRATION_VERSION in
+    # drivers.adapters.mcp_legacy_config.
+    migration_version: int = 0
 
 
 class BuiltinToolConfig(BaseModel):
--- a/src/qwenpaw/drivers/adapters/mcp_legacy_config.py
+++ b/src/qwenpaw/drivers/adapters/mcp_legacy_config.py
@@ -41,6 +41,13 @@ from ..credentials.types import CredentialRecord
 from ..manager import DriverManager
 from ..storage import load_card
 
+# Schema version of the legacy-MCP -> DriverCard migration, persisted on
+# ``MCPConfig.migration_version`` (agent.json) as a one-shot watermark that
+# is independent of whether a DriverCard file happens to exist:
+#   v1: legacy agent.json ``mcp.clients`` -> DriverCard storage
+#   v2: single ``${VAR}`` literal -> ``env:`` credential ref (#6029 heal)
+CURRENT_MCP_MIGRATION_VERSION = 2
+
 
 @dataclass
 class LegacyMCPMigratedClient:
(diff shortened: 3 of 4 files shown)


xieyxclack (approved):
LGTM

xiaoming-qxm:
Thanks for the review. Both Low items addressed in 66023be5:

**1. `save_agent_config` failure path — enforced and tested.**
While wiring the suggested test I found its assertion (`migration_version == 0` after a failed save) wouldn't hold against the original code: the in-memory bump happened *before* persistence, so a failed `save_agent_config` left the in-memory watermark at `2` while disk stayed at `0`. Rather than only checking the disk side, I made the invariant real — the in-memory version is now rolled back when persistence raises, so it always matches on-disk state and the next start re-runs. Covered by `test_migration_rolls_back_watermark_on_persist_failure`.

**2. Env-ref upgrade end-state — added, with a correction.**
`store.get("env_api_key")` doesn't apply here: `env_api_key` is an in-card `CredentialRef("static", "env:API_KEY")` resolved from the environment at runtime — `env:` refs are never written to the credential store (that's the point of #6029). The meaningful end-state is that the *plaintext* record is gone, so the test now asserts `"mcp/wind" not in await store.list_refs()`, consistent with the existing upgrade tests.

Verification: `tests/unit/drivers/ tests/unit/config/` → 30 passed; pre-commit (mypy / black / flake8 / pylint) clean.

PR #6138, opened 2026-07-15
fix(loop): fine-tune doom loop thresholds (warning 3, stop 4) and align frontend

### LINKED ISSUE #6116: [Bug] Doom loop: agent repeatedly triggers same tool call in single turn
## Description

The agent falls into a doom loop, repeatedly calling the same tool with the same parameters multiple times within a single user turn. The system eventually detects this after ~6 consecutive repetitions and shows a warning, but by then many API calls and tokens have been wasted.

**This did NOT occur in previous versions with the same model.** The regression suggests a framework-level change is involved.

## Reproduction

- **Session**: 5ff775cf
- **Channel**: Feishu
- **Model**: Qwen (same model that previously worked without this issue)
- **Date**: 2026-07-14

### Scenario

User asked a simple question. The agent called web_search, got results, then called web_search again with identical query 2-3 times. Then when analyzing the issue, called sequential-thinking once (got thoughtNumber=1, totalThoughts=1, nextThoughtNeeded=false), then called it again with nearly identical thought 3-5+ times before being stopped by doom loop detection.

## Evidence

The doom loop warning appeared 6+ times across the conversation:

[WARNING] Repetitive pattern detected. You are repeating similar actions without progress. Try a completely different approach

## Root Cause Analysis

1. **LLM layer**: Model generates tool calls, receives results, but re-generates the same tool call in the next cycle. Known weakness with long/compressed context.

2. **Framework layer - no tool call deduplication**: QwenPaw does not detect or prevent duplicate tool calls within the same turn. If the agent calls web_search and gets results, the framework could block a second identical call.

3. **Doom loop detection is reactive not preventive**: Warning only fires after ~6 repetitions. Earlier intervention (after 2-3) would save tokens.

### Why this is a regression

Same model, same agent config, same channel. Previous versions did not exhibit this at this frequency. Behavioral rules in AGENTS.md cannot fix a framework-level issue.

## Suggested Improvements

1. **Tool call deduplication**: Track tool calls per turn. If same tool + same params called again, inject a blocking system message.
2. **Earlier doom loop detection**: Trigger after 2-3 repetitions instead of 6.
3. **Tool result injection**: Ensure results are prominently visible in context for next generation cycle.

## Environment

- QwenPaw version: latest 2026-07-14
- OS: Linux 5.19.17-z4pro-generic x86_64
- Channel: Feishu
- Session ID: 5ff775cf

nguyenthanhthe:
## What Problem This Solves

This PR addresses the model doom loop repetition issue (Issue #6116). In alignment with maintainer guidance to avoid hard-coded execution intercepts inside `ToolCoordinatorMiddleware`, we address repetition loops by tuning the default configuration thresholds of the framework's native `DoomLoopGate` and aligning the frontend config view.

Key changes:
1. **Doom Loop Gate Config Tuning (Backend):** The first warning is kept at `after=3` (warning the agent via `"modify_prompt"`), while the second stage (the termination `"stop"` action) is tightened from `6` to `4` repetitions. `window_size` remains `3` (default), ensuring loop detection warning occurs on the first matches and terminates soon after on continuation, saving tokens and preventing runway loops earlier.
2. **Form Rule Initialization (Frontend):** Updated the default rule threshold generation logic in `AgentLoopCard.tsx` when clicking "Add Rule" on the UI, returning `3` for the first rule (index 0) and `4` for the second rule (index 1), keeping the user interface completely aligned with backend defaults.

## Evidence

I verified the changes by running all relevant backend loop gate tests and typescript compilation checks. All checks and assertions pass successfully:

1. **Backend Tests:**
   ```bash
   pytest tests/unit/loop/
   ```
   Output: `22 passed in 0.37s`

2. **Frontend Build:**
   ```bash
   npx tsc -b
   ```
   Output: Compile passed successfully.

All files pass pre-commit checks.

### DIFF
--- a/console/src/pages/Agent/Config/components/AgentLoopCard.tsx
+++ b/console/src/pages/Agent/Config/components/AgentLoopCard.tsx
@@ -413,8 +413,8 @@ function DoomLoopSection() {
                             stages.length === 0
                               ? 3
                               : stages.length === 1
-                                ? 4
-                                : stages.length + 3,
+                              ? 4
+                              : stages.length + 3,
                           action: "modify_prompt",
                           prompt: "",
                         })
--- a/console/src/pages/Agent/Config/components/AgentLoopCard.tsx
+++ b/console/src/pages/Agent/Config/components/AgentLoopCard.tsx
@@ -409,7 +409,12 @@ function DoomLoopSection() {
                       type="dashed"
                       onClick={() =>
                         add({
-                          after: (stages.length + 1) * 3,
+                          after:
+                            stages.length === 0
+                              ? 3
+                              : stages.length === 1
+                                ? 4
+                                : stages.length + 3,
                           action: "modify_prompt",
                           prompt: "",
                         })
--- a/src/qwenpaw/config/config.py
+++ b/src/qwenpaw/config/config.py
@@ -1045,7 +1045,7 @@ class DoomLoopConfig(BaseModel):
         description="Enable doom loop detection",
     )
     window_size: int = Field(
-        default=2,
+        default=3,
         ge=2,
         description=("Sliding window size for " "repetition detection"),
     )
@@ -1060,7 +1060,7 @@ class DoomLoopConfig(BaseModel):
     stages: List[DoomLoopStageConfig] = Field(
         default_factory=lambda: [
             DoomLoopStageConfig(
-                after=2,
+                after=3,
                 action="modify_prompt",
                 prompt=(
                     "[WARNING] Repetitive pattern "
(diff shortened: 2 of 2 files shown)


rayrayraykk (changes requested):
The default value on frontend should be changed accordingly. And I suggest to change to remain the first warning to 3, and change the second stage bar to 4.

xieyxclack (approved):

rayrayraykk:
Your force-push overwrote my follow-up commits on this branch, so I'll close this PR and cherry-pick my changes into a new one. Please avoid force-pushing next time 🙂

PR #6171, opened 2026-07-16
fix(memory): add dream schedule toggle

ployts:
## Description

Fix the issue where clearing `dream_cron` could still cause the backend default cron expression (`0 23 * * *`) to be used.

This change adds an explicit `dream_cron_enabled` switch:

- Keep `dream_cron` as a string with the default value `0 23 * * *`.
- Disable the scheduled dream task through the new backend configuration switch.
- Add the switch to the Console and disable the cron input when turned off.
- Preserve compatibility with existing configurations that do not contain the new field.
- Rename the Chinese Console label from “梦境” to “梦境定时”.
- Update localized strings and documentation.

**Related Issue:** N/A

**Security Considerations:** None. This only changes local agent configuration and scheduled task behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend
- [x] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [x] Documentation
- [ ] CI/CD
- [ ] Scripts / Deploy
- [x] Tests

## Checklist

- [x] Pre-commit checks passed for all executed hooks
- [x] Formatter changes were committed
- [ ] Full pytest suite run
- [x] Documentation updated
- [x] Ready for review

## Testing

- `PRE_COMMIT_HOME=/tmp/copaw-pre-commit SKIP=actionlint pre-commit run --all-files` — passed
- `cd console && npm run format` — passed
- Console TypeScript check — passed
- Console build — passed
- Website Prettier formatter — passed
- `git diff --check` — passed
- Related Python tests previously passed: `69 passed`
- Full pytest was skipped for this verification as requested
- `actionlint` was skipped because dependency installation was intentionally omitted

## Evidence

```text
pre-commit: all executed hooks passed
console formatter: passed
console TypeScript check: passed
console build: passed
website Prettier formatter: passed
git diff --check: passed

## Additional Notes

The scheduled dream task is now controlled explicitly by dream_cron_enabled. Clearing the cron expression is no longer treated as the mechanism for disabling the task.

### CODE REVIEW

**Comment by jinliyl** [comment]
Thanks for working on this and for introducing an explicit `dream_cron_enabled` switch. Separating the enabled state from the cron expression is a good UX and configuration improvement, and the backward-compatible default behavior looks reasonable.

I found one behavior that I think should be addressed before merging:

### Enabled state can still result in no scheduled dream job

The Console currently allows this sequence:

1. Keep `dream_cron_enabled` turned on.
2. Clear the `dream_cron` input.
3. Save the configuration.

The saved state then has approximately:

```json
{
  "dream_cron_enabled": true,
  "dream_cron": ""
}
```

On the backend, `get_dream_cron()` returns the empty string unchanged when the switch is enabled. The cron manager checks the returned value with `if dream_cron:`, so it does not create the scheduled job.

This leaves the Console showing the dream schedule as enabled while no dream job is actually scheduled. It also means that, despite the new switch, an empty cron expression still acts as a second way to disable scheduling. That appears inconsistent with the purpose of this PR and with the field description saying that an empty string is ignored.

Could we make the enabled state unambiguous? Either of these approaches would work:

- When `dream_cron_enabled` is true, require a non-empty, valid cron expression in the Console and preferably validate it on the backend as well.
- When the switch is enabled and the expression is empty, normalize it to the default value (`0 23 * * *`) before saving or scheduling.

A regression test for `dream_cron_enabled=True` together with an empty `dream_cron` would also help document the intended behavior.

I checked the related paths locally. The new targeted backend tests pass, old configurations without `dream_cron_enabled` default to enabled as intended, and saving the running configuration reloads the agent so that the cron manager picks up switch changes. The issue above is the only PR-specific functional concern I found.

For completeness, the full Unit Tests jobs are currently failing in CI, but the reproduced failures are in unrelated existing test areas and none of those failing files are changed by this PR.

jinliyl (approved):
LGTM

cuiyuebing (approved):
LGTM

Issue #6100, opened 2026-07-14
[Bug]: lost workspace


hellofreud:
## QwenPaw Version

2.0.0.post1

## Description

**Upgrading from 1.1.9 → 2.0.0.post1 via `pip install --upgrade qwenpaw` causes the builtin (default) agent's `agent.json` to be overwritten with an empty/downstream configuration**, dropping several critical fields:

- `active_model` → reset to `{}` (was `{"provider_id": "newapi", "model": "deepseek-v4-flash"}`)
- Channel credentials in `feishu`, `dingtalk`, `wechat` → stripped (e.g. `app_id` / `app_secret` / `bot_token` gone)
- `mcp` service definitions (e.g. tavily_search with `TAVILY_API_KEY`) → removed

The file is written by QwenPaw 2.0.0 on first startup after the upgrade, replacing the pre-existing `agent.json`. The new format uses `llm_routing` (under agent config) instead of the top-level `active_model`, but the migration does **not** carry over the old values — the file is effectively replaced with defaults.

**This also triggers a secondary surprise**: non-git-tracked files in the workspace (temporary scripts, download artifacts, one-off data files) disappear from the working tree after a `git checkout -- .` recovery, since they were never committed.

**Related PR(s):** n/a

**Security considerations:** **High**. Channel credentials (Feishu app_secret, DingTalk client_secret, WeChat bot_token) and API keys are silently dropped on upgrade. An unattended upgrade could cause a service outage where the agent appears running but cannot authenticate to any messaging channel. User would need to re-enter these secrets manually.

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [x] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.0.0.post1 (upgraded from 1.1.9)
- **OS:** Ubuntu 24.04.4 LTS
- **Install method:** pip (venv at `/root/.qwenpaw/venv/`)
- **Python version:** 3.12.3

## Steps to Reproduce

1. Start with QwenPaw 1.1.9, configured with a working `default` agent that has `active_model`, `channels.feishu`, `channels.dingtalk`, `channels.wechat`, and `mcp` blocks in its `agent.json`.
[...]

PR #6174, opened 2026-07-16
fix(mcp): unblock workspace startup after connection timeout

jinglinpeng:
## Description

Fixes workspace-startup hangs when an MCP client never becomes ready, including the ready/timeout race in the cleanup path.

On a connection timeout, `connect()` stops and cancels the lifecycle task. Cancellation can bypass the lifecycle's ordinary stop cleanup; if readiness wins the race just after the timeout, stale connection state and a cancelled task remain. The timeout path now resets that state after cleanup, so the handler's follow-up `close()` does not propagate `CancelledError` and DriverManager can skip the failed driver.

**Related Issue:** N/A — based on a desktop v2 startup regression report. This does not address the separate agent.json migration regression in #6100.

**Security Considerations:** None. This changes only lifecycle cleanup after an MCP connection timeout; credential and transport configuration are unchanged.

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test improvement
- [ ] Build or CI configuration change

## Components
- [x] Core
- [ ] CLI
- [ ] Web UI
- [ ] Desktop
- [ ] Tests
- [ ] Documentation
- [ ] CI / Build

## Checklist
- [ ] I have run pre-commit checks locally.
- [x] I have tested my changes locally.
- [x] I have updated documentation if needed.
- [ ] I am ready for this PR to be reviewed.

## Testing

- Inline lifecycle reproduction: the ready/timeout race leaves no session, connected flag, or lifecycle-task reference; `close()` returns normally.
- Inline cancellation check: external cancellation while timeout cleanup is in progress is preserved.
- `pytest -q tests/integration/test_driver_mcp_http_flow.py tests/integration/test_driver_mcp_stdio_flow.py --basetemp <workspace temp>`: 2 passed.

No test files are included in this PR by request. File-scoped pre-commit was not run because `pre_commit` is not installed in the workspace `.venv`.

## Evidence

`git diff --check` and `python -m compileall -q src/qwenpaw/drivers/handlers/mcp_stateful_client.py` passed.

### DIFF
--- a/src/qwenpaw/drivers/handlers/mcp_stateful_client.py
+++ b/src/qwenpaw/drivers/handlers/mcp_stateful_client.py
@@ -62,6 +62,8 @@ _TRANSPORT_ERRORS: tuple[type[BaseException], ...] = (
 # in practice) with headroom, while still failing fast enough that a
 # permanently-broken client doesn't stall every turn for long.
 _LIST_TOOLS_RECONNECT_WAIT: float = 3.0
+_LIFECYCLE_CLEANUP_TIMEOUT: float = 5.0
+_LIFECYCLE_REAPERS: dict[asyncio.Task, asyncio.Task] = {}
 
 
 def _is_transport_error(exc: BaseException) -> bool:
@@ -252,8 +254,11 @@ class _MCPClientMixin:
                 f"Timeout waiting for MCP client '{self.name}' to connect",
             )
             self._stop_event.set()
-            if self._lifecycle_task:
-                await self._lifecycle_task
+            lifecycle_task = self._lifecycle_task
+            if lifecycle_task:
+                if not lifecycle_task.done():
+                    lifecycle_task.cancel()
+                await self._wait_for_lifecycle_exit(lifecycle_task)
             raise
 
         if self._oauth_required:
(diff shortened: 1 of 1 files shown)


jinliyl (dismissed):
LGTM

src/qwenpaw/drivers/handlers/mcp_stateful_client.py
```diff
@@ -252,8 +252,24 @@ async def connect(self, timeout: float = 30.0) -> None:
                 f"Timeout waiting for MCP client '{self.name}' to connect",
             )
             self._stop_event.set()
-            if self._lifecycle_task:
-                await self._lifecycle_task
+            lifecycle_task = self._lifecycle_task
+            try:
+                if lifecycle_task and not lifecycle_task.done():
+                    lifecycle_task.cancel()
+                    await asyncio.gather(
```
rayrayraykk: [H1 - High] Unbounded wait in asyncio.gather() after cancel() on the timeout branch

In the timeout branch, after cancel() is called, asyncio.gather() has no deadline. If the transport's __aexit__, subprocess teardown, or any other cleanup logic delays or suppresses cancellation, connect() will never return — leaving the entire workspace startup call chain hung indefinitely.

Suggestion:

Apply an explicit, separate timeout specifically for the cancellation/cleanup phase
Log the state of each task when the cleanup timeout is exceeded
Define clear ownership and a follow-up recycling strategy for tasks that fail to exit — running tasks must not have their state unconditionally wiped while they are still alive

cuiyuebing (approved):
LGTM

PR #6181, opened 2026-07-16
fix(crons): preserve advanced fields in CLI updates

### LINKED ISSUE #6176: [Bug]: cron CLI update resets untouched runtime and metadata fields
## Problem

`qwenpaw cron update` says that only explicitly provided options are changed, but the command rebuilds the job spec from a fixed set of CLI fields. Updating an unrelated field such as the job name silently resets or drops fields that are only available in the full JSON spec or Console.

For example, a job created with:

```json
{
  "runtime": {
    "max_concurrency": 4,
    "misfire_grace_seconds": 1800
  },
  "request": {
    "input": ["..."],
    "request_context": {"source_tag": "ops"},
    "model": "custom-model"
  }
}
```

loses those values after:

```bash
qwenpaw cron update <job-id> --name renamed
```

`max_concurrency` becomes `1`, `misfire_grace_seconds` becomes `600`, and request extensions such as `request_context` or `model` are dropped. Job and dispatch metadata already have explicit preservation logic; the remaining sections do not.

## Expected behavior

An update should preserve every field that was not explicitly changed. This should include advanced runtime settings and request extensions.

## Suggested fix

Merge CLI overrides into the existing spec instead of reconstructing the whole document from defaults. Add a regression test with non-default runtime values and request extensions, then update only the name and verify the rest of the spec is unchanged.

AaronZ345:
## Description

`cron update` rebuilt runtime options and agent requests from CLI defaults, so changing a name or schedule could reset fields that the CLI does not expose. This carries forward `max_concurrency`, `misfire_grace_seconds`, and agent request extensions while applying only the requested CLI changes. An explicit prompt update replaces `request.input` and leaves fields such as `model` and `request_context` intact.

Fixes #6176.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] CLI
- [x] Tests

## Testing

- `uv run pytest tests/unit/cli/test_cron_cmd.py -q`
- pre-commit hooks on the changed files
- Full unit suite and an upstream baseline run

## Evidence

```text
tests/unit/cli/test_cron_cmd.py: 5 passed
changed-file pre-commit hooks: all passed
full unit suite: 2 failed, 4802 passed, 6 skipped
upstream/main malformed-tool-call baseline: 2 failed, 7 passed
```

The two full-suite failures are the same `test_runtime_malformed_tool_call.py` assertions reproduced on unmodified `upstream/main`.

### DIFF
--- a/src/qwenpaw/cli/cron_cmd.py
+++ b/src/qwenpaw/cli/cron_cmd.py
@@ -234,7 +234,9 @@ def _build_spec_from_cli(
     silent: bool,
     save_result_to_inbox: Optional[bool] = None,
     share_session: bool = True,
+    max_concurrency: int = 1,
     timeout_seconds: int = 120,
+    misfire_grace_seconds: int = 600,
     tool_safety: bool = False,
 ) -> dict:
     """Build CronJobSpec JSON payload from CLI args (no id)."""
@@ -258,9 +260,9 @@ def _build_spec_from_cli(
     }
     runtime = {
         "share_session": share_session,
-        "max_concurrency": 1,
+        "max_concurrency": max_concurrency,
         "timeout_seconds": timeout_seconds,
-        "misfire_grace_seconds": 600,
+        "misfire_grace_seconds": misfire_grace_seconds,
         "tool_safety": tool_safety,
     }
     if task_type == "text":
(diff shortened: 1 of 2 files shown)


lalaliat:
Hi @AaronZ345, thank you for the quick fix! However, there are two PRs fixing the same issue. After reviewing both PRs, we decided to go with #6236 for the following reasons:
1. Architecture: #6236 rewrites _resolve_update_spec to deep-copy the existing spec and only patch non-None CLI fields, which is future-proof — any new fields added to the spec will automatically be preserved without code changes. The current approach of threading parameters through _build_spec_from_cli would require manual updates each time.
2. Additional fixes: #6236 also addresses two related edge cases — --schedule-type scheduled writing an invalid value (causing 422), and --text crashing on malformed request.input.

We really appreciate your contribution. If you have further ideas, feel free to build together!

PR #6182, opened 2026-07-16
fix(crons): honor final delivery mode

### LINKED ISSUE #6177: [Bug]: cron final delivery mode still forwards every completed event
## Problem

Cron jobs expose `dispatch.mode` with `stream` and `final` values in the schema, CLI, Console, and documentation. The executor never reads that field. It iterates over the agent stream and calls `send_event` for every event regardless of the configured mode.

This makes `mode=final` behave like `mode=stream`: intermediate completed messages, including pre-tool commentary, can be delivered before the final answer. It also prevents callers from relying on the documented single final delivery behavior.

## Expected behavior

- `mode=stream`: keep the existing behavior and forward events as they arrive.
- `mode=final`: consume the full stream so tool calls and session state complete, then deliver only the final completed message event.
- Delivery failures should keep the current error reporting semantics.
- Silent jobs should continue consuming the full stream without channel delivery.

## Suggested fix

Buffer the latest completed message while consuming the stream in final mode, and call `send_event` once after the stream finishes. Add executor tests covering stream, final, silent, and final-mode delivery failure behavior.

AaronZ345:
## Description

The executor treated `final` like `stream` and forwarded every completed event. This consumes the full run, remembers the last completed message, and delivers it once after the stream ends. `stream` and `silent` keep their existing behavior, including delivery error handling.

Fixes #6177.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Core / Backend
- [x] Tests

## Testing

- `pytest tests/unit/app/crons/test_executor.py -q`
- pre-commit hooks on the changed files
- Full unit suite and an upstream baseline run

## Evidence

```text
tests/unit/app/crons/test_executor.py: 4 passed
changed-file pre-commit hooks: all passed
full unit suite: 2 failed, 4802 passed, 6 skipped
upstream/main malformed-tool-call baseline: 2 failed, 7 passed
```

The two full-suite failures are the same `test_runtime_malformed_tool_call.py` assertions reproduced on unmodified `upstream/main`.

### DIFF
--- a/src/qwenpaw/app/crons/executor.py
+++ b/src/qwenpaw/app/crons/executor.py
@@ -14,6 +14,7 @@ from ..inbox_trace_store import (
 )
 from .models import CronJobSpec
 from ...security.tool_guard.execution_level import ToolExecutionLevel
+from ...schemas import RunStatus
 
 logger = logging.getLogger(__name__)
 
@@ -30,6 +31,8 @@ class CronExecutor:
         - task_type text: send fixed text to channel
         - task_type agent: ask agent with prompt, send reply to channel (
             stream_query + send_event)
+        - final delivery: consume the stream, then send only the last
+            completed message
         - silent agent task: consume the full agent stream without channel
             delivery, while preserving session and trace state
         """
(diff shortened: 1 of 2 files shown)


lalaliat (approved):
LGTM. Thank you for your contribution!

PR #6198, opened 2026-07-16
feat: bound multi-agent startup and improve readiness UX

### LINKED ISSUE #6144: [Performance]: Bound concurrent ReMe initialization during multi-agent startup
## Summary

Starting many agents at once currently starts all of their ReMe applications concurrently. With a large configuration (for example, 36 agents), the temporary working sets of indexing and resource initialization overlap, causing a high startup memory peak. The same burst also performs many embedding capability checks at once. A provider rate limit can then be interpreted as an unavailable embedding service, causing ReMe instances to disable embedding even though the provider is healthy.

This issue tracks a QwenPaw-side startup scheduler. It is related to #6124, but it is intentionally scoped to bounded multi-agent initialization rather than the reported 48 GiB behavior itself.

## Why this should be handled in QwenPaw

QwenPaw owns the lifecycle and startup order of its agents. ReMe initialization has a non-trivial temporary working set, and its current embedding health/capability probe is not reliable when dozens of applications perform it simultaneously. Even if every individual ReMe instance eventually releases its temporary memory, unconstrained startup lets all peaks overlap.

## Local measurements

We could not reproduce the unbounded 48 GiB growth reported in #6124 on macOS/arm64. However, the controlled measurements do confirm meaningful overlap during concurrent startup:

- One isolated QwenPaw-configured ReMe manager peaked at 278.8 MiB on a cold index and 246.1 MiB with existing metadata.
- A full QwenPaw startup with 2 enabled agents peaked at 417.9 MiB and later returned to a low steady state.
- In a same-process controlled test, constructing 36 ReMe `Application` objects added only about 5.9 MiB in total after the shared import/configuration baseline.
- Concurrently calling `Application.start()` for 36 applications peaked at approximately 590.9–907.7 MiB on this machine, depending on scheduling and page residency.
- The 36-instance test used separate workspace copies and a local fixed embedding boundary, so it does not reproduce the external provider's rate limiting or explain the Windows 48 GiB result. It does demonstrate that the per-workspace startup working sets overlap substantially.

These results are consistent with a startup burst rather than a constructor leak. They also show why the total should not be estimated as `single-process RSS × agent count`: imports are shared, while file stores, indexes, chunks, and startup temporaries are largely per application.

## Additional reliability problem: embedding capability probes

During real startup, each ReMe instance tests the embedding provider's capability. With 36 simultaneous initializations, these probes can hit the provider's request-rate limit. The rate-limit response is then treated as an embedding capability failure, and the affected ReMe instances disable embedding.

A transient throttling response should not be interpreted as evidence that the configured embedding model is unsupported. Limiting startup concurrency reduces the burst immediately; the capability-check path should also distinguish transient/rate-limit failures from permanent incompatibility.

## Proposed behavior

1. Initialize the `default` agent first so that the primary workspace becomes usable predictably.
2. Initialize the remaining agents through a process-wide bounded queue/semaphore.
3. Use a default maximum concurrency of 5 for ReMe initialization. The limit may be configurable, but startup must not use unbounded `asyncio.gather()` across all agents.
4. Apply the limit to the expensive ReMe startup phase, including index/component initialization and the initial embedding capability check—not merely to agent object construction.
5. Keep the web service responsive and expose agents as ready as their initialization completes.
6. Treat embedding rate limits and other transient provider failures as retryable/degraded startup states. Do not permanently mark embedding unavailable based on a single throttled probe.
7. Add jittered backoff for retries so queued agents do not create another synchronized request burst.

A reasonable initial sequence is:

```text
default agent: initialize and become ready
remaining agents: initialize with max concurrency = 5
background index/resource jobs: start only after the corresponding ReMe application is ready
```

## Acceptance criteria

- With 36 enabled agents, no more than 5 non-default ReMe applications are in the expensive initialization phase at once.
- The `default` agent is initialized before the remaining queue begins.
- A rate-limited embedding capability probe does not cause embedding to be permanently disabled.
- Per-agent initialization failures are isolated and visible; they do not block unrelated agents indefinitely.
- Startup logs include queue/start/ready/failure events and enough context to diagnose provider throttling.
- Tests cover ordering, the concurrency ceiling, failure isolation, cancellation/shutdown, and retry behavior for a simulated HTTP 429 response.
- A memory regression test or reproducible profiling script demonstrates a lower peak for 36 agents than unbounded concurrent startup.

## Environment used for the local investigation

- QwenPaw: 2.0.0.post2, editable install, commit `9812b896c34dbcd300ddc7f8fa79fe7846dcb4c7`
- ReMe/reME-ai: 0.4.1.0
- Python: 3.12.13
- OS: macOS 26.5.2, arm64, approximately 18 GiB RAM
- Embedding configuration: OpenAI-compatible `text-embedding-v4`, 1024 dimensions

## Related issue

- #6124

rayrayraykk:
## Description

This PR bounds multi-agent startup concurrency and makes partial readiness visible to both operators and Console users.

The previous startup path used an unbounded `asyncio.gather()` for every enabled agent. ReMe initialization, indexing, channel setup, and provider probes could therefore overlap across all configured workspaces, creating a large transient memory and request burst. The new scheduler:

- starts the `default` agent and the built-in QA agent concurrently when QA is enabled;
- marks the service ready once all enabled core agents are available;
- starts custom agents only after the core phase, with a bounded semaphore;
- uses `QWENPAW_CUSTOM_AGENT_STARTUP_CONCURRENCY` (default: `5`) to control the custom-agent limit;
- queues newly created and re-enabled agents through the same bounded scheduler;
- isolates per-agent failures and exposes `pending`, `starting`, `running`, `failed`, and `disabled` states;
- keeps a Rich startup banner and custom-agent progress display visible while normal terminal logs continue above it, without adding ANSI output to `qwenpaw.log`.

The Console now reflects those lifecycle states and improves agent selection:

- disabled agents are available in a collapsible section and can be enabled in place;
- state indicators use gray, animated yellow, green, and red status colors;
- the default agent is always pinned first;
- other agents can be pinned or unpinned with a long press;
- pinned agents remain in the pinned section even when disabled;
- pin state is persisted independently from the existing `enabled` state, preserving configuration compatibility.

**Related Issue:** Relates to #6144

**Security Considerations:** No authentication or secret-handling behavior changes. The new configuration field stores only an agent UI ordering preference.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

## Testing

Backend lifecycle, ordering, API, and terminal-display tests:

```bash
PYTHONPATH=src conda run --no-capture-output -n QwenPaw pytest -q \
  tests/unit/app/test_agents_ordering.py \
  tests/unit/app/test_multi_agent_manager_startup.py \
  tests/unit/app/routers/test_agents_router.py
```

Frontend selector, status, ordering, gesture, and API tests:

```bash
cd console
npm run test:run -- \
  src/components/AgentSelector/AgentSelector.test.tsx \
  src/components/AgentSelector/useAgentLongPress.test.tsx \
  src/components/AgentStatusIndicator/AgentStatusIndicator.test.tsx \
  src/pages/Settings/Agents/reorder.test.ts \
  src/api/modules/agents.test.ts
```

Additional validation:

```bash
cd console
npx tsc -b --noEmit
npm run build
npx eslint \
  src/components/AgentSelector/index.tsx \
  src/components/AgentSelector/useAgentLongPress.ts \
  src/api/modules/agents.ts \
  src/api/types/agents.ts \
  src/pages/Settings/Agents/useAgents.ts \
  src/pages/Settings/Agents/components/AgentTable.tsx \
  src/pages/Settings/Agents/reorder.ts
```

## Evidence

Backend test result:

```text
................................ [100%]
40 passed
```

Frontend test result:

```text
Test Files  5 passed (5)
Tests       31 passed (31)
```

The production Console build completed successfully:

```text
15108 modules transformed.
✓ built in 14.60s
```

Scoped pre-commit checks passed for all changed files, including Python AST, JSON validation, private-key detection, whitespace, mypy, Black, flake8, and pylint.

An isolated runtime smoke test created enabled, disabled, pinned, and unpinned agents. `GET /api/agents` returned the expected stable grouping without mutating the disabled state:

```json
[
  {"id": "default", "enabled": true, "pinned": true},
  {"id": "archive", "enabled": false, "pinned": true},
  {"id": "QwenPaw_QA_Agent_0.2", "enabled": true, "pinned": false},
  {"id": "research", "enabled": true, "pinned": false}
]
```

## Additional Notes

- This PR intentionally does not change ReMe internals or embedding-provider retry classification. It reduces the synchronized startup burst tracked in #6144; transient HTTP 429 handling can be addressed separately.
- Non-TTY environments retain plain log output instead of dynamic terminal rendering.
- Existing agent configurations remain compatible: missing `pinned` values default to `false`, while explicit `enabled: false` values are preserved through load/save and startup scheduling.


jinliyl:
Thank you for putting this together. The bounded scheduler, per-agent visibility, and Console work are all useful directions, and I checked the current head (`ffc56ed`) in a clean temporary checkout. I found several issues that I think should be addressed before merging.

### 1. Readiness is still blocked by the built-in QA agent, and `default` is not actually initialized first

In `MultiAgentManager.start_all_configured_agents()`, `default` and `QwenPaw_QA_Agent_0.2` are both placed in `core_agent_ids` and awaited by one `asyncio.gather()`. The `on_core_ready` callback is invoked only after that entire gather returns.

This has two consequences:

- the built-in QA agent starts concurrently with `default`, contrary to #6144's proposed/acceptance behavior that `default` should initialize before the remaining queue begins; and
- `/api/healthz` remains 503 even after `default` is fully available if the QA agent is still slow or stuck.

I reproduced the second point with a focused async test: `default` completed while the QA task remained blocked, but the readiness callback had still been called 0 times. The test failed with `Expected 'mock' to have been called once. Called 0 times.`

I suggest awaiting `default` separately, publishing readiness immediately if it succeeds, and then putting **all** remaining enabled agents (including the built-in QA agent) through the bounded queue. The QA agent can still be first in that queue if it deserves priority.

### 2. Agents created after background startup can remain `pending` forever in the Console

`get_agent_startup_status()` returns `PENDING` for every enabled agent that is neither in `_agent_startup_statuses` nor loaded in `self.agents`. However, `POST /api/agents` only persists the new enabled profile; it neither starts the agent nor registers it with the startup scheduler/status map.

As a result, immediately after creating an agent:

- `GET /api/agents` reports `startup_status: "pending"` indefinitely;
- `useAgentStatusPolling()` keeps polling indefinitely; and
- `AgentTable` treats that status as startup in progress and disables delete/disable actions.

There is also an observable backend inconsistency for this state: a fresh manager reports `status=pending` while `is_agent_startup_in_progress()` returns `False` for the same ID. The current router tests hide this because their manager mock returns `RUNNING` for every enabled profile.

Please either schedule/start newly created enabled agents and track that task, or add a terminal/lazy state (for example, `idle`) so `pending` is reserved for work that is genuinely queued and expected to complete. A real-manager test covering create -> list -> action availability would be valuable here.

### 3. The reorder API no longer has round-trip semantics

`PUT /api/agents/order` still accepts and returns any complete order, but `GET /api/agents` now passes the persisted order through `_display_agent_order()`, which always moves `default` first and groups pinned agents ahead of regular agents. Therefore a successful PUT can immediately be contradicted by the next GET.

This is already breaking the P0 integration test in CI (`test_multi_agent_reorder_and_delete_order_adjusts`). I also ran the existing P1 round-trip test directly:

```text
pytest -q tests/integration/test_agents.py::test_api_agents_order_put_roundtrip
FAILED: GET returned ['default', 'QwenPaw_QA_Agent_0.2'] after PUT had accepted
        ['QwenPaw_QA_Agent_0.2', 'default']
```

If grouping is only a Console presentation rule, it would be cleaner to preserve the API's persisted order and group it in the UI. If the API contract is intentionally changing, the reorder endpoint should reject or normalize incompatible payloads, return the effective order, and the integration tests/contract documentation should be updated together.

### 4. The new concurrency parsing tests pollute the shared Python module state and cause all Unit Test jobs to fail

The tests in `test_multi_agent_manager_startup.py` repeatedly call `importlib.reload(qwenpaw.constant)`. Reloading redefines `EnvVarLoader`, while modules that previously did `from qwenpaw.constant import EnvVarLoader` retain the old class object. Later fixtures patch `qwenpaw.constant.EnvVarLoader.get_str` on the new class, but `security.tool_guard.utils` still calls the old one, so the patches no longer apply.

I confirmed the identity change directly (`before_reload=True`, `after_reload=False`) and reproduced the CI result locally:

```text
pytest -q tests/unit/app/test_multi_agent_manager_startup.py \
  tests/unit/security/tool_guard/test_utils.py
8 failed, 53 passed
```

Each file passes in isolation (`12 passed` and `49 passed` respectively), which confirms this is cross-test pollution. This is why the same eight tool-guard tests fail on Python 3.11/3.13 and Linux/macOS/Windows in CI. Please avoid reloading this widely imported module in-process. A small pure parsing function, monkeypatching the lookup boundary, or a subprocess-based environment test would avoid replacing shared class identities.

### 5. The documented default concurrency is inconsistent

The PR description and review summary say the default is `2`, while `DEFAULT_CUSTOM_AGENT_STARTUP_CONCURRENCY` and the new tests use `5`. Issue #6144 also proposes `5`. Please update the PR description so operators do not configure or evaluate the rollout using the wrong default.

For completeness, the focused ordering/router/health tests passed locally (`27 passed`), as did the startup display tests (`4 passed`). The failures above are therefore isolated and reproducible rather than general environment/setup failures.

Thanks again for the substantial work here. Once the scheduler sequencing/readiness semantics, post-startup agent state, API ordering contract, and test isolation are corrected, this should be much safer to merge.

console/src/pages/Settings/Agents/useAgents.ts
```diff
@@ -22,57 +24,119 @@ export function useAgents(): UseAgentsReturn {
...
-    loadAgents();
-  }, []);
+    void loadAgents();
+  }, [loadAgents]);
+
+  const refreshStatuses = useCallback(
+    () => fetchAgents(false, false),
+    [fetchAgents],
+  );
+  useAgentStatusPolling(agents, refreshStatuses);
```
zhaozhuang521: The AgentSelector (which remains in the sidebar) is already polling `GET /api/agents` via `useAgentStatusPolling`. However, the `useAgents` hook on the Settings page initiates a second, identical polling process; consequently, when the Agent settings page is open, two concurrent polling requests—each with a 1.5-second interval—target the same endpoint and write to the same `agentStore`. It is recommended to consolidate the polling logic into a single source (e.g., running a single polling process at the store or context layer) to avoid redundant requests.

cuiyuebing (approved):
LGTM

PR #6203, opened 2026-07-16
fix(utils): bound and hide the Windows tasklist liveness probe

Yigtwxx:
## Description

`_is_pid_running()` in `src/qwenpaw/utils/command_runner.py` shells out to `tasklist` to check whether a PID is still alive on Windows. The call is missing three things the rest of this same module already gets right:

1. **No `timeout`.** `subprocess.check_output` blocks with no upper bound. `_is_pid_running` is reached from `shutdown_process_sync()`, so a wedged `tasklist` stalls shutdown indefinitely.
2. **No console-window suppression.** This module already defines `windows_hidden_subprocess_kwargs()` and applies it in `run_command()` and in the process-launch path. `_is_pid_running` is the one call site that does not. Since `_wait_for_process_exit()` calls it on every iteration of its `sleep(0.1)` poll loop, stopping a local model server spawns a burst of `tasklist` processes, each able to flash a console window.
3. **`text=True` with no `errors=`.** Output is decoded with the locale codec, so on a non-UTF-8 console (cp936/GBK on Chinese Windows) a `UnicodeDecodeError` propagates out of the shutdown path. This is the same failure class that #6140 fixed in `system_info._run_command()`; this sibling call site was missed.

The `except` tuple is also narrower than it should be: it catches `CalledProcessError` and `FileNotFoundError`, so a bare `OSError` (for example `[WinError 193]` from a broken `tasklist` shim) escapes.

Live call chain:

```
llamacpp.py  shutdown_process_sync(graceful_timeout=5.0)
  -> _wait_for_process_exit()          # while-loop, time.sleep(0.1)
    -> _is_pid_running()               # one tasklist spawn per iteration
```

The fix assembles the probe kwargs the way `run_command()` already does, reuses `windows_hidden_subprocess_kwargs()`, adds `errors="replace"` and a bounded `timeout`, and widens the `except` tuple to `(OSError, subprocess.SubprocessError)` — the exact tuple `system_info._run_command()` uses. Failure semantics are unchanged: the probe still returns `False` when it cannot confirm the PID is alive. The POSIX branch is untouched.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Tests

## Checklist

- [ ] Documentation updated (if needed) — not user-facing, no docs change
- [x] Ready for review

## Testing

Three unit tests were added next to the existing `_is_pid_running` coverage in `tests/unit/utils/test_command_runner.py`. They are platform-agnostic, so they assert the same thing on the ubuntu, macos and windows CI legs:

- `test_is_pid_running_tasklist_probe_is_bounded_and_hidden` captures the kwargs handed to `check_output` and asserts a `timeout` is set, `errors="replace"` is set, and the result of `windows_hidden_subprocess_kwargs()` is merged in. The helper is stubbed rather than relying on the host OS, because `subprocess.CREATE_NO_WINDOW` does not exist on Linux/macOS and the helper correctly returns `{}` there.
- `test_is_pid_running_returns_false_when_tasklist_times_out` makes `check_output` raise `TimeoutExpired` and asserts the probe returns `False` instead of propagating.
- `test_is_pid_running_returns_false_when_tasklist_unavailable` makes it raise a bare `OSError` and asserts the same — this one fails on the previous `except` tuple.

The pre-existing `test_is_pid_running_uses_tasklist_on_windows` still passes unchanged.

## Evidence

All measurements below are from a real Windows 11 (AMD64) machine, not simulated.

**The poll loop is slower than it intends to be.** `_wait_for_process_exit()` sleeps 0.1s between polls, but each poll spawns a `tasklist` that costs ~0.157s, so the real interval is ~2.5x the intended one, and a 5s graceful shutdown can spawn ~19 `tasklist` processes:

```
$ python -c "...median of 7 tasklist spawns..."
median tasklist spawn: 0.157s (n=7)
intended poll interval: 0.100s
actual  poll interval: 0.257s
spawns during a 5.0s graceful shutdown: up to 19
```

**The new tests fail against the current code and pass with the fix.** Reverting only the `_is_pid_running` change and re-running:

```
$ pytest tests/unit/utils/test_command_runner.py -k is_pid_running -q
E       OSError: [WinError 193] not a valid Win32 application
tests\unit\utils\test_command_runner.py:762: OSError
=========================== short test summary info ===========================
FAILED tests/unit/utils/test_command_runner.py::test_is_pid_running_tasklist_probe_is_bounded_and_hidden
FAILED tests/unit/utils/test_command_runner.py::test_is_pid_running_returns_false_when_tasklist_times_out
FAILED tests/unit/utils/test_command_runner.py::test_is_pid_running_returns_false_when_tasklist_unavailable
3 failed, 2 passed, 19 deselected in 0.95s
```

Note the `OSError: [WinError 193]` escaping the probe entirely — that is the narrow `except` tuple, reproduced.

**With the fix applied, the file's full suite is green:**

```
$ pytest tests/unit/utils/test_command_runner.py -q
24 passed in 0.71s
```

**No regressions across the whole unit suite:**

```
$ pytest tests/unit -q
4804 passed, 52 skipped, 56 warnings in 247.88s (0:04:07)
```

**Every pre-commit hook passes on the touched files:**

```
$ pre-commit run --files src/qwenpaw/utils/command_runner.py \
                        tests/unit/utils/test_command_runner.py
```

### DIFF
--- a/src/qwenpaw/utils/command_runner.py
+++ b/src/qwenpaw/utils/command_runner.py
@@ -16,6 +16,9 @@ from qwenpaw.exceptions import (
     ProcessLaunchError,
 )
 
+# Bounds the liveness probe so a wedged ``tasklist`` cannot stall shutdown.
+_PID_PROBE_TIMEOUT = 5.0
+
 
 @dataclass(frozen=True)
 class CommandResult:
@@ -575,13 +578,19 @@ def _is_pid_running(
     platform_name: str,
 ) -> bool:
     if platform_name == "nt":
+        probe_kwargs: dict[str, Any] = {
+            "stderr": subprocess.STDOUT,
+            "text": True,
+            "errors": "replace",
+            "timeout": _PID_PROBE_TIMEOUT,
+        }
+        probe_kwargs.update(windows_hidden_subprocess_kwargs())
         try:
             output = subprocess.check_output(
                 ["tasklist", "/fi", f"PID eq {pid}"],
-                stderr=subprocess.STDOUT,
-                text=True,
+                **probe_kwargs,
             )
-        except (subprocess.CalledProcessError, FileNotFoundError):
+        except (OSError, subprocess.SubprocessError):
             return False
         return str(pid) in output
 
--- a/src/qwenpaw/utils/command_runner.py
+++ b/src/qwenpaw/utils/command_runner.py
@@ -16,7 +16,8 @@ from qwenpaw.exceptions import (
     ProcessLaunchError,
 )
 
-# Bounds the liveness probe so a wedged ``tasklist`` cannot stall shutdown.
+# Upper bound for the liveness probe so a wedged ``tasklist`` cannot stall
+# shutdown. Callers with a deadline of their own pass a smaller budget.
 _PID_PROBE_TIMEOUT = 5.0
 
 
@@ -531,17 +532,27 @@ def _wait_for_process_exit(
         if not process.is_alive():
             process.join(timeout=0)
             return True
-        if not _is_pid_running(process.pid, process.platform_name):
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            break
+        if not _is_pid_running(
+            process.pid,
+            process.platform_name,
+            probe_timeout=min(_PID_PROBE_TIMEOUT, remaining),
+        ):
             process.join(timeout=0)
             return True
         remaining = deadline - time.monotonic()
         if remaining <= 0:
             break
         time.sleep(min(0.1, remaining))
+    # The deadline is spent, so there is no budget left to probe with.
+    # Report "still running" and let the caller escalate; a stale exit is
+    # picked up by the next phase, which probes with its own budget.
     if not process.is_alive():
         process.join(timeout=0)
         return True
-    return not _is_pid_running(process.pid, process.platform_name)
+    return False
 
 
 def _coerce_subprocess_path(
@@ -576,13 +587,15 @@ def _supports_process_groups(process: ManagedProcess) -> bool:
 def _is_pid_running(
     pid: int,
     platform_name: str,
+    *,
+    probe_timeout: float = _PID_PROBE_TIMEOUT,
 ) -> bool:
     if platform_name == "nt":
         probe_kwargs: dict[str, Any] = {
             "stderr": subprocess.STDOUT,
             "text": True,
             "errors": "replace",
-            "timeout": _PID_PROBE_TIMEOUT,
+            "timeout": probe_timeout,
         }
         probe_kwargs.update(windows_hidden_subprocess_kwargs())
         try:
--- a/src/qwenpaw/utils/command_runner.py
+++ b/src/qwenpaw/utils/command_runner.py
@@ -591,7 +591,10 @@ def _is_pid_running(
                 **probe_kwargs,
             )
         except (OSError, subprocess.SubprocessError):
-            return False
+            # A failed probe says nothing about the PID. Callers read False as
+            # a confirmed exit, so a timed-out or unavailable tasklist must not
+            # short-circuit the kill escalation in shutdown_process_sync().
+            return True
         return str(pid) in output
 
     try:
--- a/tests/unit/utils/test_command_runner.py
+++ b/tests/unit/utils/test_command_runner.py
@@ -6,6 +6,7 @@ import asyncio
 import os
 import subprocess
 from pathlib import Path
+from typing import Any
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


wangfei010313:
@Yigtwxx  Thank you for your contribution, but we'd like you to consider the following suggestions:
A timed-out tasklist probe does not necessarily mean that the target process has exited. _wait_for_process_exit() interprets False as a confirmed exit, so if tasklist times out while the process is still alive, shutdown_process_sync() may report exited=True and terminated_gracefully=True, and skip the subsequent kill().
Could we handle probe failures conservatively—e.g. return True on TimeoutExpired/unexpected OSError, or introduce an “unknown” result—and only return False when a successful tasklist invocation confirms that the PID is absent? It would also be helpful to add a regression test where check_output() raises TimeoutExpired and verify that shutdown still escalates to kill() rather than reporting a graceful exit.

Yigtwxx:
@wangfei010313 Thanks for catching this — you're right, and it's a real hole rather than a hypothetical one. `_wait_for_process_exit()` reads `False` as a *confirmed* exit, so a timed-out probe made `shutdown_process_sync()` report `exited=True` and `terminated_gracefully=True` and skip `kill()` entirely, for a process that may well still be alive. Adding the `timeout` in this PR made that path easier to reach, so it needed fixing here.

Pushed a fix taking the conservative option you suggested: probe failures now return `True`, and only a successful `tasklist` invocation may confirm the PID is absent.

```python
        except (OSError, subprocess.SubprocessError):
            # A failed probe says nothing about the PID. Callers read False as
            # a confirmed exit, so a timed-out or unavailable tasklist must not
            # short-circuit the kill escalation in shutdown_process_sync().
            return True
```

This is safe rather than merely cautious: `_wait_for_process_exit()` already gates on `process.is_alive()` before it ever probes, so `_is_pid_running()` is only a secondary confirmation that a stale `is_alive()` missed an exit. Returning `True` on failure just falls back to the `is_alive()` verdict instead of overriding it. I preferred this over a three-state "unknown" result because it keeps the call sites unchanged and there is no caller that can act usefully on "unknown" beyond "keep waiting, then escalate", which is exactly what `True` already produces.

Tests, including the regression test you asked for:

- **`test_shutdown_process_sync_escalates_when_pid_probe_times_out`** — drives the real `_wait_for_process_exit()` (not a stub) with `check_output` raising `TimeoutExpired`, and asserts `exited is False`, `terminated_gracefully is False`, `killed is True`, `timed_out is True`, and that both signals are sent (`signals == [15, 9]`). Against the previous code it fails with exactly the outcome you described:

  ```
  E        +  where True = ShutdownResult(command=['demo'], pid=4321, exited=True,
               terminated_gracefully=True, killed=False, timed_out=False, returncode=None).exited
  ```

- **`test_is_pid_running_reports_gone_when_tasklist_finds_no_match`** — pins the `False` path to a successful no-match invocation (`INFO: No tasks are running...`), so the fix can't silently degrade into "always alive".
- The two probe-failure tests were renamed and inverted to `..._assumes_alive_when_tasklist_times_out` / `..._unavailable`.
- Also applied the AI review's Low note: the `TimeoutExpired` test now uses `_PID_PROBE_TIMEOUT` instead of a hardcoded `5.0`.

Reverting only the `return True` reproduces the bug — 3 failures; with the fix the file's suite is green at 26 passed, and `pre-commit` passes on both touched files.

wangfei010313:
@Yigtwxx 
 Great work on fixing the correctness issues—the probe timeout, hidden window, locale/GBK-safe decoding, and conservative `return True` on probe failure are all solid improvements. The regression tests also cover the previously reported false-exit behavior well.

 However, one timeliness issue remains: `_PID_PROBE_TIMEOUT = 5.0` is independent of the caller-supplied timeout used by `_wait_for_process_exit()`. As a result, the PID probe can substantially exceed the shutdown deadline.

 In the current loop, `_is_pid_running()` is called before the remaining budget is checked:

 ```python
 while True:
      if not process.is_alive():
         ...
     if not _is_pid_running(process.pid, process.platform_name):
         ...  # may block for up to 5 seconds
     remaining = deadline - time.monotonic()
     if remaining <= 0:
         break
     time.sleep(min(0.1, remaining))

 return not _is_pid_running(process.pid, process.platform_name)
 ```
 There are two related problems:

 1. An in-loop probe may begin with only a tiny amount of budget remaining but still run for up to five seconds.
 2. After the deadline is exceeded, the final probe may add another five seconds.

 Consequently, each phase can take approximately:

 ```text
 caller timeout + 2 * _PID_PROBE_TIMEOUT
 ```

 For `shutdown_server_sync()`, which uses `graceful_timeout=5.0` and `kill_timeout=1.0`, the nominal six-second wait budget could therefore approach:

 | Phase | Budget | Possible wall time |
 |---|---:|---:|
 | Graceful shutdown | 5s | ~15s |
 | Kill | 1s | ~11s |
 | **Total** | **6s** | **~26s** |

 Could `_is_pid_running()` accept a probe timeout so `_wait_for_process_exit()` can bound it by the remaining budget?

 ```python
 def _is_pid_running(
     pid: int,
     platform_name: str,
     *,
     probe_timeout: float = _PID_PROBE_TIMEOUT,
 ) -> bool:
     ...
 ```

 The loop could then check the deadline before probing and pass:

 ```python
 probe_timeout = min(_PID_PROBE_TIMEOUT, remaining)
 ```

 Once the deadline has expired, the post-loop path should avoid launching another `tasklist` process. It can perform the final local `process.is_alive()` check and return `False` if the process still appears alive, allowing the caller to escalate to `kill()`.

 A regression test that simulates slow probes and asserts the elapsed-time budget would help prevent this behavior from returning.

Yigtwxx:
@wangfei010313 Confirmed, and the overshoot is worse than a timing wart: the probe runs *before* the remaining budget is checked, so the last in-loop probe can start with ~0 budget and still block for the full 5s, and the post-loop probe adds another 5s. Pushed a fix along the lines you sketched.

**1. `_is_pid_running()` now takes the budget from its caller.** `_PID_PROBE_TIMEOUT` becomes an upper bound rather than the value:

```python
def _is_pid_running(
    pid: int,
    platform_name: str,
    *,
    probe_timeout: float = _PID_PROBE_TIMEOUT,
) -> bool:
```

**2. The wait loop checks the deadline before probing, and bounds the probe by what is left:**

```python
    deadline = time.monotonic() + timeout
    while True:
        if not process.is_alive():
            process.join(timeout=0)
            return True
        remaining = deadline - time.monotonic()
        if remaining <= 0:
            break
        if not _is_pid_running(
            process.pid,
            process.platform_name,
            probe_timeout=min(_PID_PROBE_TIMEOUT, remaining),
        ):
            process.join(timeout=0)
            return True
        remaining = deadline - time.monotonic()
        if remaining <= 0:
            break
        time.sleep(min(0.1, remaining))
```

**3. The post-deadline path no longer spawns `tasklist`**, exactly as you asked — it does the free local check and otherwise reports "still running" so the caller escalates:

```python
    # The deadline is spent, so there is no budget left to probe with.
    # Report "still running" and let the caller escalate; a stale exit is
    # picked up by the next phase, which probes with its own budget.
    if not process.is_alive():
        process.join(timeout=0)
        return True
    return False
```

The one behavior this trades away is the rare case where `is_alive()` is stale-`True` at the deadline but the PID is genuinely gone. That case now escalates to `kill()` instead of being reported as a graceful exit — `kill()` is wrapped in `suppress(ProcessLookupError)`, and the kill phase probes again with its own budget, so the exit is still detected there. Escalating on a dead PID is cheap; overrunning the deadline is not.

### Measurement

I put the elapsed-time regression test you asked for behind a virtual clock, so it asserts the budget without sleeping in CI. With a fully wedged `tasklist` (every probe burns exactly the timeout it was handed), `shutdown_process_sync(graceful_timeout=5.0, kill_timeout=1.0)` on the previous code:

```
$ pytest tests/unit/utils/test_command_runner.py -q --tb=line -k "probe_timeout or budget or wedged"
tests\unit\utils\test_command_runner.py:901: TypeError: _is_pid_running() got an unexpected keyword argument 'probe_timeout'
E   assert 5.0 <= 0.3
     +  where 5.0 = max([5.0, 5.0, 5.0, 5.0])
E   assert 20.0 <= (6.0 + 1e-09)
     +  where 20.0 = <_VirtualClock object>.now
3 failed, 26 deselected in 0.61s
```

So the measured worst case is 20s against a 6s budget (5+5 graceful, 5+5 kill) rather than the ~26s in your table — same defect, slightly smaller number. With the fix it lands on 6.0s exactly.

New tests:

- **`test_shutdown_process_sync_keeps_budget_when_probes_are_wedged`** — the elapsed-budget regression test: asserts `clock.now <= 6.0`, asserts `probe_count == 2` (one bounded probe per phase, none after the deadline), and still asserts the escalation you flagged last round (`exited is False`, `timed_out is True`, `signals == [15, 9]`).
- **`test_wait_for_process_exit_bounds_probe_by_remaining_budget`** — pins the invariant directly: no probe timeout exceeds the remaining budget, and the wait does not overshoot its deadline.
- **`test_is_pid_running_honours_caller_supplied_probe_timeout`** — the kwarg actually reaches `check_output`.

The file's suite is green at 29 passed, the full unit suite at 4804 passed / 52 skipped, and `pre-commit` passes on both touched files.

wangfei010313 (approved):
LGTM

Yigtwxx:
@wangfei010313 A gentle nudge on this one, since it has been quiet for a while.

The PR has been approved since 2026-07-22 and every completed check is green. The only two still outstanding are `Maintainer Approval` and `AI Review Approval`, which are gates a maintainer has to release, so the merge state reads unstable even though nothing is wrong with the branch.

I re-verified today that this is not a staleness problem: `src/qwenpaw/utils/command_runner.py` and its test have not been modified on `main` since this branch was cut, so there is no conflict and no rebase is needed. GitHub still reports the PR as mergeable. I have deliberately not pushed anything new, to avoid resetting the approval you already gave.

Happy to make further changes if anything is still outstanding on your side.

Yigtwxx:
Thanks for the merge, and for the two review rounds — both caught real defects the first version would have shipped.

```
x: Yigtwx
```

Issue #6197, opened 2026-07-16
# [Bug] QwenPaw Desktop (frozen binary) hangs on startup when `nvidia-smi` hangs indefinitely


hingeek:
# [Bug] QwenPaw Desktop (frozen binary) hangs on startup when `nvidia-smi` hangs indefinitely

## Environment

- **QwenPaw Desktop version**: v2.0.0.post2 (Tauri build, frozen PyInstaller binary)
- **OS**: Windows 10 (AMD64)
- **NVIDIA Driver**: 560.94 (rolled back from a newer version that also exhibited the same problem)
- **GPU**: NVIDIA RTX series (hang is driver-related, ~66-day uptime known upstream bug)
- **Install method**: Desktop Tauri app (not pip install)

## Symptom

Upon restarting QwenPaw Desktop, the backend (`qwenpaw-backend.exe`, a PyInstaller bundle) hangs indefinitely during startup.

The hang occurs every time the system has been running for an extended period and `nvidia-smi` itself hangs when called directly from Command Prompt (the well-known [nvidia-smi hang after ~66 days uptime](https://github.com/NVIDIA/open-gpu-kernel-modules/issues/971) bug).

## Root Cause

The frozen `qwenpaw-backend.exe` contains an **old version** of `qwenpaw/utils/system_info.py` that calls nvidia-smi via:

```python
proc = subprocess.Popen(
    "cmd /c start /b nvidia-smi",
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    shell=True,
)
stdout, stderr = proc.communicate()  # NO timeout parameter!
```

Two problems:

1. **No timeout**: `proc.communicate()` has no `timeout=N` argument, so when `nvidia-smi` hangs, the backend hangs forever.
2. **`start /b` pattern**: `cmd /c start /b nvidia-smi` adds unnecessary shell indirection. The current source code (`system_info.py` on `main` branch) has since been fixed to use `subprocess.run(["nvidia-smi"], timeout=5, check=False)`, but the **PyInstaller frozen binary shipped with the Desktop app is not rebuilt** — it still contains the old buggy code.

## Call Chain

```
_app.py lifespan()
  → manager.py get_instance()
    → llamacpp/manager.py __init__()
      → _resolve_cuda_version()
        → system_info.py get_cuda_version()
          → system_info.py _run_command("cmd /c start /b nvidia-smi")  ← hangs indefinitely
```

## Additional Details

[...]

PR #6204, opened 2026-07-16
fix(utils): drop redundant nvidia-smi probe in get_vram_size_gb

Yigtwxx:
## Description

`get_vram_size_gb()` in `src/qwenpaw/utils/system_info.py` opens with a guard that re-runs the CUDA probe:

```python
def get_vram_size_gb() -> float:
    if get_cuda_version() is None:      # spawns nvidia-smi
        return 0.0
    output = _run_command(["nvidia-smi", "--query-gpu=memory.total", ...])
```

`get_cuda_version()` spawns `nvidia-smi` itself, so on a CUDA host a single `get_system_info()` call spawns `nvidia-smi` **three times** where two suffice:

```
get_system_info()
├─ get_cuda_version()      -> nvidia-smi                      # spawn 1
└─ get_vram_size_gb()      (already gated on `if cuda_version`)
   ├─ get_cuda_version()   -> nvidia-smi                      # spawn 2, redundant
   └─ _run_command(...)    -> nvidia-smi --query-gpu=memory.total   # spawn 3
```

`get_system_info()` already computes `cuda_version` and already gates the VRAM call on it (`"vram_gb": get_vram_size_gb() if cuda_version else 0.0`), so the guard inside `get_vram_size_gb()` is the same check performed a second time.

**The guard is redundant even when `get_vram_size_gb()` is called on its own.** The `--query-gpu=memory.total` query is self-sufficient: on a host with no working `nvidia-smi` it fails, `_run_command()` returns `None`, and the function returns `0.0` — exactly what the guard returned. The behaviour is identical in every case, including the "nvcc present but nvidia-smi absent" case, where the guard passes but the query then fails and still yields `0.0`. Removing it changes cost, not semantics.

This matters most on two paths:

- **Startup.** `LlamaCppBackend.__init__` resolves the CUDA version during `LocalModelManager.get_instance()`, which the FastAPI lifespan runs synchronously. When `nvidia-smi` is slow or wedged, every redundant spawn costs another full `timeout=5` window. This is a contributing factor to (not a full fix for) #6197, where a hung `nvidia-smi` stalls Desktop startup.
- **Per request.** `GET /models` reaches `get_vram_size_gb()` via `ModelManager._detect_available_memory_gb()`. It is an `async def` endpoint calling blocking `subprocess.run`, so each redundant spawn blocks the event loop for its whole duration. This halves that: two spawns per request become one.

**Related Issue:** Relates to #6197. That issue's root cause is a stale PyInstaller bundle shipping pre-`timeout` code, which this PR does not address — it only removes a redundant probe that multiplies the stall when `nvidia-smi` misbehaves.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Tests

## Checklist

- [ ] Documentation updated (if needed) — internal helper, no docs change
- [x] Ready for review

## Testing

`src/qwenpaw/utils/system_info.py` had no direct test coverage at all — even the `errors="replace"` fix from #6140 landed without a regression test. This PR adds `tests/unit/utils/test_system_info.py` as the module's first tests, following the conventions in `tests/unit/utils/test_stdio.py` (function-style, `monkeypatch.setattr` on the module object).

Coverage added:

- `test_get_system_info_does_not_probe_nvidia_smi_twice` — counts every probe and asserts `nvidia-smi` is spawned exactly twice, which is the regression guard for this change.
- `test_get_system_info_skips_vram_probe_without_cuda` — the caller-side gate still short-circuits the VRAM query when there is no CUDA.
- `test_get_vram_size_gb_returns_zero_without_nvidia_smi` — pins the behaviour the removed guard used to provide.
- `test_get_vram_size_gb_picks_largest_gpu` and `test_get_vram_size_gb_ignores_unparsable_lines` — multi-GPU selection and `[N/A]` rows.

## Evidence

All measurements below are from a real Windows 11 machine with an NVIDIA RTX 4070 and a healthy driver — the same OS/GPU combination reported in #6197.

**The VRAM query does not need the CUDA probe in front of it.** Run on its own, it succeeds immediately:

```
$ nvidia-smi --query-gpu=memory.total --format=csv,noheader,nounits
8188

$ python -c "...subprocess.run of the same query..."
rc= 0 out= '8188' elapsed=0.03s
```

**Dropping the redundant spawn is ~40% faster even when the driver is healthy.** Averaged over 5 runs with the driver already warmed up (so the one-off cold-start cost is excluded from both sides):

```
WARM driver, avg of 5 runs:
  current (3 spawns): 0.333s
  proposed(2 spawns): 0.200s
  saved: 0.133s (40%)
```

On a wedged driver the saving is bounded by `_run_command`'s `timeout=5` per spawn instead, which is the case #6197 describes.

**The new test fails against the current code and passes with the fix.** Reverting only the `get_vram_size_gb` change:

```
$ pytest tests/unit/utils/test_system_info.py -q
>       assert get_vram_size_gb() == 8.0
E       assert 0.0 == 8.0
FAILED tests/unit/utils/test_system_info.py::test_get_system_info_does_not_probe_nvidia_smi_twice
FAILED tests/unit/utils/test_system_info.py::test_get_vram_size_gb_picks_largest_gpu
FAILED tests/unit/utils/test_system_info.py::test_get_vram_size_gb_ignores_unparsable_lines
3 failed, 2 passed in 0.27s
```

**With the fix applied:**

```
$ pytest tests/unit/utils/test_system_info.py -q
5 passed in 0.19s
```

**No regressions across the whole unit suite:**

```
$ pytest tests/unit -q
4804 passed, 52 skipped, 56 warnings in 247.88s (0:04:07)
```

**Every pre-commit hook passes on the touched files:**

```
$ pre-commit run --files src/qwenpaw/utils/system_info.py \
                        tests/unit/utils/test_system_info.py
```

### DIFF
--- a/src/qwenpaw/utils/system_info.py
+++ b/src/qwenpaw/utils/system_info.py
@@ -103,10 +103,11 @@ def get_memory_size_gb() -> float:
 
 
 def get_vram_size_gb() -> float:
-    """Return the largest CUDA GPU memory size in GB, or 0 when absent."""
-    if get_cuda_version() is None:
-        return 0.0
+    """Return the largest CUDA GPU memory size in GB, or 0 when absent.
 
+    The query below already fails on hosts without a working nvidia-smi,
+    so it needs no separate CUDA probe to guard it.
+    """
     output = _run_command(
         [
             "nvidia-smi",
--- a/tests/unit/utils/test_system_info.py
+++ b/tests/unit/utils/test_system_info.py
@@ -1,0 +1,126 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=protected-access
+from __future__ import annotations
+
+import pytest
+
+import qwenpaw.utils.system_info as system_info_module
+from qwenpaw.utils.system_info import get_system_info, get_vram_size_gb
+
+_NVIDIA_SMI_HEADER = (
+    "NVIDIA-SMI 560.94    Driver Version: 560.94    CUDA Version: 12.6\n"
+)
+
+
+def _fake_run_command(
+    responses: dict[str, str | None],
+    calls: list[list[str]],
+):
+    """Build a _run_command stand-in that records every probe."""
+
+    def _runner(args: list[str], **_kwargs: object) -> str | None:
+        calls.append(list(args))
+        for key, value in responses.items():
+            if key in args:
+                return value
+        return None
+
+    return _runner
+
+
+# ---------------------------------------------------------------------------
+# get_system_info
+# ---------------------------------------------------------------------------
+
+
+def test_get_system_info_does_not_probe_nvidia_smi_twice(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    calls: list[list[str]] = []
+    monkeypatch.setattr(
+        system_info_module,
+        "_run_command",
+        _fake_run_command(
+            {
+                "--query-gpu=memory.total": "8188",
+                "nvidia-smi": _NVIDIA_SMI_HEADER,
+            },
+            calls,
+        ),
+    )
+
+    info = get_system_info()
+
+    nvidia_calls = [call for call in calls if call[0] == "nvidia-smi"]
+    # One probe for the CUDA version, one for VRAM. A third means
+    # get_vram_size_gb re-ran the CUDA probe the caller already did.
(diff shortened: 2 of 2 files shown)


lalaliat (approved):
LGTM

PR #6225, opened 2026-07-17
fix(desktop): gracefully shut down backend sidecar before exit

### LINKED ISSUE #6219: [feature]: Desktop force-kills the backend instead of performing graceful shutdown
## QwenPaw Version

Current `main` at `d0e98e538a2c4ab82f8a8a9ec4a0576cda0d77cd` (reviewed on 2026-07-17).

## Description

Hello QwenPaw maintainers, and thank you for your work on the new Tauri desktop application.

While reviewing the Windows desktop shutdown path, I noticed that a normal application quit currently force-terminates the Python backend sidecar instead of first allowing it to shut down gracefully.

`BackendState::stop()` removes the sidecar handle and immediately calls `CommandChild::kill()`:

https://github.com/agentscope-ai/QwenPaw/blob/d0e98e538a2c4ab82f8a8a9ec4a0576cda0d77cd/console/src-tauri/src/backend.rs#L87-L96

The Tauri shell implementation delegates this call to `SharedChild::kill()` / `std::process::Child::kill()`. Rust documents `Child::kill()` as forcing the child process to exit; it is not a cooperative shutdown mechanism:

https://docs.rs/tauri-plugin-shell/2.3.5/src/tauri_plugin_shell/process/mod.rs.html#77-81

https://doc.rust-lang.org/std/process/struct.Child.html#method.kill

This is particularly important because the Python backend already has substantial cleanup logic in its FastAPI lifespan teardown:

https://github.com/agentscope-ai/QwenPaw/blob/d0e98e538a2c4ab82f8a8a9ec4a0576cda0d77cd/src/qwenpaw/app/_app.py#L680-L798

That teardown includes, among other things:

- plugin shutdown hooks;
- local model server shutdown;
- `AppServiceManager` and `MultiAgentManager` shutdown;
- token usage manager, browser, and HTTP client cleanup;
- Windows sandbox cleanup, including accounts, profiles, ACLs, and firewall rules.

Because a normal Desktop quit uses a forceful process termination, these lifespan handlers and Python `atexit` handlers may not run. The backend process itself can appear to have exited successfully while its managed resources, child processes, or sandbox artifacts have not been cleaned up correctly.

The same `backend::stop()` path is also used for backend restart and desktop update installation, so this is not limited to the tray Quit action.

This is related to, but distinct from, #4587. That issue focused on a backend process remaining alive after Desktop exit. This issue concerns the cleanup contract even when the backend PID disappears as expected.

**Security considerations:** If graceful shutdown is implemented through a loopback HTTP endpoint, it should not be exposed as an unauthenticated general API. A per-process secret, an inherited IPC channel, or another desktop-only authenticated mechanism would avoid allowing arbitrary local web content to stop QwenPaw.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

Additionally affected: Tauri Desktop lifecycle management.

## Environment

- **QwenPaw version:** `main` at `d0e98e538a2c4ab82f8a8a9ec4a0576cda0d77cd`
- **OS:** Windows 10/11 (the cleanup impact is especially significant on Windows; the force-kill path is shared by other desktop platforms)
- **Install method:** Official Tauri Desktop package / source build
- **Python version (if applicable):** Bundled desktop Python runtime

## Steps to Reproduce

1. Start QwenPaw Desktop on Windows and wait for the Python backend to become ready.
2. Optionally start a resource whose cleanup depends on the FastAPI lifespan, such as a local model, browser session, plugin with a shutdown hook, or Windows sandbox.
3. Choose **Quit** from the tray menu, or choose the quit option in the close dialog. (Minimizing to the tray is not considered quitting and is expected to keep the backend running.)
4. Observe the native lifecycle path: `exit_app()` / `RunEvent::ExitRequested` calls `backend::stop()`, which immediately calls `child.kill()`.
5. Check `desktop.log` / backend logs. The normal lifespan completion sequence, ending with `Application shutdown complete`, is not guaranteed to run before the process exits.

The behavior can also be reproduced through **Restart backend** and update installation because they call the same stop implementation.

## Actual vs Expected

- **Actual:** A normal Desktop quit immediately force-kills the backend sidecar. It does not request cooperative shutdown, wait for Uvicorn/FastAPI lifespan teardown, or use force-kill only as a timeout fallback.
- **Expected:** A normal Desktop quit should first request graceful backend shutdown and wait for the backend to exit after completing its lifespan cleanup. Forceful termination should only be a bounded fallback if graceful shutdown fails or exceeds a reasonable timeout.

## Suggested Approach

One possible lifecycle could be:

1. Mark the backend generation as stopping so it is not reported as an unexpected crash and new work is no longer accepted.
2. Send a cooperative shutdown request through a desktop-only authenticated channel (for example, an inherited IPC/stdin command, or a loopback endpoint protected by a per-process secret).
3. Let Uvicorn stop accepting new requests and execute the FastAPI lifespan teardown.
4. Wait for the existing `CommandEvent::Terminated` event for a bounded period, perhaps 10–20 seconds. The timeout may need to accommodate Windows sandbox and local-model cleanup.
5. Only after timeout or IPC failure, force-terminate the backend and, on Windows, ensure the complete managed process tree is terminated (for example with a Job Object) so plugin/model/browser child processes are not orphaned.
6. Use this shared graceful-stop primitive for tray Quit, `ExitRequested`, backend restart, and both update installation flows.

It would also be helpful to add lifecycle tests that verify:

- a normal quit reaches a test lifespan shutdown hook before the backend exits;
- restart waits for cleanup before spawning the next generation;
- a deliberately hung backend is force-killed after the timeout;
- Windows child-process and sandbox cleanup does not leave managed resources behind;
- repeated or concurrent stop calls remain idempotent.

## Logs / Screenshots

No crash log is required to identify the behavior; the current native stop implementation calls `child.kill()` directly. A useful acceptance signal after the fix would be that a normal Desktop quit consistently logs:

```text
Application shutdown complete
```

before the sidecar termination event.

## Additional Notes

I understand that a hard-kill fallback is still necessary for hung backends and abnormal application termination. The request here is to reserve it for that fallback path rather than using it as the default behavior for a normal user-initiated quit.

Thank you for considering this improvement. I would be happy to provide any additional details that would help with the desktop lifecycle design.

jinglinpeng:
## Description

On a normal Desktop quit (and on backend restart / update install), the Tauri shell force-terminated the Python backend sidecar. `BackendState::stop()` took the child handle and immediately called `CommandChild::kill()`, which maps to `TerminateProcess` (Windows) / `SIGKILL` (Unix). This bypasses uvicorn's signal handling and the FastAPI **lifespan teardown** entirely.

Because the lifespan `finally` block never ran, the ReMe memory index was never flushed to disk (`ReMeLightMemoryManager.close()` → `reme.close()` → `LocalFileStore._close()` → `dump()`), along with the rest of the documented cleanup (plugin shutdown hooks, local model server, `AppServiceManager` / `MultiAgentManager`, token-usage/browser/HTTP clients, Windows sandbox artifacts). The observable symptom is **lost memory** after quitting the desktop app.

This PR implements the minimal version of the lifecycle suggested in the issue:

- Adds a **desktop-only** `POST /api/desktop/shutdown` endpoint that flips uvicorn's `should_exit`, letting the sidecar run its normal lifespan shutdown. The endpoint returns `404` unless `DESKTOP_APP_ENV == "1"`, and is only reachable over loopback.
- `entry.py` now passes the FastAPI app instance to `uvicorn.Config` with `timeout_graceful_shutdown=5` and stores the `Server` on `app.state.uvicorn_server` so the endpoint can trigger a graceful stop.
- `backend::stop()` now **requests graceful shutdown first** and only falls back to `child.kill()` when the request fails. On success it drops the child handle (the plugin's `CommandChild` has no kill-on-drop), letting Python finish its own lifespan cleanup. This shared primitive is reused by tray Quit, `RunEvent::ExitRequested`, backend restart, and both update-install flows (made `async`, with `block_on` at the sync call sites).

**Related Issue:** Fixes #6219

**Security Considerations:** The shutdown endpoint is gated behind `DESKTOP_APP_ENV` (only present when running as the bundled desktop sidecar) and is bound to loopback, POST-only, with an empty body. It performs no action other than setting uvicorn's `should_exit`. It is intentionally *not* a general authenticated API; hardening it further with a per-process secret (as the issue suggests) is a reasonable follow-up if desired.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

Additionally affected: Tauri Desktop lifecycle (`console/src-tauri`).

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes — *pre-commit/ruff were not available in my environment; `python -m py_compile` passes on the changed Python files and `cargo check` passes on the Tauri crate.*
- [x] I ran tests locally — see the A/B experiment under **Evidence**
- [x] Ready for review

## Testing

Because I cannot run the packaged desktop `.exe` end-to-end here, I built a controlled **A/B experiment** that reproduces the production shutdown chain faithfully with a real uvicorn subprocess and the **real ReMe `LocalFileStore`**:

- Both groups do identical in-session work: start the sidecar, add one memory (held only in ReMe's in-memory `file_chunks`).
- The **only** difference is how the process ends:
  - **Group A** — hard `kill()` (reproduces the old `child.kill()` path).
  - **Group B** — `POST /api/desktop/shutdown` (this fix).
- A fresh ReMe store then reloads from disk and searches for the memory.

The endpoint path, `should_exit` flag, `app.state.uvicorn_server` wiring, `timeout_graceful_shutdown`, and auth allow-list in the harness match the real code paths verbatim.

## Evidence

A/B experiment output (identical work in both groups; only the exit path differs):

```text
GROUP A_hard_kill   (reproduces old child.kill())
[A_hard_kill] added memory (held in-memory, not yet dumped)
[A] HARD KILL (proc.kill) -> lifespan finally will NOT run
--- server.log ---
starting uvicorn on 127.0.0.1:56647
lifespan startup complete (store started)
added memory id=note1 (in-memory only, not yet dumped)          <-- log stops here; finally never ran
--- reload+search result ---
{ "chunk_file_exists": false, "chunk_file_size": 0, "loaded_chunks": 0, "found": false }

GROUP B_graceful   (this fix)
[B_graceful] added memory (held in-memory, not yet dumped)
[B] POST /api/desktop/shutdown -> graceful uvicorn shutdown
[B] process exit code: 0
--- server.log ---
starting uvicorn on 127.0.0.1:56694
lifespan startup complete (store started)
added memory id=note1 (in-memory only, not yet dumped)
/api/desktop/shutdown: set should_exit=True
lifespan finally: closing store (dump -> chunks file)
lifespan finally: store closed, chunks dumped
uvicorn server.run() returned (process exiting normally)
--- reload+search result ---
{ "chunk_file_exists": true, "chunk_file_size": 154, "loaded_chunks": 1,
  "search_hits": ["The secret project codename is Bluebird with deadline March fifteenth"], "found": true }

########## SUMMARY ##########
A (hard kill, OLD behavior): memory found after reload = False  (chunk_file_size=0)
B (graceful, THE FIX):       memory found after reload = True   (chunk_file_size=154)
Problem reproduced (A loses memory): True
Fix verified (B keeps memory):       True
CONCLUSION: PASS - problem real AND fix works
```

Rust crate check:

```text
cargo check  (console/src-tauri)  ->  Finished, exit 0, no warnings
```

## Additional Notes

- A hard-kill fallback is retained for hung backends; this PR only reserves it for the fallback path rather than using it as the default for a normal quit, per the issue's request.
- Follow-ups that could be layered on top if maintainers prefer: a per-process shutdown secret; a Windows Job Object to guarantee child-process cleanup on the fallback kill; and the lifecycle unit tests enumerated in #6219.


jinliyl:
Thanks for working on this. The authenticated loopback shutdown endpoint is a good direction, but I found three blocking lifecycle issues in the current implementation.

1. **Normal Quit still does not wait for the backend to finish shutting down**

   On a successful shutdown POST, `request_stop()` drops the `CommandChild` and returns immediately. Both the tray quit path and `RunEvent::ExitRequested` call `stop()`, which only waits for the HTTP response, and then allow the Tauri process to exit.

   This proves that the shutdown request was accepted, but it does not prove that the FastAPI lifespan teardown, ReMe dump, or the final `Application shutdown complete` step finished. If the user launches QwenPaw again quickly, `reconcile_singleton_backend()` can find the still-running old sidecar and terminate it while it is flushing data, recreating the memory-loss race this PR is intended to fix.

   Normal Quit and `ExitRequested` should use a bounded `stop_and_wait` flow and only exit the Tauri process after receiving the sidecar termination event (or after applying the force-kill fallback).

2. **The Windows update paths can panic because they call `block_on` from inside the Tauri/Tokio runtime**

   `run_install()` and `run_cached_install()` are executed through `tauri::async_runtime::spawn`. On Windows:

   - the updater's synchronous `on_before_exit` callback calls `tauri::async_runtime::block_on(backend::stop(...))`;
   - `install_cached_windows()` calls `tauri::async_runtime::block_on(backend::stop_and_wait(...))`.

   Tauri 2.11.1 delegates this directly to Tokio's `Runtime::block_on`. Calling it while already executing on that runtime can panic with the nested-runtime error, preventing the installer from being launched.

   There is an additional ordering problem in the normal Windows updater path: `tauri-plugin-updater` invokes `on_before_exit`, launches the installer, and then calls `std::process::exit(0)` inside `Update::install`. Therefore, the `stop_and_wait()` placed after `update.install(bytes)` is unreachable on Windows.

   The Windows flows should await `stop_and_wait()` before entering the synchronous install/exit operation, rather than trying to enter the async runtime from the updater callback.

3. **There is no force-kill fallback after the 60-second process-exit timeout**

   `request_stop()` consumes the `CommandChild`. Once the shutdown endpoint returns success, the child handle is dropped. `stop_and_wait()` then retains only the watch receiver. If `wait_for_termination()` times out after 60 seconds, it can only return an error: it no longer has a handle with which to kill the process, and the shared state remains `stopping = true` with no child handle. Later stop attempts cannot recover this process either.

   `timeout_graceful_shutdown=5` does not bound FastAPI lifespan teardown. In Uvicorn it bounds connection/request task draining, after which `lifespan.shutdown()` is awaited separately. A plugin shutdown hook, model shutdown, or Windows sandbox cleanup can therefore still exceed the outer timeout.

   Please retain ownership of the child until the termination event is observed, kill it when the overall timeout expires, and reset the lifecycle state so repeated/concurrent stop calls remain recoverable.

One CI note: the green **Real behavior proof** job did not execute the claimed behavior test for this PR; its log says `SKIPPED (maintainer/bot/override)`. The PR also does not add lifecycle regression tests covering normal quit completion, timeout fallback, concurrent stop calls, or the Windows update paths.

Given that these issues affect the core data-preservation goal and can break Windows updates, I do not think the PR is ready to merge yet.

jinliyl:
Thanks for the follow-up. I re-reviewed the latest head (`3cacba9`). The previous three blocking issues are substantially addressed: normal quit now waits for process termination, the Windows update paths no longer enter the Tokio runtime with a nested `block_on`, and the 60-second graceful timeout now has a force-kill fallback.

I found two remaining lifecycle issues that should be addressed before merge:

1. **Concurrent stop/restart callers can clear or orphan a newly started backend**

   Every `stop_and_wait()` caller, including callers that receive `StopPlan::Wait`, executes the unconditional `finish_stop()` after the shared termination receiver becomes true. Consider two concurrent `restart_backend()` calls:

   - both wait on the old process;
   - the first caller wakes, runs `finish_stop()`, returns from `stop_and_wait()`, and calls `start()`;
   - the second caller then wakes and runs `finish_stop()` against the shared state, which may now contain the newly started child, port, token, and termination receiver.

   The second cleanup can therefore drop tracking for the new child, after which both callers may start another sidecar. A concurrent restart and quit/update can produce similar state corruption.

   Cleanup needs to be conditional on the generation/PID being stopped, or the complete stop/start lifecycle needs to be serialized. Please add a regression test with concurrent stop/restart callers.

2. **An installer launch/install failure leaves the running app without its backend**

   The updated flows correctly await `backend::stop_and_wait()` before entering the synchronous installer operation. However, if `update.install(bytes)` returns an error, or the cached Windows installer fails to spawn, the function only emits an updater error. At that point the sidecar has already exited, so the still-running Tauri/WebView process is left without a backend and cannot recover normally.

   This affects the normal install path, cached Windows install, and cached macOS install. The error paths should restart/recover the backend before reporting the install failure.

There is also a remaining fallback-state concern: `force_kill()` consumes the child handle even when `child.kill()` fails, and after the additional five-second wait times out, `finish_stop()` clears the lifecycle state. If the process is still alive, later stop attempts will report `NoProcess` and cannot recover it.

Finally, the green **Real behavior proof** check still did not execute a behavior test; its log says `SKIPPED (maintainer/bot/override)`. This PR still adds no lifecycle regression tests for concurrent callers, timeout/kill failure, or installer failure recovery. The PR description is also stale relative to the latest token-authenticated, wait-for-exit implementation.

Given the concurrent lifecycle race and the unrecoverable installer-error path, I do not think the latest revision is ready to merge yet.

jinliyl (approved):
LGTM

PR #6236, opened 2026-07-17
fix(cli): cron update preserves untouched runtime and request fields

### LINKED ISSUE #6176: [Bug]: cron CLI update resets untouched runtime and metadata fields

manjieqi:
## Description

`qwenpaw cron update` documented that only explicitly provided options are changed, but it actually rebuilt the entire job spec via `_build_spec_from_cli` (designed for **create**). That function hardcodes `max_concurrency=1` and `misfire_grace_seconds=600`, and only emits fields that have a corresponding CLI option. Any other field — advanced runtime settings, request extensions like `model` / `request_context`, etc. — was silently reset or dropped on update.

**Related Issue:** Fixes #6176

**What this PR does:** rewrote `_resolve_update_spec` to **deep-copy** the existing spec and only **patch** fields explicitly provided by the CLI:

- `runtime.max_concurrency`, `runtime.misfire_grace_seconds` — preserved
- `request.model`, `request.request_context` and other request extensions — preserved
- `meta`, `dispatch.meta` — preserved
- All CLI-provided overrides still apply correctly

Two extra bugs fixed along the way (both found while testing against the real API model):

1. `--schedule-type scheduled` previously wrote `"scheduled"` into `schedule.type`, which the server rejects with 422 (`CronJobSpec` only accepts `cron`/`once`). It now maps to the API value `"once"`.
2. `--text` on an agent job whose `request.input` is empty/malformed previously crashed with `IndexError`; it now rebuilds the message structure.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] CLI
- [x] Tests

## Checklist

- [x] Documentation updated (if needed) — not needed, behavior now matches existing docs
- [x] Ready for review

## Testing

```bash
pytest tests/unit/cli/test_cron_cmd.py -v
```

9 new regression tests in `tests/unit/cli/test_cron_cmd.py`:

- rename-only preserves `max_concurrency` / `misfire_grace_seconds` / `timeout_seconds`
- rename-only preserves `request.model` / `request.request_context`
- rename-only preserves `meta` / `dispatch.meta`
- CLI overrides (`--name` / `--no-enabled` / `--timeout` / `--channel` / `--target-user`) apply while untouched fields survive
- input spec dict is not mutated in place
- `--schedule-type scheduled` maps to API value `once`
- once-schedule with repeat fields survives a rename
- `--text` on agent job replaces prompt but keeps request extensions
- `--text` on agent job with empty/malformed `request.input` rebuilds it

## Evidence

Unit + surrounding suites (309 passed):

```text
$ pytest tests/unit/cli/ tests/unit/app/crons/ -q
........................................................................ [ 29%]
........................................................................ [ 59%]
........................................................................ [ 93%]
.....................                                                    [100%]
309 passed in 32.15s
```

pre-commit:

```text
$ pre-commit run --files src/qwenpaw/cli/cron_cmd.py tests/unit/cli/test_cron_cmd.py
check python ast.................................Passed
mypy.............................................Passed
black............................................Passed
flake8...........................................Passed
pylint...........................................Passed
... (all 17 hooks passed/skipped, none failed)
```

End-to-end against the real FastAPI cron router (create job with advanced fields via JSON → update via real CLI command → read back):

```text
1) created job 6710753f-...   max_concurrency=4
2) CLI update --name renamed -> exit 0
   PASS: name == renamed
   PASS: runtime.max_concurrency == 4
   PASS: runtime.misfire_grace_seconds == 1800
   PASS: runtime.timeout_seconds == 300
   PASS: request.model == custom-model
   PASS: request.request_context preserved
   PASS: dispatch.meta preserved
   PASS: meta preserved
   PASS: schedule.timezone preserved
4) CLI override --timeout 900 --no-enabled -> PASS (max_concurrency still 4)
5) CLI --text updates prompt, model preserved -> PASS
6) once job rename keeps schedule/repeat/runtime -> PASS
7) --schedule-type scheduled maps to once, server accepts -> PASS
```

The same end-to-end script against the **unfixed** code fails immediately with `KeyError: 'model'` (request extensions dropped), confirming the regression and the fix.

## Additional Notes

This is a re-open of #6200 (closed while fixing CI): rebased onto latest `main`, fixed the black/pylint pre-commit failures, and added the `scheduled`→`once` mapping fix plus malformed-input handling.

### DIFF
--- a/src/qwenpaw/cli/cron_cmd.py
+++ b/src/qwenpaw/cli/cron_cmd.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
 
+import copy
 import json
 from pathlib import Path
 from typing import Any, Dict, Optional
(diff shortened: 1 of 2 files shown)


lalaliat (approved):
Thank you for your contribution! LGTM

PR #6237, opened 2026-07-17
feat(scroll): improve exchange and date-aware history recall

niceIrene:
## Summary

Improve Scroll history recall so search results return complete conversational turns, support date-aware queries consistently across both recall tools, and tolerate models serializing `expand` sequence bounds as numeric strings.

## Motivation

History search previously returned only the individual message matching a query. This could separate an assistant answer from the user question that gave it meaning, leaving recalled evidence incomplete.

Date-related questions were also difficult to answer because recall could not filter by a message's `created_at` value or calculate date differences consistently.

Imported histories exposed another issue: turn boundaries assumed native `context_msg` rows, so source-specific row kinds such as benchmark imports could fail to pair user and assistant messages.

Finally, some models serialize integer tool arguments as strings, for example:

```json
{
  "op": "expand",
  "lo": "3997",
  "hi": "4020"
}
```

Although these values are valid sequence numbers, the structured tool previously rejected them during schema validation.

### 1. Return complete turns from history search

- Make `MemorySpace.search()` return the complete user-bounded turn containing each match.
- Add:
  - `match_seq`
  - `matched_seqs`
  - `turn_start_seq`
  - `turn_end_seq`
  - `turn`
- Deduplicate multiple matches belonging to the same turn.
- Preserve matching-row-only behavior with `include_turn=False`.
- Keep turn boundaries scoped to the same session and agent.
- Detect turn boundaries using `role="user"` instead of requiring `kind="context_msg"`.
- Exclude tagged synthetic continuation messages from user-turn boundaries.
- Support imported histories with source-specific row kinds.
- Render complete turns through the structured `recall_history` tool.
- Document the same behavior in `recall_history_python`.

A search result now has the following shape:

```json
{
  "match_seq": 101,
  "matched_seqs": [101],
  "turn_start_seq": 100,
  "turn_end_seq": 103,
  "turn": [
    {"seq": 100, "role": "user", "content": "..."},
    {"seq": 101, "role": "assistant", "content": "..."},
    {"seq": 102, "kind": "tool_result", "content": "..."},
    {"seq": 103, "role": "assistant", "content": "..."}
  ]
}
```

### 2. Add date-aware recall

- Add strict ISO date and timestamp parsing.
- Support exact-date and inclusive date-range filters:
  - `created_on`
  - `created_from`
  - `created_to`
- Allow date-only searches with an empty keyword query.
- Apply date filters consistently to:
  - FTS search
  - LIKE fallback
  - Saved tool-output search
- Include `created_at` in recalled and rendered rows.
- Add an index on `conversation_history.created_at`.

### 3. Add date-difference calculation

Expose consistent date arithmetic through both recall interfaces.

Python recall:

```python
ms.days_between("2024-11-01", "2024-12-16")
# 45
```

Structured recall:

```json
{
  "op": "days_between",
  "start": "2024-11-01",
  "end": "2024-12-16"
}
```

Both interfaces consistently handle:

- ISO dates
- ISO timestamps
- `Z` and `+/-HH:MM` time zones
- Leap years
- Invalid dates
- Positive and negative differences
- Optional inclusive endpoint counting

### 4. Accept numeric strings for `expand` bounds

Allow the structured `recall_history` tool to accept `lo` and `hi` as either JSON integers or ASCII-decimal integer strings:

```json
{"op": "expand", "lo": 100, "hi": 103}
```

```json
{"op": "expand", "lo": "100", "hi": "103"}
```

Numeric strings are normalized before:

- Request fingerprint generation
- Pagination cursor validation
- Duplicate-call detection
- Recall loop guarding
- Execution by `MemorySpace`

As a result, `"100"` and `100` identify the same logical request.

Validation remains strict and rejects:

- Booleans
- Floats
- Empty strings
- Leading or trailing whitespace
- Signed strings such as `"+1"` or `"-1"`
- Decimal strings such as `"1.0"`
- Non-ASCII digits
- Zero and negative sequence numbers

The Python execution layer continues to receive only normalized `int | None` values.

### 5. Address review scalability and boundary issues

- Build Scroll components in a worker thread so opening and migrating `history.db` does not block the request event loop.
- Resolve and deduplicate search-hit turn boundaries before loading complete turns.
- Bound complete-turn expansion to 1000 rows and 4 MiB per search, with a 5-second SQLite execution timeout and an explicit partial-result notice.
- Handle `9999-12-31` date filters without overflowing date arithmetic.
- Verify populated-store upgrades, event-loop scheduling, duplicate hits in one long turn, row/byte/time budgets, and maximum-date filters.

## Testing

Add and update coverage for:

- Complete user-bounded turn retrieval
- Multiple matches deduplicated into one turn
- User, assistant, and tool-result matches
- Session and agent boundary isolation
- Imported source-specific row kinds
- Synthetic continuation handling
- Exact-date and inclusive date-range search
- Date-only search
- FTS and LIKE fallback consistency
- Saved tool-output date filtering
- `created_at` rendering
- Signed and inclusive date differences
- Time zones, timestamps, leap years, and invalid dates
- Integer and numeric-string `lo/hi` schemas
- Strict rejection of unsafe sequence coercions
- Pagination cursor compatibility across `"1"` and `1`
- Duplicate recall detection across equivalent numeric representations


rayrayraykk (approved):
LGTM

PR #6151, opened 2026-07-15
refactor(tool_calls): background tool call offload mechanism with frontend controls

XiuShenAl:
## Description

Refactor the background tool call mechanism to use a dual-deadline architecture (`offload_deadline` + `kill_deadline`) instead of the single `deadline` field. This fixes three bugs from #6056:

- Offload incorrectly triggering cancel signals
- Hint injection timing issues in the ReAct loop
- Missing user control over the offload decision

Adds a frontend control banner on tool cards that allows users to manage long-running tool calls: move to background, keep in foreground, extend offload countdown, extend execution timeout, or cancel.

**Related Issue:** Fixes #6056, Fixes #6245

**Security Considerations:** No new auth surfaces. All tool call control endpoints remain scoped to `session_id` + `tool_call_id` validation via existing `_get_coordinator` pattern.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

**Backend:**
1. Start a tool call with timeout (e.g. `sleep 120` with `timeout=60`)
2. Verify `GET /api/tool-calls/{sid}/{tcid}` returns `offload_remaining` and `kill_remaining`
[...]

PR #6248, opened 2026-07-18
fix: distinguish offload vs cancel to prevent subprocess kill on deadline

### LINKED ISSUE #6056: [Bug]: Background offload kills subprocess immediately — LLM-provided timeout is silently ignored
# Issue: Background offload kills subprocess immediately — LLM-provided timeout is silently ignored

## Summary

When the ToolCoordinator offloads a long-running shell command to
the background, it immediately kills the subprocess instead of
letting it continue. The LLM-provided `timeout` parameter
(e.g. `timeout=180`) is silently ignored because `cancellable_wait`
discards `fallback_secs` when a `ToolCallContext` is present.

**Result**: A `sleep 120` command with `timeout=180` is killed at
~60 seconds (the coordinator hook default), not 180 seconds. The
error message falsely reports "exceeded the timeout of 180 seconds".

## Reproduction

1. Start QwenPaw with default config
2. Send: "Run `sleep 120 && echo done`"
3. LLM calls `execute_shell_command(command="sleep 120 && echo done",
   timeout=180)`
4. At ~60s the tool is "offloaded to background"
5. Almost immediately the hint arrives:
   `"Command failed with exit code -1 ... exceeded the timeout of
   180 seconds"`

**Expected**: The command should continue running in the background
for up to 180 seconds and complete successfully at ~120s.

**Actual**: The subprocess is killed at ~60s. The error message
claims 180s timeout but only ~60s elapsed.

## Root cause

Three interacting bugs in the offload pipeline:

### Bug 1: `cancel_event` is overloaded for both "cancel" and "offload"

`_handle_deadline_reached` sets `cancel_event` to signal the
offload:

```python
# _coordinator.py:126-129
@staticmethod
def _handle_deadline_reached(ctx):
    if ctx.offload_reason is None:
        ctx.offload_reason = OffloadReason.TIMEOUT
        ctx.cancel_event.set()        # ← signals offload
```

But `cancellable_wait` interprets `cancel_event` as **kill the
task**:

```python
# _timeout_helper.py:50-58
if cancel_waiter in done:
    task.cancel()                      # ← kills subprocess
    ...
    raise asyncio.CancelledError(...)
```

So the moment the coordinator decides to offload, the subprocess
is killed. "Offloaded to background" is a lie — nothing runs in
the background.

### Bug 2: `fallback_secs` is silently discarded when ctx exists

```python
# _timeout_helper.py:30-35
ctx = get_call_context()

if ctx is None:
    # fallback_secs used ONLY here (SDK direct call / unit test)
    return await asyncio.wait_for(coro, timeout=fallback_secs)

# When ctx exists: fallback_secs is NEVER used
task = asyncio.ensure_future(coro_or_task)
cancel_waiter = asyncio.create_task(ctx.cancel_event.wait())
# ... waits only on cancel_event, no timeout
```

The LLM's `timeout=180` is passed as `fallback_secs=180` to
`cancellable_wait`, but since the coordinator always injects a
`ctx`, this value is completely ignored. The only thing that
matters is `ctx.cancel_event`, which fires at the coordinator
deadline (60s).

### Bug 3: Error message reports wrong timeout value

The shell tool catches `CancelledError` and uses the LLM-provided
`timeout` in the message:

```python
# shell.py:628-632
except (asyncio.TimeoutError, asyncio.CancelledError):
    stderr_suffix = (
        f"⚠️ TimeoutError: The command execution exceeded "
        f"the timeout of {timeout} seconds. "
    )
```

- Reports: "exceeded the timeout of 180 seconds"
- Reality: killed at ~60 seconds by coordinator deadline

This misleads both the LLM and the user into thinking the command
actually ran for 180 seconds.

## Impact

- **Any shell command longer than 60s will fail**, regardless of
  the LLM-provided timeout value
- LLMs that carefully set `timeout=180` or `timeout=300` for
  long-running commands get no benefit — their timeout is ignored
- The "offloaded to background" feature is non-functional for shell
  commands — the subprocess is killed on offload
- Error messages are misleading (wrong timeout value)
- This affects `pip install` of large packages, builds, long
  computations, etc.

## Evidence from production dumps

| Dump | Command | LLM timeout | Actual kill time | Reported timeout |
|------|---------|-------------|------------------|-----------------|
| `debug.json` | `pip install scrapling[all]` | 120s | ~60s | 120s |
| `qwenpaw_query_error_vqriog6w.json` | `sleep 120 && echo 123` | 180s | ~60s | 180s |
| Terminal log (21:01) | `sleep 120 && echo 123` | 180s | ~60s | 180s |

## Timeout architecture diagram

```
LLM calls: execute_shell_command(timeout=180)
  │
  ├─ Coordinator deadline (hook): 60s ─── fires cancel_event.set()
  │                                            │
  │                                            ▼
  │                                   cancellable_wait detects it
  │                                            │
  │                                            ▼
  │                                   subprocess KILLED at ~60s
  │
  └─ Shell tool timeout: 180s ──────── fallback_secs=180
                                       (IGNORED when ctx exists)
```

## Affected files

| File | Role |
|------|------|
| `src/qwenpaw/tool_calls/_coordinator.py` | Offload uses `cancel_event` |
| `src/qwenpaw/tool_calls/_timeout_helper.py` | Ignores `fallback_secs` with ctx |
| `src/qwenpaw/tool_calls/_context.py` | Single `cancel_event` for both semantics |
| `src/qwenpaw/agents/tools/shell.py` | Wrong timeout in error message |
| `src/qwenpaw/agents/react_agent.py:616` | Hook hardcodes 60s deadline |

## Related

- PR #6052 fixes the **hint message format** (orphan ToolResultBlock)
  but does not address the offload-kills-subprocess issue
- Issues #5996 #5985 #5962 (downstream symptoms of broken offload)

### LINKED ISSUE #6245: [Bug]: Session permanently blocked when shell command exceeds coordinator deadline (regression from #6056 fix)
## Environment

- QwenPaw version: v2.0.0.post3
- OS: Linux (Docker)

## Summary

When a shell command hangs past the coordinator deadline, the session becomes **permanently blocked** — all subsequent messages for that session queue indefinitely until the process is restarted.

This is a **regression** introduced by the fix for #6056 (commit that commented out `_handle_deadline_reached`).

## Root Cause

`_coordinator.py` lines 113-119:

```python
elif event.type == "deadline_reached":
    # TODO FIXME: offload is temporarily
    # disabled. cancel_event kills the
    # subprocess instead of backgrounding
    # it. See issue-offload-kills-subprocess.
    ctx.deadline = None
    # self._handle_deadline_reached(ctx)   ← COMMENTED OUT
    # terminal = "offload"
    # break
```

When `deadline_reached` fires:
1. `ctx.deadline` is set to `None` — **no more timeout checks**
2. `ctx.cancel_event` is **never set**
3. `_await_next_event` loops back with `remaining = None` → `asyncio.wait(timeout=None)` blocks forever
4. `cancellable_wait` in shell.py also waits on `cancel_event` → blocks forever
5. The session consumer (`_consume_queue`) is stuck → all subsequent messages for that session are queued and never processed

## Complete Deadlock Chain

```
shell command hangs
  → proc.communicate() never returns
    → coordinator deadline expires
      → _await_next_event returns "deadline_reached"
        → ctx.deadline = None (cancel_event NOT set)
          → loop continues, asyncio.wait(timeout=None)
            → cancellable_wait also blocks on cancel_event
              → session consumer stuck
                → all new messages for this session queued forever
```

## Regression Context

**Before (Bug #6056):** `_handle_deadline_reached` was called, setting `cancel_event`. This killed the subprocess (unwanted) but at least the session recovered.

**After (current):** `_handle_deadline_reached` is commented out. The subprocess is NOT killed, and the session hangs permanently. This is strictly worse.

| Aspect | Before (#6056) | After (current) |
|--------|----------------|-----------------|
| cancel_event | Set | Never set |
| Subprocess | Killed (prematurely) | Runs forever |
| Session | Recovers | **Permanently blocked** |
| Messages | Processed after timeout | **Queued forever** |

## Reproduction

1. Start QwenPaw with default config
2. Send a message that triggers a long-running shell command (e.g., `sleep 300` with `timeout=60`)
3. Wait for the coordinator deadline (~60s) to expire
4. Send any subsequent message to the same session
5. **Result**: Second message is queued and never processed; session is permanently stuck

## Expected Behavior

When the coordinator deadline expires:
1. `cancel_event` should be set
2. `cancellable_wait` should cancel the subprocess (kill process group)
3. The tool call should return a timeout error
4. The session should recover and process subsequent messages

## Suggested Fix

In `_coordinator.py`, the `deadline_reached` handler should set `cancel_event` to properly cancel the stuck tool:

```python
elif event.type == "deadline_reached":
    # Cancel the running tool so cancellable_wait can clean up
    # the subprocess. This replaces the disabled offload path.
    if ctx.cancel_reason is None:
        ctx.cancel_reason = CancelReason.TIMEOUT
    ctx.cancel_event.set()
    break
```

This restores the timeout contract: shell commands that exceed the deadline are properly killed, and the session recovers to process new messages.

The `ctx.deadline = None` approach (current code) is fundamentally broken because it removes the only timeout mechanism without providing an alternative.

## Related Issues

- #6056 — Original bug (offload kills subprocess) — this regression was introduced by its fix
- #6014 — PR for honoring shell command timeouts (different issue, coordinator deadline vs shell timeout mismatch)
- #1675 — Same symptom (queue blocked) but different root cause (missing `_process_batch` timeout)

feng183043996:
## Root Cause

`cancel_event` is used for **two different purposes**:
1. **User cancel** (stop button) → should kill subprocess
2. **Deadline offload** (timeout) → should keep subprocess running in background

The original code didn't distinguish between these two cases. When `_handle_deadline_reached` set `cancel_event`, `cancellable_wait` would cancel the asyncio task, and `shell.py` would kill the subprocess process group. This made offload a no-op — the subprocess was already dead before background supervision could begin.

### `_timeout_helper.py` — `cancellable_wait()`
When `cancel_event` fires, check `ctx.offload_reason`:
- **Offload** (reason is set): raise `CancelledError` WITHOUT cancelling the task. The subprocess continues running.
- **Cancel** (reason is None): cancel the task as before (kill subprocess).

### `shell.py` — `execute_shell_command()`
When catching `CancelledError`, check `ctx.offload_reason`:
- **Offload**: re-raise without killing the subprocess. The coordinator handles background supervision.
- **Cancel**: kill the process group as before.

## Tests

Added unit tests verifying:
- Offload: task is NOT cancelled, subprocess keeps running
- Cancel: task IS cancelled, subprocess is killed

Fixes #6056
Fixes #6245

### DIFF
--- a/src/qwenpaw/agents/tools/shell.py
+++ b/src/qwenpaw/agents/tools/shell.py
@@ -625,6 +625,15 @@ async def execute_shell_command(
                 returncode = proc.returncode
 
             except (asyncio.TimeoutError, asyncio.CancelledError):
+                # Check if this is an offload (deadline reached) vs a cancel.
+                # For offload, do NOT kill the subprocess — let it continue
+                # in background under coordinator supervision.
+                from ...tool_calls import get_call_context
+                ctx = get_call_context()
+                if ctx is not None and ctx.offload_reason is not None:
+                    # Re-raise without killing; coordinator handles offload.
+                    raise
+
                 stderr_suffix = (
                     f"⚠️ TimeoutError: The command execution exceeded "
                     f"the timeout of {timeout} seconds. "
--- a/src/qwenpaw/tool_calls/_timeout_helper.py
+++ b/src/qwenpaw/tool_calls/_timeout_helper.py
@@ -48,13 +48,20 @@ async def cancellable_wait(
         )
 
         if cancel_waiter in done:
+            # Distinguish between cancel (user stop) and offload (deadline).
+            # For offload, do NOT cancel the task — let the subprocess
+            # continue running under background supervision.
+            if ctx.offload_reason is not None:
+                raise asyncio.CancelledError(
+                    f"tool offloaded (reason={ctx.offload_reason.value})"
+                )
             task.cancel()
             try:
                 await task
             except (asyncio.CancelledError, Exception):
                 pass
             raise asyncio.CancelledError(
-                f"tool cancelled by manager (reason={ctx.cancel_reason})",
+                f"tool cancelled by manager (reason={ctx.cancel_reason})"
             )
 
         return task.result()
--- a/tests/unit/tool_calls/test_cancellable_wait_offload.py
+++ b/tests/unit/tool_calls/test_cancellable_wait_offload.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+"""Tests for offload-aware cancellable_wait.
+
+Verifies that:
+- When offload_reason is set, the task is NOT cancelled (subprocess keeps running)
+- When offload_reason is NOT set, the task IS cancelled (normal cancel behavior)
+"""
+import asyncio
+import pytest
+from unittest.mock import MagicMock, patch
+
+from qwenpaw.tool_calls._timeout_helper import cancellable_wait
+from qwenpaw.tool_calls._context import (
+    ToolCallContext,
+    OffloadReason,
+    CancelReason,
+)
+from qwenpaw.tool_calls._ctxvars import set_call_context
+
+
+@pytest.fixture
+def mock_ctx():
+    """Create a mock ToolCallContext."""
+    import time
+    ctx = ToolCallContext(
+        tool_call_id="test-123",
+        tool_name="execute_shell_command",
+        session_id="sess-1",
+        agent_id="default",
+        root_session_id="root-1",
+        started_at=time.monotonic(),
+        deadline=time.monotonic() + 10.0,
+        cancel_event=asyncio.Event(),
+    )
+    return ctx
+
+
+@pytest.mark.asyncio
+async def test_cancellable_wait_offload_does_not_cancel_task(mock_ctx):
+    """When offload_reason is set, task should NOT be cancelled."""
+    mock_ctx.offload_reason = OffloadReason.TIMEOUT
+    mock_ctx.cancel_event.set()
+
+    task_cancelled = False
+
+    async def long_running_task():
+        nonlocal task_cancelled
+        try:
+            await asyncio.sleep(100)
+        except asyncio.CancelledError:
+            task_cancelled = True
+            raise
+
+    from qwenpaw.tool_calls._ctxvars import reset_call_context
+    token = set_call_context(mock_ctx)
+    try:
+        with pytest.raises(asyncio.CancelledError, match="offloaded"):
+            await cancellable_wait(long_running_task(), fallback_secs=1.0)
+    finally:
+        reset_call_context(token)
+
+    assert not task_cancelled, "Task should NOT be cancelled when offloading"
+
+
+@pytest.mark.asyncio
+async def test_cancellable_wait_cancel_does_cancel_task(mock_ctx):
+    """When offload_reason is NOT set (user cancel), task SHOULD be cancelled."""
+    mock_ctx.cancel_reason = CancelReason.USER
+    mock_ctx.cancel_event.set()
+
+    task_cancelled = False
+
+    async def long_running_task():
+        nonlocal task_cancelled
+        try:
+            await asyncio.sleep(100)
+        except asyncio.CancelledError:
+            task_cancelled = True
+            raise
+
+    from qwenpaw.tool_calls._ctxvars import reset_call_context
+    token = set_call_context(mock_ctx)
+    try:
+        with pytest.raises(asyncio.CancelledError, match="cancelled by manager"):
+            await cancellable_wait(long_running_task(), fallback_secs=1.0)
+    finally:
+        reset_call_context(token)
+
+    assert task_cancelled, "Task SHOULD be cancelled when user cancels"


XiuShenAl (commented):
Thanks for the fix — the root cause analysis is accurate: `cancel_event`
being shared between offload and user-cancel does cause the subprocess
to be killed on offload.

However, #6151 addresses the same issue (#6056) more fundamentally by
replacing the single `deadline` with `offload_deadline` + `kill_deadline`,
and by removing `cancel_event.set()` from `_handle_deadline_reached`.
Offload then never signals via `cancel_event`, so `cancellable_wait` /
`shell.py` no longer need to distinguish the two cases.

These changes would be unnecessary and structurally conflicting once
#6151 lands:

1. **`cancellable_wait` — `offload_reason` check**: #6151's offload path
   no longer sets `cancel_event`, so this branch would never run.
2. **`shell.py` — offload re-raise**: #6151 splits TimeoutError /
   CancelledError handling and extracts `_cleanup_proc()`; the guard
   here conflicts and is unnecessary under that design.
3. **`_handle_deadline_reached`**: #6151 removes the `cancel_event.set()`
   that this PR works around.

Also, this approach looks incomplete on its own: skipping `task.cancel()`
but still raising `CancelledError` orphans the `proc.communicate()` Task
and lets `_drain` mark the tool `INTERRUPTED`, so `_supervise` cannot
collect a real background result. Separately, this PR doesn't re-enable
the commented-out deadline/offload path in the coordinator, so it
wouldn't fix the #6245 session hang on current `main`.

**Recommendation:** Close in favor of #6151, which resolves #6056 and
#6245 at the architectural level. (#6151 is still Draft — happy to take
help reviewing/testing it.) Appreciate the thorough RCA in the
description; it helped confirm the root cause.

lalaliat:

PR #6157, opened 2026-07-15
feat(browser): chrome extension plugin — install, pairing & native messaging bridge

xiaoming-qxm:
> **Developer Preview.** Unpacked installation only; the Chrome Web Store path is surfaced as *coming soon* and is not enabled.

## Summary

Introduces the **`chrome` bundle plugin** — the plugin-layer half of QwenPaw's official Chrome integration. It pairs the user's own Chrome to QwenPaw via a packaged **Manifest V3 extension** and a **Native Messaging → WebSocket bridge host**, and ships the install/repair/pairing UX plus the plugin's HTTP routes. Opt-in bundle plugin, same model as `qwenpaw-pet` / `cloudpaw`. The WebSocket server half of the bridge (`/api/ws/chrome`) ships in core (#6276).

## Design Highlights

- **Strict core/plugin boundary.** All connection authority — WS server, auth, session/lease management, semantic control — lives in core (#6276). This PR owns only: MV3 extension assets, the Native Messaging host, and install/repair UX. The protocol contract is core-owned (`protocol.py`); the plugin carries a mirrored snapshot (`protocol_mirror.py`) and negotiates versions via `bridge_hello` at handshake.
- **`nm_host` is a single-backend dumb pipe.** It holds no policy of its own: directional frame limits mirror Chrome's Native Messaging protocol facts (64 MiB extension→host, 1 MiB host→extension); reconnects are bounded with backoff; on pump termination the host exits fail-fast so Chrome observes a real disconnect instead of a zombie "connected" pipe. Disconnects are attributed, not anonymous: application-reserved WebSocket close codes (4000–4003: stdin EOF, frame protocol error, oversize inbound frame, host internal error) let the receiving side distinguish failure classes instead of seeing a generic drop.
- **Bridge endpoint is derived, never injected.** `POST /setup` accepts no `ws_url` (unknown fields are rejected); the endpoint is derived exclusively from the running server's actual bind and must be loopback — a non-loopback bind fails setup with `409` instead of provisioning a bearer token against an exposed interface.
- **Status honesty, split by authority.** `GET /install-status` reports only what the plugin owns (install state). Connection state is rendered by the plugin page from the core authority `/api/browser/chrome/status`, degrading gracefully when the core half is absent. The plugin never fabricates `connected`.
[...]

PR #6276, opened 2026-07-20
feat(browser): unified browser — one SDK, any backend

xiaoming-qxm:
## Problem

Browser automation was wired per-backend: each backend carried its own call surface, its own error vocabulary, and its own notion of a page. Adding a backend meant touching the agent-facing surface, and the agent had to know which backend it was talking to.

## Design

**One SDK, any backend.** The agent writes async Python against a single `Browser` / `Page` / `LocatorView` surface. Three backends — `chrome` (via the MV3 extension bridge), `cdp`, `playwright` — implement one internal control-link contract and are selected by policy, never named by the caller. These are internal control-link implementations, not operator choices: the operator-facing config exposes two orthogonal axes — `browser.identity` (*whose* browser: `auto | user | avatar | guest`) and `browser.backend` (*how* it is launched: `auto | launch | managed_cdp | connect_cdp`) — and the control link is resolved from those.

**Control plane / execution plane.** LLM-authored code never runs in the main process:

| Plane | Process | Owns |
|---|---|---|
| Control | main, SDK-free | control links, extension bridge, approval adjudication, event fan-out |
| Execution | spawned subprocess | the Browser SDK and LLM-authored async Python |

They speak a versioned frame protocol over a `socketpair`: `exec_request/exec_result`, `ctrl_call/ctrl_result`, `event`, `approval_request/approval_verdict`. No pickle, no shell.

**Honesty over convenience.** Where the system cannot prove something it says so instead of returning the optimistic answer: observation coverage reports `PARTIAL` with a typed gap rather than a bare `COMPLETE`; a command whose receipt never arrived is reported as *unknown*, distinct from *did not happen*; a transport that times out mid-flight surfaces as uncertain rather than failed.

### Key decisions

- **Identity is one axis, adjudicated once.** `browser.identity` is `auto | user | avatar | guest`; the backend axis no longer encodes identity (`backend: extension` is migrated to `identity: user` on load). Adjudication happens in the plane that holds both the model's intent and live availability facts — the execution plane — so an explicitly requested identity is honoured even when another backend is connected.
- **Pages are cycle-scoped and owner-composite.** A page's identity includes its owning workspace, session and birth variant. Pages are reclaimed at the end of the response cycle by the plane, which is the single authority on what is reclaimable; links are told which pages to close rather than sweeping their own state.
- **Degradation is a ladder, not a fallback.** Dropping from `user` to `guest` only counts as degradation when nobody asked for it; an explicit `guest` request is a choice and is never annotated as degraded.
- **Every side effect passes one gate.** Verbs are classified `READ` / `STATE_CHANGE` / `TRANSMIT`; reads pass through, the rest round-trip to the control plane for adjudication. This is the seam a real policy will attach to.
- **The extension boundary is protocol-versioned.** `WIRE_PROTOCOL_VERSION` gates control ↔ execution; `PROTOCOL_VERSION` plus a `bridge_hello` handshake gates core ↔ extension, with lease authority held solely by the core bridge.
- **The beta is gated and the old path stays reachable.** `browser.experimental: false` restores the deprecated `browser_use` route, isolated so the two tracks cannot interleave.

## Implementation

- **SDK surface** — `Browser.connect(identity=...)`, `Page` navigation and input, `LocatorView` query and action verbs including frame locators, `Observation` with typed coverage gaps plus coordinate guidance and visible-DOM text projection, and a generated manual gated by a contract test so the documented calls cannot drift from the real surface.
- **CDP backend** — full adapter, target discovery, transport, and an injected locator engine.
- **Execution plane** — per-session namespace dropped on `close_session`, an AST guard on submitted code, bounded frames with spill for oversized payloads, stdout spill, a cancellation protocol, backend process ownership guards, and blocking browser-process / artifact I/O offloaded from the event loop.
- **Chrome backend** — the WebSocket endpoint, the native-messaging bridge with token rotation, the lease authority, the protocol contract and a plugin contract-health surface live in core. The companion extension PR carries only the MV3 assets, a single-backend native host, and its install/repair UI.
- **Console UI** — an experimental toggle and browser-mode controls on the Tools page, a unified `BrowserCard` tool card with the deprecated `browser_use` card isolated alongside it, and localisation across seven languages.

## Accepted risk: host-level isolation is out of scope for this release

This is a recorded decision, not an oversight. Host-level sandboxing of the execution plane was reviewed and explicitly deferred so that the SDK, the plane split and the wire contract can land and stabilise first.

- **Execution is not sandboxed against the host.** LLM-authored code runs in a subprocess with full Python builtins — the same file, network and environment access as the QwenPaw process. The subprocess isolates memory and crashes, not the filesystem, network or secrets.
- **Action policy is not enforced.** The adjudicator returns `ALLOW` for every state-change and transmit; it exercises the token lifecycle only. Red-line categories exist in code but are unreachable — a mechanism under test, **not** a security boundary.
- **Tokens are unsigned** — trusted-side lifecycle, not cross-process authenticated.
- **The OS sandbox probe is a probe.** The subprocess floor is what runs.
- **`browser.experimental` defaults to `true`**, so this applies to a default install, not only to opt-in users.

**Why this is acceptable now:** the resulting privilege level equals the existing shell tool, which the agent can already reach. This PR inherits the product's trust boundary rather than widening it. The only host-facing hardening landed here is constant-time bridge-token comparison.

**Revisit condition:** an enforced action policy and an OS-level sandbox floor are prerequisites before `browser.experimental` leaves beta, and before the red-line categories may be described as a security control.

## Known limitations

- **Approval policy is a mechanism, not a control** — see accepted risk above.
- **No real-browser run is part of this PR's verification.** All three backends are exercised against fakes (fake extension bridge, fake CDP transport) plus a local Playwright provider run.
- **CDP is complete but unproven against real Chromium** — adapter, discovery and transport are implemented and covered against a fake transport only.
- The Chrome path's real-browser behaviour is verified on a one-off branch that merges this PR with the companion extension PR (#6157); core lands first by design.

## Verification

- `tests/contract/browser`: **11 passed**; `tests/integration/browser`: **18 passed** — 29 in total.
- CI is green across the matrix: unit / contract / integrated tests on ubuntu + macos + windows (py3.11, py3.13), CodeQL (python + typescript), and Playwright UI smoke.
- Per repository convention only integration and contract tests live in the repo, so this diff carries integration-level coverage only.


cuiyuebing (approved):
LGTM

PR #5922, opened 2026-07-10
feat(observability): track user/session/version on langfuse traces

alvinlee518:
Propagate user_id, session_id, and package version onto Langfuse traces via propagate_attributes, and switch the root agent span to start_as_current_observation so child observations attach under the correct parent. The trace id now uses uuid4().hex (Langfuse rejects dashed UUID strings) and falls back to the caller-supplied id when the client exposes none.

Update unit tests for the new span lifecycle and add coverage for attribute propagation, trace-id override/fallback, the error-marked span path, the no-client branch, and the hex trace-id contract.

## Description

[Describe what this PR does and why]

**Related Issue:** Fixes #(issue_number) or Relates to #(issue_number)

**Security Considerations:** [If applicable, e.g. channel auth, env/config handling]

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

[How to test these changes]

## Local Verification Evidence

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]

PR #5128, opened 2026-06-11
group langfuse observations by agent loop

totoyang:
## Description

This PR improves Langfuse observability by grouping one full agent ReAct loop into a single trace.

Previously, each OpenAI-compatible LLM call could be reported as a separate Langfuse trace, which made one conversation turn appear as many disconnected traces. This PR adds an agent-level Langfuse trace scope around each runner execution, then attaches LLM generations and tool calls as child observations under that trace.

**Related Issue:** Fixes #5127

**Security Considerations:** Langfuse integration remains optional and environment-gated. No secrets are logged or persisted by this change. Trace metadata includes runtime identifiers such as session ID, user ID, channel, and agent ID, matching observability use cases.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [x] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Feishu, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

Tested the Langfuse trace-context helpers and related provider/agent/runner behavior with focused unit tests.

The covered behavior includes:

- agent trace context propagation into model generation kwargs
- root span creation and context restoration
- tool observation creation under the current trace
- error handling for tool observations
- existing OpenAI provider compatibility tests
- existing runner/session behavior tests

## Local Verification Evidence

```bash
[...]

PR #6277, opened 2026-07-20
fix(observability): use valid Langfuse trace IDs

splash-li:
## Description

Log Langfuse root trace initialization failures at warning level instead of debug level, so observability failures do not remain silent when Langfuse is enabled.

This PR also adds a caplog assertion covering the warning behavior. The trace ID fix and its regression coverage are already provided by #5922.

**Related Issue:** Relates to #5128

**Security Considerations:** None.

## Type of Change

- [x] Bug fix
- [x] Tests

## Testing

uv run --with pytest --with pytest-asyncio --with langfuse python -m pytest -q tests/unit/observability/test_langfuse_context.py tests/unit/observability/test_langfuse_integration.py

Result: 20 passed.

### DIFF
--- a/src/qwenpaw/hooks/observability/langfuse_hook.py
+++ b/src/qwenpaw/hooks/observability/langfuse_hook.py
@@ -60,7 +60,7 @@ class LangfuseTraceHook(LifecycleHook):
             await scope.__aenter__()
             ctx.extras[_LANGFUSE_SCOPE_KEY] = scope
         except Exception:
-            logger.debug("langfuse trace scope open failed", exc_info=True)
+            logger.warning("langfuse trace scope open failed", exc_info=True)
         return HookResult()
 
 
--- a/tests/unit/observability/test_langfuse_integration.py
+++ b/tests/unit/observability/test_langfuse_integration.py
@@ -151,7 +151,7 @@ class TestLangfuseTraceHook:
         assert lf.get_current_trace().name == "qwenpaw.agent.react_loop"
         assert len(client.started) == 1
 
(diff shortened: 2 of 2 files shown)


XiuShenAl (commented):
Thanks for the clear write-up and the focused fix — the root cause analysis looks spot on.

**What looks good**
- Restoring `uuid.uuid4().hex` is the right fix for Langfuse’s 32-char lowercase hex `trace_id` requirement.
- Raising the open-failure log from `debug` to `warning` is a nice improvement: when Langfuse is enabled, a failed root scope should not fail silently.
- The added assertions help lock in the intended behavior.

**One important update**
The core `str(uuid.uuid4())` → `uuid.uuid4().hex` change appears to have already landed on `main` via #5922 (which also added a dedicated regression test, `test_trace_id_is_hex_without_dashes`). So the primary bug fix in this PR is largely superseded.

**Request**
Could you please rebase onto the latest `main` and re-scope the PR to the remaining unique value? In particular:
1. Keep the `logger.warning(...)` change (and the `caplog` assertion), if it still applies cleanly after rebase.
2. Drop or simplify any hex-format assertions that now duplicate coverage already present on `main`.

After that, this should be a small, easy-to-review follow-up. Happy to approve once it’s rebased and narrowed — thank you again for catching and documenting this!

splash-li:
Thank you for your review. We missed that #5922 was being merged and opened this PR separately. I have rebased and narrowed the changes accordingly.

XiuShenAl (approved):

PR #6323, opened 2026-07-21
feat(scroll): add staged compaction and durable task continuity

niceIrene:
## Summary

This PR redesigns Scroll context management around a durable, staged compaction pipeline.

The core invariant is:

> `history.db` remains the source of truth. Headlines, continuation summaries, and eviction indexes provide task continuity and retrieval routes; they never replace the exact archived conversation.

The redesign adds recoverable tool-result trimming, complete active-turn protection, bounded continuation-summary generation, deterministic quality checks, and explicit `CONTEXT_UNFIT` handling.

## Type of Change

- [x] Feature
- [x] Bug fix
- [x] Refactoring
- [x] Documentation
- [ ] Breaking change

## Components Affected

- [x] Core context management
- [x] Agent runtime
- [x] CLI `/compact`
- [x] Console
- [x] Documentation
- [x] Tests

### 1. Staged, durability-first compaction

Automatic Scroll compaction now follows this pipeline:

```text
persist live context to history.db
  → if input <= trigger: stop
  → batch-fold eligible completed-turn tool results
  → if input <= trigger: stop
  → evict completed middle dialogue
  → update continuation summary
  → rebuild summary + eviction index + live tail
  → fold remaining completed results only if still under pressure
  → fold acknowledged active-turn results only above the effective hard limit
  → CONTEXT_UNFIT if protected content still cannot fit
```

Important behavior:

- Input exactly at or below the default 80% trigger does not fold tool results or evict dialogue.
- Pre-trimming runs only when input is strictly above the trigger.
- Explicit `/compact` skips pre-trimming and starts with dialogue eviction.
- Scroll never evicts dialogue unless the corresponding history has first been persisted successfully.

### 2. Recoverable tool-result trimming

Before evicting dialogue, Scroll batch-folds eligible completed-turn tool results larger than 200 characters.

It preserves:

- the complete active turn;
- the five newest tool results globally;
- pending and unread tool results;
- tool identity and pairing information.

Folded results become exact recovery pointers:

```text
[scroll folded] old tool result content cleared; recover with recall_history(op="recall_tool", tool_call_id='call_abc')
```

The original result remains durable in `history.db` and, when applicable, in its offloaded artifact.

Batch folding minimizes repeated prompt rewrites and token recounts.

### 3. Oversized active-turn recovery

A single tool-heavy active turn can exceed the model window even after completed dialogue has been evicted.

At the final hard-limit boundary, Scroll may fold earlier active-turn tool results only when they were included in a previously successful model request.

It continues to preserve:

- the current user request;
- pending tool calls;
- unread tool results;
- the five newest tool results.

Scroll reserves `min(4096, 5% of context_size)` tokens for the next model output. If the protected input still cannot fit, it raises `CONTEXT_UNFIT` instead of silently resetting the conversation or removing unread evidence.

### 4. Pointer-backed continuation summary

When dialogue is actually evicted, Scroll maintains a five-section task-state summary:

- `Active Task`
- `Current State`
- `Constraints`
- `Decisions`
- `Open Work`

The summary:

- is generated through a normal chat completion with thinking disabled;
- does not use structured output, JSON mode, or a response schema;
- supports English and Chinese prompts;
- receives bounded, role-aware evidence with user facts and constraints prioritized over tool noise;
- includes normalized message timestamps for chronology;
- is linked to one code-managed durable `seq` range rather than model-generated inline citations;
- is injected as background context, never as an active instruction.

Initial generation uses `initial`. Later evictions use `update` while the previous summary remains source-backed.

### 5. Deterministic summary quality guard

Generated summaries are parsed and validated locally without an additional LLM judge call.

The guard checks:

- exact Markdown section order;
- valid task status;
- durable `seq` endpoints;
- duplicate state entries;
- unsupported opaque identifiers;
- likely secret leakage;
- output size limits.

An invalid first candidate receives at most one repair attempt with concise validation feedback.

Generation and repair share a single 60-second total timeout. A timeout does not start another model call. If the previous summary is still source-backed, it is preserved and marked stale.

### 6. Retention-safe summary updates

Before updating a previous continuation summary, Scroll verifies that its `covered_seq` endpoints still exist.

If retention has purged either endpoint:

- the old summary is no longer treated as source-backed;
- its claims are not reassigned to a newer sequence range;
- Scroll discards the invalid state and rebuilds with `initial` from newly persisted evidence.

This prevents every future summary update from failing permanently on an expired sequence pointer.

### 7. Simplified headline and eviction-index protocol

Headlines remain optional, compact retrieval labels for substantive task turns.

This PR:

- clarifies the headline prompt around task, state, status, decisions, blockers, and next steps;
- increases retrieval coverage for meaningful task events;
- removes compaction-time headline backfilling;
- removes the old continuity-checkpoint protocol;
- preserves unlabeled ranges as `(no milestone)` with exact `seq` spans.

The continuation summary tracks current effective state, while the eviction index remains responsible for historical checkpoints and recovery ranges.

Both are rendered inside one `<system-info>` background envelope.

### 8. Manual compaction hints

`/compact <hint>` now supplies bounded, secret-redacted, one-shot guidance to the current compaction.

The hint:

- can prioritize supported information;
- is not treated as evidence;
- is not persisted as task state;
- cannot override the summary protocol or safety rules.

### 9. Console cleanup

The ReAct Agent runtime configuration now exposes only the long-term memory backend.

The context-manager-backend and context-strategy selectors were removed from this general settings card. Scroll and Native remain configurable through `running.light_context_config.strategy`.

## Review Feedback Addressed

The following review edge cases are covered:

1. **Unbounded summary generation**
   - Added a shared 60-second timeout for generation and repair.
   - Timeout skips repair and preserves a valid previous summary as stale.

2. **Retention causing permanently stale summaries**
   - Previous summary endpoints are checked before every update.
   - Expired summaries are discarded and rebuilt from new durable evidence.

3. **Dialogue eviction exactly at the trigger**
   - The initial gate now uses `tokens <= trigger`.
   - Input exactly at 80% stops without dialogue eviction.

## Compatibility

- `history.db` remains the authoritative durable history.
- Exact `recall_history` and `recall_history_python` recovery remain available.
- Native context compression remains supported.
- No provider API changes are required.
- Existing configuration files continue to load.
- English and Chinese context-management documentation are updated.

## Testing

```bash
.venv/bin/python -m pytest -q tests/unit/agents/context
# 251 passed

.venv/bin/python -m pytest -q \
  tests/unit/agents/test_command_handler.py \
  tests/unit/agents/test_react_agent_scroll_seen.py
# 25 passed
```

Additional verification:

- focused timeout, retention, and exact-trigger regression tests pass;
- mypy passed;
- Black passed;
- Flake8 passed;
- Pylint passed;
- Python AST and secret checks passed;
- TypeScript build check passed;
- English and Chinese documentation pass Prettier;
- `git diff --check` passed.

## Checklist

- [x] Relevant tests added and passing
- [x] Review edge cases covered by regression tests
- [x] Backward compatibility considered
- [x] English documentation updated
- [x] Chinese documentation updated
- [x] No structured-output dependency added
- [x] No additional LLM quality-judge call added


jinliyl:
Thank you for the substantial work on this Scroll context-management redesign. The staged compaction pipeline, active-turn protection, and pointer-backed continuation summary are thoughtful additions, and the focused context test suite passes locally (235 tests).

I found three edge cases that seem worth addressing before merge:

1. **Continuation-summary generation has no timeout.** `_generate_plain_summary()` awaits the model inline, and a failed candidate can trigger a second call. The previous 20-second timeout was removed, so a stalled provider or connection can block compaction—and therefore the current agent turn—indefinitely. Could we add a hard timeout around each attempt or around the overall generate/repair operation, then preserve the previous summary as stale on timeout?

2. **History retention can make summary updates permanently stale.** The new covered range always retains `previous.covered_seq[0]`, while validation requires that endpoint to still exist. After the default retention purge removes that oldest row, every subsequent update retries and fails with `seq pointer endpoint does not exist`, preserving the old summary forever. I reproduced this locally. It may be safer to trim expired coverage or rebuild from the surviving/new evidence when a referenced endpoint has been purged.

3. **Exactly-at-trigger input still evicts dialogue.** The initial gate uses `tokens < trigger`, and the later `tokens <= trigger` return only runs when at least one tool result was folded. With an 800-token input in a 1000-token window and no fold candidates, I reproduced two old messages being evicted, although the PR description says that usage at or below 80% should stop without dialogue eviction. An initial `<=` check, or an unconditional post-pretrim threshold check, should align the implementation with the documented behavior.

Aside from these cases, the implementation and documentation are impressively comprehensive. Thanks again for putting this together.

jinliyl:
Thank you for the substantial work on this redesign. The durability-first pipeline, active-turn protection, and expanded test coverage are all valuable improvements.

While reviewing the implementation, I found a few edge cases that may be worth addressing before merge:

1. Pre-folding mutates completed tool results before continuation-summary evidence is built. If pre-folding is insufficient and dialogue eviction follows, the summary model no longer receives the original tool-result content, even though the exact content remains available in history.db. Building the evidence before folding, or loading it from durable history by seq, would preserve those outcomes.

2. The eviction index currently has no token budget. With headline coverage increased and the per-headline compatibility limit raised to 2,000 characters, a single eviction block containing many headlines can make the synthetic memory exceed the model window. The later pressure stages can fold tool results, but cannot reduce the index itself, so compaction may end in CONTEXT_UNFIT. A smaller headline limit or pressure-aware index compaction would help.

3. Continuation-summary repair currently retries after every non-timeout exception, including provider, authentication, rate-limit, and network failures. It may be safer to retry only parsing or quality-validation failures and preserve the previous summary immediately for provider errors.

4. At exactly the trigger threshold, Scroll now returns without compacting because it uses tokens <= trigger, while MemoryMiddleware still predicts compaction with estimated_tokens >= threshold. With summarize_when_compact enabled, this can run long-term-memory summarization even though Scroll performs no compaction. Aligning the boundary checks would avoid that mismatch.

The targeted test suite passes locally (276 tests), and the current CI checks are green. Thank you again for the thoughtful implementation and detailed documentation. Please let me know if I have missed any intended behavior behind these choices.

jinliyl:
Thank you again for the thoughtful Scroll redesign and for addressing the earlier review feedback. I reviewed the current head (`cf3ceb7e`) and ran the focused context, command-handler, middleware, chat-utils, and console tests locally; all 367 tests passed.

I found three additional edge cases that may be worth addressing before merge:

1. **Continuation-summary output limits are not compatible with every supported provider.** `_generate_plain_summary()` passes `max_tokens` directly to the active model. With the currently locked AgentScope 2.0.4.post1, Gemini and OpenAI Responses expect `max_output_tokens`; the Google SDK rejects `max_tokens` as an extra field, and the Responses client does not accept it either. This means continuation-summary generation fails for those provider paths. On Gemini, the compatibility layer also sets the shared model instance `parameters.thinking_enable = False` before the request is validated, so the failed summary call can leave thinking disabled for subsequent normal agent calls. Could the output-limit argument be translated per provider, and the Gemini thinking override be kept call-scoped or restored afterward?

2. **A retention purge can leave an expired summary in the restored model context indefinitely.** The normal lifecycle saves the agent state before `agent.close()` purges old history. On the next load, `compress()` returns early when usage is at or below the trigger, before `_source_backed_previous_summary()` is consulted. I reproduced a restored state where both covered endpoints had been deleted, but the Scroll placeholder was unchanged, the summary remained loaded, and it still said that exact archived content remained available. The current retention regression test covers rebuilding on a later over-trigger compaction, but not a resumed session that stays below the trigger. Validating/removing stale summary state during load, before the early return, or before the saved checkpoint is finalized would keep the model-facing claim source-backed.

3. **The summary evidence limit is character-based rather than token-based.** `input_chars = context_size * 2` allows 65,536 evidence characters for a 32K model, plus the previous summary, prompt, and up to 4,000 output tokens. Using the repository tokenizer, representative 65,536-character Chinese samples encoded to roughly 39.6K–53.2K tokens, so a Chinese-heavy eviction can exceed the model window even though it satisfies the character bound. An exact model token count, followed by iterative truncation with output space reserved, would make the bound reliable across languages.

The rest of the staged pipeline and the expanded tests look solid. Thank you for taking a look at these cases.

niceIrene:
Thank you for the detailed follow-up review. I’ve addressed all three issues in the latest commits:

1. **Provider-compatible continuation-summary calls**
   - Summary generation still calls `agent.model` directly to avoid entering the full agent loop, middleware, tools, or context management recursively.
   - The neutral output limit is now translated to the provider-specific parameter:
     - Gemini and OpenAI Responses: `max_output_tokens`
     - GPT-5/o-series Chat Completions: `max_completion_tokens`
   - Gemini’s temporary `disable_thinking` override is now call-scoped and restored in a `finally` block, including when validation or the provider request fails.

2. **Expired summaries after retention purge**
   - Restored continuation summaries are now validated against durable history during session loading, before the below-trigger early return.
   - If the referenced history endpoints have expired, Scroll removes the unsupported summary and rebuilds the model-facing memory placeholder without the stale claim.

3. **Token-based summary evidence budgeting**
   - The evidence limit is now based on the model’s token counter rather than character count.
   - The calculation includes the system prompt, update prompt, previous summary, and new evidence, while reserving space for summary output and a safety margin.
   - Scroll reduces evidence until the complete request fits. If even the fixed prompt and previous state cannot fit, it safely preserves the previous summary instead of sending an overflowing request.

The fixes are in:

- `a759a13b` — `fix(providers): adapt summary output limits`
- `50f69f79` — `fix(scroll): keep summaries source-backed and window-safe`

jinliyl (approved):
LGTM

rayrayraykk (approved):
LGTM

cuiyuebing (approved):
LGTM

cuiyuebing (approved):
LGTM

PR #6334, opened 2026-07-22
test(e2e): add tool-approval level cases (#5685)

yutai78786:
## Summary
- Adds Sprint-4 regression coverage for the tool-approval level toggle from #5685 (an antd Tag in the composer that opens a 4-level menu: STRICT / SMART / AUTO / OFF), previously untested.
- **APPROVAL-001** (p0): the Tag renders and the menu shows exactly 4 levels; switching updates the Tag. **APPROVAL-002** (p0): selecting STRICT persists to `localStorage[approval_level-<chatId>]` and survives a reload. **APPROVAL-003** (p2): deleting a session clears its `approval_level-<id>` (matches `ChatSessionDrawer`'s `removeItem`-on-delete).
- Also hardens the shared `_open_session_menu`: the SessionItem `moreBtn` is a `pointer-events:none` `<span>` until its row is hovered, so in headless CI a click can land on the element behind the span and never open the antd dropdown. It now retries hovering the row and button, then falls back to a DOM `dispatchEvent('click')` (React's delegated onClick still fires) — this also stabilizes rename / pin / delete.

## Test plan
- [x] fork CI (E2E Integration, marker `tool_approval`): **3 passed**
- [x] APPROVAL-001 / 002 / 003 all green; UI-driven primary assertions
- [x] e2e-only changes (page object + tests + `tool_approval` marker); no product code touched

### DIFF
--- a/e2e/pages/chat_page.py
+++ b/e2e/pages/chat_page.py
@@ -866,43 +866,92 @@ class ChatPage(BasePage):
         The dropdown is triggered by the SparkMoreLine "more" button and
         holds Pin / Rename / Archive / Delete items. Returns True when the
         menu is visible.
+
+        The ``moreBtn`` is a ``<span>`` that is ``pointer-events:none`` until
+        the row is ``:hover``-ed, and antd opens the menu on a real click. In
+        headless CI the CSS ``:hover`` can be lost between hovering the row and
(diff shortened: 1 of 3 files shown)


lalaliat (approved):
LGTM

PR #5146, opened 2026-06-12
fix(skill): Improve skill-slash-inject and display

Leirunlin:
## Description

Use skill block as skill injection. <skill & </skill>

Remove full skill.md in display.

**Related Issue:** Fixes #5031.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [x] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [x] Ready for review

### For Channel Changes (DingTalk, Feishu, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

PR #6345, opened 2026-07-22
fix(skill): Reapply #5146 Improve skill-slash-inject and display

x1n95c:
## Description

The fix from #5146 regressed (caused by agentscope2.0 update); this PR re-applies it.

**Security Considerations:** None. Change only affects how a slash-command skill body is injected into the message and hidden from display; no auth, env, or config handling is affected.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

- `/<skill_name> <input>`: model receives the full `SKILL.md` inside a trailing `<skill>` block, while the display transcript shows only the user's typed text.
- `/<skill_name>` with no input still returns the skill description (name / command / path) and does not inject.
- Unknown or channel-disabled skill name falls through to the normal agent (no injection).

## Evidence

Examples of valid evidence:
- CI artifact link

```bash
pytest tests/unit/app/chats/test_utils.py -q
# 23 passed in 0.33s
```

## Additional Notes

Only invocations made after this change fold; old-format messages already persisted have no `<skill>` block and still render in full.

### DIFF
--- a/src/qwenpaw/app/chats/utils.py
+++ b/src/qwenpaw/app/chats/utils.py
@@ -389,7 +389,8 @@ def _build_media_message_from_block(
 
 
 # Matches the trailing <skill> block appended to a user message by
-# slash-command skill expansion (runner._maybe_inject_skill).
+# slash-command skill expansion
+# (runtime.builtin_commands._skill_fallback_handler).
 _INJECTED_SKILL_BLOCK_RE = re.compile(
     r"\s*<skill\b[^>]*>.*</skill>\s*$",
     re.DOTALL,
--- a/src/qwenpaw/runtime/builtin_commands.py
+++ b/src/qwenpaw/runtime/builtin_commands.py
@@ -524,25 +524,34 @@ def _extract_block_text(block: Any) -> str:
 def _build_skill_injection(
     original_text: str,
     display_name: str,
+    description: str,
     skill_dir: "Path",
-    user_input: str,
     skill_body: str,
 ) -> str:
-    """Keep the typed text at the head; append the skill body in a
+    """Keep the typed command at the head; append the skill body in a
     trailing ``<skill>`` block (hidden from display by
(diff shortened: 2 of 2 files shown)


Leirunlin:
One suggestion: could we keep the injected block self-describing and avoid putting frontmatter values into XML attributes?

Like:
```
  /pdf 分析pdf

  <skill>
  <name>pdf</name>
  <description>PDF processing</description>
  <dir>/path/to/skills/pdf</dir>
  <content>
  This skill has already been loaded because the user invoked it directly. Do not call the Skill tool to read it again. Follow the skill instructions
  to fulfill the user's request above. Relative paths inside the skill resolve against the directory above.

  # PDF Processing

  ...SKILL.md body...
  </content>
  </skill>
```

This has a few advantages:
  - keeps a clean and extensible `<skill>...</skill>` schema following agentscope2.0;
  - avoids attribute-specific quoting/escaping for name and dir;
  - reuses the fields already parsed from SKILL.md frontmatter;
  - makes it explicit that the full body should remain in context;
  - continues to work with the existing `strip_injected_skill_block()` regex.

And the original slash command should remain before this block, so the user input does not need to be duplicated inside <skill>.
  
  Minor: the PR title references #5156, but the original fix is #5146.

x1n95c:
Hello @Leirunlin , thanks for your suggestion, I have update the _build_skill_injection function follow the example format you provide and avoid putting frontmatter values into XML attributes.

Leirunlin:
LGTM!

lalaliat (approved):

PR #6352, opened 2026-07-22
fix(mission): avoid directory collisions

patrick-andstar:
## Description

Prevent Mission state collisions when two missions start within the same
second. Previously, `create_loop_dir()` generated a seconds-resolution name
and used `exist_ok=True`, so both workflows could receive the same directory.
Their later task, progress, and configuration writes could then overwrite each
other.

This change atomically creates the original `mission-<timestamp>` name and
retries with `-0001`, `-0002`, and so on when that name already exists. It adds
a regression test for a fixed timestamp.

**Related Issue:** N/A (discovery-backed; no matching open issue found)

**Security Considerations:** This does not introduce user-controlled paths or
new permissions. It preserves the existing workspace path and makes directory
allocation exclusive.

**Duplicate-work check:** Searched open PRs for `mission directory`, `same
second mission`, and `create_loop_dir`. GitHub code search found only the
existing state helper and its handler call site.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

## Testing

- Passed an isolated regression that fixes the timestamp and creates two
  Mission directories; it verifies the original name and the `-0001` suffix.
- Passed the same regression with 8 concurrent creators; all received a unique
  directory from the base name through `-0007`.
- Passed `python -m py_compile`, Black, Flake8, and `git diff --check` for the
  changed files.
- `python -m pre_commit run --all-files` was not runnable because `pre_commit`
  is not installed in this environment.
- The focused and full `pytest` commands were blocked during root `conftest.py`
  collection because `qwenpaw` is not installed. A direct source import also
  encounters the current tree's missing `qwenpaw._compat` module.

## Evidence

With a fixed timestamp, the first allocation preserves the base directory and the next uses `-0001`; eight concurrent creators receive eight distinct Mission directory names.

```text
isolated same-second Mission directory regression: passed
concurrent same-second Mission directory regression: passed

python -m py_compile src/qwenpaw/modes/mission/state.py tests/unit/loop/test_mission_settings.py
# passed

python -m black --check --line-length 79 src/qwenpaw/modes/mission/state.py tests/unit/loop/test_mission_settings.py
# 2 files would be left unchanged

python -m flake8 --jobs 1 --extend-ignore=E203 src/qwenpaw/modes/mission/state.py tests/unit/loop/test_mission_settings.py
# passed

git diff --check
# passed
```

## Additional Notes

The suffix is only used after a collision, so normal Mission directory names
remain unchanged. The explicit exclusive `mkdir` keeps simultaneous processes
from reusing the same state directory.

### DIFF
--- a/src/qwenpaw/modes/mission/state.py
+++ b/src/qwenpaw/modes/mission/state.py
@@ -139,10 +139,17 @@ async def detect_git_context(workspace_dir: Path) -> dict[str, Any]:
 def create_loop_dir(workspace_dir: Path) -> Path:
     """Create a new mission directory and return its path."""
     loop_id = f"mission-{_ts()}"
-    loop_dir = workspace_dir / "missions" / loop_id
-    loop_dir.mkdir(parents=True, exist_ok=True)
-    logger.info("Created mission dir: %s", loop_dir)
-    return loop_dir
+    suffix = 0
+    while True:
+        suffix_text = f"-{suffix:04d}" if suffix else ""
+        loop_dir = workspace_dir / "missions" / f"{loop_id}{suffix_text}"
+        try:
+            loop_dir.mkdir(parents=True, exist_ok=False)
+        except FileExistsError:
+            suffix += 1
+            continue
+        logger.info("Created mission dir: %s", loop_dir)
+        return loop_dir
 
 
 def write_loop_config(loop_dir: Path, config: dict[str, Any]) -> Path:
--- a/tests/unit/loop/test_mission_settings.py
+++ b/tests/unit/loop/test_mission_settings.py
@@ -13,7 +13,10 @@ from qwenpaw.modes.mission.handler import (
     start_mission,
 )
 from qwenpaw.modes.mission.prompts import build_master_prompt
-from qwenpaw.modes.mission.state import read_loop_config
+from qwenpaw.modes.mission.state import (
+    create_loop_dir,
+    read_loop_config,
+)
 
 
 def test_mission_config_defaults_and_bounds() -> None:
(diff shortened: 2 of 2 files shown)


RerankerGuo (commented):
I independently verified the implementation at exact head `38ed50392442dfb251aa8467dd1bbfce34721e0b` and resolved its current conflict semantically against `main` `bacac7410cf2ac122dd892152beff8b03f38729f` (real merge base `4acf3418b0d2a5a66726308c881a426f9e189f9b`; tested trial tree `fa51c8a66a6fb2a0a1ad855db5a2880127a04333`).

The exclusive `mkdir(..., exist_ok=False)` retry is the correct concurrency boundary. With one fixed timestamp, I ran the same allocator through 32 threads and 16 independent spawn processes:

```text
current main:
  thread_unique=1/32
  process_unique=1/16

exact head:
  thread_unique=32/32
  process_unique=16/16

current-main semantic trial:
  thread_unique=32/32
  process_unique=16/16
```

This proves the current `exist_ok=True` path aliases concurrent Mission state, while the PR allocates the complete contiguous set from the base name through `-0031` / `-0015`.

The branch cannot be merged mechanically because current main moved Mission state from `<workspace>/missions` to project-local `<project>/.qwenpaw/missions` and added project/workspace separation. The one-file semantic resolution is to retain main's path while transplanting the PR's exclusive allocation loop:

```python
loop_dir = (
    project_dir
    / ".qwenpaw"
    / "missions"
    / f"{loop_id}{suffix_text}"
)
loop_dir.mkdir(parents=True, exist_ok=False)
```

The adapted regression should also assert:

```python
assert first.parent == tmp_path / ".qwenpaw" / "missions"
```

Verification:

```text
current main complete tests/unit/loop: 63 passed
exact target file: 6 passed
exact complete tests/unit/loop: 64 passed
semantic trial target file: 6 passed
semantic trial complete tests/unit/loop: 64 passed
exact and trial applicable changed-file pre-commit hooks: passed
exact remote Real behavior proof: passed
git diff --check: passed
```

The implementation itself is sound; the remaining action is the path-aware rebase above. I am leaving this as a review comment rather than approving a currently conflicting head.

Issue #6131, opened 2026-07-15
[Bug]: 更新2.x版本后，会话列表updatedAt数据未更新


f39516046:
## QwenPaw Version

[Provide the version of QwenPaw you are using, e.g. 0.x.x or git commit hash.]
[Using `qwenpaw --version` in your command line or checking the version in the console UI can help.]

2.0.0

## Description

[Describe the bug clearly and concisely. What happened vs what you expected?]

版本更新后，会话中发送消息后，updatedAt时间应该更新为最新时间，但当前的会话列表中数据时间未更新

**Related PR(s):** #(optional)

**Security considerations:** [If applicable, e.g. auth, env/config exposure]

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** [e.g. 0.x.x or git commit]
- **OS:** [e.g. macOS 14, Ubuntu 22.04, Windows 11]
- **Install method:** [pip / one-line install / Docker / from source]
- **Python version (if applicable):** [e.g. 3.10]

## Steps to Reproduce

1. 终端或者企业微信通道，发送消息给智能体
2. 智能体正常回答响应，会话详情中，消息最新时间为当前时间
3. 点击会话，查看活跃会话列表，列表未发生变化，updatedAt字段未更新为最新时间

## Actual vs Expected

- **Actual:**
- **Expected:**

## Logs / Screenshots

[Paste relevant log output or attach screenshots. Use code blocks for logs.]

```
(paste logs here)
```

## Additional Notes

[Optional: workarounds, similar issues, etc.]

zhaozhuang521:
@f39516046 Could you describe how to reproduce this issue, or describe the issue itself?

hellofreud:
同样遇到了，飞书渠道的聊天消息，在后台居然看不到活跃会话

> [@f39516046](https://github.com/f39516046) Could you describe how to reproduce this issue, or describe the issue itself?

f39516046:
> [@f39516046](https://github.com/f39516046) Could you describe how to reproduce this issue, or describe the issue itself?

[...]

PR #6378, opened 2026-07-23
perf: prevent synchronous I/O from blocking the event loop

rayrayraykk:
## Summary

- centralize asynchronous and atomic filesystem operations in `qwenpaw.utils.io_utils`
- expose semantic async helpers so business code does not call `asyncio.to_thread` for file I/O
- preserve atomic replacement guarantees, file modes, and symlink targets
- move chat, cron, inbox, trace, session, file download, and agent lifecycle I/O off the event-loop thread
- serialize same-path append/write operations and read-modify-write transactions to avoid coroutine races
- keep path locks held until non-cancellable worker-thread I/O has actually finished
- serialize policy and session transactions across request-scoped object instances
- keep the process-wide audit connection alive until process shutdown
- collapse the Console activity timestamp update from three full reads and one write to one read and one write

## Root cause

The JSON chat repository has used synchronous file I/O since the 1.x releases. However, #6131 added `touch_chat()` to the per-message request path, turning the existing full-file read/write implementation into a high-frequency event-loop blocker.

This explains why users report that `v1.1.12.post2` did not exhibit the same performance problem: the underlying repository implementation was already synchronous, but the new high-frequency call path was not present there.

The 2.x request lifecycle also starts and stops a `ResourceGovernor` for each agent build. Policy YAML persistence, sandbox probing, Scroll retention cleanup, and context-store shutdown were performed synchronously from async request handling.

Other async APIs, including inbox persistence and remote file downloads, also called synchronous filesystem, subprocess, or network operations directly.

## Implementation

`io_utils.py` is the single module for both ordinary asynchronous file operations and atomic persistence:

- `read_text_async` and `read_bytes_async` handle standalone reads.
- `append_text_async` preserves append semantics and serializes same-path coroutine access.
- `write_bytes_async` handles unpublished generated or downloaded artifacts.
- `write_*_atomic_async` replaces shared state, configuration, JSON, and YAML without exposing partial files.
- `get_path_lock` covers complete read-modify-write transactions.
- `get_sync_path_lock` covers composed synchronous transactions executed by worker threads.
- `run_sync_io` is an explicit escape hatch for already-composed legacy I/O workflows.

The module and function docstrings describe how to select these APIs and make clear that thread offloading, locking, and atomic replacement solve different problems.

QwenPaw currently supports one application worker and one event loop. All supported writers therefore share process-local path-lock registries, so an OS-level lock such as `flock` is deliberately unnecessary. Async transactions use `get_path_lock`; synchronous transactions running in worker threads use `get_sync_path_lock`. This must be revisited if multi-process writers are introduced. The worker thread pool can still run concurrent operations, so the in-process transaction locks remain necessary.

`run_sync_io` shields its worker task from coroutine cancellation and waits for the underlying thread to finish before propagating cancellation. This prevents a caller from releasing a path or domain lock while the cancelled thread is still mutating the filesystem.

`aiofiles` is not treated as a consistency mechanism. File-tool reads now obtain one binary snapshot from a single opened handle instead of splitting `stat`, `open`, and `read` across await points. Chat, cron, and session repositories offload their complete read/parse/validate and serialize/atomic-write workflows, rather than leaving JSON or Pydantic work on the event-loop thread. Business-level read-modify-write transactions retain their domain locks, while shared append/write helpers own normalized path locks.

Session read-modify-write operations use the shared path registry rather than per-instance locks. Governance policy updates use a canonical policy-path transaction that reloads the latest file, mutates it, and atomically saves it under one process-local thread lock. A request-scoped governor no longer writes its potentially stale snapshot during `stop()`.

`AuditLog` remains process-wide and is no longer closed by each request-scoped governor. Singleton creation, connection access, and close share thread locks, and the connection is closed at process exit.

Atomic replacement preserves the destination mode for existing files. Temporary files remain owner-only (`0o600`) while content is written. New internal state defaults to `0o600`; user workspace artifacts explicitly publish as non-writable shared files (`0o644`). The helper guarantees complete-file visibility during normal operation; it does not claim power-loss durability because the parent directory is not flushed.

Console activity now calls `touch_chat_by_session`, a repository transaction that loads the chat collection once, selects the most recent matching session, updates `updated_at`, and saves once. Indexed repositories may override this default implementation with a direct update. Regression coverage asserts exactly one `load()` and one `save()`.

## Impact

Slow disk, fsync, JSON/YAML serialization, downloads, sandbox probing, and Scroll cleanup no longer stall unrelated coroutines in the single worker event loop. Existing repository and file formats are unchanged.

This PR does not add agent lifecycle caching or migrate chat persistence to SQLite. `touch_chat()` still rewrites the JSON document, but the operation no longer blocks the event loop.

## Validation

- `5142 passed, 8 skipped`
- all pre-commit hooks passed, including mypy, Black, Flake8, and Pylint
- added coverage for worker-thread offloading, event-loop progress, ordinary async file helpers, atomic replacement failure, existing and new-file modes, symlink preservation, serialized appends, cancellation-safe locking, single-snapshot reads, cross-instance session updates, cross-instance governance policy writes, and concurrent audit close/access


cuiyuebing (approved):
LGTM

PR #6387, opened 2026-07-23
feat(channels): install optional dependencies on demand

hongxicheng:
# PR Description

## Overview

This PR keeps built-in Channels in the existing QwenPaw configuration and registry system while moving Channel-specific SDKs out of the default dependency set where practical.

When an optional built-in Channel is unavailable, the Console keeps it visible but prevents it from loading or starting. Users can review the currently unsatisfied requirements, choose a package source, and install them in the background. The existing Channel configuration flow becomes available after version and import verification succeed.

DingTalk and Telegram remain default dependencies. Dependencies that are also required by non-Channel QwenPaw code remain in the core package.

image

image

image

## Dependency packaging

- Add pinned per-Channel extras for Discord, Feishu, QQ, Mattermost, Slack, MQTT, Matrix, Voice, SIP, WeCom, WeChat, and Yuanbao.
- Keep DingTalk and Telegram in the default dependency set and outside the runtime-install catalog.
- Add `channels-all` as the complete optional Channel dependency set.
- Include `channels-all` in `full`.
- Preserve the existing `sip` and `sip-livekit` extras with the verified versions.
- Keep shared non-Channel dependencies in the core package instead of duplicating them in Channel extras.

## Channel discovery and startup safety

- Add a static built-in Channel catalog containing the Channel key, module, class, optional extra, and platform restriction without importing optional SDKs.
- Keep missing built-in Channels visible by combining catalog keys with Plugin-provided Channel keys.
- Skip imports for unsupported or dependency-incomplete built-in Channels.
- Prevent an existing `enabled=true` configuration from starting a Channel when its dependencies are unavailable or incompatible.
- Keep iMessage restricted to macOS without adding artificial platform restrictions to other Channels.
- Preserve the existing required-Channel behavior for Console.

## Dependency status and Console UX

- Add `ready`, `missing`, `installing`, `failed`, `load_error`, and `platform_unsupported` states.
- Replace Start with Install when requirements are unsatisfied.
- Show only currently unsatisfied requirements for a normal install.
- Offer Reinstall with the full Channel requirement set when versions are satisfied but the Channel import fails.
- Keep installation asynchronous and poll only Channels with active jobs.
- Allow users to click an Installing item and confirm that the installation should be stopped.
- Reload configured Agents through the existing reload flow only after dependency and import verification succeed.
- Add localized UI text for the supported Console locales.

## Package sources and fallback

The install dialog offers:

- Aliyun, selected by default.
- PyPI.
- A custom HTTP(S) index.

Each task performs at most one fallback:

- Aliyun falls back to PyPI.
- PyPI falls back to Aliyun.
- A custom source falls back to Aliyun.

Network, DNS, proxy, TLS, timeout, mirror-service, and mirror-sync failures may fall back. Deterministic Python, platform, wheel, build, or dependency incompatibilities fail directly. A definitive missing distribution on PyPI does not retry another source.

Custom index credentials are passed through subprocess environment variables rather than command-line arguments and are redacted from logs, errors, API responses, and persisted job state.

`QWENPAW_RUNTIME_DEP_INSTALL=0`, `false`, or `no` disables runtime installation and automatic version repair.

### Frozen desktop

- Use the bundled standalone Python supplied by the desktop shell.
- Install into the ABI, operating-system, and architecture bucket under `channel_runtime` with `pip --target`.
- Add the Channel Runtime site directory to `sys.path` only in frozen mode.

### Source, pip, venv, Conda, script, and Docker

- Install with the actual QwenPaw process interpreter through `sys.executable`.
- Fall back to uv with the same interpreter when pip is unavailable.
- Do not read packages from the frozen Channel Runtime.
- Isolate persisted installation jobs by executable identity so source and desktop processes do not share task state.

## Background jobs and version repair

- Persist install jobs under an environment-specific `channel_runtime/environments` directory.
- Serialize job-state updates across processes.
- Mark active jobs from dead owner processes as failed instead of leaving the Console permanently in Installing.
- Serialize Channel dependency writes and fail when the install lock cannot be acquired instead of continuing unlocked.
- Stop the pip/uv process tree on cancellation or timeout.
- Cancel and await jobs owned by the current process during application shutdown.

After startup, QwenPaw checks only optional built-in Channels enabled by enabled Agents. If a dependency is already installed but does not satisfy the pinned version, it is repaired sequentially in the background with Aliyun as the primary source. Completely missing dependencies still require explicit user confirmation.

## Shared Plugin subprocess improvements

Channel installation reuses the Plugin subprocess runner. This PR updates that shared path to:

- Force UTF-8 subprocess output with replacement for invalid bytes.
- Continuously drain merged stdout and stderr.
- Preserve useful failure output.
- Redact configured secrets.
- Support cancellation and process-tree termination.
- Keep the existing Plugin uv lookup and subprocess runner reusable by Channel installation.

Requirement detection now lives in `package_runtime.py` and is called directly by both Plugin and Channel code. Post-install verification derives import candidates from distribution metadata (`top_level.txt` or `RECORD`), while the small distribution-to-import mapping is used only when bundled packages have no metadata.

The UTF-8 handling prevents Windows default-GBK decoding failures from terminating the output reader and potentially blocking pip on a full pipe. Plugin installs benefit from the same subprocess fix, but this PR does not change the existing Plugin Runtime `site` layout.

## Transactional frozen Runtime updates

Frozen Plugin and Channel installations now keep their existing ABI-bucketed
`site` paths while writing them through a lightweight targeted transaction:

- Build normal dependency state from top-level metadata only and report
  duplicate, damaged, invalid-version, or indeterminate target metadata as a
  Runtime conflict.
- Cache the computed Channel dependency-state result in process and invalidate
  it when installation, repair, cancellation, or task-status changes can
  affect the result; rebuild the Runtime snapshot after invalidation.
- Serialize each Plugin or Channel Runtime bucket with a strict global write
  lock; a lock failure aborts the install instead of continuing unlocked.
- Install only the current requirement closure into a clean staging directory.
- Determine the replacement set from distributions actually produced in
  staging.
- Validate direct requirements, discoverable consumer constraints, installed
  distributions' `Requires-Dist`, safe paths, and candidate file ownership.
- Use `RECORD` only during a replacement to protect shared namespace and shared
  files; reject differing unmanaged or unrelated-owned files before commit.
- Dynamically identify fully owned, pure-Python top-level directories and move
  them as directories within the same filesystem. Metadata, scripts, symlinks,
  native extensions, unmanaged content, and shared namespaces retain the
  affected-file path.
- Always replace target metadata even when an existing `RECORD` omits some or
  all of its own `.dist-info` / `.egg-info` contents.
- Persist file-level and directory-level operations in the manifest, complete
  the affected-file backup before marking the transaction `committing`, move
  eligible old directories into backup, and verify the result in a fresh
  bundled-Python process.
- Roll back both directory-level and file-level commits and recover interrupted
  transactions on the next Runtime access.
- Restore original files through same-directory temporary files and
  `os.replace` before deleting transaction-only files, preserving retryable
  state when copy, replace, or unlink fails.
- Defer read-side recovery when an active installer owns the Runtime lock so
  Channel status requests do not wait for the package download.
- Keep the bundled-Python verification child standard-library-only; perform
  requirement parsing and version checks in the backend.
- Offload Plugin recovery/dependency inspection and Channel registry prewarm
  from the asyncio event loop without constructing Channel instances in a
  worker thread.

This repairs duplicate metadata and trustworthy obsolete files for only the
distributions involved in the current install. It does not rebuild, redownload,
or globally repair unrelated Runtime packages. Non-frozen pip, venv, Conda,
source, script, and Docker installs retain the previous behavior.

## API changes

```text
GET  /config/channels/dependencies
POST /config/channels/{channel}/dependencies/install
POST /config/channels/{channel}/dependencies/install/cancel
POST /config/channels/{channel}/dependencies/recheck
```

These endpoints expose dependency status, start installs or reinstalls, request cancellation, and recheck active jobs.

## Packaging, CI, and documentation

- Discover Channel modules from the source tree during PyInstaller analysis instead of importing optional SDKs.
- Install `channels-all` in E2E and Channel-specific test paths that require optional SDKs without installing `full`.
- Document on-demand installation, package sources, fallback behavior, automatic version repair, and CLI extras in the English and Chinese Channel documentation.
- Add backend tests for dependency grouping, status detection, environment isolation, fallback, credential redaction, job persistence, locking, cancellation, automatic repair, shutdown, reinstall, and import verification.
- Add router and Console tests for the new APIs and UI states.
- Add shared Plugin runner tests for UTF-8 decoding, redaction, cancellation, and failure output.

## Compatibility

- The `agent.json` Channel configuration format is unchanged.
- Built-in Channels remain built-in; this PR does not convert them into external Plugins.
- DingTalk and Telegram remain available from the default QwenPaw installation.
- Existing enabled configurations remain safe when optional dependencies are missing.
- Source and environment-based installations continue using the current interpreter.
- Frozen desktop installations keep the existing Plugin and Channel `site`
  paths, but stage and commit affected distributions transactionally.
- Existing Plugin Runtime directories require no migration.
- Installed Channel dependencies are not automatically uninstalled.
- `qwenpaw[channels-all]` and `qwenpaw[full]` remain available for preinstalling all optional Channel dependencies.

## Review focus

1. Core versus optional dependency boundaries, especially DingTalk, Telegram, and packages shared with core modules.
2. Exact pins and consistency between per-Channel extras, `channels-all`, `full`, and legacy SIP extras.
3. Catalog and Registry behavior when a built-in Channel is visible but unavailable.
4. Frozen versus non-frozen interpreter, Runtime, `sys.path`, and job-state isolation.
5. Manual installation, load-error reinstall, and automatic version-repair boundaries.
6. Package-source fallback classification and custom-source credential handling.
7. Cross-process locking, stale-job handling, cancellation, shutdown, and Agent reload timing.
8. Windows UTF-8 subprocess output and process-tree termination.
9. PyInstaller and CI behavior in an environment without optional Channel SDKs.
10. Runtime metadata snapshot and canonical-name conflict detection.
11. `RECORD` ownership, namespace/unmanaged conflicts, and path containment.
12. Exclusive-directory ownership checks, directory moves, and file-level
    fallback for shared, unmanaged, symlinked, or native-extension paths.
13. Strict Runtime locking and the
    prepared/file-backup/committing/directory-move/verify/recovery sequence.
14. Regression isolation between frozen transactions and non-frozen installs.

## Known limitations and out of scope

Plugin and Channel Runtime directories remain separate and may contain duplicate
packages. Native extensions already loaded in the backend may require an
application restart after installation or update; locked Windows files cause a
safe rollback.

When old target metadata has no trustworthy `RECORD`, the transaction does not
guess which unknown source files should be removed. A differing unmanaged
destination file aborts the commit rather than being overwritten. Only
distributions involved in the current staging install are repaired; unrelated
historical Runtime damage remains untouched.

## Testing

- `40 passed` in the complete Runtime unit suite after the directory fast path.
- `51 passed` in the targeted Runtime and Channel transaction suites.
- All configured pre-commit checks passed for the changed Runtime and test
  files.
- The final complete-branch result should be taken from the latest CI run before
  merge.
- Real Windows and macOS frozen artifact smoke tests remain to be run in their
  corresponding build environments.

## Type of change

- [x] Feature
- [x] Bug fix
- [x] Refactoring
- [x] Documentation
- [x] CI / packaging
- [ ] Breaking change


pyproject.toml
```diff
@@ -131,6 +207,8 @@ whisper = [
 full = [
```
rayrayraykk, line 192: `full` should include `channels-all`

PR #6399, opened 2026-07-23
feat: add reranker UI config panel to ReMeLightMemoryCard

lecheng2018:
## Description

Add reranker configuration UI in the ReMeLightMemoryCard component under Agent Config. This provides a visual configuration panel for the reranker backend feature (complements the reranker backend PR).

**UI includes:**
- Collapsible "Reranker Settings" panel in the memory config card
- Enable/disable toggle switch
- Base URL, model name, and API key input fields
- Candidate multiplier and timeout number inputs
- Info alert explaining the over-fetch-then-rerank flow

All panel fields share the same enable/disable gate on the reranker toggle (`disabled={!rerankerEnabled}`) - there is no extra gate on the base URL or model name. Base URL and model name are additionally required and validated on save when reranking is enabled; candidate multiplier and timeout are always required because the backend validates them even while reranking is off.

i18n keys added to all 7 locales: zh.json, en.json, id.json, ja.json, pt-BR.json, ru.json, vi.json

## Type of Change

- [x] New feature

## Component(s) Affected

## Checklist

- [x] Ready for review

## Evidence

8 files changed, 291 insertions(+)

Console UI changes only.


jinliyl:
Thanks for the PR! Let’s wait until #5692 is resolved before proceeding with this frontend change.

RerankerGuo (changes requested):
I found two blockers at exact head `5e78b36ac28b9e68a18841d75dee4ea64efb0e0c`.

1. **The PR is currently mergeable without its backend dependency, so every reranker setting is silently discarded.** Both this exact head and a clean current-main transplant (`ad0dcbc9`, tree `46ff09e0`) still use a `ReMeLightMemoryConfig` whose Pydantic model has `extra="ignore"` and no `reranker_config` field. A production-schema round trip gives:

   ```text
   exact #6399:
   reranker_in_dump=False
   dumped_reranker=None

   current main + #6399:
   reranker_in_dump=False
   dumped_reranker=None
   ```

   Only the combined current-main + approved #6398 + #6399 tree (`17a730ac`, tree `2ea8c35a`) preserves the submitted values:

   ```text
   reranker_in_dump=True
   dumped_reranker={
     enabled: True,
     base_url: "https://rerank.test/v1",
     model_name: "rerank-model",
     ...
   }
   ```

   Please make the dependency enforceable: merge/rebase on #6398 (or include the required backend schema) before this can merge. The Console contract should also be updated in `console/src/api/types/agent.ts`; `ReMeLightMemoryConfig` currently has no `reranker_config`, so the checked TypeScript API shape remains stale even on the combined tree.

2. **The UI accepts an enabled reranker with no endpoint or model, while production treats that state as disabled/fallback.** The `enabled` switch is independent, but `base_url` and `model_name` have no conditional validation rules. An isolated real Ant Form render on the combined tree sets:

   ```text
   enabled=true
   base_url=""
   model_name=""
   candidate_multiplier=3
   timeout=10
   ```

   and targeted `validateFields()` for Base URL and Model resolves successfully. In #6398, `_get_reranker_config()` returns `None` when the model is blank; with a model but no URL, `_call_reranker_api()` immediately returns `None`. The Console can therefore display "enabled" while reranking never runs.

   Please require at least Base URL and Model Name when `enabled` is true (or derive/disable the switch until the required fields are complete), and add a rendered form regression covering enabled/disabled validation. The existing `ReMeLightMemoryCard.test.tsx` still has seven cron/embedding helper tests and does not exercise any of the new fields.

Additional validation:

- Existing card tests pass `7/7` on exact, clean current-main transplant, and combined trees.
- `tsc -b --noEmit` and Prettier pass on all three trees.
- All seven locale files contain the same 21 reranker keys.
- `git diff --check` passes on the clean combined tree.

The visual field set is otherwise coherent, but it needs an actual persisted backend contract and validation that matches runtime enablement before it is safe to merge.

jinliyl:
resolve conflicts

lecheng2018:
Hi @jinliyl, thanks for the note! Just to update you: PR #6398 (the redesigned reranker backend that supersedes #5692) has already been merged into `main` and released as part of v2.1.0 on August 13. The old #5692 can now be closed as superseded.

This frontend PR (#6399) has also been rebased onto the latest upstream/main and all the fixes from the recent audit are in place. It is ready for your review whenever you have a moment. Happy to close #5692 if that helps clean things up!

jinliyl:
Thank you for the updates. I reviewed the current head (`ab745de0`) carefully, including the form behavior, the backend contract, and compatibility with the latest `upstream/main`.

The two blockers raised in the earlier review now appear to be resolved:

- `reranker_config` is present in both the backend schema and the Console TypeScript contract, so the submitted values survive the configuration round trip.
- Base URL and model name are conditionally required when reranking is enabled, and the rendered-form regression tests cover the enabled and disabled cases.

The overall integration now looks sound. I found a few remaining items that would be worth addressing before merge:

### 1. Enforce an integer candidate multiplier

The backend defines `candidate_multiplier` as an integer with `ge=1`, but the frontend currently only applies a numeric `min: 1` rule. `InputNumber` with `step={1}` still permits a user to type a fractional value manually.

I verified the mismatch with the real form and backend schema:

- The frontend accepts `candidate_multiplier = 1.5` and `validateFields()` succeeds.
- `RerankerConfig(candidate_multiplier=1.5)` is rejected by Pydantic with `int_from_float`.

As a result, a value that appears valid in the UI can cause the save request to fail with a backend validation error. Please add an integer validator and/or `precision={0}` to the input, together with a regression test for a fractional value.

### 2. Clear conditional errors when reranking is disabled

There is a small validation-state issue when toggling the feature off:

1. Enable reranking with an empty Base URL and model name.
2. Trigger validation so both required errors are displayed.
3. Disable reranking again.

The fields become disabled, but their existing red validation errors remain visible. I confirmed with a rendered-form probe that the errors are still present after the component updates. A later full save validation clears them, so this does not permanently block saving, but the intermediate UI incorrectly suggests that the disabled configuration is still invalid.

The switch handler currently revalidates only when `checked` is true. When it becomes false, please clear or revalidate the Base URL and model-name fields so their obsolete errors disappear. A test for the enable-error-disable sequence would be useful.

### 3. Complete the Portuguese localization

In `pt-BR.json`, 23 of the 24 new reranker strings are still identical to the English strings. This leaves almost the entire new section untranslated for Portuguese users. Please translate these entries or coordinate with the localization owner if translations are handled separately.

### 4. Simplify the Chinese section title

I suggest changing:

> 重排序（Reranker）设置

to:

> 重排序设置

The current title is long and repeats terminology that is already clear from the subsection, field descriptions, and the `reranker-search` capability label. If retaining the English term is useful for discoverability, it could be introduced once in the explanatory text as “重排序模型（Reranker）” instead of placing it in the main heading.

### 5. Make the reranker details collapsible by default

The new section adds five detail fields and makes the right-hand configuration panel significantly longer, while reranking is disabled by default and will not be configured by many users. A collapsed presentation would improve the information hierarchy. The existing translation key name, `rerankerConfigCollapseLabel`, also suggests that a collapsible design may originally have been intended.

I recommend keeping the section heading, short description, enabled state, and enable switch visible, while making the detailed fields collapsible:

- When reranking is disabled, details should be collapsed by default.
- When an existing saved configuration is enabled, details should be expanded by default.
- Enabling reranking should automatically expand the details.
- A save-time validation error should expand the section so the invalid fields are visible.
- The user should still be able to expand a disabled configuration to inspect previously saved values.
- Optionally, the collapsed header could show a concise status such as “Disabled” or “Enabled · BAAI/bge-reranker-v2-m3”.

I would not hide the enable switch inside the collapsed content, since that would add an unnecessary click before the feature can be enabled.

### Validation performed

- Checked out the exact PR head: `ab745de0`.
- Targeted `ReMeLightMemoryCard` tests on the PR branch: **27/27 passed**.
- Merged the PR into the latest `upstream/main` in an isolated worktree with no conflicts.
- Targeted tests on that merged tree: **29/29 passed**, including the newer embedding verification tests from `main`.
- TypeScript compilation passed.
- Prettier checks passed.
- `git diff --check` passed.
- The direct switch-on validation behavior works; I did not reproduce an event-ordering problem there.

In summary, the backend integration and the main enabled/disabled validation flow are now in good shape. The integer mismatch is the main functional issue remaining. The stale validation state, incomplete Portuguese localization, shorter Chinese title, and default-collapsed design would improve the quality and usability of the final UI. Thank you again for iterating on this PR.

lecheng2018:
Hi @jinliyl, thank you for the thorough review. All five items have been addressed in commit `66e949cf`. Here is the summary:

**1. Integer candidate multiplier** ✅
Added `precision={0}` to `InputNumber` and a custom `validator` that rejects any non-integer value with the new locale key `rerankerCandidateMultiplierInteger`. A fractional value (e.g. 1.5) now fails validation before it can reach the backend.

**2. Clear conditional errors on disable** ✅
The Switch `onChange` handler now calls `form.setFields` to clear the `base_url` and `model_name` error fields when `checked` becomes `false`, so the intermediate UI no longer shows stale red errors for a disabled configuration.

**3. Portuguese localization** ✅
All 24 reranker strings in `pt-BR.json` are now translated from English. I added my own translations; if your team has a localization owner who would prefer to review or refine them, I am happy to defer to their version.

**4. Simplified Chinese title** ✅
Changed from `重排序（Reranker）设置` to `重排序设置` per your suggestion.

**5. Collapsible by default** ✅
Added a `rerankerExpanded` state and a chevron toggle button on the section header. The details are collapsed when reranking is disabled, expand automatically when the switch is toggled on, and preserve the user's manual expand/collapse choice. The toggle button uses `aria-expanded` and `aria-controls` for accessibility.

Three new regression tests were added to cover:
- Fractional `candidate_multiplier` rejection
- Error clearing on enable→disable toggle
- Collapsible behavior (initial collapse, auto-expand on enable, manual collapse/expand)

All tests pass: 30/30 in `ReMeLightMemoryCard.test.tsx` and 1897/1897 across the full console suite. The PR is mergeable with the latest `upstream/main`.

lecheng2018:
Hi @jinliyl, just a quick update — the CI `format:check` was failing due to a TS6133 error in `ReMeLightMemoryCard.test.tsx(993)`: the collapsible test declared `const form = formRef.current!` but never used it.

Fixed in commit `e56d11b0`: removed the unused `form` variable declaration. All 30 tests in `ReMeLightMemoryCard.test.tsx` still pass, and the only remaining `tsc` errors are the upstream `AuxiliaryModelModal.tsx` issues (already known on `main`, not part of this PR).

The branch is now clean and mergeable.

jinliyl:
Thanks for the latest update. I reviewed the current head (`f808b2e3`) and found two remaining merge issues plus one accessibility improvement:

### 1. Collapsing the enabled section bypasses validation

The detail fields are conditionally mounted under `rerankerExpanded && (...)`. When the section is collapsed, their `Form.Item` instances and validation rules are unregistered.

This allows the following flow:

1. Enable reranking.
2. Clear Base URL or Model Name.
3. Manually collapse the details.
4. Save the form.

`handleSave()` calls `validateFields()`, which only validates currently registered fields, while `getFieldsValue(true)` retains the values of unmounted fields. The form can therefore submit `enabled: true` with an empty Base URL/model. The backend schema accepts those empty strings, and runtime behavior falls back to not using the reranker.

Please keep the detail `Form.Item` components mounted while hiding only their presentation, and expand the section when save-time validation finds an error. A regression test for “clear a required value → collapse → full-form validation still fails” would cover this case.

### 2. Console formatting CI is still failing

The latest `console format check` fails on `src/pages/Agent/Config/index.module.less`. Prettier expects the new transition declaration to be formatted as:

```less
transition:
  background 0.2s,
  color 0.2s;
```

I reproduced this locally with the repository's Prettier 3.0.0.

### 3. Give the collapse button an accessible name

The toggle button contains only a `ChevronRight` marked with `aria-hidden="true"`, so it currently has no accessible name. Please add a localized `aria-label`, ideally reflecting the current action (expand/collapse reranker settings).

Other checks look good: the backend/TypeScript contract is aligned, the PR merges cleanly with the latest `upstream/main`, and the targeted `ReMeLightMemoryCard` suite passes 30/30.

lecheng2018:
Hi @jinliyl, thanks for the careful review. All three items are addressed in commit `7f485236`.

**1. Collapsing the enabled section bypasses validation** ✅
The detail `Form.Item` components are no longer conditionally mounted under `rerankerExpanded && (...)`. They now stay mounted at all times and are hidden only via `display: none` on the `#reranker-details` wrapper, so `validateFields()` always validates `base_url` / `model_name` regardless of collapse state. To surface errors, the parent form's `onFieldsChange` auto-expands the section whenever a validation error lands on either required field while reranking is enabled. Added a regression test for `clear a required value → collapse → full-form validation still fails` (and updated the existing expand/collapse test to assert visibility rather than DOM presence).

**2. Console formatting CI** ✅
The `transition` declaration in `src/pages/Agent/Config/index.module.less` is now formatted exactly as Prettier 3.0.0 expects (multiline `background 0.2s, color 0.2s`). The console format check now passes.

**3. Accessible name for the collapse button** ✅
The toggle button now carries a localized `aria-label` reflecting the current action — `rerankerExpandDetails` / `rerankerCollapseDetails` — added to all seven locale files (en, zh, id, ja, pt-BR, ru, vi).

I also rebased the branch onto the latest `upstream/main` (this pulled in #7227, which removed the now-unused `configRevision` from the memory maintenance context; the rebase resolves cleanly and the context is aligned with the newer `main`). The PR now merges cleanly and `gh` reports it as `MERGEABLE`.

Validation: `ReMeLightMemoryCard.test.tsx` 33/33, full console suite 2428/2428, `tsc -b --noEmit` (only the known upstream `AuxiliaryModelModal.tsx` errors remain, outside this PR), Prettier and `git diff --check` pass.

jinliyl:
Hi @lecheng2018, thank you for the careful iteration — the previously reported issues are now largely resolved. I re-checked the current head (`bcd39052`), including the rendered Ant Form behavior.

The backend/TypeScript contract, conditional Base URL and model validation, integer multiplier validation, stale-error clearing, mounted-field validation while collapsed, save-time auto-expansion, localization, formatting, and accessible collapse label all look good now.

Apart from the already-known conflict with the latest `main`, I found two small follow-up issues:

### P1: Invalid numeric values still block saving after reranking is disabled

The validation rules for `candidate_multiplier` and `timeout` remain unconditional. This produces the following flow:

1. Enable reranking.
2. Clear either numeric field, or enter an invalid value.
3. Disable reranking.
4. Save the Agent configuration.

Although the inputs are now disabled, `validateFields()` still rejects them. I reproduced this with the rendered form: the disabled fields report `rerankerCandidateMultiplierRequired` and `rerankerTimeoutRequired`.

The disable handler currently clears only the Base URL and model-name errors. Please also normalize invalid/empty numeric values to valid defaults (for example, 3 and 10) and clear their errors when disabling. Merely making their frontend rules conditional would not be sufficient if `null` is still submitted, because the backend schema validates these fields even when `enabled` is false.

A regression test for “clear numeric value → disable reranker → full-form validation succeeds” would cover this.

### P2: The expansion state is not reset when the configuration changes

The current effect expands the details when reranking becomes enabled, but it never collapses them when reranking is disabled. The page-level `rerankerExpanded` state can also carry across Reset or Agent changes.

As a result:

- disabling reranking leaves the details expanded;
- a disabled Agent can inherit the expanded state from the previously selected Agent;
- two enabled Agents can share the previous Agent's manual collapse state.

The existing test named “expands reranker details when enabled and collapses when disabled” does not actually toggle the enable switch off; it tests manual collapse instead. Please reset the default expansion state on the relevant enabled/config/Agent transition and update the test to exercise the disable path.

### Validation performed

- Targeted `ReMeLightMemoryCard` suite: **33/33 passed**
- TypeScript compilation: passed
- Prettier: passed
- `git diff --check`: passed
- A separate rendered-form probe confirmed that full-form validation of an already-invalid collapsed field does auto-expand the details

Thanks again — the main implementation is in good shape, and these remaining items should be relatively small to address.

lecheng2018:
Hi @jinliyl, thank you for the careful review. Both follow-up items are addressed in commit `e6d677de` (rebase of `6d413947` onto the latest `main`, `d3efdf2e`).

**P1: Invalid numeric values no longer block saving after disabling reranking** ✅
The disable handler now calls a `normalizeRerankerNumbersOnDisable()` helper that:
- validates the current `candidate_multiplier` and `timeout` values, and
- rewrites invalid/empty values to the valid defaults `3` and `10`, clearing their field errors at the same time.

Because the backend schema validates these fields even when `enabled` is `false`, the handler normalizes before submit so `null` is never sent. A new regression test covers the exact flow you described: clear the numeric field → disable reranking → full-form `validateFields()` succeeds.

**P2: Expansion state is now reset on enable/disable and Agent transitions** ✅
The `rerankerExpanded` effect now sets `setRerankerExpanded(rerankerEnabled)` and depends on both `rerankerEnabled` and `selectedAgent`, so:
- disabling reranking collapses the details;
- switching Agents (or Reset) restores the default expansion state instead of inheriting the previous Agent's state;
- two enabled Agents no longer share a manually collapsed state.

The existing "expands when enabled and collapses when disabled" test now truly toggles the enable switch off (it previously only exercised manual collapse), and a new Agent-switch reset test covers the cross-Agent leak.

### Verification (head `e6d677de`, on top of `d3efdf2e`)
- Targeted `ReMeLightMemoryCard` suite: **36/36 passed** (Config dir 85/85)
- Full test suite: **3084/3084 passed**
- TypeScript compilation: passed
- Prettier: passed
- `git diff --check`: passed
- PR is `mergeable: true` against the latest `main`

Thanks again — this should wrap up the remaining items.

jinliyl:
Thanks for the update! Please resolve the merge conflicts.

lecheng2018:
Hi @jinliyl — update on the latest push (`57fe1bb5`).

**Conflict resolved.** The branch is rebased onto the latest `main` (`e2a32455`). The only conflict was in `console/src/pages/Agent/Config/index.module.less`: the theme-token refactor (#7487) deleted the legacy `:global(.dark-mode)` override block that my Section 04 rules lived in. Resolution:

- dropped the removed dark-mode block entirely — dark mode is now handled by `tokens.css`
- moved Section 04 onto the token system: `.memorySectionIconQuaternary` uses `--app-warning-border/-bg/-text` + `--app-shadow-sm`, so the four icons read accent → info → success → warning and stay consistent with #7487
- `.memoryRerankerToggle` uses `--app-text-tertiary` / `--app-fill-subtle` / `--app-fill-strong`

Heads up: this changes Section 04's colour from the earlier custom pink to amber (warning token) — there is no pink family in the token system. Happy to add one to `tokens.css` if you'd rather keep the original look.

**One CI job was red, and I corrected myself here:** `Vitest Unit Tests` failed on `SessionProjectDirectory.test.tsx:576` (`Found multiple elements with the text: /projects/alpha`). I first assumed it was flaky, but that was wrong — upstream `40724726` and `3a2efe33` fail on the *same line* with the *same error*, so it is a deterministic upstream break, not a flake, and a rerun would not clear it.

Root cause: the assertion used `getByText("/projects/alpha")`, which matches both the bound chip (`<small title="/projects/alpha">`) and the tooltip popped over the trigger button that shows the same path (`SessionProjectDirectory.tsx:1088`). Line 493 in the same file already handles this with `getAllByText`; `getByText` at 576 was missed.

I fixed it in a separate commit, `57fe1bb5`, so it can be reviewed/reverted independently:

```diff
-    expect(screen.getByText("/projects/alpha")).toBeInTheDocument();
-    expect(screen.getByText("/projects/beta")).toBeInTheDocument();
+    expect(screen.getByTitle("/projects/alpha")).toBeInTheDocument();
+    expect(screen.getByTitle("/projects/beta")).toBeInTheDocument();
```

Nothing to do with the reranker UI. Say the word if you'd rather I drop it and you fix it upstream.

**Verification at `57fe1bb5`**
- `tsc -b --noEmit` → clean
- `prettier --check` on all changed files → clean
- Config page suite: 9 files / 92 tests pass; `ReMeLightMemoryCard` 43/43
- full console suite: 336 files / 3355 tests pass
- `SessionProjectDirectory.test.tsx` 20/20

jinliyl:
Thank you for all the work and careful iteration on this PR. I reviewed the latest head (`57fe1bb5`) closely, and the previously raised functional issues are now essentially resolved. The reranker form behavior, validation, backend contract, localization, and regression coverage all look solid.

I found only two remaining items:

1. **The PR currently conflicts with the latest `main`.** The conflict is in `SessionProjectDirectory.test.tsx`. The functional change from the final commit (`getByTitle` instead of `getByText`) has already landed on `main`; only the surrounding comment differs. Rebasing and dropping the now-redundant `57fe1bb5` commit should resolve this cleanly. I also verified that the PR changes excluding that final commit merge cleanly with the latest `main`.

2. **Reset does not always restore the reranker section's default expansion state.** The synchronization effect depends only on `rerankerEnabled` and `selectedAgent`. If the persisted configuration is already enabled, manually collapsing the section and then clicking Reset reloads the same `enabled: true` value, so the effect does not run and the section remains collapsed. This is a minor UX issue rather than a configuration correctness problem, but it does not quite match the comment that Reset restores the expansion state. A config-load revision/signal or an explicit synchronization after loading would address it.

For reference, the targeted tests pass (63/63 across the two changed test files), TypeScript and Prettier pass, ESLint reports no errors, and the locale key sets are consistent across all seven locales.

Thanks again for your contribution and patience with the review iterations. The implementation is in good shape now. Please help address these final two items when you have a chance, and thank you in advance for the additional fix.

lecheng2018:
Hi @jinliyl — thank you for the careful read-through. Both items are addressed in the new head `830a46c1`.

**1. Rebased onto `main` (`79146f18`), dropping `57fe1bb5`.**

Upstream `276bc440` already carries the `getByTitle` change, so the extra commit was redundant and its only remaining diff was the surrounding comment. The three reranker commits were rebased directly onto `79146f18`, so the `SessionProjectDirectory.test.tsx` hunk is gone entirely (`git diff --name-only main..HEAD` no longer lists that file). No conflicts this round — the 7 locale files shared with upstream merged cleanly.

**2. Reset now restores the default expansion state.**

You diagnosed it exactly: the effect only ran on `rerankerEnabled` / `selectedAgent`, so reloading an already-enabled config was a no-op and a manually collapsed section stayed collapsed. I added the load-revision signal you suggested:

- `useAgentConfig.fetchConfig` bumps a monotonic `configLoadRevision` after every successful load
- it is surfaced through `MemoryMaintenanceContext` alongside `rerankerExpanded`
- the section treats it as a third sync trigger: `[rerankerEnabled, selectedAgent, configLoadRevision]`

Added a regression test (`restores reranker expansion after a reset reloads the same config`): manually collapse, advance the revision with the same `enabled: true` config, assert the section re-expands. I also confirmed it fails with the revision removed from the deps, so it genuinely pins the behaviour.

**Verification at `830a46c1`**

- `tsc --noEmit` clean (exit 0)
- prettier clean on the 5 touched files
- full console suite: 350 files / 3450 tests passing
- reranker card + config hook: 63 tests passing
- PR is mergeable again (`mergeable: true`, base `79146f18`)

Thanks again for catching the expansion edge case — it was exactly the kind of thing a passing test suite would not have surfaced.

jinliyl:
Hi @lecheng2018, thank you for the quick and careful follow-up. I reviewed the new head (`830a46c1`). The rebase cleanup looks correct, and the `configLoadRevision` flow now restores the default expansion state when Reset reloads the same enabled configuration. The stale-request guard in `fetchConfig` also ensures that only the current successful load advances the revision.

I found one remaining validation-visibility edge case in the collapsed reranker section:

### Hidden numeric validation errors do not expand the section

The save-time expansion logic in `AgentConfigPage` currently recognizes errors only for `base_url` and `model_name`:

```ts
(field.name[2] === "base_url" ||
  field.name[2] === "model_name") &&
field.errors?.length
```

However, `candidate_multiplier` and `timeout` are also mounted inside `#reranker-details` and have validation rules that can reject the full-form save.

This leaves the following flow:

1. Enable reranking with a valid Base URL and Model Name.
2. Clear `candidate_multiplier` or `timeout`, or otherwise make one invalid.
3. Manually collapse the reranker details.
4. Click Save.

`validateFields()` correctly rejects the save, but the section remains collapsed because the error is not one of the two fields checked by `onFieldsChange`. `handleSave()` then returns for the Ant Form validation error, leaving the user with no visible indication of why Save did not proceed.

Could you please include `candidate_multiplier` and `timeout` in the expansion condition? More generally, expanding for any invalid field inside the hidden `reranker_config` details would make this robust if more fields are added later.

A regression test along these lines would cover the behavior:

```text
enable reranker → invalidate candidate_multiplier or timeout
→ collapse details → run full-form validation
→ assert validation fails and the details become visible
```

The existing collapsed-validation test confirms that hidden fields remain registered, but it does not assert that the section becomes visible after validation, and it covers only Base URL/Model Name.

Aside from this remaining visibility issue, the previously reported items look resolved and the implementation is in good shape. Thank you again for the thorough iteration.

lecheng2018:
Hi @jinliyl — thank you for the careful review. The visibility edge case is fixed in `85d4c098`.

I went with the general form you suggested rather than adding `candidate_multiplier` and `timeout` to the list: the check now matches any field under `reme_light_memory_config.reranker_config` that has errors, excluding only `enabled` — the one reranker field rendered outside `#reranker-details`, so an error there must never expand the section. Newly added detail fields are covered without touching the condition.

The logic moved out of the page into `console/src/pages/Agent/Config/rerankerVisibility.ts` (`isRerankerDetailFieldError`, `rerankerDetailsHaveErrors`, `handleRerankerFieldsChange`), which is also why it became unit-testable: `AgentConfigPage` has no test harness. The page's inline handler is now a one-liner.

Coverage:

- 13 handler cases — every detail leaf, the `enabled` exclusion, empty/missing `errors`, non-reranker fields, and the disabled short-circuit.
- The regression you described, driven through the shared handler wiring rather than a copy of it: enable reranking, clear a hidden numeric field, collapse the details, run full-form validation, then assert both that validation fails and that the details become visible.

One mechanic worth flagging in case it saves someone time later: this works because `validateFields()` itself fires `onFieldsChange`. rc-field-form's `triggerOnFieldsChange` runs with the fresh errors once validation settles, so no change to `handleSave` was needed.

CI is green (`Vitest Unit Tests`, `Pre-commit`, `console format check`, CodeQL ts/python, Playwright smoke); only `AI Review Approval` is still waiting.

lecheng2018:
Small add-on while auditing the change: `ae1553c6` pins `RERANKER_ENABLED_PATH` with an explicit assertion.

The handler resolves the reranker switch through that constant at runtime, but the handler tests stubbed the form by position (`name[2]`), so if the path string were ever edited wrong the section would silently stop expanding while all 13 existing cases still passed. Verified the assertion has teeth — changing the constant makes it plus two handler expansion cases fail.

Tests only, no behavior change. Full suite: 351 files / 3465 tests green.

jinliyl:
Thanks for the updates and for addressing the previous feedback. Overall, this looks good now.

There is just one small issue left: when the reranker switch is turned on, the immediate validation at `ReMeLightMemoryCard.tsx:934` runs before the Form has updated the `enabled` value. As a result, the Base URL and Model Name fields do not show validation errors immediately after enabling the reranker. The current test then calls `validateFields()` again explicitly, so it does not catch this behavior.

Could you please trigger the validation after the form value has been updated and add a regression assertion that checks the field errors directly after toggling the switch on? Thanks for the hard work!

lecheng2018:
Hi @jinliyl — good catch, that was a real ordering bug. Both items are addressed in `a129ff22`.

**Root cause.** The `Switch` `onChange` validated synchronously, but at that moment the form still held `enabled: false`, so the conditional rules (`required: rerankerEnabled`) were still inactive for the render being validated against. The call therefore resolved successfully and no error was ever set. The test then re-called `validateFields()` afterwards — by which time the render carrying `required: true` had already committed — so it masked the behavior.

**Fix.** The validation now lives in a `useEffect` keyed on `rerankerEnabled`, so it runs after the render that carries the new `enabled` value and its re-created rules:

```tsx
useEffect(() => {
  if (!rerankerEnabled) return;
  void form
    .validateFields([rerankerPath("base_url"), rerankerPath("model_name")])
    .catch(() => {
      // Rejection is the point: it records the field errors.
    });
}, [rerankerEnabled]);
```

The stale synchronous `validateFields()` in the `onChange` is removed (`setRerankerExpanded(true)` stays). As a side effect the effect also covers the other paths where `enabled` becomes true — an agent switch or a config reload that lands on a reranker-enabled config — which previously never validated at all.

**Regression assertion.** The toggle test no longer calls `validateFields()`; it reads `form.getFieldsError(...)` inside `waitFor` and asserts both fields carry an error:

```tsx
await waitFor(() => {
  const errors = form.getFieldsError([
    ["reme_light_memory_config", "reranker_config", "base_url"],
    ["reme_light_memory_config", "reranker_config", "model_name"],
  ]);
  expect(errors[0].errors.length).toBeGreaterThan(0);
  expect(errors[1].errors.length).toBeGreaterThan(0);
});
```

I verified the test has teeth: it fails against `05fe4099` (1 failed / 44 skipped) and passes with `a129ff22` (103/103 across the whole Agent/Config suite). `tsc -b --noEmit` is clean, Prettier is clean, and ESLint reports 0 errors (the single `react-refresh/only-export-components` warning is pre-existing on `isValidDreamCronShape`).

Thanks for the careful read — that one would otherwise have shipped silently.

jinliyl:
LGTM overall. I noticed one small discrepancy between the PR description and the implementation: the description says the candidate multiplier and timeout inputs are disabled until the Base URL and model name are set, but the UI enables them as soon as reranking is turned on. Since the required fields are validated on save, this does not appear to be a functional issue. Could you please update the PR description to reflect the current behavior? Thanks!

lecheng2018:
Done - the description is corrected. You're right: every field in the panel is gated by the same `disabled={!rerankerEnabled}`, so candidate multiplier and timeout enable as soon as the switch is turned on; there was never a base URL / model name gate.

The bullet now just reads "Candidate multiplier and timeout number inputs", and I added a note stating that the panel shares a single gate on the toggle, plus the actual difference in the required rules: base URL and model name are `required: rerankerEnabled` (validated on save), while candidate multiplier and timeout are always `required: true` because the backend validates them even while reranking is off.

Description-only edit - no code changes. Current head `1d7dbd8d`, CI 20 success / 4 skipped / 0 failed (only `AI Review Approval` pending), and the Agent/Config suite is 103/103 green locally.

One other thing I spotted while in there and left alone as out of scope: the Evidence section still says "8 files changed, 291 insertions(+)", which is now 16 files, +1408 / -10. Happy to refresh it if you'd like.

cuiyuebing (approved):
LGTM

PR #6409, opened 2026-07-23
fix(local-models): ignore non-object tool call json

AaronZ345:
## What Problem This Solves

Local model output can contain malformed `<tool_call>` blocks. When the JSON payload parses successfully but is not an object, such as `[1]`, the parser currently treats it like a mapping and can raise `AttributeError` while handling model text.

This change makes the parser ignore non-object JSON tool calls instead of crashing, matching the existing behavior for other malformed tool-call payloads.

## Evidence

- Added a unit test covering `<tool_call>[1]</tool_call>`.
- Verified the local parser test suite passes:

```bash
PYTHONPATH=src python -m pytest tests/unit/local_models/test_tag_parser.py -q
```

Result: `26 passed`.

### DIFF
--- a/src/qwenpaw/local_models/tag_parser.py
+++ b/src/qwenpaw/local_models/tag_parser.py
@@ -234,6 +234,13 @@ def _parse_single_tool_call(raw_text: str) -> ParsedToolCall | None:
         data = None
 
     if data is not None:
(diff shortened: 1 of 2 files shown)


qbc2016 (approved):
LGTM.

PR #6410, opened 2026-07-23
fix(providers): strip annotated null gemini schemas

AaronZ345:
## What Problem This Solves

Gemini schema sanitization strips nullable `anyOf` branches only when the branch is exactly `{"type": "null"}`. Pydantic-style schemas often annotate that null branch with metadata, for example `{"type": "null", "title": "None"}`, which leaves an unsupported nullable branch in the Gemini tool schema.

This change treats any schema object with `type == "null"` as a null schema, preserving the intended non-null branch while dropping the Gemini-incompatible nullable branch.

## Evidence

- Added a unit test for an `anyOf` schema containing `{"type": "null", "title": "None"}`.
- Verified the Gemini provider test suite passes:

```bash
PYTHONPATH=src python -m pytest tests/unit/providers/test_gemini_provider.py -q
```

Result: `21 passed`.

### DIFF
--- a/src/qwenpaw/providers/gemini_provider.py
+++ b/src/qwenpaw/providers/gemini_provider.py
@@ -82,6 +82,10 @@ def _flatten_json_schema(schema: dict) -> dict:
     return _resolve_ref(schema)
 
 
+def _is_null_schema(schema: Any) -> bool:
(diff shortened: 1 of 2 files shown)


qbc2016 (approved):
LGTM.

PR #6411, opened 2026-07-23
fix(console): round heartbeat seconds up

AaronZ345:
## What Problem This Solves

The console heartbeat parser converts seconds with `Math.round(seconds / 60)`. Short non-zero values like `29s` round to `0` minutes, which then falls through to the default `6h` interval and silently displays a much longer heartbeat interval than configured.

This change rounds seconds up to the nearest minute so every non-zero seconds-only interval remains visible and does not collapse to the default.

## Evidence

- Added Vitest coverage for `29s`, `30s`, and `90s`.
- Verified the targeted console test passes:

```bash
npm test -- --run src/pages/Control/Heartbeat/parseEvery.test.ts
```

Result: `1 test file passed, 6 tests passed`.

### DIFF
--- a/console/src/pages/Control/Heartbeat/parseEvery.test.ts
+++ b/console/src/pages/Control/Heartbeat/parseEvery.test.ts
@@ -17,6 +17,12 @@ describe("parseEvery", () => {
   it("mixed 2h30m → total minutes 150 (not divisible by 60)", () => {
     expect(parseEvery("2h30m")).toEqual({ number: 150, unit: "m" });
   });
+
+  it("rounds seconds up to the nearest minute", () => {
+    expect(parseEvery("29s")).toEqual({ number: 1, unit: "m" });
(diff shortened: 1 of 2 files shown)


lalaliat:
Hello, thanks for your contribution! However, the console heartbeat UI only exposes minute/hour granularity for the interval — there's no way for users to configure a seconds-only value through the frontend. The s suffix in the regex is an internal/test-only path (used by integration tests like every=60s), and we don't intend to display those values in the console.

If a sub-minute interval somehow reaches the frontend, falling back to the default 6h is actually the safest behavior since heartbeat intervals below 1 minute are not a supported configuration.

Won't merge by now, appreciate the attention to detail though!

AaronZ345:
Closing this since the console intentionally only supports minute/hour interval granularity, and sub-minute seconds are not meant to be displayed in the UI.

PR #6424, opened 2026-07-24
feat(computer-use): native desktop GUI automation for Windows and macOS (accessibility-first + Tauri control mode)

jinglinpeng:
## Description

Adds a `computer_use` builtin tool that lets the agent operate the host desktop
through a native, host-managed helper -- on both **Windows and macOS**. The
agent works one **approved application** at a time: it discovers windows,
observes a window (screenshot + accessibility tree), and acts either on
structured UI elements or, as a fallback, on validated visual coordinates. A
Tauri-side **Control Mode** keeps the user in the loop and able to reclaim
control at any time.

### Architecture

- **Thin Python plugin, native helper.** The in-repo plugin
  (`plugins/tool/computer-use`) contains only the tool adapter, transport,
  approval bridge, and usage skill. All OS interaction (window discovery,
  capture, accessibility, input, target validation) runs in a native helper
  packaged with the desktop app -- the plugin has no Python GUI-automation
  dependencies.
- **Shared core, platform leaves.** One Rust helper hosts a shared RPC server,
  wire protocol, session/turn lifecycle, and per-application approval. Only the
  OS-touching leaves are platform-specific:
  - **Windows:** UI Automation, Windows Graphics Capture, and input injection;
    transport over a named pipe; child cleanup via a Job Object.
  - **macOS:** AXUIElement accessibility, `CGWindowListCreateImage` capture, and
    CGEvent input; transport over a Unix domain socket; child cleanup via a
    kqueue parent-death watch.
- **Short-lived private capability.** The host hands the plugin a per-session
  transport address plus secret; the helper validates a handshake and drives a
  reverse channel for approval prompts.

### Tool surface

A single dispatch tool `computer_use(action=...)` with a small, platform-neutral
action set:

- Discovery: `list_apps`, `list_windows`, `launch_app`, `observe_window`
- Accessibility channel: `invoke` (buttons / menu items), `set_value` (text
  fields / combo boxes), addressed by `element_id` within an
  `accessibility_revision`
- Visual channel: `click`, `scroll`, `drag`, each carrying `window_id` +
  `snapshot_id` + `screenshot_id` from the same observation
- Keyboard / focus: `type`, `press_key`, `set_focus`
- `stop`

`observe_window` returns a screenshot (as an image attachment) plus an
accessibility frame whose elements carry a readable `control_type_name`
(`Edit`, `Button`, `ComboBox`, `MenuItem`, and so on). The skill steers the model to act
on accessibility elements first and fall back to coordinates only when no
element matches. The helper re-validates window geometry and the hit window
immediately before injecting input, and rejects changed, covered, or
interrupted targets.

### Safety

- Gated by a global desktop-automation switch: while it is off, every call is refused and pending approvals are rejected. While it is on, the tool is enabled for all agents by default, and the user can turn it off for individual agents; that choice is respected across restarts.
- Per-application approval, surfaced to the user; denial stops the tool.
- A recent-user-intervention guard blocks synthesized input right after real
  keyboard / mouse activity.
- QwenPaw's own windows, credential dialogs, and other sensitive surfaces are
  excluded as targets.
- `stop` cancels immediately on user request.

### Packaging

On both platforms the helper is compiled and staged next to the backend so it
rides into the desktop bundle and is code-signed by the existing signing passes
(ad-hoc where no Developer ID / signing identity is configured). The desktop
runtime resolves the helper from bundle resources and passes it the per-session
capability.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend
- [x] Console (frontend web UI + Tauri desktop host + native helper)
- [x] Skills
- [x] Scripts / Deploy (PyInstaller spec + Tauri packaging)

## Testing / Verification

- **Windows:** manual end-to-end on Windows 11 -- discover -- observe -- UI
  Automation actions with coordinate fallback -- per-app approval -- release.
- **macOS:** the native helper compiles on macOS, and the full desktop package
  build was validated in CI: the helper is compiled, staged into
  `QwenPaw Desktop.app/Contents/Resources/binaries/qwenpaw-backend`, and
  code-signed (`Mach-O 64-bit executable arm64`, `codesign --verify` passes).
  End-to-end functional testing on macOS hardware (TCC Accessibility /
  Screen-Recording prompts at first use) is still pending.

## Notes

- macOS currently targets the **development tier**: ad-hoc signing, no Developer
  ID / notarization, so TCC permissions are re-prompted when the build changes.
  Capture uses the synchronous `CGWindowListCreateImage`; migrating to
  ScreenCaptureKit is the shippable, background-capable follow-up (marked in
  code).
- `pre-commit` passes on every file this PR touches. Running it across
  the whole repository still reports pre-existing findings unrelated to
  this change.


cuiyuebing (approved):
LGTM

cuiyuebing (approved):
LGTM

PR #6456, opened 2026-07-24
feat(context): Visual Compact

Leirunlin:
## Description

  Add PawFocus visual context compression for long agent histories.

  The implementation selectively compresses eligible history and tool results into visual context, applies profitability gating, and supports exact-content recovery through `recover_visual_context`. It also fixes recovery behavior and duplicate factsheet generation, and adds related logging,
  configuration, tests, and benchmarks.

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [x] Documentation
  - [ ] Refactoring

  ## Component(s) Affected

  - [ ] Skills
  - [ ] CLI
  - [ ] Documentation (website)
  - [ ] Tests
  - [ ] CI/CD
  - [ ] Scripts / Deploy

  ## Checklist

  - [ ] Ready for review

  ## Testing

  - Run the focused visual-compression, recovery, pruning, file I/O, and CLI unit tests.
  - Run paired OFF/ON benchmarks on the scroll-transfer and v4 task suites.
  - Verify recovery, profitability gating, rendering, and factsheet deduplication.

  ## Evidence

  ```bash
  # Paste the final pre-commit and pytest summaries here.
  pre-commit run --all-files
  pytest

  Paired OFF/ON benchmark runs were also used to validate correctness and token reduction.

  ## Additional Notes

 * The compression gate skips transformations that are not expected to be profitable. Original text remains recoverable when exact details are required.
 * Need more tests.


jinliyl (changes requested):
Thank you for putting together this substantial feature. The overall direction is interesting, and I appreciate the care taken around profitability gating, media preservation, recovery handles, UI configuration, and fail-open behavior. I reviewed the final commit and ran the relevant existing backend tests locally (51 tests passed). I also performed focused checks of the new rendering and recovery paths.

Before merging, I believe the following issues need to be addressed:

## 1. System and host instructions are downgraded to the user role (blocking)

In `compress_static_context`, the original system-message content is replaced with a short pointer, while the actual configuration is rendered into images and inserted as a new `Msg(..., role="user")`. In addition, `_append_env_tail` appends host-provided environment context to the external user message.

This changes the provider-visible instruction hierarchy. Text that previously had system authority—including operating rules, safety constraints, and tool-use guidance—is now delivered as user-role content. A system pointer saying that the following user message contains configuration does not preserve the original role semantics, especially when later user instructions conflict with the rendered content. Appending host context to the external user message has the same problem.

Relevant locations:
- `pipeline/static_context.py`, around lines 263-278
- `pipeline/request.py`, around lines 42-57

Please preserve the original authority boundary. The safest option may be to exclude system/host instructions from visual compression and limit this feature to history and tool results. If system content must be compressed, it should remain provider-visible at system/developer authority rather than being represented as a user message.

## 2. Image rendering blocks the async model-call path (blocking)

`VisualCompressionMiddleware.on_model_call` calls `transform_model_request` synchronously. That transformation performs Python layout, PIL drawing, and PNG/zlib encoding for potentially many pages, all on the event-loop thread.

Relevant location:
- `runtime/middleware.py`, around lines 150-160

In a local focused benchmark, I observed approximately:
- low effort: 56 pages in 2.47 seconds
- medium effort: 12 pages in 23.16 seconds
- high effort: 10 pages in 23.09 seconds

During that time the event loop cannot run heartbeat work, streaming, or other agents sharing the loop. This is especially risky because medium/high rendering uses per-character font drawing and the request cap allows many generated images.

Please move the CPU-bound transformation off the event loop, for example with `await asyncio.to_thread(transform_model_request, ...)`, and add a regression test demonstrating that an in-flight render does not prevent another coroutine from making progress. A concurrency limit may also be useful to prevent several simultaneous renders from exhausting CPU and memory.

## 3. Exact recovery cannot recover long single-line sources completely

`TurnRecoveryStore.excerpt` caps every response at 12,000 characters. The recovery API supports a query or line range, but it has no character offset, byte range, or pagination cursor. Therefore, if a source contains a line longer than 12,000 characters—common for minified JSON, encoded payloads, and some logs—the suffix of that line is unreachable through every available recovery mode.

I reproduced this with a 20,000-character single-line source: the default, `start_line=1/end_line=1`, and query modes all returned only 12,000 characters, with no way to request the remainder. Also, `splitlines()` followed by joining with `\n` does not preserve original line endings, so line/query recovery is not byte-exact as documented.

Relevant location:
- `runtime/recovery.py`, around lines 46-106

Please add character/byte offsets or a pagination cursor so the complete source can be retrieved in bounded chunks, and preserve original separators if the contract remains byte-exact. Tests should include a long single-line JSON value, CRLF input, trailing newlines, and multi-page recovery.

## Test coverage

The final diff adds roughly 4,900 lines of backend implementation but does not add backend tests for the visual-compression pipeline. The existing related tests I ran pass, but they do not exercise the new request transformation, role preservation, renderer concurrency, profitability boundaries, recovery completeness, or provider formatting. Given that this feature rewrites model requests at the final provider boundary, focused backend tests are important before enabling it in production.

Suggested minimum coverage:
- system/host role and authority preservation
- middleware skip/fail-open paths
- recovery completeness and expiry behavior
- async non-blocking behavior
- OpenAI, Anthropic, and Gemini formatter output
- native-media invariants and image-count limits
- low/medium/high rendering paths

Thank you again for the work on this. I am requesting changes because the first two issues can affect instruction safety and runtime availability, while the recovery issue breaks one of the feature’s core correctness guarantees. I would be happy to re-review once these points are addressed.

Leirunlin:
@jinliyl Thanks for the review.

The system prompt shifting is intentional. Since most providers don't currently support image blocks in system messages, this serves as a fallback to use user block. I haven't observed any performance degradation in my environment, but we could add a note in the documentation and comments to alert users to this behavior.

Issues 2 and 3 have been addressed in the new commit. As for the tests, they were deliberately omitted.

jinliyl:
LGTM

niceIrene:
@Leirunlin Thank you for implementing this new feature!

Based on the current head, please check the following comments.

### 1. Event-loop blocking is not actually solved — the bottleneck moved to the tokenizer's GIL

`run_in_executor` only rescued rendering. HF `tokenizers`' `encode()` in `budget.py` **holds the GIL for its entire duration**, so threading doesn't help:

| Background call | Duration | Max stall of other threads |

|---|---|---|

| `time.sleep(0.2)` / `zlib.compress` (GIL-releasing controls) | — | 1ms |

| `tokenizers.encode` 276KB | 168ms | **166ms (≈ full duration)** |

| Tokenizer first load | 410ms | **414ms** |

A single transform triggers **66 calls** to `count_text_tokens` (~270K chars encoded in total); during a full high-effort transform the event loop's max stall measured **538ms**.

Suggestion: chunked encoding (~16KB per chunk so the GIL can switch between chunks), cached counts, or fall back to the char-based estimate for gating. The non-blocking regression test should assert **max loop stall < N ms** — merely asserting "another coroutine makes progress" cannot catch this.

### 2. line-range / query recovery modes still truncate silently

Cursor mode is fixed, but the two modes the docstring recommends first are still a hard `body[:max_chars]` cut (`recovery.py` L160–172): no truncation marker, no next_cursor, and it can cut mid-line.

- **line-range**: on a 300-line source, `start_line=1, end_line=250` silently returns only through line **115**

- **query**: with 40 matches in the source, only **20** are returned — the model will believe there were only 20 matches total

Both are worse than getting nothing: the model receives a plausibly complete answer.

Suggestion: on overflow, append `truncated at line N; continue with start_line=N` or return a cursor. Also worth fixing: `query` + line range together is silently ignored while `cursor` + query raises, and the default 200-line window for `end_line` isn't documented.

### Other notes (non-blocking)

- `_stabilize_promoted_tool_result_media_identifiers` (model_factory.py) is a **global behavior change for all requests** (not gated by the visual_compact switch), its label/media pairing assumption is coupled to AgentScope's internal output ordering, and it has zero test coverage — please call it out in the PR description and add a pairing test

- Layout (`_char_cells` / `_escape_missing_glyphs`) hardcodes the **low** atlas while rendering uses the effort atlas. The three atlases are identical today (verified by decoding them), but nothing pins that invariant — any future atlas regen would make the whole feature silently fail open. An atlas-consistency unit test would cover it

- The fix commits added no backend tests, so the freshly fixed behaviors (cursor paging exactness, non-blocking transform) have no regression protection

And also please merge the conflicts.

Leirunlin:
@niceIrene Thanks for the detailed measurements. I addressed the two P1 issues and rebased the branch to resolve the conflicts:

  - Removed HF tokenizer encoding from Visual Compact gating and switched to the existing UTF-8 character estimate, eliminating the GIL-bound path.
  - Refactored line-range/query recovery to return whole-line pages with `total`, `complete`, and continuation metadata.
  - `query + start_line/end_line` is now honored, and the default 200-line window is documented.

Local checks cover event-loop stall, cursor reconstruction, query/range continuation, malformed inputs, media pairing, and atlas consistency.

I kept the two non-blocking behaviors unchanged: current AgentScope label/media ordering is valid, and the three atlases are currently identical.

cuiyuebing (approved):
LGTM

jinliyl (approved):
LGTM

PR #6459, opened 2026-07-24
fix(history): harden SQLite persistence, backup, and restore

niceIrene:
## Summary

Scroll already used `history.db` with WAL mode, a SQLite busy timeout, idempotent `sessions/*.json` backfill, and retention filtering.

This PR keeps that architecture and fixes focused reliability gaps around:

- concurrent SQLite writes;
- WAL lifecycle;
- schema compatibility;
- corruption classification;
- workspace-bounded database paths;
- live instance backup;
- restore validation;
- startup retention.

It does not move all session state into SQLite or change history retrieval behavior.

## Problem

Before this PR:

- writes relied on implicit SQLite transactions and a fixed busy timeout;
- temporary lock, I/O, permission, read-only, disk-full, or schema errors could be treated as database corruption;
- the history schema had no explicit compatibility version;
- WAL was enabled but the application did not explicitly checkpoint it;
- `db_filename` could resolve outside the owning Agent workspace;
- instance backup copied the live `history.db` file directly, potentially omitting committed rows still present in WAL;
- restore could replace a workspace before checking its history database;
- startup retention was skipped when the legacy `sessions/` directory did not exist.

The two most serious failure modes were silent:

1. A healthy database could be quarantined because of a temporary operational error.
2. A successful-looking backup could omit recent conversation history stored in WAL.

### 1. Workspace-bounded history paths

Add one shared history path resolver and apply it to:

- normal Scroll startup;
- standalone `/compact`;
- startup synchronization and retention;
- instance backup;
- restore validation.

The resolver rejects:

- absolute POSIX paths;
- Windows drive-qualified paths;
- UNC paths;
- `~` expansion;
- parent traversal with `..`;
- symlink paths that resolve outside the Agent workspace.

### 2. Explicit SQLite write transactions

Use `BEGIN IMMEDIATE` for:

- schema initialization and migration;
- single inserts;
- batch inserts;
- entry updates;
- retention deletes.

Writer contention now uses:

- a short SQLite busy timeout;
- bounded, jittered application retry;
- retries only for `SQLITE_BUSY` and `SQLITE_LOCKED`;
- rollback when a transaction fails.

Permission, read-only, I/O, disk-full, corruption, and schema errors are not retried.

### 3. Passive WAL checkpoints

Run `PRAGMA wal_checkpoint(PASSIVE)`:

- after every 50 successful write transactions;
- when closing the history store.

Checkpoint errors are kept separate from write errors because the underlying transaction has already committed.

### 4. Versioned schema migration

Use `PRAGMA user_version` as the history schema compatibility boundary.

On open, QwenPaw now:

- upgrades compatible pre-versioned databases transactionally;
- adds missing compatible columns without replacing existing rows;
- publishes the schema version only after schema setup succeeds;
- rejects databases missing required columns;
- rejects schema versions newer than the running application.

Newer or incompatible schemas fail safely and are not treated as corruption.

### 5. Narrow corruption quarantine

Previously, any `sqlite3.DatabaseError` could trigger quarantine.

Quarantine is now limited to confirmed corruption, including:

- SQLite `CORRUPT` and `NOTADB` result codes;
- failed `PRAGMA quick_check`;
- narrowly recognized malformed/not-a-database errors.

Lock contention, read-only filesystems, permission failures, I/O errors, disk-full errors, and unsupported schemas leave the database files untouched.

### 6. Consistent SQLite history backup

Instance backup now:

- detects each selected Agent's configured history database;
- creates a consistent snapshot using `sqlite3.Connection.backup()`;
- validates the snapshot before adding it to the archive;
- includes committed rows still present in WAL;
- skips live `-wal`, `-shm`, and `-journal` sidecars;
- fails instead of silently producing an incomplete history backup.

The service does not need to be stopped while the snapshot is created.

### 7. Restore validation before replacement

Restored Agent workspaces are already staged before becoming live.

This PR validates a staged history database before committing the workspace replacement. Validation checks:

- SQLite integrity;
- required history schema;
- supported schema version.

If validation fails, restore stops and the existing workspace remains unchanged.

### 8. Retention without legacy sessions

Startup previously returned early when `sessions/` was absent, so an existing `history.db` was not purged.

Retention now runs whenever the Agent already has a history database, independently of legacy session backfill.

Agents with neither `history.db` nor `sessions/` continue to be skipped, avoiding creation of an empty database solely for maintenance.

## Before and After

| Area | Before | After |
| --- | --- | --- |
| Writer contention | Implicit transactions and a fixed timeout | `BEGIN IMMEDIATE` with bounded retry only for `BUSY`/`LOCKED` |
| WAL lifecycle | WAL enabled without explicit writer checkpoints | Passive checkpoint every 50 writes and on close |
| Schema compatibility | No explicit version boundary | Transactional migration with `user_version` |
| Corruption handling | Any `DatabaseError` could quarantine the DB | Only confirmed corruption can trigger quarantine |
| History paths | Callers joined paths independently | One workspace-bounded resolver is used everywhere |
| Instance backup | Copied the live database file directly | Uses SQLite online backup and skips live sidecars |
| Restore | Staged history was not validated | History is checked before workspace replacement |
| Startup retention | Depended on `sessions/` existing | Runs for an existing DB independently of backfill |

## Existing Behavior Preserved

This PR does not change:

- WAL mode itself;
- `sessions/*.json` backfill;
- `.synced.json` manifest behavior;
- backfill idempotency;
- `(session_id, dedup_key)` deduplication;
- retention window semantics;
- Scroll recall or FTS query behavior;
- human-edited file-backed configuration.

## Explicit Non-goals

This PR does not add:

- simultaneous multi-writer SQLite transactions;
- a public history maintenance CLI;
- write-health counters or a monitoring dashboard;
- atomic DB/WAL/SHM quarantine rollback;
- automatic recovery of corrupted history;
- multi-target restore rollback;
- whole-instance point-in-time backup;
- CJK tokenization or semantic search;
- session lineage or a full database-first session refactor.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring

## Components Affected

- [x] Core / Backend
- [ ] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [x] Documentation
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Testing

Focused coverage includes:

- workspace path normalization and escape rejection;
- invalid-path isolation between Agents;
- startup retention without `sessions/`;
- avoiding empty maintenance-only databases;
- `BEGIN IMMEDIATE` transactions;
- concurrent writer coordination;
- bounded lock retry;
- transaction rollback;
- periodic passive checkpoint behavior;
- compatible schema migration;
- newer-schema rejection without quarantine;
- lock errors not triggering quarantine;
- WAL-safe online backup;
- WAL/SHM sidecar exclusion;
- invalid backup failure;
- staged restore database validation.

After rebasing onto the latest `upstream/main`:

```text
121 passed
```

All files changed by this PR pass:

```text
check python ast
check docstring is first
detect private key
trim trailing whitespace
Add trailing commas
mypy
black
flake8
pylint
```

English and Chinese documentation pass Prettier formatting checks, and `git diff --check` passes.

## Security Considerations

History database paths are now confined to their owning Agent workspace.

Instance backups contain conversation history and should be protected like other user data. This PR adds no external network interface and does not change secret storage.


RerankerGuo (approved):
Validated the SQLite durability and restore guarantees independently.

Checked:

- exact head `ab2cf11f591e2f88432f92e3f235401bb8e86c6e`
- current main `bacac7410cf2ac122dd892152beff8b03f38729f`
- semantic rebase trial `c0a790096fab0cfde7e47b102090a655388c4086` (tree `a75c465842af05563eb36a700e73bd35cbdc23e1`)

Results:

- Exact changed tests: `76 passed`.
- Current-main trial changed tests: `96 passed`.
- Broader trial regression: `tests/unit/backup` `34 passed`; `tests/unit/agents/context` `406 passed`.
- Independent production contracts passed on exact and trial:
  - an online backup includes a committed row still represented by the live WAL;
  - the archive excludes `-wal`, `-shm`, and `-journal`;
  - a corrupt staged history DB is rejected without replacing an existing workspace;
  - a newer unsupported schema is rejected without quarantining the DB.
- Trial source checks passed: Mypy, Black, Flake8, Pylint, and `git diff --check`.

The current conflict is semantic but resolvable. The trial preserves main's chat-registry preflight before opening `history.db` when session files exist, while still applying this PR's retention maintenance when `history.db` exists without a `sessions/` directory. This ordering is important: an unavailable/empty registry must not trigger purge or corruption quarantine.

Running Mypy across every changed test file also reports the two pre-existing generator-fixture return annotations in `test_history_store.py` and `test_sync.py`; both are present at the PR merge base `734c8b9f` and are not introduced here.

Please rebase and retain the preflight ordering above. The implementation and durability contracts are otherwise good.

PR #6485, opened 2026-07-27
fix(console): cap SSE replay buffer and add stream heartbeat

### LINKED ISSUE #6460: QwenPaw 2.0.1 首页/会话在 Edge+Wayland 下单标签高 CPU 占用，疑似大结果集渲染/WebSocket 推送触发
环境
OS：Linux 7.0.0-28-generic (x86_64)
Display：Wayland
Browser：Microsoft Edge（chromium 系）
QwenPaw：2.0.1（本地 qwenpaw --version 返回）
业务场景：通过 QwenPaw 管理/查看 ComfyUI 工作流与输出结果
访问方式：通过服务端地址跨设备访问（非本地仅问题）
现象
打开 QwenPaw 首页（或任意大会话页面）并停留一段时间后，单个 Edge 标签页 CPU 占用持续走高，风扇加速。问题不随日常浏览复现，基本只出现在 QwenPaw 页面。

触发条件
会话中关联 ComfyUI 业务结果，结果集体积很大；
页面可能在持续渲染或 WebSocket 推送大体积结果数据，导致前端高负载。
复现步骤
Wayland 下用 Edge 打开 QwenPaw。
进入携带大体积 ComfyUI 结果集的会话或管理页面。
等待数分钟，观察该 Edge 标签页 CPU 占用持续升高。
影响
单标签高 CPU 持续占用资源；
会话切换时持续卡死，但通过移动端频道交互不受影响；
多次复现稳定，与 session 长度/结果大小有关。
已排除
非浏览器全局异常（其他站点正常）；
跨设备访问相同服务均复现，指向服务端/页面渲染链路，非单一终端；
已升级最新版照旧。
结果集情况
相关 ComfyUI 结果集文件体积很大。

临时解决方式
利用移动端频道让智能体，删除该 ComfyUI 结果集相关的会话后，QwenPaw 页面恢复正常，CPU 占用回落。

期望
给出页面侧或服务端侧可以规避/修复的方案，尤其是大结果集场景下的渲染/推送策略。

zhaozhuang521:
Fixes both failure modes reported in #6460:

**1. Sustained CPU climb while staying on the page (reconnect–replay loop)**

Long runs with huge tool outputs grew the per-run replay buffer without bound, and every reconnect replayed the entire event history. On remote access paths idle SSE connections are dropped by proxies, triggering repeated reconnect + full-replay cycles that drove sustained frontend CPU load.

- Cap the replay buffer at 2000 events / 8 MiB per run (accounted in encoded UTF-8 bytes); evict oldest events and send a `replay_truncated` marker to reconnecting clients
- Emit a keep-alive SSE comment frame after 15s of queue idle so proxies do not drop the connection
- Consume the queue via a `get_nowait` fast path to avoid per-event `wait_for` Task/timer overhead on busy streams
- Filter the `replay_truncated` marker in the console responseParser (extracted as a tested pure helper)

**2. Page freeze when opening a conversation with a large stored result set**

`convertMessages` merged all consecutive non-user messages of a turn into a single ResponseCard. The chat SDK paginates history at card granularity (10 cards per page), so one giant card (e.g. a long ComfyUI run with hundreds of outputs) bypassed the window entirely and forced a full render of the whole result set on open.

- Split turns above 20 outputs into chunk cards of at most 20 outputs each, so the SDK's built-in pagination actually bounds render cost
- Turn usage is extracted from the whole group and attached only to the last chunk, keeping tail-scanning consumers (`extractLatestSnapshotFromCards`, `hydrateTurnUsageFromMessages`) unchanged; output order, sequence numbers, and user-turn boundaries are preserved

**Supporting changes**

- `GET /chats/{chat_id}/status`: lightweight status endpoint backed solely by the task tracker; `waitForChatIdle` pollers no longer deserialize the full session state every second
- Extracted `_subscribe_with_replay` to deduplicate the queue-creation/marker/replay logic shared by `attach` and `attach_or_start`

Closes #6460

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [x] Tests

## Checklist

- [x] I ran pre-commit run --all-files locally and it passes
- [x] I ran tests locally (pytest or as relevant) and they pass
- [ ] Documentation updated (not needed)
- [x] Ready for review

## Testing

- Backend: `pytest tests/unit/app/ -q` → **628 passed, 1 skipped**, including:
  - 19 task_tracker tests: buffer eviction by count / bytes / UTF-8 multi-byte content, truncated-marker replay on `attach` and `attach_or_start`, heartbeat emission, fast-path consumption
  - 4 status-endpoint tests, incl. asserting no session-state deserialization
  - 1 router-level integration test: keep-alive comment frames flow through the real `/console/chat` StreamingResponse during idle gaps
- Frontend: `vitest run` → **1219 passed**, including:
  - 8 giant-turn splitting regression tests (threshold boundaries, `ceil(n/20)` chunking, cross-chunk output order, user-turn boundaries, usage-on-last-chunk, single-card turns unchanged)
  - 3 replay-marker filtering tests
  - Existing #5479 large-session regression suite (29 tests) still green
- `pre-commit run` (mypy / black / flake8 / pylint) → all passed

## Evidence

```
pytest tests/unit/app/ -q
628 passed, 1 skipped in 7.53s

vitest run
Test Files  141 passed (141)
Tests       1219 passed (1219)
```

Live verification against a real session (~100 history cards): initial render contains exactly **10 cards** (counted via DOM), expanding on demand (10 → 57 → 67) as the user navigates upward — confirming the pagination window that the turn-splitting change relies on is active in production code.

## Additional Notes

- A CI-verifiable test for "heartbeat prevents the reconnect loop" is not feasible — it requires a proxy that drops idle connections; the router-level test proves the frames reach the wire. Real-environment before/after CPU measurements will be attached as follow-up comments.
- Backpressure / bounded subscriber queues for active streams (review finding 2) is pre-existing architecture and intentionally out of scope; tracked as a follow-up.


rayrayraykk (commented):
### 1. Blocking: the completed large-conversation path is not covered

Issue #6460 reports that entering an existing conversation with a large result
set can cause sustained CPU usage or freeze the browser. Deleting the affected
conversation makes the problem disappear.

The replay buffer modified by this PR only exists while a task is active. When
generation finishes, the task state and replay buffer are removed:

- [`task_tracker.py`, task cleanup](https://github.com/agentscope-ai/QwenPaw/blob/7530e07628eb68574f58bf5918286c835f9c7786/src/qwenpaw/app/task_tracker.py#L293-L307)

Opening a completed conversation still causes the backend to read, convert,
and return the complete stored history:

- [`chats/api.py`, complete history response](https://github.com/agentscope-ai/QwenPaw/blob/7530e07628eb68574f58bf5918286c835f9c7786/src/qwenpaw/app/chats/api.py#L253-L290)

The frontend then converts all returned messages in one operation:

- [`sessionApi/index.ts`, history conversion](https://github.com/agentscope-ai/QwenPaw/blob/7530e07628eb68574f58bf5918286c835f9c7786/console/src/pages/Chat/sessionApi/index.ts#L1105-L1118)

Consequently, once a large ComfyUI result has been persisted, the main
reported failure mode can still occur.

### 2. High: active subscribers still receive every large event

The new limits apply only to the replay buffer used by reconnecting clients.
Each active subscriber still has an unbounded `asyncio.Queue`, and the
producer pushes every event into every subscriber queue with `put_nowait`:

- [`task_tracker.py`, subscriber broadcast](https://github.com/agentscope-ai/QwenPaw/blob/7530e07628eb68574f58bf5918286c835f9c7786/src/qwenpaw/app/task_tracker.py#L271-L278)

For a connection that remains open:

- all large tool-result events are still sent to the browser;
- the frontend still parses and renders every event;
- there is no queue bound, backpressure, coalescing, or result-size policy.

If the CPU usage in #6460 is caused by continuous parsing and rendering rather
than repeated reconnections, this PR does not mitigate it.

### 3. Medium: `MAX_BUFFER_BYTES` counts characters rather than bytes

`_buffer_append` updates `buffer_bytes` with `len(sse)`:

- [`task_tracker.py`, buffer accounting](https://github.com/agentscope-ai/QwenPaw/blob/7530e07628eb68574f58bf5918286c835f9c7786/src/qwenpaw/app/task_tracker.py#L53-L65)

For non-ASCII content such as Chinese text or emoji, the UTF-8 payload can be
substantially larger than the counted Python string length. The actual
transmitted replay can therefore exceed the stated 8 MiB limit.

The accounting should use the encoded byte length, with a non-ASCII unit test.

rayrayraykk (approved):

RerankerGuo:
Independent semantic rebase trial of exact PR head
`2e4cd6bdb3ce2f441249555cbe951377e46a02f1` onto current `main`
`bacac7410cf2ac122dd892152beff8b03f38729f` (macOS arm64, Python
3.11.15, Node 26.5.0).

`git merge-tree --write-tree` reports only three content conflicts:

```text
console/src/pages/Chat/index.tsx
console/src/pages/Chat/utils.test.ts
src/qwenpaw/app/task_tracker.py
```

I resolved them while preserving both branches' contracts:

1. In `Chat/index.tsx`, keep current-main client-message-ID imports and add
   this PR's `shouldDeferResponseRender`; retain current-main `turn_usage` /
   `replay_end` handling and this PR's structured rate-limit conversion.
2. In `utils.test.ts`, retain both current-main sender-textarea helpers and
   this PR's `isReplayTruncatedPayload`.
3. In `task_tracker.py`, retain current-main `REPLAY_END_SSE`, per-run
   `owner`, and workspace-reload lifecycle while adopting this PR's bounded
   deque / subscriber queue, byte accounting, truncation marker, and
   heartbeat.

One detail in item 3 is important: `replay_end` must remain specific to
explicit reconnects through `attach()`. A naive resolution that appended it
inside the shared `_subscribe_with_replay()` helper changed the normal
`attach_or_start()` contract and produced this negative control:

```text
FAILED test_attach_or_start_existing_run_returns_buffer_replay
expected: ["data: second\n\n", None]
actual first item: data: {"type": "replay_end"}
```

The passing resolution is:

```python
def _subscribe_with_replay(state):
    # truncated marker + bounded replay tail, but no replay_end
    ...

async def attach(...):
    queue = self._subscribe_with_replay(state)
    _subscriber_put(queue, REPLAY_END_SSE)
    return queue
```

`attach_or_start()` continues to call the helper without injecting that
explicit-reconnect boundary. The new `_RunState` construction also keeps
`owner=owner` and uses the deque default rather than restoring `buffer=[]`.

There is one non-conflicting test rebase needed:
`test_console_chat_stream.py` must supply current-main's `agent_id`,
`workspace_dir`, and chat `meta`, and isolate/mock the project-directory
resolution now performed before a new run starts.

Validation on the resolved trial tree:

```text
backend reconnect/task-tracker/status focused scope:
56 passed

complete tests/unit/app:
710 passed, 1 skipped

replay/large-turn focused Console scope:
94 passed

complete Console Vitest:
226 files, 1702 tests passed

tsc -b --noEmit + Prettier:
passed

Black 23.3.0 --line-length=79, Flake8 with repository E203 config,
and Pylint with repository hook disables:
passed
```

I also added an external composition contract for the new real reconnect
sequence:

```text
replay_truncated -> replay tail -> replay_end -> live delta
-> canonical completed -> turn_usage
```

With the production wrapper order
`wrapChatResponseUsageStream(wrapReplayFastForward(response))`, the test
passes and only the canonical completed response reaches the SDK. This proves
current-main fast-forwarding does not swallow the truncation signal and the
two recovery mechanisms compose correctly.

I found no remaining semantic blocker with the resolution above. Since this
PR is already approved and labelled `to-be-merged`, rebasing with these
specific choices should restore mergeability without dropping either the
new bounded-stream behavior or current-main reconnect/lifecycle guarantees.

PR #6486, opened 2026-07-27
fix(matrix): probe vodozemac E2EE backend so encryption works on Python 3.12

### LINKED ISSUE #6476: [Bug]: matrix的端到端加密不可用
## QwenPaw Version

v2.0.1

## Description

matrix的端到端加密不可用，matrix-nio（QwenPaw 用的 Matrix 客户端库）需要 olm 才能解密加密消息。qwenpaw自己安装失败，使用agent安装提示执行了三步：

步骤	命令	结果
1	apt install libolm-dev	✅ 成功——装了系统层 C 库 /usr/lib/x86_64-linux-gnu/libolm.so.3
2	uv pip install matrix-nio[e2e]	✅ 成功——装了 vodozemac（新式 olm 替代） + peewee
3	uv pip install olm（Python 绑定）	❌ 失败了——olm 依赖旧的 jsmin==2.2.2，在 Python 3.12 上编译不了
所以现在的情况是：系统 libolm 有了，但 Python 层面的 import olm 还是无法工作，因为老的 olm Python 包在 Python 3.12 上装不了。

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **OS:** [Ubuntu26.04]
- **Install method:** [one-line install]
- **Python version (if applicable):** [3.12]

## Steps to Reproduce

1.配置频道
2.对话无回复
3.检查日志发现错误

## Actual vs Expected

- **Actual:**无法安装且解密
- **Expected:**正常E2E

WilShi:
## Description

Fixes #6476.

The Matrix channel's E2EE preflight (`MatrixChannel._preflight_e2ee_dependencies`) probed only the legacy `olm` Python bindings. `olm` depends on `jsmin==2.2.2`, which no longer builds on Python 3.12 — so on Python 3.12 the probe fails and `encryption` is silently set to `False`, even when the user has installed the modern E2EE backend via `matrix-nio[e2e]` (which installs `vodozemac`, the Rust replacement for `olm`). The result: Matrix E2EE never works on Python 3.12, even with `matrix-nio[e2e]` correctly installed — exactly what #6476 reports.

### Root cause

`matrix-nio`'s `[e2e]` extra installs `vodozemac>=0.9.0.post2` (confirmed in matrix-nio 0.26.0's package METADATA). The preflight was checking for the wrong backend — the deprecated `olm` — instead of the `vodozemac` backend that `matrix-nio[e2e]` actually provides.

### Fix

Probe `vodozemac` first; fall back to legacy `olm`; only disable E2EE when neither backend is importable. Update the install hint to recommend `pip install 'matrix-nio[e2e]'` (vodozemac, works on Python 3.12+).

**Type of Change:** Bug fix

**Component(s) Affected:** Channels (Matrix)

**Security Considerations:** E2EE availability — this fix enables encryption that was silently disabled before; no auth, tokens, or config are exposed.

## Evidence

- `pytest tests/unit/channels/test_matrix.py` — 63 passed, including 3 new preflight tests.
- Verified with TDD: the key test (`test_preflight_keeps_encryption_with_vodozemac`) **fails on the old code** (probes `olm`, finds it absent → `encryption=False` → assertion fails) and **passes after the fix** (probes `vodozemac` → `encryption` stays `True`). This is real proof the preflight now uses the correct backend.

```text
$ pytest tests/unit/channels/test_matrix.py -q
...
63 passed, 5 warnings in 0.73s

$ pytest tests/unit/channels/test_matrix.py -k preflight
test_preflight_keeps_encryption_with_vodozemac PASSED   # failed before the fix
test_preflight_falls_back_to_olm_when_no_vodozemac PASSED
test_preflight_disables_encryption_when_no_backend PASSED
```

- `pre-commit run --files` on both changed files: black / mypy / flake8 / pylint / trailing-whitespace all Passed.

### Verification scope

The unit test covers the preflight backend-selection logic (the actual bug). End-to-end decryption requires a Matrix homeserver + an encrypted room + a `matrix-nio[e2e]` install; the reporter's environment (#6476) is the natural place to confirm the full E2EE path now stays enabled. The fix is grounded in matrix-nio's own `[e2e]` extra metadata (it installs `vodozemac`), so probing `vodozemac` is correct by construction.

## Checklist

- [x] I ran tests locally (`pytest tests/unit/channels/test_matrix.py`) and they pass (63 passed)
- [x] `pre-commit run --files` on changed files passes (black/mypy/flake8/pylint)
- [x] Documentation: n/a — no user-facing API change; the updated install hint in the log message is self-documenting. (A docs note recommending `matrix-nio[e2e]` for E2EE could be added to `website/public/docs/channels.*.md` in a follow-up if maintainers want.)
- [x] Ready for review

### DIFF
--- a/src/qwenpaw/app/channels/matrix/channel.py
+++ b/src/qwenpaw/app/channels/matrix/channel.py
@@ -424,16 +424,11 @@ class MatrixChannel(BaseChannel):
         )
 
     def _preflight_e2ee_dependencies(self) -> None:
-        """Disable E2EE if matrix-nio has no crypto backend available.
+        """Disable E2EE when matrix-nio has no crypto backend.
 
-        Check ``nio.crypto.ENCRYPTION_ENABLED`` — the exact flag
-        ``AsyncClientConfig`` validates against — rather than probing
-        module names, so this can never disagree with what matrix-nio
-        actually does across versions (0.25 checks ``olm``, 0.26 checks
-        ``vodozemac``). Probing module names risks enabling encryption
-        when the installed matrix-nio can't use the backend, which makes
-        ``AsyncClientConfig.__post_init__`` raise ``ImportWarning`` and
-        crashes channel startup (see #6476 review feedback).
+        Checks ``nio.crypto.ENCRYPTION_ENABLED`` — the same flag
+        ``AsyncClientConfig`` validates — instead of probing backend
+        module names, which are version-dependent (#6476).
         """
         if not self.encryption:
             return
--- a/src/qwenpaw/app/channels/matrix/channel.py
+++ b/src/qwenpaw/app/channels/matrix/channel.py
@@ -424,18 +424,32 @@ class MatrixChannel(BaseChannel):
         )
 
     def _preflight_e2ee_dependencies(self) -> None:
-        """Probe olm before creating AsyncClientConfig;
-        disable E2EE if absent."""
+        """Probe an E2EE crypto backend before creating AsyncClientConfig;
+        disable E2EE if none is available.
+
+        matrix-nio's modern E2EE uses ``vodozemac`` (installed via the
+        ``matrix-nio[e2e]`` extra) and works on Python 3.12+. The legacy
+        ``olm`` Python bindings are also supported but depend on ``jsmin``,
+        which no longer builds on Python 3.12 (see #6476), so probe
+        vodozemac first and fall back to olm.
+        """
         if not self.encryption:
             return
         try:
+            importlib.import_module("vodozemac")
+            return
+        except ImportError:
+            pass
+        try:
             importlib.import_module("olm")
+            return
         except ImportError:
             logger.error(
-                "MatrixChannel: olm not installed — falling back to "
-                "non-encrypted mode. "
-                "To enable E2EE: pip install matrix-nio[e2e] && "
-                "apt/dnf install libolm-dev",
+                "MatrixChannel: no E2EE crypto backend installed — "
+                "falling back to non-encrypted mode. To enable E2EE: "
+                "pip install 'matrix-nio[e2e]' (vodozemac; works on "
+                "Python 3.12+). The legacy `olm` backend is also "
+                "supported but is broken on Python 3.12+.",
             )
             self.encryption = False
 
--- a/src/qwenpaw/app/channels/matrix/channel.py
+++ b/src/qwenpaw/app/channels/matrix/channel.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 
 import asyncio
 import html
-import importlib
 import inspect
 import io
 import logging
@@ -51,6 +50,7 @@ from nio import (
     ToDeviceError,
     UploadResponse,
 )
+from nio.crypto import ENCRYPTION_ENABLED
 from nio.event_builders.direct_messages import ToDeviceMessage
 from nio.events.to_device import RoomKeyRequest, RoomKeyRequestCancellation
 from nio.responses import (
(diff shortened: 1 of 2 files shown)


hongxicheng:
Thanks for the fix — the root-cause analysis is spot on. One concern
with the approach, though:

## Problem

Probing `vodozemac` / `olm` module names duplicates matrix-nio's
internal backend detection, which is version-dependent and exclusive:

- **0.25.x**: `ENCRYPTION_ENABLED = package_installed("olm")`
- **0.26.0**: `ENCRYPTION_ENABLED = package_installed("vodozemac")`
  (olm no longer accepted)

Since we pin `matrix-nio>=0.25.0`, a mismatch is reachable (e.g. nio
0.25 + only vodozemac, or nio 0.26 + only olm). In those cases the
new preflight keeps `encryption=True`, but
`AsyncClientConfig.__post_init__` then raises `ImportWarning` — turning
the previous silent fallback into a channel startup crash.

## Suggested fix

Probe the exact flag that `AsyncClientConfig` validates against — it
is importable in both versions, with or without a backend installed:

```python
# top of channel.py (importlib becomes unused, can be removed)
from nio.crypto import ENCRYPTION_ENABLED
```

```python
def _preflight_e2ee_dependencies(self) -> None:
    if not self.encryption:
        return
    if ENCRYPTION_ENABLED:
        return
    logger.error(
        "MatrixChannel: matrix-nio has no E2EE backend "
        "available — falling back to non-encrypted mode. "
        "To enable E2EE: pip install 'matrix-nio[e2e]' "
        "(matrix-nio >= 0.26 on Python 3.12+).",
    )
    self.encryption = False
```

This can never disagree with what nio actually does, on any version,
and keeps the graceful fallback intact. The rest of the channel needs
no changes — I checked every nio API it uses against both 0.25.2 and
0.26.0; they are identical.

For the tests: patch
`qwenpaw.app.channels.matrix.channel.ENCRYPTION_ENABLED` instead of
`importlib.import_module`, so they test the real contract. Two cases
(flag True / False) are enough.

The direction is right — just anchor the probe to nio's own flag
instead of hardcoded module names.

WilShi:
Pushed in 0b3dfba3: the preflight now checks `nio.crypto.ENCRYPTION_ENABLED` (the flag `AsyncClientConfig` validates against) instead of probing module names — so it tracks matrix-nio's version-dependent backend detection and can't enable encryption when nio can't use the backend (no more `ImportWarning` crash on a version mismatch). Dropped the now-unused `importlib`; tests patch `qwenpaw.app.channels.matrix.channel.ENCRYPTION_ENABLED` (True / False, 2 cases) instead of `importlib.import_module`.

hongxicheng:
Verified the updated fix locally on both ends of the version range:

- **matrix-nio 0.25.2, no backend**: preflight disables E2EE with the
  new hint; `AsyncClientConfig` creation no longer hits the
  `ImportWarning` crash path.
- **matrix-nio 0.26.0 + vodozemac** (fresh venv, same setup as the
  #6476 reporter): `ENCRYPTION_ENABLED` is `True`, preflight keeps
  encryption enabled, and the full matrix unit suite passes (62/62)
  against the real 0.26 API. Also did a real vodozemac
  Megolm/Olm encrypt–decrypt round trip and a nio client bootstrap
  with `encryption_enabled=True` + a real store — all good.

LGTM.

nit (non-blocking): the docstring could be tighter — the rationale
for using `ENCRYPTION_ENABLED` is worth keeping, but the
failure-mode narrative of the previous approach belongs in the
commit message rather than the code. Something like:

```python
"""Disable E2EE when matrix-nio has no crypto backend.

Checks ``nio.crypto.ENCRYPTION_ENABLED`` — the same flag
``AsyncClientConfig`` validates — instead of probing backend
module names, which are version-dependent (#6476).
"""
```

Fine to address in this PR.

WilShi:
Nit addressed in 131e894a: docstring tightened to the suggested version; the previous-approach failure-mode narrative stays in commit 0b3dfba3.

WilShi:
Nit addressed in 131e894a: docstring tightened to the suggested version; the previous-approach failure-mode narrative stays in commit 0b3dfba3.

lalaliat (approved):

PR #6495, opened 2026-07-27
fix(video): deliver video data to models across OpenAI Response, Anthropic, and Chat Completions providers

### LINKED ISSUE #6474: [Bug]: `view_video` returns "Video loaded" but video DataBlock is silently dropped before reaching the model — no code path actually delivers video to the LLM
## QwenPaw Version

2.0.1

## Description

`view_video` reports success and the model's provider config has `supports_video: true`, but **no formatter in the pipeline actually serializes a `video/*` DataBlock into the API request**, so the model never receives the video bytes. The agent therefore cannot perceive any video content even though `view_video` appears to have "loaded" it.

There are two compounding problems:

1. **All OpenAI/Anthropic formatters silently drop video blocks.**  
   `OpenAIResponseFormatter`, `OpenAIChatFormatter`, `OpenAIChatMultiAgentFormatter`, `AnthropicChatFormatter`, and `AnthropicMultiAgentFormatter` all have `input_types` that omit `video/*`. Their shared `_format_openai_data_block` only handles `main_type == "image"` and `"audio"`, logging `Unsupported main media type video` and returning `None`. The Anthropic formatter even hard-rejects any non-`image/*` media type. Only `_DashScopeFormatterBase` supports video.

2. **Capability probing and real inference use two different code paths.**  
   `OpenAIProvider._try_video_url` probes video support by directly calling `client.chat.completions.create(...)` with a raw `{"type": "video_url", "video_url": {"url": ...}}` content part — completely bypassing the formatter. So the probe returns `supports_video: true` for providers like Volcengine Ark that accept `video_url` (e.g. `doubao-seed-evolving`), but once the conversation actually runs, the formatter throws the video block away. The flag says "video supported" while the wire format never sends it.

3. **`_CappingOpenAIResponseFormatter` has no `_format_video_source` override** (only `_format_image_source`), so even if video support were added to the base formatter the capping layer would not know how to handle / cap video. The cap default (`MAX_INLINE_MEDIA_BYTES = 2 MiB`) is also too small for typical MP4 files (tens of MB), so even with a correct formatter most real videos would be replaced by the `[video omitted ...]` placeholder.

The net effect: the built-in `view_video` tool is non-functional on the OpenAI-compatible Response/Chat providers (Volcengine Ark in particular), which is surprising because the multi-modal probe claims video works.

**Security considerations:** None. This is a data-pipeline bug; no credentials or user data are exposed.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

Affected files:
- `src/qwenpaw/agents/tools/view_media.py`
- `src/qwenpaw/providers/openai_provider.py` (`_try_video_url`, `_probe_video_support`)
- `src/qwenpaw/providers/openai_response_provider.py`
- `src/qwenpaw/providers/capping_formatter.py`
- (upstream agentscope) `formatter/_openai_formatter.py`, `formatter/_openai_response_formatter.py`, `formatter/_anthropic_formatter.py`

## Environment

- **QwenPaw version:** 2.0.1 (installed via Docker, package at `/app/venv/lib/python3.11/site-packages`)
- **OS:** Debian GNU/Linux 12 (bookworm) (Docker container)
- **Install method:** Docker
- **Python version:** 3.11.2
- **Provider tested:** Volcengine Ark Agent Plan (`ark-token-plan`), model `doubao-seed-evolving` with `supports_video: true, probe_source: "probed"`

## Steps to Reproduce

1. Configure a provider whose OpenAI-compatible endpoint accepts `video_url` input (e.g. Volcengine Ark / BytePlus ModelArk with `doubao-seed-evolving`), confirm `supports_video: true` is set on the model.
2. In a chat session, ask the agent to analyze a local `.mp4` file using the `view_video` tool (or let the agent call it on its own).
3. Observe the tool returns:
   ```
   Video loaded: <filename>.mp4
   ```
   with no fallback/warning hint (because `_check_multimodal_support("video")` returns True from the probe).
4. Ask the model what it sees in the video.

## Actual vs Expected

- **Actual:**
  - The model cannot see the video. It only receives the literal text `"Video loaded: <filename>.mp4"`.
  - In debug logs the formatter emits: `Unsupported media type video/mp4 for OpenAI API. Supported types: text/plain, image/*` (for the Chat formatter) or silently drops the block on the Response/Anthropic path.
  - The capping formatter `_CappingOpenAIResponseFormatter` also has no `_format_video_source`, so even a patched base formatter would fail here.
  - Result: agents either hallucinate video content (as we observed on 2026-07-20 with an Arknights fan video) or honestly report they cannot see anything.

- **Expected:**
  - When a model is flagged `supports_video: true` and the provider's OpenAI-compatible endpoint accepts `video_url` content parts, `view_video` should actually deliver the video bytes (base64 data URI or HTTP URL) to the model inside the formatted request, the same way `view_image` delivers images.
  - The formatter(s) used by the OpenAI-compatible providers (at least `OpenAIResponseFormatter` / `OpenAIChatFormatter` and their capping subclasses) should support a `video/*` input type and emit `{"type": "video_url", "video_url": {"url": ...}}` (or the Responses API's equivalent `input_video` if/when applicable).
  - The capping layer should implement `_format_video_source` and use a cap appropriate for video sizes (or a strategy that avoids inlining large videos, e.g. refusing to inline above the cap and returning a clear hint, or serving via a local HTTP URL).
  - The video capability probe should exercise the **same formatter/pipeline** used for real inference (or the formatter must be aligned with the probe) so that `supports_video` does not advertise a capability that the runtime cannot actually deliver.

## Logs / Screenshots

From `agentscope/formatter/_openai_formatter.py::_format_openai_data_block` (the code that actually drops the block):

```python
# _openai_formatter.py lines 51-68
if not any(
    fnmatch(block.source.media_type, pattern)
    for pattern in self.supported_input_media_types
):
    logger.warning(
        "Unsupported media type %s for OpenAI API. "
        "Supported types: %s. This block will be skipped.",
        block.source.media_type,
        ", ".join(self.supported_input_media_types),
    )
    return None

main_type = block.source.media_type.split("/")[0]
if main_type == "image":
    return self._format_image_source(block.source)
if main_type == "audio":
    return self._format_audio_source(block.source)
# video falls through here → warning + None → dropped
logger.warning(
    "Unsupported main media type %s for OpenAI API. "
    "This block will be skipped.",
    main_type,
)
return None
```

From `_openai_response_formatter.py` (note the default `input_types` omits video):

```python
input_types: list[str] = Field(
    default_factory=lambda: ["text/plain", "image/*"],
    description='Defaults to ``["text/plain", "image/*"]``. '
                'Audio is not supported by the Responses API.',
)
```

From `src/qwenpaw/providers/capping_formatter.py` (no `_format_video_source` for the OpenAI Response capping formatter):

```python
class _CappingOpenAIResponseFormatter(
    OpenAIResponseFormatter, CappingMixin
):
    def _format_image_source(self, source): ...
    # _format_video_source is MISSING
```

From `src/qwenpaw/providers/openai_provider.py::_try_video_url` (the probe path that bypasses the formatter and therefore succeeds, producing a misleading `supports_video: true`):

```python
res = await client.chat.completions.create(
    model=model_id,
    messages=[{
        "role": "user",
        "content": [
            {"type": "video_url", "video_url": {"url": video_url}},
            {"type": "text", "text": "What is the single dominant color ..."},
        ],
    }],
    timeout=req_timeout,
    **_token_limit_kwargs(model_id, 200),
)
```

Volcengine Ark / BytePlus ModelArk explicitly documents that its Responses API supports video input for visual models version 250615 and later (https://docs.byteplus.com/ja/docs/ModelArk/1958521), so the probe's belief that `doubao-seed-evolving` supports video is correct — the failure is purely in the QwenPaw/agentscope formatter layer.

### Why this is hard for users to self-diagnose

- The tool returns a success message (`Video loaded: ...`) instead of the multimodal-fallback hint, because `_check_multimodal_support("video")` returns True from the cached probe.
- The video block is dropped at `WARNING` level, not `ERROR`, so in default log levels it is easy to miss.
- The agent then either hallucinates (as we hit earlier) or gives a vague answer; nothing on the surface tells the user the video never reached the model.

### Suggested fix direction

1. Add `video/*` to the accepted `input_types` for `OpenAIChatFormatter` (and its multi-agent variant), and implement `_format_video_source` mirroring `_format_image_source` so it emits `{"type": "video_url", "video_url": {"url": ...}}` (base64-inlining local file:// URLs, the same way images are inlined). For the Responses API path, confirm the correct content-part name with Volcengine (OpenAI official Responses API does not support video, but Volcengine's extended Responses API does — likely `input_video` or `video_url`); gate the support behind a provider capability flag so stock OpenAI endpoints don't receive unknown fields.
2. Add `_format_video_source` to `_CappingOpenAIResponseFormatter` (and the OpenAI Chat capping formatter) and choose a cap policy for video — either a much larger cap (e.g. 50–100 MiB) or a strategy that replaces over-cap local videos with a clear text hint instead of silently base64-ing 50 MB into every turn.
3. Either (a) route the video probe through the same formatter + `model._call_api` path so probe success implies runtime success, or (b) validate that the formatter's `supported_input_media_types` matches the probed capability and warn/override when they diverge.
4. Same video support should be added to `_CappingOpenAIFormatter` (Chat completions) so the fix covers both API styles.

### Workaround

Until this is fixed, the only way to "watch" a video is to use `ffmpeg` to extract keyframes and pass them through `view_image`, which defeats the purpose of the dedicated `view_video` tool.

Reference: DashScope already implements this end-to-end correctly (`_DashScopeFormatterBase._format_video_source` + `_CappingDashScopeFormatter._format_video_source`), so that implementation can serve as a template for the OpenAI-compatible providers.

qbc2016:
## Description

[Describe what this PR does and why]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

## Testing

[How to test these changes]

## Evidence

Examples of valid evidence:
- CI artifact link

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]

### DIFF
--- a/src/qwenpaw/agents/model_factory.py
+++ b/src/qwenpaw/agents/model_factory.py
@@ -178,21 +178,34 @@ def _anthropic_media_dedup_key(source: Any) -> str | None:
     return None
 
 
-def _video_oversize_placeholder(size: int) -> dict:
-    """Text placeholder substituted for a video that exceeds the inline cap.
+def _video_oversize_placeholder(
+    size: int,
+    *,
+    response_api: bool = False,
+) -> dict:
+    """Text placeholder substituted for a video that exceeds
+    the inline cap.
 
     Mirrors the wording used by ``capping_formatter``'s
-    ``CappingFormatterMixin._placeholder_text`` so oversized-video messages
-    are consistent across every provider path.  Tool-result videos inline
-    through these helpers bypass the capping formatters (which only see
-    ``_format_*_source``), so the cap is enforced here instead.
+    ``CappingFormatterMixin._placeholder_text`` so
+    oversized-video messages are consistent across every
+    provider path.  Tool-result videos inline through these
+    helpers bypass the capping formatters (which only see
+    ``_format_*_source``), so the cap is enforced here
+    instead.
+
+    When *response_api* is True the block uses
+    ``input_text`` instead of ``text``, matching the
+    Responses API content-type convention.
     """
+    txt_type = "input_text" if response_api else "text"
     return {
-        "type": "text",
+        "type": txt_type,
         "text": (
-            f"[video omitted from model context: local file is "
-            f"{size} bytes, exceeds inline limit of "
-            f"{MAX_INLINE_MEDIA_BYTES} bytes]"
+            "[video omitted from model context: "
+            f"local file is {size} bytes, exceeds "
+            f"inline limit of {MAX_INLINE_MEDIA_BYTES}"
+            " bytes]"
         ),
     }
 
@@ -272,43 +285,60 @@ def _format_anthropic_video_data_block(block: Any) -> dict | None:
     return None
 
 
-def _format_openai_video_block(video_block: dict) -> dict:
+def _format_openai_video_block(
+    video_block: dict,
+    response_api: bool = False,
+) -> dict:
     """Format a video block for OpenAI-compatible API.
 
     Local files are converted to base64 data URLs; web URLs are
     passed through directly.
 
+    When ``response_api`` is True the output uses the
+    ``input_video`` content type adopted by Volcengine Ark
+    and other providers that extend the OpenAI Responses API
+    with native video support.  Official OpenAI and DashScope
+    Responses APIs do **not** support video; callers should
+    fall back gracefully (the react-agent already retries
+    without media on 400 errors).
+
     Args:
-        video_block (`dict`):
-            The video block to format.
+        video_block: The video block to format.
+        response_api: When True, emit the Responses API
+            ``input_video`` shape instead of the Chat
+            Completions ``video_url`` shape.
 
     Returns:
-        `dict`:
-            ``{"type": "video_url", "video_url": {"url": ...}}``.
+        Wire-format dict for the provider.
 
     Raises:
-        `ValueError`:
+        ModelFormatterError:
             If the source type or video format is not supported.
     """
     source = video_block["source"]
     if source["type"] == "base64":
         media_type = source["media_type"]
-        # base64 length -> approximate raw byte count.
         size = len(source.get("data") or "") * 3 // 4
         if size > MAX_INLINE_MEDIA_BYTES:
-            return _video_oversize_placeholder(size)
+            return _video_oversize_placeholder(
+                size,
+                response_api=response_api,
+            )
         url = f"data:{media_type};base64,{source['data']}"
     elif source["type"] == "url":
         raw_url = _file_url_to_path(source["url"])
-        if os.path.exists(raw_url) and os.path.isfile(raw_url):
-            # Cap oversized local files before reading/encoding the whole
-            # thing into the request body (see ``capping_formatter``).
+        if os.path.exists(raw_url) and os.path.isfile(
+            raw_url,
+        ):
             try:
                 size = os.path.getsize(raw_url)
             except OSError:
                 size = 0
             if size > MAX_INLINE_MEDIA_BYTES:
-                return _video_oversize_placeholder(size)
+                return _video_oversize_placeholder(
+                    size,
+                    response_api=response_api,
+                )
             ext = os.path.splitext(raw_url)[1].lower()
             media_type = _SUPPORTED_VIDEO_EXTENSIONS.get(ext)
             if not media_type:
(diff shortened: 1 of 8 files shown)


lalaliat (approved):
LGTM

PR #6500, opened 2026-07-27
fix(browser): make unauthenticated local CDP exposure opt-in

xiaoming-qxm:
## Description

`browser_use(action="start")` previously launched Chrome with an unauthenticated local DevTools TCP port by default. Any local process able to reach that port could attach to the browser session, inspect logged-in pages and cookies, and execute browser actions.

This PR makes managed CDP an explicit opt-in: the default uses Playwright's pipe transport and opens no TCP debugging port; passing a positive `cdp_port` remains the intentional way to share the browser with another local tool.

The review-fix round closes the remaining bypass and state-lifecycle gaps:

1. **No-debug-port default** — managed CDP is selected only for an explicit positive `cdp_port`.
2. **No prefix-form bypass** — the `browser_args` guard now rejects `--remote-debugging-port` and `--remote-debugging-address` in Chromium's `--x`, `-x`, and Windows `/x` forms, case-insensitively. This is an accident guard, not a security boundary: `browser_args` intentionally retains full local command-line power for all other arguments.
3. **Launch intent is current and explicit stops clear it** — `_expose_cdp` / `_cdp_port` are recorded after already-running returns but before every preflight check. An explicit `action="stop"` clears both keys, so a later auto-start returns to the no-debug-port default. Crash recovery retains its original async launch mode when the session was not explicitly stopped.
4. **Conflicting arguments are rejected** — `private_mode=true` plus `cdp_port` returns `ok: false` instead of silently selecting a mode.
5. **Profile directories are owner-only** — managed-CDP and async persistent-profile directories are set to `0700`.
6. **Agent-facing documentation matches shipped behavior** — the `browser_cdp` bodies now describe the no-port default, managed-CDP examples carry an explicit `cdp_port`, and both `browser_visible` bodies remove obsolete `private_mode` decision rules. All four edited skills bump `builtin_skill_version`, which is required because pool skills are version-compared snapshots; content-only edits would not update existing user copies.

Scope stays intentionally narrow: `connect_cdp`, `list_cdp_targets`, `_chromium_launch_args()`, and the Windows sync launch path are unchanged. No new configuration surface is added.

**Related Issue:** N/A (external security report, no issue filed)

**Security Considerations:** The default unauthenticated TCP DevTools surface is removed. Explicit `cdp_port`, external CDP attachment, and port scanning remain available only when deliberately requested.

## Breaking Changes / Migration

- Sharing the agent browser with another local tool now requires an explicit positive `cdp_port`; `cdp_port=0` opens no port and does not auto-select one.
- `browser_args` containing either blocked remote-debugging switch now fails with guidance to use `cdp_port`.
- On macOS, **every switch between the default pipe path and an explicit `cdp_port` can drop login state**. Both paths share the profile directory but use different cookie-encryption backends, so users may need to sign in after each mode switch.

## Known Limitations

- The Windows sync-Playwright recovery path calls `_sync_browser_launch` without `cdp_port`, so recovery fails closed by not re-exposing that port. It is outside this PR; the recovery-inheritance claim applies to the async path.
- On Windows, `shlex.split(posix=False)` preserves quotes, so a quoted `"-remote-debugging-port=9222"` does not match a switch prefix. Chrome also receives the leading quote and does not treat it as a switch, so this is not a bypass.

## Not Adopted

The real-browser smoke probe is deliberately not added to the automated suite: it requires CI-side Chrome, process spawning, and listening-port inspection. Manual probe evidence is included below.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran relevant focused tests locally and they pass
- [ ] Ready for review

Not applicable.

## Testing

- Focused launch-security regression suite: **17 passed** after the implementation.
- Reverse verification against the pre-fix source: **7 failed, 10 passed**; failures covered the single-dash, uppercase, and Windows-slash guard forms plus stale port intent.
- `.venv/bin/pre-commit run --all-files` passed before every review-fix commit.
- Real Chrome probe: `browser_use(action="start", browser_args="-remote-debugging-port=9411")` was rejected with `cdp_port` guidance, and `lsof` confirmed 0 listeners on 9411. A normal start used Playwright pipe transport and had 0 `--remote-debugging-port` processes.
- The prescribed `tests/unit/agents` regression reached 37% without a test failure, then blocked for more than six minutes waiting for unrelated `mini_mcp_server.py` and `stdio_echo_server.py` fixture processes. The stalled run was stopped and is documented rather than reported as passing.

## Evidence

```text
pre-fix reverse verification
7 failed, 10 passed in 0.86s

post-fix focused suite
17 passed in 0.22s

manual single-dash probe
browser_use(start, browser_args="-remote-debugging-port=9411")
{ "ok": false, "error": "-remote-debugging-port is not allowed in browser_args; use the cdp_port parameter instead." }
PORT_9411_LISTEN_LINES=0
DEFAULT launch_mode=playwright
PIPE_PROCESS_COUNT=3
TCP_DEBUG_PORT_PROCESS_COUNT=0

original reviewer reproduction before this fix
-remote-debugging-port=9411 opened 127.0.0.1:9411 and /json/list returned an unauthenticated webSocketDebuggerUrl.
```

## Additional Notes

- The focused regression suite lives in the repository's intentionally untracked `tests/local` path, so it is not included in the PR diff.
- The macOS authenticated-session probe confirmed the encryption-backend migration behavior described above.

### DIFF
--- a/src/qwenpaw/agents/skills/browser_cdp-en/SKILL.md
+++ b/src/qwenpaw/agents/skills/browser_cdp-en/SKILL.md
@@ -2,7 +2,7 @@
 name: browser_cdp
 description: "Use this skill when the user explicitly wants to connect to a running Chrome browser, scan local CDP ports, specify a `cdp_port`, or share a single browser across multiple agents/tools. By default browser_use opens no debugging port; pass an explicit `cdp_port` only when the user wants another local tool to attach."
 metadata:
-  builtin_skill_version: "1.2"
+  builtin_skill_version: "1.3"
   qwenpaw:
     emoji: "🔌"
     requires: {}
@@ -10,7 +10,7 @@ metadata:
 
 # Browser CDP Reference
 
-By default, **browser_use** launches and manages the local Chrome/Chromium via **managed CDP**, but this does not mean the CDP port should be exposed to the user or other tools every time.
+By default, **browser_use** opens no debugging port: the browser is managed directly by Playwright over a pipe. This skill covers the cases where a CDP port is wanted on purpose.
 
 This skill focuses on more "explicit" CDP usage:
 
@@ -21,12 +21,12 @@ This skill focuses on more "explicit" CDP usage:
 
 In other words:
 
-- The default `start` uses managed CDP under the hood, but users typically do not need to understand or be aware of CDP details
+- The default `start` opens no debugging port, so users typically do not need to understand or be aware of CDP details
 - Only enter the scope of this skill when the user explicitly mentions "connect to an existing browser / scan ports / specify a port / share a browser"
 
 > **Privacy Recommendation**
 >
-> If the user does not want to expose browser history, cookies, page content, or session data, recommend using `private_mode=true`, which switches to Playwright-managed mode.
+> The default `start` exposes nothing — no debugging port is opened. Only pass `cdp_port` when the user explicitly wants another local tool to attach, and warn about the exposure first.
 
 > **Warning: One browser instance per workspace**
 >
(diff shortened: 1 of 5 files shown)


cuiyuebing (approved):
LGTM

PR #6522, opened 2026-07-28
fix: retain dirty flag on token usage flush failure

### LINKED ISSUE #6374: [Bug]: token usage persistence does not retry after a transient write failure
## QwenPaw Version

Current `main` commit: `4acf3418b0d2a5a66726308c881a426f9e189f9b`.

## Description

`TokenUsageBuffer._flush_once()` clears `_dirty` before calling
`save_data_sync()`. The storage helper catches `OSError`, logs it, and returns
without signalling failure. As a result, a transient atomic-write failure marks
the in-memory data clean even though nothing reached disk. Later periodic
flushes skip the data until a new usage event happens to set `_dirty` again.

## Steps to Reproduce

1. Seed a `TokenUsageBuffer` and add a processed usage event to its cache.
2. Make the first `os.replace()` inside `save_data_sync()` raise `OSError`.
3. Run `_flush_once()`, restore `os.replace()`, then run `_flush_once()` again
   without enqueueing another event.

## Actual vs Expected

- **Actual:** the first failure is logged, but `_dirty` becomes `False`; the
  next flush performs no write and the usage file remains absent.
- **Expected:** a failed atomic write must leave the buffer dirty so a later
  periodic flush retries it. A successful write must still preserve dirty state
  set by events arriving while the snapshot is being written.

## Logs / Screenshots

```text
baseline dirty_after_failed_flush=False
baseline write_attempts_after_retry=1
baseline file_exists_after_retry=False
```

## Additional Notes

A narrow fix can have `save_data_sync()` report write success and restore the
dirty flag when persistence fails. It need not change token accounting, file
format, flush interval, or public configuration.

RerankerGuo:
## Summary

This PR fixes #6374 by ensuring that failed token usage persistence flushes do not silently lose data.

## Problem

In `TokenUsageBuffer._flush_once()`, the `_dirty` flag was cleared **before** calling `save_data_sync()`. When a transient I/O failure occurred:
1. `_dirty` was already set to `False`
2. `save_data_sync()` caught the `OSError`, logged it, and returned without signaling failure
3. Subsequent periodic flushes saw `_dirty=False` and skipped the write
4. Token usage data was lost until a new event happened

## Changes

- **`storage.py`**: `save_data_sync()` now returns `bool` — `True` on success, `False` on failure
- **`buffer.py`**: `_flush_once()` only clears `_dirty` after a successful write. Failed writes retain the dirty flag so the next periodic flush retries automatically.
- **`test_token_usage.py`**: Added two regression tests:
  - `test_flush_once_retains_dirty_flag_on_failure` — verifies dirty flag persists after failed flush and clears after retry
  - `test_flush_once_success_clears_dirty_flag` — verifies normal successful flush still works

## Testing

- [x] Manual async test verified: dirty=True after failed flush, dirty=False after retry
- [x] Existing token usage tests still pass (tested via import and direct execution)
- [x] No public API changes — `save_data_sync()` return type changed from `None` to `bool`, but it was not used as a return value anywhere else

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/qwenpaw/token_usage/buffer.py
+++ b/src/qwenpaw/token_usage/buffer.py
@@ -146,19 +146,25 @@ class TokenUsageBuffer:
                     break
 
     async def _flush_once(self, force: bool = False) -> None:
-        """Write ``_disk_cache`` to disk if dirty."""
+        """Write ``_disk_cache`` to disk if dirty.
+
+        The dirty flag is cleared **only** after a successful write so that
+        transient I/O failures are retried by the next periodic flush.
+        """
         if not self._cache_loaded:
-            # The cache was never seeded, so no event can have reached
-            # ``_disk_cache`` and it still holds the initial empty dict.
-            # Writing it out would replace the stored history with ``{}``.
             return
         if not self._dirty and not force:
             return
-        self._dirty = False
 
         snapshot = copy.deepcopy(self._disk_cache)
-        await asyncio.to_thread(save_data_sync, self._path, snapshot)
-        logger.debug("token_usage: flushed cache to disk")
+        ok = await asyncio.to_thread(save_data_sync, self._path, snapshot)
+        if ok:
+            self._dirty = False
+            logger.debug("token_usage: flushed cache to disk")
+        else:
+            logger.warning(
+                "token_usage: flush failed, retaining dirty flag for retry",
+            )
 
     async def _flush_loop(self) -> None:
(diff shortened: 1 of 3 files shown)


yuanxs21:
This issue has been addressed by an earlier pr #6402 similarly. Thank you for the contribution!

Issue #6520, opened 2026-07-28
agent.json systematic corruption: BOM, missing quotes, double-encoding


easyaha:
## Summary

`agent.json` suffered **systemic, distributed corruption** across ~20+ fields, causing complete system failure. Three damage types found: BOM header, missing closing quotes in string values, and double-encoded Chinese text.

## Environment

- **QwenPaw version**: 2.0.1
- **OS**: Windows 10 (AMD64)
- **Agent**: `default` (id=`default`, name=`药老(YaoLao)`)
- **First failure**: 2026-07-27 16:18 CST

## Symptoms

When user tried to switch models in the UI:

1. **Error 1**: `Failed to save active model to agent config`
2. **Error 2** (in logs): `Unexpected UTF-8 BOM (decode using utf-8-sig): line 1 column 1 (char 0)`
3. **System crash**: runtime.py:171 repeated unhandled errors, workspace.py:504 agent failed to start

## Timeline (from qwenpaw.log)

| Time | Event |
|------|-------|
| 12:07 | `config.1cb6c90e.channel-display-migrate.bak` created (clean, valid JSON) |
| 12:13 | `agent.a65ca840.channel-display-migrate.bak` created (clean, valid JSON, 23,266 B) |
| 16:18 | **First BOM error** at `runtime.py:171` — system crash onset |
| 16:25 | `providers.py:675` Failed to get agent-specific model (BOM) |
| 16:25 | `providers.py:760` Failed to save active model (BOM) |
| 16:26 | `workspace.py:504` Failed to start agent instance (BOM) |
| (next day) | User found QwenPaw broken, used external tool to diagnose and fix |

### File size comparison

| File | Size | BOM? | JSON valid? |
|------|------|------|-------------|
| Migration backup (`agent.a65ca840.*`) | 23,266 B | ❌ No | ✅ Yes |
| Config migration backup (`config.1cb6c90e.*`) | 26,451 B | ❌ No | ✅ Yes |
| Corrupted file (backup) | 26,549 B | ✅ **Yes** (EF BB BF) | ❌ **No** |
| Current (rebuilt) | 24,811 B | ❌ No | ✅ Yes |

### Damage types found in corrupted file

1. **UTF-8 BOM** at position 0 (`EF BB BF`) — Python's `json.loads` throws immediately
2. **~20 `icon` fields** missing closing `"` — string terminator lost, subsequent structure corrupted
3. **`description` field** missing closing `"` — with literal newlines inside value
[...]

PR #6528, opened 2026-07-28
fix: resolve agent.json corruption (#6520)

mohitdebian:
## What Problem This Solves
Relative paths specified in agent.json were being resolved relative to the current working directory (CWD) where the app was launched from, which caused file-not-found errors when the app was launched from different directories. 

## Evidence
After this fix, paths are correctly resolved relative to the location of the agent.json file itself, regardless of where the app is launched from.


RerankerGuo:
Independent review at exact head `fc51d4dc6e7e66c2c138153f6324db296731e0b5` on macOS arm64 / Python 3.11.15.

The BOM path works as intended, and the focused suites are green:

```text
exact head:
tests/unit/utils/test_io_utils.py       18 passed
tests/unit/config                       13 passed

current-main bacac741 trial merge:
utils + config                          31 passed

Black 23.3.0 (the repository-pinned version), Flake8, and
git diff --check                        passed
```

I found one data-safety gap in the new backup path. `write_text_atomic()` catches every `OSError` from `shutil.copy2()` and then unconditionally publishes the new file with `os.replace()`:

```python
try:
    shutil.copy2(target, target.with_suffix(".bak.1"))
except OSError:
    pass

os.replace(temp_path, target)
```

An external failure-injection contract makes the observable result explicit:

```python
path.write_text('{"version": 1}', encoding="utf-8")
with patch("shutil.copy2", side_effect=OSError("backup unavailable")):
    write_json_atomic(path, {"version": 2}, backup_count=3)
```

```text
expected:
OSError; target remains {"version": 1}

exact head:
no exception
target={"version": 2}
backup_exists=False

current-main trial merge:
no exception
target={"version": 2}
backup_exists=False
```

This is the failure mode where the backup is most needed (for example disk-full, ACL, antivirus, or filesystem errors), but the caller receives success and the only current config is replaced without any recoverable previous state. It contradicts the PR's stated guarantee that previous valid states are preserved before replacement.

I recommend making creation of `.bak.1` fail-closed when `backup_count > 0`: if the backup copy or required rotation cannot complete, propagate the error and leave `target` untouched. A regression should inject `copy2` failure and assert both the exception and original target contents. If best-effort backups are intentional instead, the PR should not present them as recovery protection, but that would leave the issue's rollback requirement unresolved.

The branch is 79 main commits behind, but a current-main trial merge is clean and reproduces the same behavior, so this is not stale-base drift.

mohitdebian:
Fixed the silent OSError catch. It now fails closed and leaves the original file intact if the backup fails. Added the failure-injection test case too.

RerankerGuo (approved):
Revalidated the author's fail-closed fix at exact head `c424aa44cec2a422eba39a224e686eb057855d17`.

The previous blocker is resolved: backup rotation and `.bak.1` creation errors now propagate before the final `os.replace(temp_path, target)`, so the currently valid config is not replaced after a backup failure.

Independent verification on macOS arm64 / Python 3.11.15:

```text
exact head:
  focused io/config + external failure contracts   34 passed
  affected config-save callers                    113 passed

current-main bacac741 trial tree 78d22ca3:
  focused io/config + external failure contracts   34 passed

repository-pinned changed-file hooks:
  AST, encoding, private-key, whitespace, commas,
  Mypy, Black, Flake8, Pylint                      passed

git diff --check                                   passed
```

The external contracts independently inject both `shutil.copy2()` failure and a failure during `.bak.N` rotation. In both exact and current-main trial trees, the exception is propagated, the original target remains byte-for-byte unchanged, and the atomic temporary file is cleaned up. Normal backup rotation, BOM loading, JSON validation, and the affected router/session/workspace save paths remain green.

I found no remaining blocker in the exercised scope.

PR #6530, opened 2026-07-28
Fix editable per-tool call limit names

rayrayraykk:
## Summary

- allow per-tool call limit names to be edited before committing on blur
- preserve trimming and reject empty or duplicate tool names
- add a regression test that verifies the renamed key is written back to the form

## Root cause

The tool-name input was controlled by the existing object key but had no
`onChange` handler. React therefore restored the old key after every keystroke,
making the field appear read-only.

## User impact

Users can now replace the generated `tool-name` placeholder with the actual
tool name when configuring a tool-call budget.

## Validation

- `npm run test:run -- src/pages/Agent/Config/components/AgentLoopCard.render.test.tsx src/pages/Agent/Config/components/AgentLoopCard.test.ts`
- `npx tsc -b --noEmit`
- targeted Prettier and ESLint checks


Osier-Yi (approved):
LGTM

Issue #6299, opened 2026-07-21
[Bug]: Deleted session records persist in history.db, causing seq collision and cross-session context contamination


arcol:
QwenPaw 2.0.0.post2
Podman 容器部署，console + onebot 频道，scroll 上下文模式

从1.0升级上来，保留了大量旧版对话内容，出现串会话、新对话丢失、F5刷新出空白页、网页无响应（CPU满）的情况，且在界面上删除会话无效。后经AI修复处理，已恢复正常，以下是AI的分析总结。

症状：

删除一个会话后再新建会话，新会话的上下文恢复时可能串入已删除会话的对话内容。同时在流式生成过程中刷新页面，该轮回复会丢失且无法恢复。

根因：

conversation_history 表的 seq 是全局自增的，但删除会话时只清理了 chats.json 索引，没有清理 history.db 中对应 session_id 的记录，也没有清理 sessions/ 下的 session 存储文件。随着会话不断创建和删除，旧记录累积导致：

seq 区间跨会话重叠——多个会话（含已删除的）占用相同的 seq 区间，上下文恢复按 seq 排序时发生串读
history.db 持续膨胀——已删除会话的记录永远残留
相关源码位置：

runtime/builder.py:674 — history.db 跨会话共享，按 session_id 区分
agents/context/scroll/memoryspace.py:461 — 按 session_id 过滤后取 MAX(seq) 恢复上下文，但 seq 本身已重叠
会话删除路径 — 未发现对 conversation_history 或 sessions/ 的清理逻辑
期望：

删除会话时同步清理 history.db 中该 session_id 的记录和 sessions/ 下的文件
或在上下文恢复时严格按 session_id + created_at 排序，不依赖可能重叠的全局 seq
流式生成被中断（如浏览器刷新）后，已生成的部分应能落库或优雅恢复，避免界面空白

wananing:
Hi, I traced this against the current `main` implementation before starting any code changes.

A few findings may help separate the symptoms:

1. `conversation_history.seq` is an `INTEGER PRIMARY KEY AUTOINCREMENT`, so sequence values are globally unique and do not overlap between sessions.
2. The current delete endpoints intentionally remove only the `ChatSpec` entry. Both `app/chats/api.py` and `app/chats/manager.py` explicitly state that persisted session state is not deleted. As a result, `sessions/` data and scroll rows remain, and the normal cross-session history search can still recall a session that disappeared from the chat index.
[...]

PR #6536, opened 2026-07-28
fix(chats): clean up persisted data on deletion

niceIrene:
## Summary

Make chat deletion clean up its unreferenced persisted data instead of only removing the entry from `chats.json`.

This addresses the deletion-related persistence issue described in #6299.

## Root cause

The existing delete endpoints only removed the corresponding `ChatSpec` from `chats.json`.

The associated data remained in:

- `sessions/` session state files
- Scroll's shared `history.db`
- the full-text search index

As a result, deleted conversations could remain searchable, consume storage, and be imported again by startup session synchronization.

## Changes

- Add a crash-recoverable chat deletion workflow using durable tombstones in `chats.json`.
- Remove session files only when no retained chat references the same physical path.
- Account for sanitized filename and channel-path collisions when checking shared references.
- Stage session files under `.trash/<deletion-id>/` before permanently removing them.
- Delete matching Scroll history rows and their FTS entries transactionally.
- Record per-session `seq` watermarks so rows written after deletion preparation are not accidentally removed.
- Block deletion while any selected chat is running.
- Prevent new chats from reusing session storage while its deletion is pending.
- Recover unfinished deletions during startup before session-to-history synchronization.
- Skip startup synchronization for workspaces whose pending deletion could not be recovered.
- Apply the same behavior to single and batch deletion endpoints.

## Failure handling

Deletion uses `chats.json` as the durable commit point:

1. Move the selected chats into a pending-deletion tombstone.
2. Stage their unreferenced session files.
3. Delete bounded history rows and FTS entries.
4. Remove staged files.
5. Finalize the tombstone.

If cleanup fails at any point, the tombstone and staged files remain available for an idempotent retry or startup recovery.

## Safety guarantees

- Deleting from an empty registry does not touch orphaned session or history data.
- Shared session files and shared history are preserved while referenced by another chat.
- Different logical session identities that resolve to the same sanitized path do not delete each other's files.
- Corrupt history databases are not quarantined or recreated by the deletion path.
- Missing history databases are not created during deletion.
- Running chats cause deletion to return a conflict without mutating persisted data.

## Validation

- `946 passed, 1 skipped`
- All pre-commit checks passed:
  - AST and format checks
  - mypy
  - black
  - flake8
  - pylint

## Out of scope

This PR does not change global `seq` allocation, `MemorySpace.expand()` isolation, or interrupted-stream recovery. Those are independent from persisted chat deletion.


RerankerGuo:
I tested this head against current `main` and found a semantic rebase blocker in
the new checkpoint persistence path.

**Versions**

- PR head: `e98bf5186743e4e1a386b9df6440415f266dec42`
- current main: `bacac7410cf2ac122dd892152beff8b03f38729f`
- merge-tree: `7f440a126a31b5872d54e5922dfe62824e5cdbc6`
- macOS, Python 3.11.15, real shadow-Git checkpoint store

Current main now calls
`CHECKPOINT_RUNTIME.delete_session_checkpoints(...)` from both single and
batch chat deletion. This head replaces those endpoint bodies with
`ChatDeletionService.delete()`, but the durable tombstone only contains session
paths and Scroll history watermarks. It has no checkpoint identity or cleanup
stage.

I replayed the deletion service on current main, created a real auto checkpoint,
then deleted the owning chat:

```text
after_tombstone_service: timeline=1 head=True
after_checkpoint_runtime: deleted_refs=1 timeline=0 head=False
1 passed
```

So the service removes the chat registry entry/session file and finalizes its
tombstone while the deleted conversation remains addressable through the
checkpoint ref and `heads.json`. Calling the current-main runtime cleanup is
what removes both.

Focused trial results:

```text
pytest -q tests/unit/app/chats/test_deletion.py \
  tests/unit/checkpoints/test_checkpoint_basic.py::test_delete_session_cancels_pending_auto_snapshot \
  tests/unit/checkpoints/test_checkpoint_basic.py::test_delete_session_does_not_create_unused_checkpoint_store

9 passed in 0.66s
```

I do not think an endpoint-only call after `ChatDeletionService.delete()` fully
preserves this PR's crash-recovery guarantee: a crash after tombstone
finalization but before that call leaves checkpoint state with no durable
deletion intent for startup recovery.

A safe rebase would likely:

1. Store the unreferenced `(session_id, user_id, channel)` identities in
   `PendingChatDeletion` (excluding identities still referenced by retained
   chats).
2. Run the idempotent runtime checkpoint cleanup before finalizing the
   tombstone, so its auto-snapshot cancellation behavior is retained.
3. Cover failure/restart: checkpoint cleanup failure leaves the tombstone, and
   startup recovery retries checkpoint cleanup before finalization.

The existing session/history deletion tests remain green; this is specifically
the checkpoint store added after the PR's base.

RerankerGuo (changes requested):
作者意图：用 durable tombstone 将 chat registry、session files 与 Scroll history 删除组成可崩溃恢复事务。

current main 后续加入了 checkpoint persistence cleanup，但本 PR 的 `PendingChatDeletion` 没有 checkpoint identity/stage。机械 rebase 后，service 会先 finalize tombstone，而 checkpoint refs 与 `heads.json` 仍保留，删除会话仍可通过 checkpoint 恢复。

```mermaid
flowchart LR
 A[Delete chat] --> B[Remove registry/session/history]
 B --> C[Finalize tombstone]
 C --> D[Checkpoint ref still exists]
 D --> E[Deleted conversation remains recoverable]
```

真实 shadow-Git checkpoint store：service 删除后 `timeline=1 head=True`；调用 current-main `delete_session_checkpoints()` 后才变为 `timeline=0 head=False`。endpoint 在 finalize 后补调用也不满足 crash recovery，二者之间崩溃会永久丢失删除意图。

rebase 必须：
1. 在 tombstone 保存未被 retained chats 引用的 `(session_id,user_id,channel)` checkpoint identities；
2. finalize 前执行幂等 runtime checkpoint cleanup，保留 pending auto-snapshot cancellation；
3. cleanup 失败时保留 tombstone，startup recovery 重试 cleanup 后再 finalize；
4. 覆盖 single/batch deletion、shared identities 与 failure/restart。

PR #6373, opened 2026-07-22
fix(channels): preserve recreated queue state

patrick-andstar:
## Description

Prevent idle queue cleanup from removing a `QueueState` recreated for the same
key while the previous consumer is finishing cancellation. A consumer now
removes its queue mapping only when the mapping still belongs to that consumer
task, so a live replacement consumer remains tracked and later enqueues reuse
it instead of creating a duplicate consumer.

**Related Issue:** Fixes #6372

**Security Considerations:** None. This changes only internal queue ownership
tracking and does not add input, permissions, or network behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [x] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [x] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

- Passed: an isolated schedule against the parent commit confirms the bug: the
  replacement state exists before the old consumer exits and is absent after it
  exits.
- Passed: the same schedule against this commit keeps the replacement state
  registered after the old consumer exits.
- Passed: `PYTHONPATH=src python -m pytest --confcutdir=tests/unit/app/channels tests/unit/app/channels/test_unified_queue_manager.py -q` (`19 passed`).
- Passed: the new race regression ran five consecutive times (`1 passed` each run).
[...]

PR #6539, opened 2026-07-29
fix(unified_queue): preserve recreated queue state

### LINKED ISSUE #6372: [Bug]: idle cleanup can remove a newly recreated queue state
## QwenPaw Version

Current `main` commit: `4acf3418b0d2a5a66726308c881a426f9e189f9b`.

## Description

`UnifiedQueueManager._cleanup_idle_queues()` removes an idle queue entry from `_queues` before cancelling and awaiting its consumer. `_run_consumer()` then unconditionally executes `self._queues.pop(queue_key, None)` in its `finally` block.

If a new payload for the same key arrives after cleanup removes the old entry but before the old consumer finishes cancellation, `enqueue()` creates and registers a new `QueueState`. When the old consumer finally exits, its unconditional `pop()` removes that new state. The new consumer continues running but is no longer tracked, so later messages can create another consumer for the same session key.

**Related PR(s):** None.

**Security considerations:** None identified; this is queue lifecycle correctness.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** current `main` commit `4acf3418b0d2a5a66726308c881a426f9e189f9b`
- **OS:** Windows 11
- **Python version:** 3.11

## Steps to Reproduce

1. Start a `UnifiedQueueManager` with a consumer that has drained its first payload and is waiting indefinitely.
2. Let idle cleanup remove that empty queue and cancel the first consumer, but pause its cancellation cleanup before it returns.
3. Enqueue another payload for the same `(channel_id, session_id, priority_level)` key, creating a new queue state.
4. Allow the first consumer to finish cancellation.
5. Observe that the key disappears from `_queues` even though the new consumer is running.

## Actual vs Expected

- **Actual:** The old consumer removes the new queue state by key, leaving a live consumer untracked and allowing duplicate consumers for the same key later.
- **Expected:** A consumer should remove a queue entry only when it is still the `QueueState` that owns that consumer.

## Logs / Screenshots

```text
baseline new_state_before_old_exit=True new_state_after_old_exit=False
```

## Additional Notes

A narrow fix can make `_run_consumer()` remove the mapping only when the current entry still belongs to that consumer's `QueueState`. It does not change idle timeout values, queue priority, or message processing policy.

RerankerGuo:
## Description

Fixes #6372.

Idle cleanup removes an old queue entry before its consumer finishes
cancellation. If a payload recreates the same key in that interval, the old
consumer's unconditional final `pop()` removes the newly created state,
leaving its live consumer untracked.

This PR passes the owning `QueueState` into the consumer and removes the
mapping only when the current entry is still that exact state. Normal idle
cleanup behavior is unchanged.

**Related Issue:** Fixes #6372

**Security Considerations:** None; this is queue lifecycle correctness.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Applicable pre-commit hooks pass
- [x] Relevant tests pass
- [x] Documentation updated (not needed; internal race fix)
- [x] Ready for review

## Testing

The regression uses independently controlled old and replacement consumers to
reproduce the recreate-before-old-exit ordering. It also verifies normal idle
queue cleanup still removes the owning state.

## Evidence

Before the fix, the deterministic reproduction observed:

```text
new_state_before_old_exit=True
new_state_after_old_exit=False
```

After rebasing onto current `main` and removing local tracker files from the
PR diff:

```bash
.venv/bin/pytest tests/unit/app/channels/test_unified_queue_manager.py -q
# 20 passed in 1.52s

SKIP=actionlint .venv/bin/pre-commit run --files \
  src/qwenpaw/app/channels/unified_queue_manager.py \
  tests/unit/app/channels/test_unified_queue_manager.py
# Python AST, encoding, private-key detection, whitespace, trailing commas,
# mypy, black, flake8, and pylint: Passed
```

`actionlint` is not applicable to the changed Python files.

## Additional Notes

The implementation was AI-assisted and personally verified against current
`main`, Issue #6372, the deterministic race regression, and the focused suite.


RerankerGuo:
Closing in favor of the earlier active implementation #6373 from the issue reporter, which covers the same race and files. I revalidated this branch against current main (20 focused tests passed), but keeping a duplicate PR would add review burden.

PR #6545, opened 2026-07-29
fix(deps): pin Codex and Qoder SDK versions

rayrayraykk:
## Summary

- pin `openai-codex` to `0.144.4`
- pin `qoder-agent-sdk` to `1.0.9`
- update the optional dependency assertions

## Why

The unbounded SDK ranges allowed patch releases with breaking API changes to
enter CI. In particular, `qoder-agent-sdk` 1.0.10 removed the
`ToolProgressMessage` export used by the Qoder harness, causing unit-test
collection to fail.

Pinning both third-party agent SDKs keeps installs reproducible and prevents
unreviewed SDK API changes from breaking the test suite.

## Validation

- `22 passed` for the Qoder adapter, event mapper, and optional dependency tests
- pre-commit checks passed for all changed files
- `git diff --check` passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,7 +124,7 @@ test = [
     "pytest-timeout>=2.3.0",
     "pytest-xdist>=3.5.0",
     "hypothesis>=6.0.0",
-    "qoder-agent-sdk>=1.0.9,<2",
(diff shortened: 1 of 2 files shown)


cuiyuebing (approved):
LGTM

Issue #3270, opened 2026-04-11
[Feature]: Skill System Optimization and Suspected Bugs


BLUE0818:
**Environment**
- Version: v1.0.2

---

**Reports**

**1 — Skill tags cannot be reused across skills**
After adding a tag to one skill, other skills cannot select (reuse) that same tag from a dropdown or existing list — requiring the user to type it in manually each time.

**2 — Tags added in Skill Pool do not sync to Agent skills**
After adding a tag to a skill in the Skill Pool, the same skill appearing in the Agent's skill list does not reflect the tag.

**3 — Tags disappear after triggering an update on a tagged skill**
Steps to reproduce:
1. Add a tag to a skill in the Skill Pool.
2. Trigger "Update" on that skill (the UI reports the skill is already up to date, so no actual update occurs).
3. Open the corresponding `skill.md` file locally.
4. **Expected:** Tag is preserved.
5. **Actual:** Tag has disappeared.

This issue is consistently reproducible.

**4 — Built-in skills cannot be loaded from the Skill Pool into an Agent, and cannot be synced back**
In the Agent's skill window, using "Load from Skill Pool" does not allow selecting built-in skills. The reverse operation — "Sync to Skill Pool" — has the same limitation.

**5 — Long skill names cause multi-line wrapping, breaking UI layout**
In the Skill Pool (Update / Broadcast buttons) and in the Agent skill list (Load / Sync buttons), skill names that exceed a certain length wrap onto multiple lines, degrading the overall UI appearance.

**6 — Hover buttons obscure the description tooltip**
When hovering over a skill, action buttons appear and cover the description field. As a result, it is impossible to move the cursor onto the description area to trigger the full-description tooltip.

---

**Suggestions**

**For 1 & 2:**
Store created skill tag data locally so tags can be reused via a multi-select UI. Use the skill name as the matching key: if a tag has been applied to a skill, all skills sharing the same name — regardless of where they are located (Skill Pool or Agent) — should automatically receive the same tag. On every skill-related change, perform a diff and add new tags or remove tags for skills that no longer exist.

**For 3:**
Root cause is unclear, but the issue is stably reproducible. Investigation needed.

**For 4 & 5:**
[...]

PR #6552, opened 2026-07-29
fix(skills): preserve plugin-sourced skill tags across pool reconcile

### LINKED ISSUE #6537: [Bug]: Skill tags disappear on restart (regression of #3270)
## QwenPaw Version

2.0.1 (Desktop)

## Description

Skill tags set in the Skill Pool UI disappear after restarting QwenPaw. Tags are correctly saved to `skill_pool/skill.json` via the `PUT /skills/pool/{name}/tags` API endpoint, but are lost when the manifest is reconciled on startup or during builtin skill updates.

This is a **regression of #3270 (item 3)**, which was reportedly fixed by #3362. The tag-preservation logic added in #3362 covers `_register_pool_skill_entry`, `import_builtin_skills`, `_build_reconciled_pool_entry`, and `reconcile_workspace_manifest` — but there is still a code path where tags are silently dropped.

## Root Cause Analysis

The tag loss occurs in `reconcile_pool_manifest()` (`src/qwenpaw/agents/skill_system/registry.py`). The function rebuilds each skill entry by calling `_build_reconciled_pool_entry()`, which calls `build_skill_metadata()` first and then conditionally copies `tags` from the previous entry:

```python
# In _build_reconciled_pool_entry():
new_entry = build_skill_metadata(skill_name, skill_dir, source=source, protected=protected)
# ...
existing_tags = existing.get("tags")
if existing_tags is not None:
    new_entry["tags"] = existing_tags
```

`build_skill_metadata()` never includes `tags` in the new entry. Tags are restored **only if** the previous manifest entry already has them. This works correctly in most cases, **but fails for plugin-sourced skills** whose manifest entries are lightweight stubs:

```json
{
  "brainstorming": {
    "source": "plugin:superpowers",
    "protected": false
  }
}
```

When `reconcile_pool_manifest()` scans the disk for skill directories and cannot find a `SKILL.md` for a plugin skill (plugins don't have on-disk directories in the pool), the skill is **removed from the manifest entirely** along with any tags that were assigned to it. When the plugin re-registers on next startup, a fresh stub entry is created — without any tags.

The same pattern affects workspace manifests: `reconcile_workspace_manifest()` preserves tags from `existing` entries, but if a workspace entry is rebuilt from scratch (e.g., after the workspace directory is cleaned or the manifest is reset), tags are lost because there is no previous entry to copy from.

Additionally, in the `_register_pool_skill_entry()` function, the `tags` parameter defaults to `None`, and `preserve_from` is used as a fallback:

```python
def _register_pool_skill_entry(
    payload, skill_name, skill_dir, *,
    source="customized", protected=False,
    installed_from="", config=None,
    tags=None, preserve_from=None,
):
    # ...
    if tags is not None:
        entry["tags"] = tags
    elif preserve_from.get("tags") is not None:
        entry["tags"] = preserve_from["tags"]
```

When called from code paths that do not explicitly pass `tags` or `preserve_from` containing tags, the field is omitted from the new entry — silently dropping any previously assigned tags.

## Steps to Reproduce

1. Open QwenPaw Desktop v2.0.1.
2. Navigate to the Skill Pool.
3. Add one or more tags to a skill (e.g., a builtin skill like `docx` or a plugin skill like `brainstorming`).
4. Verify the tags appear in the UI.
5. Verify `skill_pool/skill.json` contains the `"tags": [...]` field for that skill.
6. Restart QwenPaw (close and reopen the application).
7. Check the Skill Pool — the tags are gone.
8. Check `skill_pool/skill.json` — the `"tags"` field has been removed from the entry.

**Expected:** Tags persist across restarts.
**Actual:** Tags are erased after restart.

## Workaround

Maintain a separate tags backup file and restore tags into the manifest after QwenPaw starts. Example script:

```python
# skill_tags_backup.py
import json
from pathlib import Path

POOL = Path.home() / ".qwenpaw" / "skill_pool" / "skill.json"
BACKUP = Path.home() / ".qwenpaw" / "skill_pool" / "tags_backup.json"

def save():
    data = json.loads(POOL.read_text(encoding="utf-8"))
    tags = {k: v["tags"] for k, v in data.get("skills", {}).items() if "tags" in v}
    BACKUP.write_text(json.dumps(tags, indent=2, ensure_ascii=False))

def restore():
    if not BACKUP.exists():
        return
    saved = json.loads(BACKUP.read_text(encoding="utf-8"))
    data = json.loads(POOL.read_text(encoding="utf-8"))
    for name, tags in saved.items():
        if name in data.get("skills", {}):
            data["skills"][name]["tags"] = tags
    POOL.write_text(json.dumps(data, indent=2, ensure_ascii=False))
```

## Related Issues

- **#3270** — Original report: tags disappear after triggering an update on a tagged skill (items 1–3).
- **#3362** — Supposed fix: preserves tags during builtin import/update and pool/workspace transfers.

The fix in #3362 addressed the explicit update/import paths, but the reconcile-on-startup path still drops tags for skills without a corresponding `SKILL.md` on disk (plugin skills) and for entries that are rebuilt from scratch.

## Component(s) Affected

- [x] Core / Backend (skill_system: registry.py, pool_service.py, store.py)
- [ ] Channels
- [x] Skills
- [ ] CLI
- [ ] Documentation
- [ ] Tests
- [ ] CI/CD

## Suggested Fix

1. **Persist tags in a sidecar file** (`skill_pool/tags.json`) alongside `skill_pool/skill.json`, similar to how `tags_backup.json` works in the workaround above. Tags should be loaded from this file during reconcile and merged into entries.

2. **OR**: In `_build_reconciled_pool_entry()`, always preserve the `tags` field from the existing entry even when the skill directory is not found on disk (currently, such skills are removed from the manifest — their tags should at minimum be preserved in a sidecar so they can be restored if the skill reappears).

3. **OR**: Ensure that plugin-sourced skill entries retain all user-editable fields (`tags`, `config`) across reconcile cycles, even when the plugin directory is temporarily unavailable.

WilShi:
## Description

Fixes #6537.

Skill tags set in the Skill Pool UI disappear after restart for plugin-sourced skills. Plugin skills are registered as lightweight manifest stubs with `source: "plugin:<id>"` and **no on-disk directory** in the pool. During startup reconcile, `reconcile_pool_manifest()` / `reconcile_workspace_manifest()` scan the pool directory for on-disk `SKILL.md` and drop any manifest entry not found on disk — which deletes plugin stubs along with their user-assigned tags. When the plugin re-registers on the next startup, a fresh tag-less stub is created.

This is a regression of #3270 (item 3): #3362 added tag preservation to the **build** path (`_build_reconciled_pool_entry` copies `tags` from the previous entry), but the **removal pass** in `reconcile_pool_manifest` / `reconcile_workspace_manifest` still drops non-disk entries unconditionally — so plugin stubs (and their tags) are deleted before the build path ever runs.

### Fix

In the removal pass of both `reconcile_pool_manifest()` and `reconcile_workspace_manifest()`, preserve entries whose `source` starts with `"plugin:"` (they have no on-disk dir by design). Genuinely-deleted customized/builtin skills (no disk dir, non-plugin source) are still dropped.

**Type of Change:** Bug fix

**Component(s) Affected:** Core / Backend (agents/skill_system)

**Security Considerations:** none — manifest bookkeeping.

## Evidence

- TDD: `test_reconcile_preserves_plugin_sourced_skill_tags` **fails on the old code** (the removal loop drops the plugin stub → `entry is None`) and **passes after the fix** (plugin stub preserved, tags intact). A second test confirms a genuinely-deleted customized skill is still removed (guards against over-preservation).
- `pre-commit run --files` on both changed files: black / mypy / flake8 / pylint / add-trailing-comma / trailing-whitespace all Passed.
- `pytest tests/integration/test_skills_pool.py tests/integration/test_skills_global.py` — 27 passed (no regression in the skill pool CRUD flow).

```text
$ pytest tests/unit/agents/test_reconcile_pool_manifest.py -v
test_reconcile_preserves_plugin_sourced_skill_tags PASSED   # failed before the fix
test_reconcile_still_removes_genuinely_deleted_customized_skill PASSED
2 passed

$ pytest tests/integration/test_skills_pool.py tests/integration/test_skills_global.py -q
27 passed
```

### Scope

This fixes the removal-pass plugin-drop in both pool and workspace reconcile (the primary reported regression). The reporter's separate "workspace tags lost when rebuilt from scratch" (no previous entry to copy tags from) is a different, larger change (would need a tag store independent of the manifest) and is out of scope here — happy to follow up if maintainers want.

## Checklist

- [x] I ran tests locally (unit + integration skill tests, all pass)
- [x] `pre-commit run --files` on changed files passes
- [x] Documentation: n/a — internal manifest reconcile, no user-facing API
- [x] Ready for review

### DIFF
--- a/src/qwenpaw/agents/skill_system/registry.py
+++ b/src/qwenpaw/agents/skill_system/registry.py
@@ -1022,6 +1022,14 @@ def reconcile_pool_manifest() -> dict[str, Any]:
 
         for skill_name in list(skills):
             if skill_name not in discovered:
+                # Plugin-sourced skills have no on-disk directory by design
+                # (registered as lightweight stubs); preserve them across
+                # reconcile so user-assigned tags survive restart. See #6537.
+                existing = normalize_skill_manifest_entry(
+                    skills.get(skill_name),
+                )
+                if str(existing.get("source") or "").startswith("plugin:"):
+                    continue
                 skills.pop(skill_name, None)
 
         return payload
@@ -1132,6 +1140,14 @@ def reconcile_workspace_manifest(workspace_dir: Path) -> dict[str, Any]:
 
         for skill_name in list(skills):
             if skill_name not in discovered:
+                # Plugin-sourced skills have no on-disk directory by design
+                # (registered as lightweight stubs); preserve them across
+                # reconcile so user-assigned tags survive restart. See #6537.
+                existing = normalize_skill_manifest_entry(
+                    skills.get(skill_name),
+                )
+                if str(existing.get("source") or "").startswith("plugin:"):
+                    continue
                 skills.pop(skill_name, None)
 
         return payload
--- a/tests/unit/agents/test_reconcile_pool_manifest.py
+++ b/tests/unit/agents/test_reconcile_pool_manifest.py
@@ -1,0 +1,104 @@
+# -*- coding: utf-8 -*-
+"""Unit tests for pool manifest reconciliation.
+
+Regression coverage for #6537: plugin-sourced skills have no on-disk
+directory in the pool, so ``reconcile_pool_manifest()`` must NOT drop them
+(and their user-assigned tags) during the "remove skills no longer on
+disk" pass.
+"""
+from __future__ import annotations
+
+import json
+
+from qwenpaw.agents.skill_system import registry as skill_registry
+from qwenpaw.agents.skill_system.registry import reconcile_pool_manifest
+
+
+def _patch_pool(monkeypatch, pool_dir, manifest_path):
+    """Redirect the registry's pool-path lookups at a temp pool."""
+    monkeypatch.setattr(skill_registry, "get_skill_pool_dir", lambda: pool_dir)
+    monkeypatch.setattr(
+        skill_registry,
+        "get_pool_skill_manifest_path",
+        lambda: manifest_path,
+    )
+    monkeypatch.setattr(
+        skill_registry,
+        "get_skill_pool_dirs",
+        lambda: [pool_dir],
+    )
+
+
+def test_reconcile_preserves_plugin_sourced_skill_tags(
+    tmp_path,
+    monkeypatch,
+):
+    """A plugin-sourced skill (no on-disk dir) keeps its tags across reconcile.
+
+    Fails on the pre-fix code: the removal loop dropped any skill absent
+    from the discovered on-disk set, which deleted plugin stubs along with
+    their tags (#6537).
+    """
+    pool_dir = tmp_path / "skill_pool"
+    pool_dir.mkdir()
+    manifest = pool_dir / "skill.json"
+    manifest.write_text(
+        json.dumps(
+            {
+                "skills": {
+                    "plugin-skill": {
+                        "source": "plugin:superpowers",
+                        "protected": False,
+                        "tags": ["favorite", "experimental"],
+                    },
+                },
+                "builtin_skill_names": [],
+            },
+        ),
+    )
+    _patch_pool(monkeypatch, pool_dir, manifest)
+
+    reconcile_pool_manifest()
+
+    on_disk = json.loads(manifest.read_text())
+    entry = on_disk["skills"].get("plugin-skill")
+    assert (
+        entry is not None
+    ), "plugin-sourced skill was dropped by reconcile (regression of #6537)"
+    assert entry.get("tags") == ["favorite", "experimental"]
+
+
+def test_reconcile_still_removes_genuinely_deleted_customized_skill(
+    tmp_path,
+    monkeypatch,
+):
+    """A deleted customized skill is still dropped (not over-preserved).
+
+    Guards that the plugin-preservation fix only retains plugin-sourced
+    stubs, not every non-disk entry.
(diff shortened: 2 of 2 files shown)


Leirunlin:
Thanks for the PR. Based on the latest investigation in #6537, the tag loss is caused by `superpowers-qwenpaw` overwriting the complete Skill Pool manifest entry during startup, rather than by the reconciliation removal pass.

Preserving missing plugin:* entries would not prevent that overwrite and may leave stale entries without a corresponding `SKILL.md`. Therefore, this change does not appear to address the confirmed root cause.

If there are other reproducible core scenarios or details we may have missed, please feel free to continue the discussion here or in #6537. We’ll be happy to follow up.

WilShi:
Acknowledged — the reconcile removal pass does drop plugin-sourced stubs (my failing test confirmed that), but per your #6537 investigation the real-world trigger is the superpowers overwrite of the manifest entry, which this fix doesn't touch; and preserving plugin:* entries unconditionally would indeed leave stale stubs after a plugin uninstall. Leaving this closed. Happy to help dig into the overwrite path in #6537 if useful.

PR #6566, opened 2026-07-30
fix(shell): newline->semicolon on Unix + tempfile redir to avoid PIPE hang

zhuzhuppx:
修复 #6565

## 两个 bug

1. **新行→空格**：`_collapse_newlines_outside_quotes` 将引号外换行替换为空格，多行命令被粘成一行 → 修复为 `"; "`（bash 命令分隔符）

2. **PIPE 卡死**：Linux 用 `stdout=PIPE`，后台进程继承 fd 导致 `communicate()` 永久阻塞 → 改为 tempfile 重定向 + `proc.wait()`（与 Windows 路径方案一致）

详情见 issue #6565。

### DIFF
--- a/src/qwenpaw/agents/tools/shell.py
+++ b/src/qwenpaw/agents/tools/shell.py
@@ -108,10 +108,14 @@ def _collapse_newlines_outside_quotes(cmd: str) -> str:
                 # Preserve newlines inside double quotes
                 result.append(char)
             else:
-                # Collapse \r\n as a single space
+                # Collapse \r\n as a single character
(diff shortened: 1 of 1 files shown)


qbc2016:
Thanks for working on #6565 and for identifying both root causes.

There is now a more comprehensive fix in #6748. It preserves Unix multiline commands as-is, handles timeout and cancellation with temporary-file output correctly, and adds regression coverage for the relevant scenarios.

To keep the review focused and avoid duplicate effort, we will close this PR in favor of #6748. Thanks again for your investigation and contribution.

PR #6492, opened 2026-07-27
fix(files): preserve uploaded filenames in hints

wananing:
## Description

Console uploads already carry the browser-provided original filename in
`DataBlock.name`, while the storage layer uses a UUID-prefixed local path. The
shared file-block processor discarded that display metadata and injected only
the generated path into the user message.

This PR propagates the display name through the current
`_process_local_data_block()` path and includes it in the localized upload
hint. It preserves the local-audio transcription/native-delivery behavior and
`file_url_to_local_path()` handling introduced by #6573.

The display name is reduced to a cross-platform basename, capped at 200 code
points, quoted, and escaped for C0/C1 controls and Unicode line separators.
Readable Unicode such as Chinese filenames remains intact. The real local path
is unchanged, and blocks without a display name keep the existing wording.

Scope is intentionally limited to AgentScope 2.0 typed `DataBlock` inputs,
which cover the reported Console flow. The legacy dict download path remains
behaviorally unchanged; its storage-filename hardening is a separate concern.

**Related Issue:** Fixes #6453

**Security Considerations:** Filename metadata is untrusted. Path components
are removed and the displayed value is bounded, quoted, and escaped before it
is added to the prompt. This does not treat the filename as trusted content or
change storage naming/file-access behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran relevant tests locally and they pass
- [x] Documentation updated (not needed; no configuration or workflow changed)
- [x] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
[...]

PR #6567, opened 2026-07-30
fix(message_processing): include original user filename (esp. CJK) in upload prompt hint (#6453)

### LINKED ISSUE #6453: [Feature]: 对话框的文件上传，是中文文件名时，在提示中尽量保持中文
## Summary

现在提示的文件名“用户上传文件，已经下载到 /C:/Users/XYZ/.qwenpaw/workspaces/default/media/884ff39f590c4472f__.docx”会把中文改成不可识别支付，太长也极不友好，中文文件名的提示建议保持中文

## Additional Context

Windows11,2.0.1桌面客户端

## Willing to Contribute

- [ ] I am willing to open a PR for this feature (after discussion).

RerankerGuo:
## Summary

Fixes #6453 (对话框的文件上传，是中文文件名时，在提示中尽量保持中文).

### Before

After uploading a Chinese-named file (e.g. `项目立项审批表.docx`), the backend injected:

> 用户上传文件，已经下载到 /workspace/media/a1b2c3d_report.docx

Only the UUID-prefixed local media path was shown — the original display name carried by the frontend was silently dropped. Since the storage layer deliberately uses safe (ASCII) UUID-prefixed filenames to avoid filesystem-encoding issues, the model had no way to recover the user-visible name. This was especially confusing on Windows installer sessions where the exact media path is rarely the artifact humans recognise.

### Root cause

`process_file_and_media_blocks_in_message` in `agents/utils/message_processing.py` collected only `(index, local_path)` tuples when it walked the incoming `Msg.content` array. The browser-provided display name that lived on:
- `DataBlock.name` for the 2.0 Pydantic upload path (Console frontend), and
- legacy dict blocks `"name"` for the 1.x wire format

was never read, so the format-string that builds the injected hint had nothing to interpolate.

### Changes

1. **Tuple shape extended** to `(i, local_path, display_name)` for both code paths (Pydantic DataBlock fast-path and legacy dict path).

2. **New sanitizer helper `_sanitize_display_filename(raw_name, fallback_path)`** — because `DataBlock.name` comes from the untrusted client and gets concatenated verbatim into the model-facing prompt, we cannot interpolate raw strings. The sanitizer is the deliberate security boundary between client input and LLM prompt:
   - `urllib.parse.unquote()` first — browsers occasionally pre-escape CJK filenames on multipart uploads without RFC 5987 encoding, resulting in raw `%E9%A1%B9` sequences reaching the backend. We decode them so the agent sees the actual glyphs.
   - Removes **all** C0/C1 control characters (`< 0x20`, `0x7F..0x9F`), RLO/LRO/LRE/RLE/PDF bidi formatting, BOM (`U+FEFF`), OBJECT REPLACEMENT CHARACTER, and the generic REPLACEMENT CHARACTER. Any of these could mislead the model or silently scramble prompt layout.
   - Collapses any interior run of whitespace to a single SP (so accidental embedded newlines / tabs / multiple spaces don't span prompt lines).
   - Caps at **120 chars** with a mid-ellipsis — a single attacker-controlled filename must never dominate the context window.
   - Falls back to `basename(fallback_path)` if the input was malformed (None / not-str / all-controls / empty after cleaning). Graceful degradation keeps storage paths intact.

3. **Prompt builder** now uses the sanitized display name when it differs from the filesystem basename:
   - zh: `用户上传文件"<NAME>"，已经下载到 <PATH>`
   - en: `User uploaded a file "<NAME>", downloaded to <PATH>`
   - When the sanitized name matches the filesystem basename (no-name DataBlock, pure ASCII uploads where the storage layer kept the name, etc.), we emit the **original** short string unchanged — zero behaviour drift for the unaffected majority.

## Verification

Unit tests added to `tests/unit/agents/utils/test_message_processing.py` (34/34 pass, `--noconftest` to avoid provider-manager heavy imports):

- **TestSanitizeDisplayFilename (12 tests)**: fallback paths (`None`, `""`, non-str input), pure ASCII passthrough, **CJK passthrough** (`项目立项审批表.docx` unchanged), %-encoded CJK decoded (`%E9%A1%B9%E7%9B%AE` → `项目立项`), invalid %-sequence safety, control/newline/tab stripping, RLO / BOM / U+FFFD stripping, long-name ellipsization, whitespace collapse, all-controls basename fallback.

- **TestProcessFileBlocksInjectsDisplayName (6 tests)**: DataBlock zh + CJK name injected correctly; DataBlock en + CJK name injected correctly; DataBlock no-name fallback uses path with no zh quotation marks; 1.x legacy dict block's `name` field preserved; %-encoded filename inside the DataBlock.name field decoded before rendering so the model never sees `%E5%AE%A1` percent sequences raw.

Lint:
- `black == 23.3.0 --line-length=79`: pass (both touched files reformatted exactly once)
- `flake8 --extend-ignore=E203`: pass
- `mypy --ignore-missing-imports`: pass (no new or existing errors touched)

## Impact Scope

- Users uploading files with non-ASCII display names (Chinese, Japanese, Korean, and generally any browser-sent display name that differs from the storage-safe UUID-prefixed media path). ~100% of the benefit accrues here.
- Zero breaking API changes; new helper is module-private; the injection is backwards-compatible because when the sanitized name matches the filesystem basename we emit exactly what prior versions emitted.
- Security posture strictly improved via the new input sanitization boundary on a previously unvalidated client field that reaches the LLM prompt.

### DIFF
--- a/src/qwenpaw/agents/utils/message_processing.py
+++ b/src/qwenpaw/agents/utils/message_processing.py
@@ -436,6 +436,60 @@ def _coerce_block_to_dict(
     return None
 
 
+def _resolve_data_block_path(block) -> tuple[str, str] | None:
+    """Return ``(local_path, display_name)`` for a Pydantic DataBlock file URL.
+
+    Extracts ``file://`` URLs from :class:`agentscope.message.DataBlock`
+    (produced by the console upload flow) and sanitizes the user-visible
+    display name via :func:`_sanitize_display_filename`.  Returns ``None``
+    for non-file DataBlocks (remote URLs, in-memory Base64Source, etc.) so
+    callers can skip non-downloadable entries without branching.
+    """
+    source = getattr(block, "source", None)
+    url = str(getattr(source, "url", "")) if source else ""
+    if not url.startswith("file://"):
+        return None
+    local_path = url.removeprefix("file://")
+    display_name = _sanitize_display_filename(
+        getattr(block, "name", None),
+        local_path,
+    )
+    return local_path, display_name
+
+
+def _resolve_legacy_block_path(
+    block: dict, local_path: str
+) -> tuple[str, str] | None:
+    """Resolve the 1.x dict-block path into ``(local_path, display_name)``."""
+    display_name = _sanitize_display_filename(
+        block.get("name"),
+        local_path,
+    )
+    return local_path, display_name
+
+
+def _build_upload_hint(local_path: str, display_name: str, lang: str) -> str:
+    """Build the user-facing "file downloaded" prompt hint.
+
+    When ``display_name`` differs from the filesystem basename (the usual
+    case for console UUID-prefixed media storage), both are shown so the
+    model has the semantic filename the user uploaded AND the concrete
+    local path to read from.  When they match, only the path is shown,
+    preserving exact historical behavior for the unaffected majority.
+    """
+    basename = Path(local_path).name
+    if lang == "zh":
+        if display_name and basename != display_name:
+            return "用户上传文件“" f"{display_name}”，已经下载到 " f"{local_path}"
+        return f"用户上传文件，已经下载到 {local_path}"
+    if display_name and basename != display_name:
+        return (
+            f'User uploaded a file "{display_name}", '
+            f"downloaded to {local_path}"
+        )
+    return f"User uploaded a file, downloaded to {local_path}"
+
+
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing this to avoid competing with the earlier active implementation in #6492, whose author had already claimed #6453. The overlap is substantial and maintaining two solutions would add review burden.

PR #6497, opened 2026-07-27
fix(plugins): remove implicit max version derivation for legacy plugins

YMG001:
## Description

Removes the implicit upper-bound derivation in `_version_compat.py` that was silently disabling legacy plugins on QwenPaw 2.0+.

When a plugin's `plugin.json` does not specify `max_version` (or `qwenpaw_version.max`), the `_derive_exclusive_max()` function derived an upper bound from `min_version` (e.g. `"0.1.0" → "0.2.0"`). On QwenPaw 2.0.1, this caused `current(2.0.1) >= max_v(0.2.0)` → incompatible, silently setting `enabled=False` and skipping `register()`. The plugin's tools never entered `ToolRegistry`, so they were invisible to the LLM — with only a WARNING log.

**Fix:** When `max` is not specified, set `max_v = None` and skip the upper-bound check entirely. This follows the universal convention (pip, npm) that `>=min` without an explicit max means no upper bound.

**Related Issue:** Fixes #6496

**Security Considerations:** None. Only affects plugin version compatibility checking logic.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [x] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

N/A — no channel changes.

## Testing

All testing was performed locally with QwenPaw 2.0.1 on Python 3.12 using `uv` for environment management and editable installs.

### End-to-end test (CLI + HTTP API)

Two identical plugins differing only in version declaration:

| Plugin | min_version | max_version |
|--------|------------|-------------|
| `normal-echo` | `2.0.0` | `3.0.0` |
| `legacy-echo` | `0.1.0` (default) | not specified |

[...]

PR #6532, opened 2026-07-28
fix(plugins): temporarily disable max version check in plugin compat

XiuShenAl:
## Description

Temporarily disable the upper-bound (`max`) check in plugin–QwenPaw version compatibility after the bump to `2.1.0b1`.

Compatibility previously used a left-closed / right-open range (`>=min, <max`), and pre-releases are normalized to their base version (`2.1.0b1` → `2.1.0`). Many plugins still declare `max: "2.1.0"` (or omit `max`, which derived `max` as the next minor), so they were treated as incompatible on the current build and failed to load.

This change keeps computing `max` / derived upper bounds, but only enforces `>= min` for now. The original full-range check is left commented in place so it can be restored later without rewriting the logic. Catalog unit tests are updated to reflect that a declared `max` no longer excludes a newer running QwenPaw. Integration plugin-type fixtures keep their original `max: "2.1.0"` values.

No per-plugin `plugin.json` updates are required for this temporary workaround.

**Related Issue:** Relates to #6518

**Security Considerations:** N/A — compatibility gating only; no auth, env, or config handling changes. Plugins that are below the required minimum version are still rejected.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [x] Documentation updated (if needed)
- [x] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
[...]

PR #6574, opened 2026-07-30
fix(plugins): permanent max-version semantics — no implicit derivation, honor explicit max

YMG001:
## Description

This PR proposes a **permanent** max-version compatibility semantics for plugins, as a follow-up to the temporary fix in #6532.

### Background

Issue #6496 reported that legacy plugins with only `min_version` (or the default `0.1.0`) were silently disabled because `_derive_exclusive_max()` implicitly derived an upper bound that rejected the current QwenPaw version.

#6532 fixed this by **temporarily disabling all upper-bound (`max`) checks** — including both the implicit derivation and explicit `max` values. Only `>= min` is enforced.

### This PR (permanent semantics)

Instead of disabling all max checks, this PR:

1. **Removes the implicit `_derive_exclusive_max()` derivation** — when no `max` is specified, there is no upper bound (consistent with pip/npm/cargo conventions).
2. **Still honors an explicit `max`** — if a plugin declares `max: "2.1.0"`, that upper bound is enforced.

### Comparison

| Case | This PR | #6532 (current `main`) |
|------|---------|------------------------|
| no `max` (legacy plugins) | ✅ compatible | ✅ compatible |
| explicit `max` below current version | ❌ incompatible (max enforced) | ✅ compatible (max ignored) |

### Why keep this PR open?

When the team is ready to restore full `>=min, <max` enforcement, this PR provides the correct long-term semantics: "no max ⇒ no upper bound", while still respecting explicit max declarations. It is a strict improvement over the temporary "ignore max entirely" approach in #6532.

## Type of Change

- [x] Bug fix (non-breaking)

## Component(s) Affected

- [x] Core
- [x] Tests

## Checklist

- [x] `black` passes
- [x] `flake8` passes
- [x] `pytest` passes (12 tests in relevant module)

## Testing

```bash
black src/qwenpaw/_version_compat.py tests/unit/plugins/test_download_catalog.py
# All done ✅

flake8 src/qwenpaw/_version_compat.py tests/unit/plugins/test_download_catalog.py
# No issues ✅

pytest tests/unit/plugins/test_download_catalog.py -v
# 12 passed ✅
```

### End-to-end verification

- Legacy plugin with only `min_version`: `enabled: true, loaded: true` ✅
- Tool count increased by 1 (28 → 29) ✅
- No WARNING about version incompatibility ✅

### DIFF
--- a/src/qwenpaw/_version_compat.py
+++ b/src/qwenpaw/_version_compat.py
@@ -2,8 +2,10 @@
 """Plugin–QwenPaw version compatibility check.
 
 Semantics: left-closed, right-open interval  ``>=min, <max``.
-When ``max`` is not specified, it is derived from ``min`` as
-``{major}.{minor+1}.0`` (all patch versions of the same minor).
+When ``max`` is not specified, no upper bound is enforced — the plugin
+is considered compatible with any QwenPaw version that satisfies ``>=min``.
+This follows the universal convention (pip, npm, cargo) where an
+unspecified upper bound means no upper bound.
 """
 
 from __future__ import annotations
@@ -23,6 +25,11 @@ def _derive_exclusive_max(min_str: str) -> Version:
     """Derive exclusive upper bound from a min version string.
 
     '1.1.6' -> Version('1.2.0')
+
+    .. deprecated::
+        This function is retained for backward compatibility but no longer
+        called by :func:`check_plugin_version_compat`.  An unspecified ``max``
+        now means *no upper bound* instead of a derived exclusive bound.
     """
     parts = min_str.split(".")
     major = int(parts[0])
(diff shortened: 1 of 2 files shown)


YMG001:
Thanks @XiuShenAl for the quick fix in #6532 and the clear write-up on #6497!

This PR (#6574) is the **permanent** version of the fix, replacing the closed #6497 (which had accumulated duplicate commits and branch conflicts).

### Why keep this as a permanent follow-up

#6532 temporarily disables **all** upper-bound checks — including explicit `max` declarations. That means a plugin with `max: "1.5.0"` would still load on QwenPaw 2.0.1, which may not be intended.

This PR takes a narrower approach:

| Case | This PR (#6574) | Current `main` (#6532) |
|------|-----------------|------------------------|
| no `max` (legacy plugins) | ✅ compatible | ✅ compatible |
| explicit `max` below current version | ❌ incompatible (max enforced) | ✅ compatible (max ignored) |

### Semantics

- **No `max` specified → no upper bound** (consistent with pip/npm/cargo conventions)
- **Explicit `max` specified → enforced** as an exclusive upper bound
- `_derive_exclusive_max()` is retained for backward compatibility but no longer called

### Conflict note

The branch is based on a slightly older fork `main`, so GitHub shows a conflict. The actual diff is clean — only 2 files changed (`src/qwenpaw/_version_compat.py` and `tests/unit/plugins/test_download_catalog.py`). Resolving the conflict is straightforward: take this branch's version of `_version_compat.py`.

Happy to rebase or adjust if the team prefers a different direction for the long-term max-semver policy.

YMG001:
New PR created: #6577 — please follow up there. (Previous #6575 was closed as it was from the wrong branch.)

PR #6575, opened 2026-07-30
fix(plugin-version): remove implicit max version derivation; keep explicit max check

YMG001:
## Summary

Follow-up to #6532 (temporary fix) — permanent max-version semantics.

When a plugin does not specify `max_version`, no upper bound should be **implicitly derived**. This matches the convention used by `pip`, `npm`, and `cargo`. Explicit `max_version` is still enforced when provided.

## Problem

`_derive_exclusive_max()` was called even when the plugin author did not specify `max_version`, producing a spurious upper bound that caused compatible plugins to be silently disabled (issue #6496).

## Fix

- **New API path** (`qwenpaw_version`): if `max_version` is `None`, skip the upper-bound check entirely.
- **Legacy path** (`min_version`/`max_version`): if `max_version` is `None`, do not call `_derive_exclusive_max()`.
- `_derive_exclusive_max()` is retained but no longer called.
- Explicit `max_version` is still checked when provided.

## Comparison with #6532

| | #6532 (merged, temporary) | This PR (permanent) |
|---|---|---|
| Implicit max derivation | ❌ disabled | ❌ disabled |
| **Explicit max check** | ❌ **disabled** | ✅ **enforced** |

#6532 disabled **all** max-version checks (including explicit ones) as an emergency fix. This PR restores the explicit `max_version` check while keeping the implicit derivation removed.

## Testing

- `black` ✅
- `flake8` ✅
- `pytest` ✅ (all existing tests pass)
- End-to-end: legacy plugin with no `max_version` → `enabled: true, loaded: true`; tool count increased by 1; no WARNING.

Supersedes #6497.

### DIFF
--- a/src/qwenpaw/_version_compat.py
+++ b/src/qwenpaw/_version_compat.py
@@ -2,8 +2,10 @@
 """Plugin–QwenPaw version compatibility check.
 
 Semantics: left-closed, right-open interval  ``>=min, <max``.
-When ``max`` is not specified, it is derived from ``min`` as
-``{major}.{minor+1}.0`` (all patch versions of the same minor).
+When ``max`` is not specified, no upper bound is enforced — the plugin
+is considered compatible with any QwenPaw version that satisfies ``>=min``.
+This follows the universal convention (pip, npm, cargo) where an
+unspecified upper bound means no upper bound.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


YMG001:
Closing in favor of #6577 which uses a dedicated feature branch. The conflict marker on this PR was due to the fork being behind upstream and workflow file permission limitations, not actual file conflicts.

PR #6577, opened 2026-07-30
fix(plugin-version): remove implicit max version derivation; keep explicit max check

YMG001:
## Summary

Follow-up to #6532 (temporary fix) — permanent max-version semantics.

When a plugin does not specify `max_version`, no upper bound should be **implicitly derived**. This matches the convention used by `pip`, `npm`, and `cargo`. Explicit `max_version` is still enforced when provided.

## What Problem This Solves

`_derive_exclusive_max()` was called even when the plugin author did not specify `max_version`, producing a spurious upper bound that caused compatible plugins to be silently disabled (issue #6496).

## Fix

- **New API path** (`qwenpaw_version`): if `max_version` is `None`, skip the upper-bound check entirely.
- **Legacy path** (`min_version`/`max_version`): if `max_version` is `None`, do not call `_derive_exclusive_max()`.
- `_derive_exclusive_max()` is retained but no longer called.
- Explicit `max_version` is still checked when provided.

## Comparison with #6532

| | #6532 (merged, temporary) | This PR (permanent) |
|---|---|---|
| Implicit max derivation | ❌ disabled | ❌ disabled |
| **Explicit max check** | ❌ **disabled** | ✅ **enforced** |

#6532 disabled **all** max-version checks (including explicit ones) as an emergency fix. This PR restores the explicit `max_version` check while keeping the implicit derivation removed.

## Testing

- `black` ✅
- `flake8` ✅
- `pytest` ✅ (all existing tests pass)
- End-to-end: legacy plugin with no `max_version` → `enabled: true, loaded: true`; tool count increased by 1; no WARNING.

Supersedes #6497 and #6574.

### DIFF
--- a/src/qwenpaw/_version_compat.py
+++ b/src/qwenpaw/_version_compat.py
@@ -2,12 +2,10 @@
 """Plugin–QwenPaw version compatibility check.
 
 Semantics: left-closed, right-open interval  ``>=min, <max``.
-When ``max`` is not specified, it is derived from ``min`` as
-``{major}.{minor+1}.0`` (all patch versions of the same minor).
-
-NOTE: Upper-bound (``max``) enforcement is temporarily disabled.
-Restore the commented check in ``check_plugin_version_compat`` when
-re-enabling full range checks.
+When ``max`` is not specified, no upper bound is enforced — the plugin
+is considered compatible with any QwenPaw version that satisfies ``>=min``.
+This follows the universal convention (pip, npm, cargo) where an
+unspecified upper bound means no upper bound.
 """
 
 from __future__ import annotations
@@ -27,6 +25,11 @@ def _derive_exclusive_max(min_str: str) -> Version:
     """Derive exclusive upper bound from a min version string.
 
     '1.1.6' -> Version('1.2.0')
+
+    .. deprecated::
+        This function is retained for backward compatibility but no longer
+        called by :func:`check_plugin_version_compat`.  An unspecified ``max``
+        now means *no upper bound* instead of a derived exclusive bound.
     """
     parts = min_str.split(".")
     major = int(parts[0])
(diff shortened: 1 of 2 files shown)


YMG001:
Thanks for the quick temporary fix in #6532! This PR proposes a permanent solution.

**About the conflict marker:** The PR shows a conflict flag, but the actual file diff is clean — only 2 files changed (`_version_compat.py` and `test_download_catalog.py`), and neither was modified in the upstream commits since the fork's base.

The conflict arises because the fork's base branch is behind upstream (the fork couldn't be synced due to workflow file permission limitations on our end). When GitHub tries to auto-merge upstream into the PR branch, it encounters workflow file changes that our token doesn't have permission to handle, resulting in a "merge conflict" error.

**Resolution for maintainers:** You should be able to merge this PR directly (e.g., squash merge), or update the branch first since you have full permissions. The only changes in this PR are the 2 files listed above — no actual file conflicts exist.

**What this PR does vs #6532:**
- #6532 (merged): disabled **all** max-version checks (both implicit and explicit) as an emergency fix
- This PR: removes only the **implicit** max derivation, **restores** the explicit `max_version` check

This means plugin authors who explicitly set `max_version` will once again have that constraint enforced, while plugins without `max_version` won't get a spurious upper bound.

Happy to make any adjustments if needed!

YMG001:
Thank you @XiuShenAl for the quick response and the temporary fix in #6532! 🙏

The conflict has been resolved — the branch is now based on the latest `main` and the PR is mergeable.

## Why keep this PR as a permanent follow-up

#6532 was a great emergency fix that immediately unblocked affected plugins. However, it disabled **all** `max_version` checks — including those explicitly set by plugin authors. This means:

| Scenario | #6532 (temporary) | This PR (#6577) |
|---|---|---|
| Plugin has **no** `max_version` | ✅ loaded (no upper bound) | ✅ loaded (no upper bound) |
| Plugin has **explicit** `max_version` above current version | ✅ loaded (**max ignored**) | ❌ rejected (**max enforced**) |
| Plugin has **explicit** `max_version` within range | ✅ loaded | ✅ loaded |

The key difference: **explicit `max_version` should be respected**. If a plugin author writes `max_version = "3.0"`, they're signaling incompatibility beyond 3.0 — ignoring that could lead to silent breakage at runtime.

### Design rationale

- **No implicit upper bound** when `max_version` is omitted — consistent with `pip`, `npm`, and `cargo` conventions
- **Explicit `max_version` still enforced** — respects author intent, prevents loading known-incompatible plugins
- `_derive_exclusive_max()` retained but no longer called — minimal footprint, easy to audit

This PR only touches 2 files (`_version_compat.py` + tests), passes `black` / `flake8` / `pytest`, and has been verified end-to-end. Happy to adjust if you'd prefer a different approach!

XiuShenAl (commented):
Thanks again for this follow-up — and for clearly comparing it with #6532. The core idea here is helpful.

### Why we’re not merging for now

After the bump to `2.1.0b1`, we temporarily disabled **all** upper-bound (`max`) checks in #6532 (including explicit ones). That was an emergency workaround: many manifests still declare `max: "2.1.0"`, and pre-releases normalize to `2.1.0`, so restoring `>=min, <max` would reject them immediately. That temporary state is still what `main` relies on today. We have **not** finished the transition around it yet — especially coordinating built-in / catalog plugin `max` values and related release timing — so we can’t restore explicit `max` enforcement on current `main` without re-breaking a large set of plugins.

So this pause is **not** a judgment on your PR quality or the usefulness of the proposal. It’s because the temporary workaround’s follow-up is still incomplete. You’re welcome to **keep this PR open** as a reference; when we revisit version-compat policy, we’ll take this discussion into account.

Really appreciate the careful write-up and your patience.

PR #6605, opened 2026-07-31
fix(providers): return typed tagged tool calls

rayrayraykk:
## Summary

- rehydrate tool calls extracted from thinking or text tags as AgentScope 2
  `ToolCallBlock` objects
- preserve raw JSON arguments as the block input
- emit each recovered tool call once with a unique stream-local ID
- cover thinking, content, streaming accumulation, and multiple tool calls

## Root cause

The compatibility parser appended legacy dictionary-shaped `tool_use` blocks
to `ChatResponse.content`. AgentScope 2 expects typed content blocks and reads
`block.id` while accumulating streaming responses, which raised:

```text
AttributeError: 'dict' object has no attribute 'id'
```

Converting the recovered calls to `ToolCallBlock` also requires one-shot
emission; otherwise subsequent stream chunks repeat the same block and append
its complete JSON input more than once.

## Validation

- `pytest -q tests/unit/providers`: 341 passed
- parser and tag-parser tests: 31 passed
- pre-commit for changed files: passed
- `git diff --check`: passed

### DIFF
--- a/src/qwenpaw/providers/openai_chat_model_compat.py
+++ b/src/qwenpaw/providers/openai_chat_model_compat.py
@@ -10,6 +10,7 @@ from datetime import datetime
 from types import SimpleNamespace
 from typing import Any, AsyncGenerator, Callable
 
+from agentscope.message import ToolCallBlock
 from agentscope.model import OpenAIChatModel
 from agentscope.model._model_response import ChatResponse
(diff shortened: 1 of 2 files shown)


qbc2016 (approved):
LGTM.

PR #6610, opened 2026-07-31
fix: shell command execution hangs and UI freezes

### LINKED ISSUE #6608: [Bug]: Long-running shell commands bypass shell_command_timeout and block the feishu session indefinitely (orphan subprocess on cancel, no per-channel total timeout)
## Summary

A long-running `execute_shell_command` (a dedup script hitting the Feishu Bitable API) blocked a feishu chat session for **1.5 hours**. The user's follow-up messages were received but never processed (queued behind the stuck task). The session only recovered after the system cancelled the task; however the spawned shell subprocess **survived the cancellation** and kept running until manually killed.

Three mechanism-level gaps are involved:

### 1. Explicit LLM-provided `timeout` bypasses `shell_command_timeout`

`agents/tools/shell.py` (~line 568):

```python
# Apply agent-configured default when the caller used the hardcoded
# default (60.0). An explicit LLM-provided value != 60.0 is kept.
if timeout == 60.0:
    configured = get_current_shell_command_timeout()
    if configured is not None:
        timeout = configured
```

If the agent explicitly passes `timeout=3600` (or larger), the configured default `shell_command_timeout: 60.0` is silently ignored. There is **no upper bound** on what an LLM can request. A single tool call can therefore run for hours.

### 2. Cancellation does not kill the spawned subprocess (orphan)

When the session task was cancelled (`cancel_cleanup: cancelled ... CancelledError`), the asyncio task was cancelled but the already-spawned `python3 dedup_bitable.py` process **kept running** (PID survived from 15:34 until manually killed at ~17:07). The `except (asyncio.TimeoutError, asyncio.CancelledError)` branch in `shell.py` only runs when the *tool-internal* timeout/cancel fires; an external session-level cancellation (error_hook `cancel_cleanup`) does not reach the subprocess. Result: orphan processes keep consuming resources after the turn is cancelled.

### 3. Feishu channel has no total task timeout fallback

`task_timeout_ms: 3600000` is configured only for the `xiaoyi` channel in `agent.json`. The feishu channel has no equivalent, so there is no last-resort upper bound on how long a single turn can block the session queue. New messages are queued with no feedback to the user.

## Impact

- A feishu chat becomes unresponsive for arbitrarily long (observed 1.5h+).
- Follow-up messages queue silently; user has no indication why.
- Cancelled turns leave orphan subprocesses running.
- Session recovery can re-run the same long task again (observed: after cancel, the agent re-invoked the same script).

## Environment

- QwenPaw version: 2.0.0.post4 (or current master)
- Channel: feishu, Linux (Debian-based container)

## Suggested fixes

1. Cap the effective shell timeout (e.g. `min(timeout, max_shell_timeout)` or clamp LLM-provided timeout to the configured default as an upper bound).
2. On session/turn cancellation, kill the process group of any spawned subprocess (align with the existing `killpg` logic used on tool-internal timeout).
3. Add a per-channel task timeout for feishu (and other chat channels) as a last-resort watchdog, with user-visible feedback when a turn is still running.

## Related

- #6565 / #6566 (PIPE-mode background process hang — different root cause but same symptom class)
- #6527 (cancellation-safe lifecycle hooks — related to cancellation semantics)

mohitdebian:
## What Problem This Solves

Fixes #6608 and #6589.

Found a couple of issues with how shell commands run that were causing freezes:
1. Hard limits for timeouts weren't working properly if the command set an arbitrarily large timeout. I capped it against the configured max (default 600s).
2. If a command spits out tens of thousands of lines, it freezes the frontend UI. I added a quick truncation step to only keep the last 500 lines for stdout and stderr to prevent UI locking up.

## Evidence

Ran local unit tests in `tests/unit/agents/tools/test_shell.py` and manually verified output truncation and timeout limits for runaway commands.

`pre-commit run --all-files`
<details>
<summary>View Output</summary>

ruff.....................................................................Passed
ruff-format..............................................................Passed

</details>

`pytest`
<details>
<summary>View Output</summary>

tests/unit/agents/tools/test_shell.py .................................. [ 38%]
......................................................                   [100%]
============================== 88 passed in 2.50s ==============================

</details>

## Type of Change

- [x] Bug fix
- [ ] New feature

## Component(s) Affected

## Checklist

- [x] Ready for review

### DIFF
--- a/src/qwenpaw/agents/tools/shell.py
+++ b/src/qwenpaw/agents/tools/shell.py
@@ -777,12 +777,12 @@ async def execute_shell_command(
         except (ValueError, TypeError):
             timeout = 60.0
 
-    # Apply agent-configured default when the caller used the hardcoded
-    # default (60.0).  An explicit LLM-provided value != 60.0 is kept.
-    if timeout == 60.0:
-        configured = get_current_shell_command_timeout()
-        if configured is not None:
-            timeout = configured
+    configured = get_current_shell_command_timeout()
+    if timeout == 60.0 and configured is not None:
+        timeout = configured
+
+    max_t = configured if configured is not None else 600.0
+    timeout = min(timeout, max_t)
 
     # Use current workspace_dir from context, fallback to WORKING_DIR
     if cwd is not None:
@@ -936,6 +936,19 @@ async def execute_shell_command(
                     proc,
                     stderr_suffix,
                 )
+
+        def _limit_lines(text: str, mx: int = 500) -> str:
+            if not text:
+                return text
+            lines = text.splitlines(keepends=True)
+            if len(lines) > mx:
+                return f"... [{len(lines) - mx} lines truncated]\n" + "".join(
+                    lines[-mx:],
+                )
+            return text
+
+        stdout_str = _limit_lines(stdout_str)
+        stderr_str = _limit_lines(stderr_str)
 
         if returncode == 0:
             if stdout_str:


lalaliat:
Thank you for your contribution. After reviewing the actual scope of this PR against #6608 and #6589, we have decided to close it.

The main reasons are:

- This PR only addresses one part of #6608: an LLM-provided timeout can exceed the configured default. It does not resolve the issue’s other two core problems. In addition, allowing the LLM to specify a timeout is intentional behavior. The original issue would be better addressed through a more appropriate channel-specific solution.

- The backend 500-line truncation added for #6589 has been solved by #6637, which provides more comprehensive protection at the Console rendering layer while preserving the complete backend output.

Thank you again for taking the time to contribute. We appreciate your effort and welcome future contributions or proposals.

PR #6617, opened 2026-07-31
fix(providers): honor the Retry-After cap on the streaming retry path

Yigtwxx:
## Description

`RetryChatModel` keeps its rate-limit policy in one place, `_handle_rate_limit_exc`, whose docstring spells out three branches:

- internal acquire timeout (`_AcquireTimeoutError`): re-raise as-is, no report, no retry
- retryable API 429 with `Retry-After` greater than `MAX_PAUSE_SECONDS`: re-raise immediately, because retrying after the capped pause would just get another 429 (the docstring names Anthropic's `FreeUsageLimitError` with `Retry-After: 51496`)
- normal 429: call `report_rate_limit()` to set the per-model pause

The non-streaming path in `__call__` calls that helper. The streaming path in `_wrap_stream` re-implements only the third branch inline and drops the cap bail-out:

```python
if _is_retryable(retry_exc) and _is_rate_limit(retry_exc):
    await limiter.report_rate_limit(
        _extract_retry_after(retry_exc),
    )
```

Streaming is the default (`stream=True`), so this is the path essentially every request takes.

What happens today when a provider returns a 429 with a large `Retry-After`: `report_rate_limit()` caps the pause at `MAX_PAUSE_SECONDS` (60 s), the caller sleeps that long inside `acquire()`, re-issues, receives the same 429, and repeats for every remaining attempt. Because the pause is installed on the per-model limiter, it also holds back every other caller of that model for the duration. `LLMRateLimiter.report_rate_limit` already documents the contract the streaming path is missing: "Callers that receive a Retry-After above the cap are raised immediately (no retry)."

There is a second, user-visible consequence visible in the log transcript below: the error that finally surfaces is not the provider's 429 but the limiter's own `_AcquireTimeoutError` ("Rate limit exceeded"), because the retry blocks on the pause this call just installed until the acquire timeout fires. The original quota error is lost.

This change routes the streaming path through the existing helper instead of keeping a second, divergent copy of the policy.

One note on the other branch, so the diff is not read as a silent behaviour change: `_AcquireTimeoutError` handling is unaffected. It subclasses `RateLimitExceededException`, and `AppBaseException.__init__` sets `error_code`, never `code`, so `extract_status_code()` finds nothing on it and returns `None`. `_is_retryable()` is therefore already `False`, and the current code already re-raises at the `if not _is_retryable(...)` line below. Control flow and the propagated exception object are identical; the only difference is one extra frame in the traceback.

**Related Issue:** none

**Security Considerations:** none

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

On the first box: I ran `pre-commit run --files` against the two changed files rather than `--all-files`. On this Windows checkout the `--all-files` run rewrites line endings across roughly a hundred untouched files, which would bury the diff. The changed-file run is included below and every hook passes.

## Testing

Two tests were added to `tests/unit/providers/test_retry_chat_model.py`:

- `test_stream_rate_limit_over_cap_raises_without_pause` builds a stream that fails on first iteration with a 429 carrying `Retry-After: 51496`. It asserts the inner model was called exactly once, the provider's own exception propagates, and the limiter recorded no pause (`total_rate_limited == 0`, `is_paused is False`).
- `test_stream_rate_limit_under_cap_still_retries` uses the same shape with a small `Retry-After` and asserts the retry and the `report_rate_limit()` call still happen (`total_rate_limited == 1`). This is the over-correction guard: simply deleting the `report_rate_limit()` call would pass the first test and fail this one.

## Evidence

With the source change reverted and only the new tests applied, the over-cap test fails. The captured log shows the defect directly: the 51496 s `Retry-After` is capped to a 60 s pause, and the exception that reaches the caller is the limiter's own "Rate limit exceeded" instead of the provider's quota error.

```
$ git stash push src/qwenpaw/providers/retry_chat_model.py
$ pytest tests/unit/providers/test_retry_chat_model.py -q \
    -k "rate_limit_over_cap or rate_limit_under_cap"

F.                                                                       [100%]
____________ test_stream_rate_limit_over_cap_raises_without_pause _____________
>           with pytest.raises(Exception, match="FreeUsageLimitError"):
E           AssertionError: Regex pattern did not match.
E             Expected regex: 'FreeUsageLimitError'
E             Actual message: 'Rate limit exceeded'

---------------------------- Captured stderr call -----------------------------
WARNING rate_limiter.py:234 | LLM rate limiter: global pause set for 60.0s (raw_retry_after=51496.0s, total_rate_limited=1)
WARNING retry_chat_model.py:628 | LLM stream failed (attempt 1/3): FreeUsageLimitError: daily quota exhausted. Retrying in 0.1s ...

FAILED tests/unit/providers/test_retry_chat_model.py::test_stream_rate_limit_over_cap_raises_without_pause
1 failed, 1 passed, 43 deselected in 11.53s
```

Note that the second test passes in both states, which is what makes it a useful guard rather than a duplicate.

With the fix applied, both tests pass and so does the rest of the provider suite:

```
$ pytest tests/unit/providers/test_retry_chat_model.py -q \
    -k "rate_limit_over_cap or rate_limit_under_cap"
2 passed, 43 deselected in 0.71s

$ pytest tests/unit/providers/ -q
341 passed in 6.19s

$ pytest tests/unit/ -q
6010 passed, 64 skipped, 58 warnings in 371.53s (0:06:11)
```

```
$ pre-commit run --files src/qwenpaw/providers/retry_chat_model.py \
    tests/unit/providers/test_retry_chat_model.py

```

## Additional Notes

None.


qbc2016:
Please merge the latest main branch and fix the conflicts.

Yigtwxx:
Done — merged the latest `main` (`bacac741`) into the branch in `56f6343c`.

The conflict came from #6721, which touched both files this PR touches:

- `src/qwenpaw/providers/retry_chat_model.py` merged automatically. #6721 rewrote the missing-`reasoning_content` branch (`_inject_reasoning_content` → `_enable_reasoning_content_fallback`) in the `except` block immediately above the line this PR changes. The two edits are independent: the streaming path still falls through to `await self._handle_rate_limit_exc(retry_exc, limiter)` after the reasoning-content `continue`.
- `tests/unit/providers/test_retry_chat_model.py` was the only real conflict, and it is purely textual — both sides appended a new block at the end of the file. Kept both, with #6721's `test_stream_recovers_agentscope_msg_via_formatter_fallback` first and this PR's "Streaming rate-limit policy" section after it. Nothing from either side was dropped or reworded.

The diff against the new base is unchanged from what was originally opened, **+100 −4** across the same 2 files (source +1 −4, tests +99 −0), so the merge did not widen the scope of the review.

Re-verified on the new base:

```
$ pytest tests/unit/providers/test_retry_chat_model.py -q
49 passed in 11.86s

$ pytest tests/unit/providers/ -q
368 passed in 6.82s

$ pre-commit run --files src/qwenpaw/providers/retry_chat_model.py \
    tests/unit/providers/test_retry_chat_model.py
```

The RED proof still reproduces identically after the merge — reverting only the one-line source change makes the over-cap test fail with the limiter's own error instead of the provider's:

```
E           AssertionError: Regex pattern did not match.
E             Expected regex: 'FreeUsageLimitError'
E             Actual message: 'Rate limit exceeded'

---------------------------- Captured stderr call -----------------------------
WARNING rate_limiter.py:234 | LLM rate limiter: global pause set for 60.0s (raw_retry_after=51496.0s, total_rate_limited=1)
WARNING retry_chat_model.py:704 | LLM stream failed (attempt 1/3): FreeUsageLimitError: daily quota exhausted. Retrying in 0.1s ...
```

Ready for another look, thanks for the review.

RerankerGuo:
Independent validation at exact head `56f6343c595aefd47dea2abf9c24b9a5c7b33b00` after the merge from current `main`.

I verified both the intended streaming rate-limit behavior and the conflict boundary with #6721.

Green results on this head:

```bash
.venv/bin/pytest tests/unit/providers/test_retry_chat_model.py -q
# 49 passed

.venv/bin/pytest tests/unit/providers/ -q
# 368 passed

PATH="/opt/homebrew/bin:$PATH" .venv/bin/pre-commit run --files \
  src/qwenpaw/providers/retry_chat_model.py \
  tests/unit/providers/test_retry_chat_model.py
# all applicable hooks passed, including mypy, black, flake8, and pylint
```

This includes `test_stream_recovers_agentscope_msg_via_formatter_fallback` from #6721, so the textual test conflict did not drop that behavior.

I also independently repeated the red proof by temporarily restoring only the old inline streaming branch while keeping the new tests:

```bash
.venv/bin/pytest tests/unit/providers/test_retry_chat_model.py -q \
  -k "rate_limit_over_cap or rate_limit_under_cap"
```

Result with the source change reverted:

```text
1 failed, 1 passed, 47 deselected
FAILED test_stream_rate_limit_over_cap_raises_without_pause
Expected: FreeUsageLimitError
Actual:   Rate limit exceeded

LLM rate limiter: global pause set for 60.0s
(raw_retry_after=51496.0s, total_rate_limited=1)
```

After restoring this PR's call to `_handle_rate_limit_exc`, the identical command reports:

```text
2 passed, 47 deselected
```

The under-cap test passes in both states, confirming the fix preserves normal 429 pause-and-retry behavior rather than bypassing rate-limit reporting wholesale. I found no regression in the updated two-file diff.

Environment: macOS, Python 3.11 virtualenv. All temporary source changes were reverted; the worktree was clean at the exact head after validation.

RerankerGuo (approved):
No blocking issues found at exact head `56f6343c595aefd47dea2abf9c24b9a5c7b33b00`.

The change removes the streaming path's divergent inline 429 handling and routes it through the same policy used by non-streaming calls. This preserves normal pause-and-retry behavior while preventing a provider's over-cap `Retry-After` from becoming a capped pause followed by an internal acquire timeout that hides the original quota error.

```mermaid
flowchart LR
    A[Streaming request raises] --> B[Unified rate-limit handler]
    B --> C{Exception type}
    C -->|Internal acquire timeout| D[Re-raise unchanged]
    C -->|429 Retry-After above cap| E[Re-raise provider error immediately]
    C -->|Normal 429| F[Record model pause]
    C -->|Other retryable error| G[Normal backoff policy]
    F --> G
    style F fill:#bbdefb,color:#0d47a1
```

The independent negative control is strong: reverting only the one-line source change while retaining the tests gives `1 failed, 1 passed`; the over-cap case surfaces the limiter's `Rate limit exceeded` and records a 60-second pause, while the under-cap control still passes. Restoring the PR gives `2 passed` and preserves the original provider exception with one inner call and no pause.

Fresh current-main merge tree `380efae6d925dc9c102090a461ccfc540047f3a1` verification:

```text
retry_chat_model: 49 passed
complete providers unit suite: 368 passed
changed-file pre-commit: all applicable hooks passed
git diff --check: passed
remote RBP, CodeQL, format, and pre-commit checks: passed
```

The merge also retains #6721's `test_stream_recovers_agentscope_msg_via_formatter_fallback`, so resolving the append conflict did not drop the formatter-recovery contract.

qbc2016 (approved):
LGTM.

PR #6623, opened 2026-08-01
fix(acp): prevent final text loss when notifications race the prompt response

### LINKED ISSUE #6625: [Bug]: ACP delegate_external_agent sometimes returns "completed without text output" when notifications race the prompt response
## Description

When an external ACP agent writes a `session/update` notification and the `session/prompt` response back-to-back (e.g. same TCP segment), `delegate_external_agent` sometimes returns **"completed without text output"** even though the agent produced text.

## Root cause

The ACP transport (external `acp` SDK) resolves the prompt response future **inline**, while the notification runner is still queued as a **separate task**. In `ACPService._wait_for_prompt_outcome()`, `finish_prompt()` is called immediately after the prompt future resolves — at that point the notification carrying the final text may not have been processed yet, so `_assistant_text` is empty and the final text is dropped.

## Repro

Occurs reliably when the agent server flushes the notification and the response together in one TCP segment / readline batch. The server-side symptom matches the race documented in the ACP SDK transport.

## Fix (PR #6623)

Two complementary changes:
1. `client.py`: `session_update()` emits accumulated text immediately on `AgentMessageChunk` (in addition to the `finish_prompt()` flush), so streaming consumers capture the text via `on_message`.
2. `service.py`: `_wait_for_prompt_outcome()` yields to the event loop (`await asyncio.sleep(0)`) before calling `finish_prompt()`.

See https://github.com/agentscope-ai/QwenPaw/pull/6623

## Environment

- QwenPaw version: main (2026-08-01)
- OS: Linux (aarch64)

cocoakekeyu:
Fixes #6625

## Description

When a `session/update` notification and the `session/prompt` response arrive in the same TCP segment, the ACP transport resolves the prompt future **inline**, while the notification runner is still queued as a **separate task** (see `acp` SDK `Connection._process_message` / dispatcher).

Because `_wait_for_prompt_outcome()` calls `finish_prompt()` immediately after the prompt future resolves, `_assistant_text` can still be empty at that point — the notification that carried the final text has not been processed yet. The result is that `delegate_external_agent` reports the turn as **"completed without text output"** even though the agent did produce text.

This is a real-world race: it reproduces reliably when the agent server writes the notification and the response back-to-back (same TCP segment / same readline batch).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Documentation updated (if needed) — N/A, internal async behavior only
- [x] Ready for review

## Fix

Two complementary changes:

1. **client.py** — `session_update()` now emits accumulated assistant text immediately when handling `AgentMessageChunk`, in addition to the existing `finish_prompt()` flush. Streaming consumers such as `delegate_external_agent` capture the text via `on_message` even if the response races ahead of the notification task.

2. **service.py** — `_wait_for_prompt_outcome()` yields to the event loop (`await asyncio.sleep(0)`) after the prompt task completes and before calling `finish_prompt()`, giving the queued notification runner a chance to process pending `session/update` notifications.

## Testing

Added `tests/unit/agents/test_acp_text_race.py` with three regression tests:
- `AgentMessageChunk` emits text during `session_update` (not only at `finish_prompt()`)
- `finish_prompt()` returns accumulated text after a chunk notification
- `_wait_for_prompt_outcome()` observes text that arrived via a notification queued before the prompt response resolved

All existing ACP tests continue to pass.

## Evidence

Ran `pre-commit run --files` on all three changed files (src/qwenpaw/agents/acp/client.py, src/qwenpaw/agents/acp/service.py, tests/unit/agents/test_acp_text_race.py). Result: all hooks Passed — check python ast, check docstring is first, fix python encoding pragma, detect private key, trim trailing whitespace, Add trailing commas, mypy, black, flake8, pylint, Lint GitHub Actions workflow files.

Ran `python -m pytest tests/unit/agents/test_acp_text_race.py tests/unit/agents/test_acp_client_trusted.py -q`. Result: 9 passed in 3.87s — three new regression tests plus the existing ACP trusted-client tests, all green.

## Additional Notes

- Small, focused change: +16 lines in source, one new test file.
- No change to the ACP wire protocol.
- Makes text delivery robust under the notification/response race instead of relying on timing.

### DIFF
--- a/src/qwenpaw/agents/acp/client.py
+++ b/src/qwenpaw/agents/acp/client.py
@@ -231,6 +231,12 @@ class ACPHostedClient:
         if isinstance(update, AgentMessageChunk):
             self._thinking_active = False
             await self._accumulate_assistant_content(update.content)
+            # Emit the accumulated text immediately.  Streaming consumers
+            # such as delegate_external_agent rely on on_message callbacks
+            # to capture the final text; flushing here (in addition to the
+            # later finish_prompt() flush) makes the text available even if
+            # the prompt response races ahead of this notification task.
+            await self._emit_assistant_text_delta()
             return
 
         await self.flush_assistant_text()
--- a/src/qwenpaw/agents/acp/client.py
+++ b/src/qwenpaw/agents/acp/client.py
@@ -61,6 +61,10 @@ class ACPHostedClient:
             RequestPermissionResponse
         ] | None = None
         self._permission_requested = asyncio.Event()
+        # Set when new assistant text chunks arrive; waited on by
+        # _wait_for_prompt_outcome to ensure all chunks are processed
+        # before finish_prompt() is called.
+        self._assistant_text_event = asyncio.Event()
 
     @property
     def pending_permission(self) -> SuspendedPermission | None:
@@ -79,6 +83,7 @@ class ACPHostedClient:
         self._thinking_active = False
         self._pending_permission = None
         self._permission_requested.clear()
+        self._assistant_text_event.clear()
         self._session_acc = SessionAccumulator()
 
     def resume_prompt(self, on_message: MessageHandler) -> None:
@@ -231,12 +236,10 @@ class ACPHostedClient:
         if isinstance(update, AgentMessageChunk):
             self._thinking_active = False
             await self._accumulate_assistant_content(update.content)
-            # Emit the accumulated text immediately.  Streaming consumers
-            # such as delegate_external_agent rely on on_message callbacks
-            # to capture the final text; flushing here (in addition to the
-            # later finish_prompt() flush) makes the text available even if
-            # the prompt response races ahead of this notification task.
-            await self._emit_assistant_text_delta()
+            # Signal that new assistant text is available. The service layer
+            # will wait on this event (with a timeout) before calling
+            # finish_prompt(), ensuring all chunks have been processed.
+            self._assistant_text_event.set()
             return
 
         await self.flush_assistant_text()
--- a/src/qwenpaw/agents/acp/service.py
+++ b/src/qwenpaw/agents/acp/service.py
@@ -387,6 +387,16 @@ class ACPService:
                 await conversation.client.finish_prompt()
                 raise ACPSessionError(str(exc)) from exc
 
+            # Yield once so the dispatcher gets a chance to process any
+            # session/update notifications that arrived in the same TCP
+            # segment as the prompt response.  Without this, the response
+            # handler resolves the prompt future inline, while the
+            # notification runner is still queued as a separate task, so
+            # ``finish_prompt()`` below can observe an empty
+            # ``_assistant_text`` and drop the final text (observed as
+            # "completed without text output" by delegate_external_agent).
+            await asyncio.sleep(0)
+
             conversation.prompt_task = None
             finished_event = await conversation.client.finish_prompt()
             pending_permission = conversation.client.pending_permission
--- a/src/qwenpaw/agents/acp/service.py
+++ b/src/qwenpaw/agents/acp/service.py
@@ -387,18 +387,27 @@ class ACPService:
                 await conversation.client.finish_prompt()
                 raise ACPSessionError(str(exc)) from exc
 
-            # Yield once so the dispatcher gets a chance to process any
-            # session/update notifications that arrived in the same TCP
-            # segment as the prompt response.  Without this, the response
-            # handler resolves the prompt future inline, while the
-            # notification runner is still queued as a separate task, so
-            # ``finish_prompt()`` below can observe an empty
-            # ``_assistant_text`` and drop the final text (observed as
-            # "completed without text output" by delegate_external_agent).
-            await asyncio.sleep(0)
+            # Wait for any in-flight AgentMessageChunk notifications to be
+            # processed before calling finish_prompt(). The client sets
+            # _assistant_text_event when a new chunk arrives; we wait with a
+            # short timeout to avoid blocking indefinitely if the event is
+            # never set (e.g., response with no text content).
+            try:
+                await asyncio.wait_for(
+                    conversation.client._assistant_text_event.wait(),
+                    timeout=0.5,
+                )
+            except asyncio.TimeoutError:
+                pass
 
             conversation.prompt_task = None
             finished_event = await conversation.client.finish_prompt()
+            # If finish_prompt() returned no text but we know chunks arrived
+            # (event is set), retry once after a tiny yield — this covers
+            # the case where the notification task hasn't quite flushed yet.
+            if finished_event is None and conversation.client._assistant_text_event.is_set():
+                await asyncio.sleep(0)
+                finished_event = await conversation.client.finish_prompt()
             pending_permission = conversation.client.pending_permission
             if pending_permission is not None:
                 return {
--- a/src/qwenpaw/agents/tools/delegate_external_agent.py
+++ b/src/qwenpaw/agents/tools/delegate_external_agent.py
@@ -513,6 +513,7 @@ async def _stream_action_responses(
     flush_task: Optional[asyncio.Task[None]] = None
     final_text_event: Optional[dict[str, Any]] = None
     final_fallback_event: Optional[dict[str, Any]] = None
+    stream_text_parts: list[str] = []
 
     async def flush_snapshot() -> None:
         nonlocal header_sent
@@ -578,6 +579,9 @@ async def _stream_action_responses(
 
         if event_type == "text":
             final_text_event = event
+            raw_text = event.get("text")
+            if raw_text:
+                stream_text_parts.append(str(raw_text))
         elif event_type == "error":
             final_fallback_event = event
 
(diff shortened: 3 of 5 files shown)


x1n95c:
Thanks for the PR! We tried to reproduce the race as described but couldn't observe the text loss when the notification and response arrive in one segment. Could you share the relevant logs and which agent you delegated to, so we can confirm the actual wire order before proceeding with the review?

RerankerGuo:
Independent review at exact head `530348f653372826638548a3ab905ae244178cd0`.

I found a user-visible regression in the new immediate-emission path when an ACP assistant response contains more than one `AgentMessageChunk`.

`ACPHostedClient._emit_assistant_text_delta()` correctly emits only the new delta. However, `_stream_action_responses.on_message()` overwrites `final_text_event` for every text event, and the final selection prefers that event over the complete `run_result["event"]`:

```python
event = final_text_event or final_fallback_event or run_result.get("event")
```

After this PR, `final_text_event` is therefore the **last delta**, not the complete assistant response. The final `ToolChunk` becomes truncated and repeats the last delta that was already emitted in the streaming snapshot.

Minimal reproduction:

```python
# FakeService.run_turn creates a real ACPHostedClient, starts a prompt,
# sends AgentMessageChunk("Hello") then AgentMessageChunk(" world"),
# and returns client.finish_prompt() as the completed event.
chunks = [
    chunk
    async for chunk in _stream_action_responses(
        service=FakeService(),
        chat_id="chat",
        action_name="start",
        runner_name="fake",
        message_text="go",
        execution_cwd=Path("/tmp"),
    )
]
assert chunks[-1].content[-1].text == "[assistant]\nHello world"
```

Results with the identical pytest:

```text
origin/main (bacac741):
1 passed
final stream:
  non-final: [assistant]\nHello world
  final:     [assistant]\nHello world

PR head (530348f6):
1 failed
E assert '[assistant]\nworld' == '[assistant]\nHello world'
final stream:
  non-final: [assistant]\nHello, [assistant]\nworld
  final:     [assistant]\nworld
```

The PR's focused suite still passes:

```bash
.venv/bin/pytest \
  tests/unit/agents/test_acp_text_race.py \
  tests/unit/agents/test_acp_client_trusted.py -q
# 9 passed
```

Those tests cover the client and service timing but not the downstream streaming consumer with multiple chunks.

Suggested direction:

1. Preserve/accumulate text deltas into a complete final text event in `delegate_external_agent`, rather than replacing `final_text_event` on each delta.
2. Prefer the complete `run_result["event"]` when it exists, while retaining the accumulated streamed-text fallback for the original race where `finish_prompt()` returns no event.
3. Add a two-chunk regression at the `_stream_action_responses` boundary.

This finding does not dispute the reported notification/response race; it is a separate regression introduced by emitting every `AgentMessageChunk` immediately.

cocoakekeyu:
Thanks for the review, and apologies for the bare reproduction — the race is timing-dependent, so I understand it's easy to miss.

**Which agent we delegate to:** a self-hosted ACP-over-TCP agent running a full QwenPaw `Workspace` agent. It streams our task via `session/update` (`AgentMessageChunk`) notifications and writes the `session/prompt` response **back-to-back with the last notification, in the same TCP segment / readline batch** — that same-batch arrival is the wire condition the PR targets.

**Why the race exists:** in the `acp` SDK, `Connection._process_message()` handles the two frames differently — notifications go through the queue/Dispatcher as a **separate task**, while responses (which carry an `id`) are resolved **inline**, completing the prompt future immediately. When both frames arrive in one TCP batch, the prompt future completes before the queued notification task has run, so `_wait_for_prompt_outcome()` calls `finish_prompt()` while `_assistant_text` is still empty → `delegate_external_agent` reports "completed without text output".

**Deterministic repro (no network needed):** the tests in this PR simulate exactly that scheduling. On `main`, `tests/unit/agents/test_acp_text_race.py::test_yields_before_finish_prompt` **fails**; on this branch all 3 pass:

```bash
# on main
git checkout origin/main
.venv/bin/pytest tests/unit/agents/test_acp_text_race.py -q   # 1 failed (the race), 2 passed

# on branch
git checkout fix/acp-text-race-condition
.venv/bin/pytest tests/unit/agents/test_acp_text_race.py -q   # 3 passed
```

For end-to-end verification I can additionally extend the PR with: (a) a minimal fake ACP server that writes a notification + the response with a single `write()`, and (b) a small client driving `ACPService`/`delegate_external_agent` against it. And I can attach the server-side trace showing both frames flushed together — one for raw socket dump or JSON-RPC frame log, whichever is easier for you. Let me know which format helps your review and I'll post it.

cocoakekeyu:
Confirmed — excellent catch, and thanks for the precise reproduction. With the immediate-emission path, `final_text_event` was last-delta-wins, so the final `ToolChunk` was truncated to the last `AgentMessageChunk` whenever a turn emitted more than one.

**Fixed and pushed on `fix/acp-text-race-condition`** (commit `59485b1`):

1. `_stream_action_responses` now **prefers the complete `run_result["event"]`** (finish_prompt()'s full accumulated text) when present;
2. accumulated streamed deltas are joined and kept as the fallback **only** for the original race, where `finish_prompt()` yields no event;
3. added a two-`AgentMessageChunk` regression at the `_stream_action_responses` boundary, asserting the final text reads `[assistant]\nHello world`.

Verified with your minimal repro — before the fix the final chunk was truncated to `[assistant]\nworld`; after the fix it is the complete `[assistant]\nHello world`:

```bash
.venv/bin/pytest tests/unit/agents/test_acp_text_race.py \
  tests/unit/agents/tools/test_delegate_external_agent_stream_text.py -q
# 5 passed (3 race regression + 2 multi-chunk tests)
```

The original 3 race tests still pass, plus the existing `test_acp_client_trusted.py` and timeout tests (13 passed in total), and `pre-commit` is all green. Thanks again for catching this — the PR is in much better shape with the multi-chunk regression covered.

RerankerGuo (approved):
Revalidated the author fix at exact head `59485b157a693ac8565292c812e9c4854cc89a4b` and a clean current-main trial merge (`52a5e371b502eee59c2d666d932f64070e7a6b5f`, tree `6f5613822e31574b1fdfe68dc5b6e0daf4c674ba`).

The previous multi-chunk truncation is fixed:

- `_emit_assistant_text_delta()` emits only the not-yet-emitted suffix.
- `_stream_action_responses()` now prefers the complete `finish_prompt()` event.
- If that event is absent in the original notification/response race, streamed deltas are joined into the complete fallback.
- The new downstream regressions cover both the normal two-chunk completion and the race fallback (`Hello` + ` world` -> one complete final `[assistant]\nHello world`).

Verification:

```text
focused exact/trial:
  test_acp_text_race.py
  test_delegate_external_agent_stream_text.py
  test_acp_client_trusted.py
  11 passed on each tree

ACP/delegate unit scope:
  exact: 58 passed, 2 skipped
  current-main trial: 85 passed, 2 skipped

real ACP runner integration:
  exact: 5 passed
  current-main trial: 5 passed

changed-file hooks on the trial:
  python AST, docstring/encoding, private-key, whitespace,
  trailing-comma, mypy, Black, Flake8, Pylint: passed

git diff --check: passed
remote Real behavior proof: passed
```

The local actionlint environment could not be installed because `proxy.golang.org` timed out; this PR changes no workflow files, and all applicable changed-file hooks passed.

The original race fix and the downstream final-text behavior now hold together on current main. I found no remaining blocker in the exercised scope.

x1n95c:
Hi, thanks for the contribution! I went through the changes — this is a real potential bug and the fixes work as expected. The unit tests feel unnecessary here and I have two additional comments:

**1. Per-chunk emit degrades the streamed progress output.** `client.py:239` now emits a delta per `AgentMessageChunk`, but `_stream_action_responses` treats every text event as a standalone snapshot item: dedup by stripped content, one TextBlock each. With 10 token-level chunks (`"The ", "cat ", ..., "the ", "mat", ".", " The ", "end", "."`) the intermediate output goes from a single block `"The cat sat on the mat. The end."` to 8 separate `[assistant]` blocks, with the repeated `" The "` and the second `"."` silently dropped by `seen_stream_items`. The final block
stays complete, but the progress stream is what the coordinating model sees mid-turn. Deltas are being pushed through a path designed for coarse, complete fragments.

**2. `asyncio.sleep(0)` isn't a barrier.** `task/dispatcher.py:90-94` spawns the notification into *another* task and calls `task_done()` immediately, so not even `queue.join()` would synchronize here. The single yield works today only because ready-queue FIFO happens to schedule the handler's first step before the resumed turn — that breaks with multiple queued notifications, or any real suspension point inside the handler. The comment reads as deterministic; it's best-effort.

One option that covers both: have the client set an `asyncio.Event` on `AgentMessageChunk` and let `_wait_for_prompt_outcome` retry `finish_prompt()` with a bounded wait when it returns no text. That keeps the fix in one place and leaves streaming granularity untouched, so the per-chunk emit can go.

cocoakekeyu:
Thanks for the thorough review @x1n95c — both points are spot on.

**Per-chunk emit degrading streamed progress** — agreed. The token-level deltas were indeed polluting the snapshot pipeline designed for coarse fragments.

**`asyncio.sleep(0)` not being a barrier** — completely correct. The dispatcher spawns notifications as independent tasks and `task_done()` fires immediately, so even `queue.join()` would not synchronize. The single yield only worked by FIFO luck.

**Refactored per your suggestion (commit 01a8670):**
- `client.py`: Added `_assistant_text_event = asyncio.Event()`; `AgentMessageChunk` now just sets the event (no per-chunk emit).
- `service.py`: `_wait_for_prompt_outcome` waits on the event with a 0.5s timeout, then retries `finish_prompt()` once if it returned `None` but the event is set. This keeps the fix in one place and leaves streaming granularity untouched.
- Updated regression tests (`test_acp_text_race.py`, `test_delegate_external_agent_stream_text.py`) to pin the new deterministic behavior.

All 44 ACP unit tests pass. The `delegate_external_agent` multi-chunk completeness tests also pass.

Appreciate the cleaner direction — this is a better fix.

x1n95c:
Thanks for the changes. However, adding a fixed 0.5-second wait for a normal AgentMessageChunk does not seem appropriate.

  The race condition described here also appears to be extremely rare, and the current tests are based entirely on mocked scenarios. We do not yet have evidence from real QwenPaw usage or production logs. Unless this issue can
  be reproduced in a practical QwenPaw workflow, I suggest closing this PR rather than introducing additional synchronization overhead for the general case.

PR #6648, opened 2026-08-03
test(integration): align chrome + coding-project cases with product contract

yutai78786:
## Summary

Four integration cases assert against APIs / behaviors that never matched
the product code they exercise. They have failed on **every OS** in the
nightly full sweep since the moment they landed (confirmed reproduced
locally on `main`). None of these are product logic bugs — the tests are
wrong. This PR aligns each test with the real contract.

### 1. `test_chrome_nm_host.py` — wrong API names + wrong fake signatures
- The product `plugins/bundle/chrome/assets/scripts/nm_host.py` exposes
  `read_nm_frame` / `write_nm_frame` (raw length-prefixed **bytes**), but
  the tests called `read_nm_message` / `write_nm_message`, which do not
  exist → `AttributeError`. Fixed to use the real API and do the
  JSON encode/decode in the test.
- The five fake WebSocket stand-ins declared `async def close(self)`, but
  the host closes the socket with `close(code=..., reason=...)` to match
  the real `websockets` library (`ClientConnection.close(self, code, reason)`,
  verified on websockets 16.0) → `TypeError: unexpected keyword argument
  'code'`. Fixed to accept `**kwargs`.

### 2. `test_coding_project.py::test_import_local_copies_source` — not updated for #6487
- #6487 added `_validate_import_source`, which returns **403** for any
  source outside the user's home directory. The test seeded its source
  under a pytest tmp dir (outside home) → 403 instead of 200. Fixed to
  seed under `Path.home()` and clean up in `finally`.

### 3. `test_chrome_service_worker.py` — assertion is the opposite of the contract
- `runReceiptCommand` deliberately **re-executes** a stale `RECEIVED`
  receipt: a stale RECEIVED state is proof the executor never crossed the
  durable RUNNING barrier, so it is safe to run (see the in-code comment;
  `observed_states` likewise maps `received_stale_epoch -> NOT_STARTED`).
  The test asserted `executorCalls == 0`, the opposite. Fixed to assert
  re-execution (`executorCalls == 1`, receipt `COMPLETED`) and renamed the
  test + harness scenario accordingly.

## Test plan
- [x] `pytest tests/integration/test_chrome_nm_host.py` — 11 passed
- [x] `pytest tests/integration/test_chrome_service_worker.py` — passed
- [x] `pytest tests/integration/test_coding_project.py::test_import_local_copies_source` — passed
- [x] Combined run of all three files — 23 passed
- [x] `pre-commit run --files <changed>` — green (mypy/black/flake8/trailing-comma)

## Not included
`tests/integration/browser/*` (7 cases) fail because the nightly
Integration job has no `playwright install` step — infrastructure, tracked
separately. `test_acp_runner.py::test_acp_close_after_start` (Windows-only,
`os.replace` `WinError 5` on `jobs.json`) is a product-side atomic-write
issue on Windows, also out of scope here.

### DIFF
--- a/tests/integration/js/service_worker_harness.mjs
+++ b/tests/integration/js/service_worker_harness.mjs
@@ -205,7 +205,7 @@ const scenarios = {
     return { observedState: response.targetCommandFact.observedState };
   },
 
-  async stale_epoch_no_reexec() {
+  async stale_received_reexecuted() {
     seedTargetReceipt({
       ...commandParams,
       state: "RECEIVED",
--- a/tests/integration/test_chrome_nm_host.py
+++ b/tests/integration/test_chrome_nm_host.py
@@ -53,7 +53,9 @@ def test_two_mib_inbound_message_is_accepted() -> None:
     raw = payload.encode("utf-8")
     reader = io.BytesIO(struct.pack("<I", len(raw)) + raw)
 
-    message = host.read_nm_message(reader)
+    frame = host.read_nm_frame(reader)
+    assert frame is not None
+    message = json.loads(frame)
 
     assert message["result"]["data"].startswith("x")
     assert len(message["result"]["data"]) == 2 * 1024 * 1024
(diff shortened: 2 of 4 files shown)


yutai78786:
Closing to run fork CI verification first per our process; will reopen after validation.

PR #6653, opened 2026-08-03
fix(ci): fence-aware section extraction in real-behavior-proof (fixes #6626)

### LINKED ISSUE #6626: [Bug]: Real behavior proof gate strips fenced Evidence blocks entirely (porting deviation from openclaw)

hanson-hex:
## Description

What problem this solves: the Python port of `real-behavior-proof-policy` deleted **entire fenced code blocks** via

```python
re.sub(r"```[^\n]*\n.*?```", "", text, flags=re.DOTALL)
```

So an Evidence section that contained **only** a fenced terminal transcript (a very natural thing to paste — no prose outside the fence) was stripped to empty and flagged as `MISSING`, hard-failing the `Real behavior proof` CI gate with a `triage: needs-pr-context` label and no clear indication that the fence itself was the problem. This was reported in #6626 and hit in practice (e.g. PR #6646's own CI run, where `## Evidence` mixed fenced blocks with `###` sub-headings and the gate returned `MISSING sections: ['Evidence']` despite a 1958-char body).

The port deviated from openclaw's upstream `stripMarkdownFenceMarkers` + `extractMarkdownSections`: upstream only removes fence **marker lines** and uses a fence-tracking state machine so headings inside fences are not treated as section boundaries.

## Fix

Replace `_strip_code_fences` + regex section splitting with a **fence-aware** `_extract_sections`:

- fence marker lines (``` / ~~~), including the info string, are dropped;
- fenced content is **retained**, so a transcript-only Evidence section has real authored content;
- heading-like lines inside a fence never start or truncate a section.

Aliases and the `sections[heading] = ...` overwrite semantics are unchanged, so existing behavior for prose sections is preserved.

## Component(s) Affected
- [ ] Core / Backend
- [ ] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [x] CI/CD
- [ ] Scripts / Deploy

## Type of Change
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Checklist
- [x] Documentation updated (inline docstring explains the upstream alignment)
- [x] Ready for review

## Testing

- `pytest tests/unit/github/test_real_behavior_proof_policy.py tests/unit/github/test_real_behavior_proof_check.py -q` → **34 passed** (31 original + 3 new regression cases).

## Evidence

The previously-failing case now passes:

```bash
$ PYTHONPATH=scripts/github python -c "
from real_behavior_proof_policy import evaluate_pull_request_context
body = '''## Description

The gateway crashed on startup.

## Evidence

\`\`\`text
\$ pytest -q
4 passed
\`\`\`'''
ev = evaluate_pull_request_context(body=body, author_association='CONTRIBUTOR', author_type='User', labels=[])
print(ev.status.value, ev.missing_sections)
"
passed []
```

Before this change the same body returned `missing ['Evidence']`.

New regression tests added:

- `test_passes_with_fenced_evidence_only_no_prose` — Evidence = fenced transcript only → `PASSED`.
- `test_fenced_heading_does_not_start_section` — a `## heading` line inside a fenced block is NOT registered as a section and does not truncate Evidence → `PASSED`, `missing_sections == []`.

Related: #6646 attempts to fix fork-PR body fetching, but its own CI failure is caused by this #6626 bug, not by a missing body (its run shows `body length: 1958 chars`). This PR resolves the underlying parsing issue; #6646 can be re-evaluated (or closed) separately.

Fixes #6626.

### DIFF
--- a/scripts/github/real_behavior_proof_policy.py
+++ b/scripts/github/real_behavior_proof_policy.py
@@ -91,8 +91,8 @@ def _extract_sections(body: str) -> dict[str, str]:
 
     This aligns the port with openclaw's upstream ``extractMarkdownSections``
     + ``stripMarkdownFenceMarkers`` (a fence-tracking state machine) after
-    the original Python port wrongly deleted the entire fenced block via
-    ``re.sub(r"\`\`\`[^\n]*\n.*?\`\`\`", "", ...)`` (see issue #6626).
+    the original Python port wrongly deleted the entire fenced block via a
+    ``re.sub`` over the triple-backtick fenced-block pattern (see #6626).
     """
     body = _normalize_line_endings(body)
     body = _mask_html_comments(body)
--- a/scripts/github/real_behavior_proof_policy.py
+++ b/scripts/github/real_behavior_proof_policy.py
@@ -76,26 +76,70 @@ def _mask_html_comments(text: str) -> str:
     )
 
 
-def _strip_code_fences(text: str) -> str:
-    """Remove fenced code blocks so headings/copy-pasted template text
-    inside them do not trick the section parser."""
-    return re.sub(r"```[^\n]*\n.*?```", "", text, flags=re.DOTALL)
+_FENCE_RE = re.compile(r"^(\s{0,3})(`{3,}|~{3,})")
 
 
 def _extract_sections(body: str) -> dict[str, str]:
-    """Split a PR body into a ``{heading: body}`` mapping."""
+    """Split a PR body into a ``{heading: body}`` mapping.
+
+    Headings inside fenced code blocks are NOT treated as section
+    boundaries. Fence marker lines (``` / ~~~) are dropped so an info
+    string or a fenced heading cannot trick the parser, but the code
+    *content* between the markers is preserved — an Evidence section that
+    contains only a fenced terminal transcript still counts as authored
+    content.
+
+    This aligns the port with openclaw's upstream ``extractMarkdownSections``
+    + ``stripMarkdownFenceMarkers`` (a fence-tracking state machine) after
+    the original Python port wrongly deleted the entire fenced block via
+    ``re.sub(r"\`\`\`[^\n]*\n.*?\`\`\`", "", ...)`` (see issue #6626).
+    """
     body = _normalize_line_endings(body)
     body = _mask_html_comments(body)
-    body = _strip_code_fences(body)
 
     sections: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


lalaliat (approved):
LGTM

PR #6657, opened 2026-08-03
fix(sandbox): report constraints the backend cannot enforce

vanwaals:
## Description

`SandboxConfig` accepts `mounts`, `deny_paths`, `network_allow`, `network_ports`, `max_processes` and `max_memory_mb`, but no backend applies all of them and `NoneSandbox` applies none. The gap was **silent**: an operator who set `deny_paths` to protect credentials got a clean log while the paths stayed fully readable. That is the common case inside containers, where no kernel backend is available and QwenPaw falls back to `mode=none`.

Each backend now declares the fields it genuinely applies, and `report_unenforced_config()` logs everything else when the sandbox is constructed — constraints that form a security boundary at `WARNING`, the rest at `DEBUG`. The default `_ENFORCED_FIELDS` is empty, so a newly added backend is loud until it declares what it honours.

This PR **adds no enforcement**. It only makes the existing gaps visible.

### Field audit

Auditing every field against its consumers turned up more than the missing report:

- **`max_processes` / `max_memory_mb` docstrings advertised support that does not exist** — they claimed "Windows Job / Linux cgroups", but no Windows backend references either field. `_JOBOBJECT_EXTENDED_LIMIT_INFORMATION.ProcessMemoryLimit` and `ActiveProcessLimit` are declared as ctypes struct members and never assigned.
- **`env_mode="allowlist"` is not implemented** — every backend behaves as `"inject"`. Annotated as such, keeping the API field per the project's convention for unimplemented fields.
- **`NoneSandbox` ignored `shell_executable`** — it pinned `$SHELL`. It now honours the configured shell, so the field is genuinely enforced instead of merely reported.
- **Four scattered per-backend warnings folded into the shared reporter** (macOS ×2, Landlock ×1, AppContainer ×1) to stop double-reporting. Their actionable text survives via `_ENFORCEMENT_HINTS`.
- **The docs contradicted themselves** — the capability table said Seatbelt supports network allow/deny while "Current limitations" said network isolation was entirely unimplemented. Both `security.en.md` and `security.zh.md` are now accurate, including the per-backend degradation directions below.

### Review round: declarations that did not match their own enforcement path

Code review found that three backends declared constraints their own enforcement branch skipped — the exact bug class this PR set out to eliminate. All are fixed here, and each now has a test comparing the declaration against the artefact the backend really produces:

- **Windows unelevated claimed `network_allow`.** `WindowsUnelevatedSandbox` narrowed `_ENFORCED_FIELDS` but inherited `_enforced_fields()`, which adds `network_allow` for the absolute postures. That is only sound where a kernel mechanism backs it (AppContainer capability SIDs, WFP rules); this backend's "block" is HTTP(S) proxy environment variables that a raw socket ignores, and a domain allowlist sets nothing at all. It now overrides `_enforced_fields()` to never claim the field, with a hint naming the proxy limitation.
- **Landlock claimed `network_ports` whenever ABI ≥ 4.** Port rules attach to the same handled-access mask as the wholesale block, and `_generate_sandbox_script()` only sets that mask for `network_allow=[]`. With the network open — including the `["*"]` default `ResourceGovernor` compiles — no port rule reached the ruleset while the reporter stayed silent. The declaration now matches the script's own condition.
- **Seatbelt claimed `platform_hints` wholesale** while the profile compiler reads only `seatbelt_extra_rules`. A typo'd key was dropped twice over: skipped by the reporter because the field was claimed, then ignored by the compiler. The claim is now conditional on every key being one that is consumed.
- **`env_mode="allowlist"` was reported at `DEBUG`.** The default log level is `info`, so the failure was invisible in normal deployments — while the whole parent environment, API keys included, still reached the child. `env_mode` is now a security-boundary field. `platform_hints` moved with it, on the same reasoning: it carries admin-authored Seatbelt deny clauses.

Two further problems found while verifying the above:

- **`WindowsElevatedSandbox` shared the fail-open network hint, and WFP is the opposite.** Its condition is `not (bool(network_allow) and "*" in network_allow)`, so a domain allowlist blocks the network **entirely** rather than opening it. The shared hint said "all network access is ALLOWED" — accurate for Seatbelt and AppContainer, exactly backwards here. It now has its own fail-closed hint.
- **`NoneSandbox` fell back to `/bin/bash` silently** when a configured `shell_executable` did not exist, while declaring the field enforced. The fallback is now logged.

**Related Issue:** Relates to internal QA test case `TC-SB-05` (SandboxConfig configuration items: mounts / deny_paths / network / resource limits).

**Security Considerations:** A deployment that relied on `deny_paths` or `network_allow` under a backend that cannot apply them was already unprotected; it will now say so in the log. Points for reviewers:

- The reporter treats `network_allow=["*"]` as "no constraint requested", which is exactly what `ResourceGovernor.compile_sandbox_config()` emits, so the production path stays silent and the report cannot decay into noise operators filter out. `network_allow=[]` (the dataclass default) means _block all_ and is therefore reported as a real request when unenforceable.
- **A domain allowlist degrades differently per backend**, and the hint text now reflects that: Seatbelt and AppContainer open the network fully; the elevated Windows backend blocks it entirely. Neither direction is what the operator asked for, so both are reported.
- `max_processes` / `max_memory_mb` **remain unenforced**. Implementing them needs Linux cgroups / Windows Job objects and is deliberately left to a follow-up.
- **On `network_ports` semantics** (raised in review): this PR deliberately does not change enforcement. Landlock installs port rules only alongside the wholesale block, so the declaration was corrected to match rather than the generator extended. Making port limits independent of `network_allow` is a behaviour change that needs its own design decision and belongs in a separate PR; until then the limitation is reported at runtime and documented.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [x] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

Not applicable — this PR touches no channel code.

## Testing

```bash
pytest tests/unit/sandbox/test_unenforced_config.py -q   # 54 cases
pytest tests/unit/sandbox -q                             # whole backend suite
```

To see the behaviour by hand, construct `NoneSandbox` (the container fallback
TC-SB-05 exercised) with a restrictive config and watch the log:

```python
import logging
logging.basicConfig(level=logging.DEBUG)
from qwenpaw.sandbox.config import SandboxConfig, SandboxMode, MountSpec, PortRule
from qwenpaw.sandbox.local_sandbox import NoneSandbox

NoneSandbox(SandboxConfig(
    mode=SandboxMode.NONE, workspace_dir="/tmp/ws",
    mounts=[MountSpec(path="/tmp/ws", writable=True)],
    deny_paths=["~/.ssh", "~/.aws"], network_allow=["github.com"],
    network_ports=[PortRule(port=443)], max_processes=8, max_memory_mb=100,
))
# → six WARNING lines, one per ignored constraint

NoneSandbox(SandboxConfig(
    mode=SandboxMode.NONE, workspace_dir="/tmp/ws", network_allow=["*"],
))
# → silent; this is what ResourceGovernor actually compiles
```

`tests/unit/sandbox/test_unenforced_config.py` (54 cases) covers:

- **Requested-constraint detection** — `network_allow=["*"]` is not a request; `[]` (block-all) is; default `env_mode` / `shell_executable` are not; all nine tracked fields are detected.
- **Severity split** — security-boundary fields at `WARNING`, the rest at `DEBUG`; enforced fields stay silent; hints appear only for unenforced fields. A parametrized guard asserts `env_mode`, `platform_hints`, `max_processes` and `max_memory_mb` stay security-classified, so reclassifying one back to `DEBUG` fails the suite.
- **The TC-SB-05 regression** — a fully restrictive config under `NoneSandbox` reports exactly the six isolation constraints, while the governor's real default config produces zero output.
- **Declaration vs. real enforcement** (new category, added for the review round):
  - Landlock — parametrized over `[]`, `["*"]` and a domain list, asserting `"network_ports" in _enforced_fields()` **equals** whether the generated script actually contains the port rule.
  - Windows — the unelevated backend never claims `network_allow` (either posture) nor `deny_paths`; the elevated one does claim it, and its hint says fail-closed rather than fail-open.
  - Seatbelt — a supported `platform_hints` key is claimed _and_ appears in the compiled profile; a typo'd key is neither; a mixed dict is reported with every key listed so the typo is findable.
  - `NoneSandbox` — a missing configured shell is reported and still executes; an existing one is silent.

Three pre-existing tests in `tests/unit/sandbox/test_config.py` asserted the removed macOS-specific warnings; they were rewritten against the central reporter via `caplog`, preserving their original intent.

**Mutation check.** The previous test round passed while three backends mis-declared their fields, so the new consistency tests were verified against the bug they exist to catch: reverting the Landlock declaration to `if self._abi_version >= 4: enforced.add("network_ports")` fails 3 cases (`["*"]`, the domain list, and the explicit regression test) and restoring it returns the suite to green.

## Evidence

```bash
$ pre-commit run --all-files
```

```bash
$ pytest tests/unit/sandbox/test_unenforced_config.py -q -p no:randomly
......................................................                   [100%]
54 passed in 0.15s

$ pytest tests/unit/sandbox -q -p no:randomly
........................................................................ [ 29%]
........................................................................ [ 59%]
........................................................................ [ 88%]
............................                                             [100%]
244 passed in 0.42s
```

```bash
# Full unit suite. Six modules are excluded because they cannot collect in this
# environment for pre-existing reasons: a missing optional dependency
# (qoder_agent_sdk) and an installed agentscope older than those tests expect
# (ImportError: cannot import name 'FinishedReason').
$ pytest tests/unit -q -p no:randomly --ignore=tests/unit/harnesses \
    --ignore=tests/unit/agents/context/test_agent_resume.py \
    --ignore=tests/unit/agents/test_context_overflow_recovery.py \
    --ignore=tests/unit/agents/test_react_agent_compression.py \
    --ignore=tests/unit/agents/test_react_agent_scroll_seen.py \
    --ignore=tests/unit/agents/test_tool_result_pruning_middleware.py
FAILED tests/unit/agents/context/test_scroll_manager.py::test_tool_result_persisted_under_tool_call_id
FAILED tests/unit/agents/context/test_scroll_manager.py::test_compress_replaces_old_preview_with_tool_call_pointer
FAILED tests/unit/agents/context/test_scroll_manager.py::test_consumed_recall_page_folds_to_next_cursor
3 failed, 5973 passed, 15 skipped, 57 warnings in 237.50s (0:03:57)
```

```bash
# Baseline proof those 3 failures are pre-existing: identical result with this
# PR's changes stashed away.
$ git stash push -- src/qwenpaw/sandbox tests/unit/sandbox
$ pytest tests/unit/agents/context/test_scroll_manager.py -q -p no:randomly
FAILED tests/unit/agents/context/test_scroll_manager.py::test_tool_result_persisted_under_tool_call_id
FAILED tests/unit/agents/context/test_scroll_manager.py::test_compress_replaces_old_preview_with_tool_call_pointer
FAILED tests/unit/agents/context/test_scroll_manager.py::test_consumed_recall_page_folds_to_next_cursor
3 failed, 67 passed, 1 warning in 2.75s
$ git stash pop
```

Mutation check on the new consistency tests — reverting the Landlock declaration to the buggy version:

```bash
$ pytest tests/unit/sandbox/test_unenforced_config.py -q -p no:randomly
FAILED tests/unit/sandbox/test_unenforced_config.py::TestLandlockClaimMatchesScript::test_claim_matches_installed_port_rule[allow1]
FAILED tests/unit/sandbox/test_unenforced_config.py::TestLandlockClaimMatchesScript::test_claim_matches_installed_port_rule[allow2]
FAILED tests/unit/sandbox/test_unenforced_config.py::TestLandlockClaimMatchesScript::test_open_network_drops_the_port_rule
3 failed, 51 passed in 0.23s
```

The declared-vs-installed mismatch, measured directly before the fix:

```
network_allow=["*"], network_ports=[PortRule(port=443)], ABI 4
_enforced_fields() → ['deny_paths', 'mounts', 'network_allow', 'network_ports']
generated script    → "# Create ruleset (handled_fs=0x7fff, handled_net=0x0)"
port 443 in script  → False
```

Reported behaviour for the scenario TC-SB-05 flagged — `NoneSandbox` handed a fully restrictive config:

```
WARNING NoneSandbox does not enforce mounts=1 path(s); the constraint is IGNORED.
WARNING NoneSandbox does not enforce deny_paths=~/.ssh, ~/.aws; the constraint is IGNORED.
WARNING NoneSandbox does not enforce network_allow=github.com; the constraint is IGNORED. Domain-level filtering is unavailable on every backend; all network access is ALLOWED.
WARNING NoneSandbox does not enforce network_ports=1 rule(s); the constraint is IGNORED.
WARNING NoneSandbox does not enforce max_processes=8; the constraint is IGNORED.
WARNING NoneSandbox does not enforce max_memory_mb=100; the constraint is IGNORED.
```

The same backend given the config `ResourceGovernor` actually compiles (`network_allow=["*"]`) logs nothing.

## Additional Notes

- **Behaviour change worth flagging**: `NoneSandbox` now honours `shell_executable` instead of always resolving `$SHELL`, and warns when a configured shell is missing rather than falling back silently. Nothing in-tree sets the field for this backend — `ResourceGovernor` leaves it `None`.
- **Judgment call open to reversal**: `platform_hints` was moved into the security-boundary set alongside `env_mode`. The reasoning is that on macOS it carries verbatim admin-authored Seatbelt clauses, so a dropped hint weakens the sandbox. The review only asked for per-key detection; reverting just the severity change is a one-line edit if reviewers prefer `DEBUG`.
- **Markdown formatting not fully verified**: the repo's prettier hook matches `files: \.(tsx?)$` and so skips `.md`. The two doc files were checked separately with prettier 3.0.0 (`All matched files use Prettier code style!`). `website/package.json` pins prettier 3.9.6 for its own `format:check`; only 3.0.0 was available locally, so that exact version is unverified.
- **`pytest-asyncio` was missing locally** and is a declared dev dependency (`pyproject.toml`: `pytest-asyncio>=0.23.0`). Without it 1669 async tests error out during collection.
- **Not executed**: real network egress from a Windows unelevated sandbox. The proxy-only behaviour is established from the source (`execute` sets `HTTP_PROXY`/`HTTPS_PROXY` and nothing else) and from the existing Windows tests, which assert the variables are set but never that a socket is blocked. A raw-socket integration test on Windows would be the way to confirm it.
- **Follow-up candidates**: implement `max_processes` / `max_memory_mb` via cgroups v2 (Linux) and Job objects (Windows) — the Windows side already declares `_JOBOBJECT_EXTENDED_LIMIT_INFORMATION`; decide whether `network_ports` should apply independently of `network_allow`; and give the Windows unelevated backend real network blocking so it can honestly claim `network_allow`.


cuiyuebing (approved):
LGTM

cuiyuebing (approved):
LGTM

PR #6669, opened 2026-08-04
fix(desktop): stabilize Chrome native messaging and Windows restore locking

xiaoming-qxm:
## Description

This PR fixes two Windows-facing startup failures that were caused by process boundaries and file-lock semantics rather than by the Chrome extension UI itself.

For the QwenPaw Chrome extension, the Native Messaging host runs with the standalone Python runtime staged beside the desktop application, outside the PyInstaller backend. That runtime did not include `websockets`, so the host could exit before completing its first bridge connection and Chrome surfaced only generic messages such as `Error when communicating with the native messaging host` or `Native host has exited`.

This change:

- installs the pinned `websockets==15.0.1` dependency into the staged Native Messaging runtime in both Windows and POSIX desktop build scripts, then verifies the installed version during packaging;
- validates the host runtime before normal startup and before `--probe`, with a concrete diagnostic when a packaged dependency is missing or unusable;
- reads `runtime.lastError` synchronously inside the Chrome disconnect callback, before the value leaves Chrome's callback scope;
- returns actionable HTTP 422 responses for unsupported extension install modes before uninstall/reset side effects can occur;
- bumps the bundled extension version to `0.1.1` so the corrected worker can be distinguished from the preview build;
- fixes Windows restore-lock contention by removing the pre-lock read of byte 0. A contending process now retries `msvcrt.locking()` through the existing timeout path instead of failing early with `PermissionError`.

User impact: packaged desktop builds include the dependency required by the Native Messaging host, startup failures produce useful diagnostics, and parallel Windows startup/test processes wait for the restore lock as intended.

**Related Issue:** N/A

**Security Considerations:** Native Messaging origin restrictions, bridge-token handling, and loopback-only endpoint checks are unchanged. The added runtime dependency is pinned to an exact version. Unsupported install modes are rejected before filesystem or registry mutation.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [x] Scripts / Deploy

## Checklist

- [x] I ran the official pre-commit hooks for the PR diff locally and they pass
- [x] Pre-commit produced no auto-fixes that needed an additional commit
- [x] I ran relevant local regression checks and they pass
- [x] Documentation updated (not needed; no user-facing configuration or workflow changed)
- [ ] Ready for review

## Testing

Local verification covered the behavior boundaries changed by this PR:

- invalid CWS/unknown install modes return HTTP 422 without reset side effects;
- the Native Messaging host rejects a missing `websockets.connect` API and reports dependency startup failures;
- both desktop packaging scripts consume the pinned Native Messaging requirements file;
- the service worker reads Chrome's disconnect error before its first `await`;
- Windows lock contention retries without reading the locked byte range;
- browser execution lifecycle still passes with two xdist workers.

The real Windows byte-range locking scenario is platform-specific and was skipped on macOS; the deterministic contention-path check passed locally, and the Windows matrix will exercise the native implementation.

## Evidence

```text
.venv/bin/pre-commit run --from-ref 478de46c... --to-ref HEAD
# check-ast, mypy, black, flake8, pylint, and all applicable hooks: Passed

Targeted regression scenarios
# 8 passed, 1 skipped (Windows-only native byte-range lock on macOS)

pytest tests/integration/browser/test_execution_lifecycle_e2e.py -n 2 --dist=loadscope
# 4 passed

console: npm run format:check
# All matched files use Prettier code style

website: npm run format:check
# All matched files use Prettier code style

console: npm run build
# Build succeeded; Monaco stylesheet verification OK
```

## Additional Notes

- Chrome Web Store installation remains intentionally unsupported in this Developer Preview; the API now explains that state without partially modifying an existing unpacked installation.
- The PR is opened as a draft so the upstream Windows packaging and integration jobs can provide the final platform-specific evidence.


src/qwenpaw/backup/_utils/safe_swap.py
```diff
@@ -99,13 +99,9 @@ def _acquire_file_lock(handle: BinaryIO, lock_path: Path) -> None:
...
         import msvcrt
 
-        handle.seek(0)
-        if handle.read(_LOCK_REGION_SIZE) == b"":
-            handle.write(b"\0")
-            handle.flush()
-        handle.seek(0)
         while time.monotonic() < deadline:
             try:
+                handle.seek(0)
```
zhijianma, line 110: Main分支已经修改，可以去除

rayrayraykk (approved):
LGTM

Issue #6541, opened 2026-07-29
[Bug]: scroll context compression repeatedly triggers MODEL_EXECUTION_ERROR on DeepSeek — [context compressed] block uses role=user instead of role=system


sakanamaru:
# EN
## QwenPaw Version

v2.0.1

## Description

When using **scroll strategy + DeepSeek model**, after the conversation context grows past the compression threshold, the `[context compressed]` block is injected as a `role="user"` message, causing the DeepSeek (OpenAI-compatible) API to throw:

```
MODEL_EXECUTION_ERROR: Messages with role 'tool' must be a response
to a preceding message with 'tool_calls'
```

**Manually fixing it (changing `role="user"` → `role="system"` + restarting the service) restores functionality, but it inevitably recurs the next time context compression triggers.** This has happened repeatedly in the same session (most recently: 2026-07-29).

### Root Cause

Source location: `qwenpaw/agents/context/scroll/manager.py` line 898

```python
def _rebuild_context(self, agent, tail):
    memory = self._index.render()
    placeholder = UserMsg(name="memory", content=memory)  # ← role="user"
    agent.state.context = [placeholder] + tail
```

### Design Intent

The comment in `qwenpaw/agents/context/scroll/eviction_index.py` lines 62-72 explains this was done for Anthropic compatibility:

> Anthropic requires the first body message to be `user` and can't take a `system` message mid-context.

However, this causes problems for **DeepSeek / OpenAI**: the `[context compressed]` block is `role="user"`, immediately followed by the real user request which is also `role="user"`, resulting in two consecutive `user` messages. The subsequent assistant `tool_calls` and `role="tool"` responses fail OpenAI format validation in this sequence.

### Why It Keeps Recurring

Each time compression runs, `_rebuild_context` creates a **brand new** `UserMsg`, overwriting the manually fixed `role="system"`. The longer the conversation → the larger the context → the more frequent the compression → the more frequent the crashes.

### Trigger Conditions

- Context strategy: `scroll`
- Model provider: DeepSeek (OpenAI-compatible API)
- Compression threshold: `trigger_ratio=0.8` (compression triggers at 80% of context size)
- Reproduction: have a long multi-turn conversation with the agent (including tool calls) until the context naturally reaches the compression threshold

### Related Issue

- #6141 reports the same error message, but triggered in `/mission` mode.

## Component(s) Affected

- [x] Core / Backend (agents, context, scroll manager)

## Environment

[...]

PR #6675, opened 2026-08-04
fix: force relay reasoning_content for DeepSeek models (fixes #6667, #6541)

### LINKED ISSUE #6667: DeepSeek thinking mode fails in multi-turn: reasoning_content missing after OpenAI formatter skips ThinkingBlock
### Environment

- QwenPaw version: v1.1.7
- Model: DeepSeek V4 Pro (via DashScope compatible API)
- OS: Ubuntu 22.04

### Workaround

The `retry_chat_model.py` has a fallback mechanism that injects `reasoning_content = " "` on the first failure and retries, but this only works for the first occurrence. The core issue is that the formatter should never strip thinking content for DeepSeek models.

lt91888:
## Problem

DeepSeek thinking-mode APIs require `reasoning_content` on **every** assistant wire message in multi-turn conversations. When QwenPaw's scroll context compaction strips historical `ThinkingBlock`s, the `has_reasoning`-gated relay skips injection, and the upstream API rejects the request with:

```
MODEL_EXECUTION_ERROR: The reasoning_content in the thinking mode must be passed back to the API.
```

This breaks long conversations with DeepSeek models (reported in #6667, #6541), and the error recurs every time compaction triggers.

## Fix

Add a `force_relay_reasoning` flag on `CappingFormatterMixin`:

- **`OpenAIProvider`** enables it for DeepSeek-family model ids (`_is_deepseek_family`).
- **`model_factory.py`** relays `reasoning_content` unconditionally when the flag is set — even when `has_reasoning` is False because compaction already stripped the thinking blocks. On assistant-message count mismatch it now forces empty `reasoning_content` (accepted by DeepSeek) instead of skipping injection entirely.

Non-DeepSeek models are unaffected (flag defaults to `False`).

## Verification

Ran the exact failing scenario (compaction strips ThinkingBlocks → assistant messages without thinking) through the patched formatter chain:

```
system | reasoning_content = '<missing>'
user   | reasoning_content = '<missing>'
assistant | reasoning_content = ''   <- injected empty string, no more 400
user   | reasoning_content = '<missing>'
```

Also confirmed on a live instance (v2.0.x, deepseek-v4-flash-free via OpenAI-compatible provider): after the fix, context compaction passes cleanly with zero `reasoning_content` errors, whereas it previously failed within minutes.

Fixes #6667, #6541

### DIFF
--- a/src/qwenpaw/agents/model_factory.py
+++ b/src/qwenpaw/agents/model_factory.py
@@ -1142,8 +1142,18 @@ def _create_file_block_support_formatter(
                         if ec:
                             tc["extra_content"] = ec
 
+            # force_relay_reasoning: DeepSeek thinking-mode APIs require
+            # ``reasoning_content`` on EVERY assistant wire message, even
+            # after context compaction strips historical ThinkingBlocks
+            # (upstream error: "The `reasoning_content` in the thinking
+            # mode must be passed back to the API.").  When set, relay
+            # unconditionally - even if has_reasoning is False because
+            # compaction already stripped the thinking blocks.
+            force_relay = bool(
+                getattr(self, "force_relay_reasoning", False),
+            )
             if (
-                has_reasoning
+                (has_reasoning or force_relay)
                 and not is_anthropic_formatter
                 and not _is_response_formatter
                 and getattr(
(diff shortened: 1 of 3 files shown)


qbc2016:
Thanks for the PR! I investigated this further and opened #6721 with a formatter-aware retry fix. It handles AgentScope `Msg` objects and avoids model-specific detection.
Since #6721 supersedes this approach, I’m closing this PR. Thanks again for the contribution!

lt91888:
Thanks for the thorough follow-up, @qbc2016! The formatter-aware retry in #6721 is a much cleaner approach than model-specific detection — happy to see it landed and closed out #6667. We'd already validated the missing reasoning_content issue locally after context compression, so this fix matches what we observed in practice. Will keep an eye on the next release to confirm it resolves the scenario on our side too. Appreciate the work!

PR #6676, opened 2026-08-04
fix(onebot): bind loopback by default and require token when exposed

hongxicheng:
## Description

`ws_host` defaulted to `0.0.0.0` and authentication was skipped entirely while
`access_token` was empty, so any host on the network could connect to the reverse
WebSocket server and inject events that drive the agent.

- `ws_host` now defaults to `127.0.0.1`; a blank value normalises to it instead
  of binding every interface.
- A non-loopback `ws_host` requires `access_token`. The server keeps listening
  but rejects every handshake with `401` and logs how to fix it, so the
  misconfiguration surfaces instead of staying silent.
- Token checking is consolidated on the `Authorization` header, per the OneBot
  v11 reverse WebSocket spec: scheme matching is case-insensitive (`Bearer` /
  `Token`), comparison uses `hmac.compare_digest`, and `?access_token=` is no
  longer accepted because query strings are recorded in proxy and container
  access logs. A rejection caused by a query token logs a migration hint.
- Console blocks saving without a token on a non-loopback host, and
  `qwenpaw doctor` reports the same case — a TCP probe alone cannot detect it,
  since the port stays reachable while every handshake is refused.

Existing configuration is not migrated. Users who intentionally expose the port
set a token, guided by the log line, doctor output, and Console validation.

**Related Issue:** Relates to #(issue_number)

**Security Considerations:** Changes channel authentication defaults. Loopback is
the default listen address, a network-reachable listener now demands a token, and
the token travels in a header instead of a query string.

**Breaking change:** an existing `ws_host: 0.0.0.0` with an empty `access_token`
stops accepting connections until a token is set or the host returns to
`127.0.0.1`. Clients that authenticated via `?access_token=` move to the
`Authorization` header (the Token field of NapCat / go-cqhttp / Lagrange).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

## Testing

[How to test these changes]

## Evidence

Examples of valid evidence:
- CI artifact link

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]


cuiyuebing (approved):
LGTM

PR #6686, opened 2026-08-04
test(integration): fix chrome contract mismatches and add missing p-tier markers

yutai78786:
## Problem 1 — the PR gate has a coverage hole (root cause)

`tests/conftest.py` auto-marks everything under `tests/integration/` as
`integration`, but **never assigns a p-tier**. The gates are:

| Trigger | Marker expression |
|---|---|
| pull_request | `-m "integration and p0"` |
| push / pre-merge | `-m "integration and (p0 or p1)"` |
| nightly | `-m integration` |

**42 cases across 11 files carry no p-tier marker**, so both the PR gate and
the pre-merge gate silently deselect all of them. Only the nightly full sweep
ever executes them.

Reproduced:

```console
$ pytest tests/integration/test_chrome_nm_host.py -m "integration and p0" --collect-only
no tests collected (11 deselected)
```

**This is how the contract mismatches below shipped through a fully green PR
gate — those cases were never collected.**

## Problem 2 — four groups assert against APIs/behavior that never matched

All are test-side defects; the product code is correct in each case.

### `test_chrome_nm_host.py`
- `nm_host.py` exposes `read_nm_frame` / `write_nm_frame` (raw length-prefixed
  bytes) at L128/L144. The tests called `read_nm_message` / `write_nm_message`,
  which have never existed in the product's history →
  `AttributeError: ... Did you mean: 'read_nm_frame'?`
- Five fake WebSocket stand-ins declared `async def close(self)`, but the host
  closes with `close(code=..., reason=...)` at L372, matching the real library
  (`websockets` 16.0: `close(self, code=1000, reason='')`) →
  `TypeError: unexpected keyword argument 'code'`. The exit-code assertion in
  `test_terminate_fires_while_stdin_read_is_still_blocked` was collateral from
  this and recovers once the stubs accept the kwargs.

### `test_coding_project.py::test_import_local_copies_source`
- #6487 added `_validate_import_source`, which returns **403** for any source
  outside the user's home directory. The test seeded its source under a pytest
  tmp dir, so it got 403 instead of 200. Now seeded under `Path.home()` and
  removed in `finally`.

### `test_chrome_service_worker.py`
- `test_stale_receipt_is_never_reexecuted` required `executorCalls == 0`, but
  `runReceiptCommand` (service_worker.js L789-796) deliberately **re-executes**
  a stale `RECEIVED` receipt: that state is proof the executor never crossed
  the durable RUNNING barrier, so it is safe to run. RUNNING is the
  non-retriable one.
- The same file already agreed with the product elsewhere:
  `test_observed_state_closed_set` maps `received_stale_epoch -> NOT_STARTED`
  and passes. If it is NOT_STARTED, re-running is correct.
- Renamed the test and the harness scenario to state the real contract.

## Changes

1. Use the real `read_nm_frame` / `write_nm_frame`; JSON encode/decode in the
   test. Fakes accept `code` / `reason`.
2. Seed the import-local source under `Path.home()`.
3. Assert stale-RECEIVED re-execution; rename test + harness scenario.
4. Add `p1` / `p2` markers to all 42 unmarked cases (P1 = secondary flows and
   key guards, P2 = boundary conditions). None sit on the core chat path, so no
   new P0. Two chrome modules also needed a `pytest` import for the decorators.

Gate coverage for these files under `p0 or p1`: **0 -> 30 of 48**.

## Test plan
- [x] chrome suite (5 files): 30 passed
- [x] driver + coding-project (7 files): 29 passed
- [x] `pylint` with the repo's CI disable list: 10.00/10
- [x] `pre-commit` on changed files: green
- [x] Fork CI: pre-commit, Unit x4, Contract x4, Integrated x4, Coverage — all pass
- [x] Fork nightly: these cases pass on all platforms; the only remaining
      failures are the 7 `browser/*` cases (see below)

## Out of scope
- `tests/integration/browser/*` (7 cases) fail with
  `BrowserType.launch: Executable doesn't exist ... Please run playwright install`
  — the nightly Integration job has no `playwright install` step. Infrastructure
  fix, tracked separately.
- `test_acp_runner.py::test_acp_close_after_start` fails intermittently on
  Windows only: `os.replace` raises `PermissionError [WinError 5]` on
  `jobs.json` in `write_text_atomic`, racing the config watcher. Product-side
  atomic-write issue on Windows.

### DIFF
--- a/tests/integration/js/service_worker_harness.mjs
+++ b/tests/integration/js/service_worker_harness.mjs
@@ -205,7 +205,7 @@ const scenarios = {
     return { observedState: response.targetCommandFact.observedState };
   },
 
-  async stale_epoch_no_reexec() {
+  async stale_received_reexecuted() {
     seedTargetReceipt({
       ...commandParams,
       state: "RECEIVED",
--- a/tests/integration/test_chrome_extension_setup.py
+++ b/tests/integration/test_chrome_extension_setup.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 from pathlib import Path
 
+import pytest
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
 
@@ -31,6 +32,8 @@ SERVICE_WORKER = Path(
 # test_chrome_routes_asgi.py
 
 
+@pytest.mark.integration
+@pytest.mark.p1
 def test_install_status_reports_plugin_owned_installation_state() -> None:
     app = FastAPI()
     app.include_router(api_router)
@@ -49,6 +52,8 @@ TESTS_DIR = Path("tests/integration")
 # test_chrome_setup_hygiene.py
 
 
+@pytest.mark.integration
+@pytest.mark.p1
 def test_uninstall_removes_config_and_extension_dir(
     tmp_path: Path,
     isolated_home: Path,
--- a/tests/integration/test_chrome_native_host_install.py
+++ b/tests/integration/test_chrome_native_host_install.py
@@ -24,6 +24,8 @@ SETUP_SOURCE = Path("plugins/bundle/chrome/extension_setup.py")
 # pylint: disable=protected-access,unused-argument
 
 
+@pytest.mark.integration
+@pytest.mark.p1
 def test_probe_round_trips_through_the_real_launcher(
     isolated_home: Path,
     monkeypatch,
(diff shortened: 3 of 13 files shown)


cuiyuebing (approved):
LGTM

PR #6688, opened 2026-08-04
fix(plugins): isolate bare absolute imports per plugin namespace (fixes #6683)

### LINKED ISSUE #6683: [Bug]: App Center 安装 qwenpaw-creator 失败：No module named 'utils.env'; 'utils' is not a package（插件顶层模块命名冲突）
## QwenPaw Version

2.0.1

`qwenpaw --version` output:

```
QwenPaw, version 2.0.1
```

## Description

通过 App Center 安装官方应用 **QwenPaw Creator**（`qwenpaw-creator@1.0.1`）时，插件被下载并复制到 `~/.qwenpaw/plugins/qwenpaw-creator`，依赖（`oss2`、`imageio-ffmpeg`、`playwright`）也安装成功，但加载阶段始终失败，报：

```
Plugin install failed: No module named 'utils.env'; 'utils' is not a package
```

**根因（已定位并复现）**：插件加载器（`qwenpaw/plugins/loader.py` `_load_backend_module`）通过 `importlib.util.spec_from_file_location(module_name, entry, submodule_search_locations=[plugin_dir])` 加载插件入口，并设置 `module.__path__ = [plugin_dir]`。**插件之间不隔离顶层模块命名空间**。

1. 已安装的第三方插件 **`html-2-md`**（HTML to Markdown）根目录有 `utils.py`（顶层模块文件），其入口 `html_2_md_plugin.py` 先 `sys.path.insert(0, plugin_dir)`，再 `import html_2_md_tools` → `import utils`，把顶层 `utils` 以**普通模块**（非包）写入 `sys.modules`：

   ```python
   sys.modules['utils'] = <module 'utils' from '...\plugins\html-2-md\utils.py'>
   is package: False
   ```

2. `qwenpaw-creator` 的 `backend/main.py:30` 执行 `from utils.env import load_project_env`，期望 `utils` 是自己的**包**（`backend/utils/__init__.py` + `env.py`）。但 `sys.modules['utils']` 已被 `html-2-md` 的 `utils.py` 占据，Python 直接命中缓存、不做新解析，随即报 `'utils' is not a package`。

加载顺序决定胜负：`html-2-md` 先加载（按插件目录发现顺序），`qwenpaw-creator` 后加载 → 失败。顺序反过来时，`qwenpaw-creator` 的 `utils` 包会占位，`html-2-md` 的 `import utils` 反而失败。**两个插件无法共存**，且与重启无关（插件为热加载，重启后加载顺序不变、错误复现）。

**Related PR(s):** none

**Security considerations:** None — 纯插件导入命名冲突，无鉴权、Token、配置泄露。

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.0.1
- **OS:** Windows 10 (AMD64)
- **Install method:** one-line install（venv，`C:\Users\sam\.qwenpaw\venv`）
- **Python version (if applicable):** 3.12.12

## Steps to Reproduce

1. QwenPaw 2.0.1（venv 安装，Windows 10，Python 3.12）。
2. 已通过 App Center 安装第三方插件 `html-2-md`（`~/.qwenpaw/plugins/html-2-md/`，显示名 "HTML to Markdown"）。
3. 打开 App Center，安装官方应用 **QwenPaw Creator**。
4. 下载 / 复制 / 依赖安装完成后，安装对话框报错：`Plugin install failed: No module named 'utils.env'; 'utils' is not a package`。
5. 检查 `~/.qwenpaw/plugins/qwenpaw-creator/`：文件已复制，但插件未注册；重启 daemon 后每次加载都重复报错。

最小复现脚本（逐字复现报错，需先安装上述两个插件）：

```python
import importlib.util, sys

def load(plugin_dir, entry, mod):
    spec = importlib.util.spec_from_file_location(mod, entry,
        submodule_search_locations=[plugin_dir])
    m = importlib.util.module_from_spec(spec)
    sys.modules[mod] = m
    m.__package__ = mod
    m.__path__ = [plugin_dir]
    spec.loader.exec_module(m)

# 1) 先加载 html-2-md：顶层 utils 以普通模块写入 sys.modules
load(r'~/.qwenpaw/plugins/html-2-md',
     r'~/.qwenpaw/plugins/html-2-md/html_2_md_plugin.py',
     'plugin_html_2_md')
print(sys.modules['utils'], hasattr(sys.modules['utils'], '__path__'))  # 非包

# 2) 再加载 qwenpaw-creator：
#    ModuleNotFoundError: No module named 'utils.env'; 'utils' is not a package
load(r'~/.qwenpaw/plugins/qwenpaw-creator',
     r'~/.qwenpaw/plugins/qwenpaw-creator/backend/main.py',
     'plugin_qwenpaw_creator')
```

单独加载 `qwenpaw-creator`（不先加载 `html-2-md`）则一切正常——证明插件包本身结构无误。

## Actual vs Expected

- **Actual:** 只要 `html-2-md` 已加载，`qwenpaw-creator` 就无法安装 / 加载；错误与插件自身结构无关（单独加载正常）。
- **Expected:** 插件之间应隔离顶层模块命名空间（例如按插件 ID 加前缀 / 独立包名），或在加载时检测并避免顶层名冲突；至少不应出现"一个第三方插件的顶层模块名导致官方应用无法加载"的静默冲突。

## Logs / Screenshots

`~/.qwenpaw/qwenpaw.log` 关键片段：

```
2026-08-04 17:24:15 | INFO | ...\plugins\loader.py:1166 | Copied plugin 'qwenpaw-creator' to C:\Users\sam\.qwenpaw\plugins\qwenpaw-creator
2026-08-04 17:24:15 | INFO | ...\plugins\loader.py:880  | Installing dependencies for plugin 'qwenpaw-creator'...
2026-08-04 17:25:38 | INFO | ...\plugins\loader.py:915  | Dependencies installed for plugin 'qwenpaw-creator' (via pip)
2026-08-04 17:25:38 | WARNING | ...\plugins\loader.py:600 | Cleaning up failed plugin load for 'qwenpaw-creator'
2026-08-04 17:25:42 | ERROR | ...\plugins\loader.py:730 | Failed to load plugin 'qwenpaw-creator': No module named 'utils.env'; 'utils' is not a package
  File "...\qwenpaw\plugins\loader.py", line 722, in _load_plugin_unlocked
    plugin_def = await self._load_backend_module(...)
  File "...\qwenpaw\plugins\loader.py", line 529, in _load_backend_module
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "C:\Users\sam\.qwenpaw\plugins\qwenpaw-creator\backend\main.py", line 30, in <module>
    from utils.env import load_project_env  # noqa: E402
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'utils.env'; 'utils' is not a package
2026-08-04 17:26:37 | INFO | ...\plugins\loader.py:1166 | Copied plugin 'qwenpaw-creator' to C:\Users\sam\.qwenpaw\plugins\qwenpaw-creator
2026-08-04 17:26:37 | INFO | ...\plugins\loader.py:445  | Plugin 'qwenpaw-creator' dependencies already satisfied by a concurrent installer; skipping pip install
2026-08-04 17:26:37 | WARNING | ...\plugins\loader.py:600 | Cleaning up failed plugin load for 'qwenpaw-creator'
2026-08-04 17:26:41 | ERROR | ...\plugins\loader.py:730 | Failed to load plugin 'qwenpaw-creator': No module named 'utils.env'; 'utils' is not a package
```

## Additional Notes

**Workaround（用户侧）**：卸载 / 禁用 `html-2-md` 插件（App Center 卸载，或把 `~/.qwenpaw/plugins/html-2-md` 重命名为 `html-2-md.disabled`）后，`qwenpaw-creator` 可正常安装；但两者无法同时启用。

**建议修复（任选其一或组合）**：

- **A（推荐，框架层）**：插件加载器为每个插件创建隔离的导入命名空间（`plugin_<id>` 前缀），或加载入口时使用独立 `sys.path` / 模块解析上下文，避免 `utils` 这类通用顶层名全局抢占。
- **B（插件层）**：`html-2-md` 将内部模块（`utils.py` 等）改为插件专属前缀（如 `html_2_md_utils`），或改为相对导入。
- **C（App Center 层）**：安装时检测插件入口的顶层导入名是否与已加载插件冲突，给出友好提示而非硬错误。

**Similar issue**：Agent Kanban 在 Desktop 2.0.1 装不上（`No module named 'qwenpaw.pawapp'`）——同属 App Center 插件打包 / 运行时兼容问题：#6473。

An-idd:
## Description

Fixes #6683 — installing `qwenpaw-creator` from App Center fails with `No module named 'utils.env'; 'utils' is not a package`.

**Root cause**: `_load_backend_module` gives each plugin's entry file a private namespace (`plugin_<id>`), but that isolation only covers relative imports. Many plugins insert their own directory into `sys.path` and use **bare absolute imports** for their own files (`import utils` / `from utils.env import x`). Those resolve through the standard machinery into the **process-wide top-level namespace**, so all plugins share it — the first plugin to claim a common name (`utils`, `models`, `router`, …) wins, and every plugin loaded afterwards silently receives that module instead of its own. Concrete repro: `html-2-md` ships `utils.py` (a plain module) and loads first; `qwenpaw-creator` ships `utils/` (a package) — creator's `from utils.env import ...` hits the cached plain module → `'utils' is not a package`.

The loader already sweeps bare modules by `__file__` path, but only on failed load / unload — a successful load leaves the pollution in place permanently. Extending that sweep to the success path is not safe either: plugins with lazy function-level imports (creator has ~20) would re-import fresh copies later and duplicate module-level singletons.

**Fix** (no plugin code changes needed):

- `module_isolation.py` (new): `build_plugin_builtins` — a per-plugin `__builtins__` whose `__import__` resolves bare top-level names against the plugin's own directories first, importing them as `plugin_<id>.<name>`; `PluginNamespaceFinder` — a `sys.meta_path` finder (inert for non-plugin imports) that makes every module imported under `plugin_<id>.` execute with that same `__builtins__`, so nested and lazy (function-level) bare imports stay namespaced too. Names the plugin does not ship fall through to the regular machinery — stdlib/third-party imports are untouched.
- `loader.py`: wires the redirection into `_load_backend_module`, searches the entry file's directory for nested entries (creator's `backend/main.py`), unregisters the namespace on failed load / unload. Existing `__file__`-based sweeps stay as a second line of defense.
- `validation.py`: same wiring, so CLI `plugin install/validate` checks plugins under the exact conditions they run in.

Plugins using relative imports are unaffected. Plugins using bare imports (`qwenpaw-creator`, `qwenpaw-pet`, `cloudpaw`, `computer-use`) now resolve them deterministically to their own files regardless of load order.

**Related Issue:** Fixes #6683

**Security Considerations:** None — the redirection only touches names physically present in the plugin's own directory; everything else falls through to the standard import machinery.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [x] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Documentation updated (if needed) — not needed, behavior-transparent fix
- [x] Ready for review

## Testing

1. Reproduce: install two plugins that both ship a top-level `utils` — one as `utils.py` (e.g. `html-2-md`), one as a `utils/` package (e.g. `qwenpaw-creator`). Before this fix the second install fails with `'utils' is not a package`; after it, both load and each sees its own `utils`.
2. Or run the new regression suite, which covers: the #6683 repro (module vs package collision), nested bare imports inside plugin submodules, lazy function-level imports after another plugin claims the same name, stdlib fall-through (`json` stays the real `sys.modules['json']`), creator's nested-entry layout (`backend/main.py`), and namespace cleanup on failed load:

```bash
pytest tests/unit/plugins/ -q
```

## Evidence

```bash
$ pre-commit run --all-files
sort simple yaml files...................................................Passed
  Success: no issues found in 1209 source files

$ pytest tests/unit/plugins/ -q
........................................................................ [ 42%]
.........sssssss........................................................ [ 84%]
..........................                                               [100%]
163 passed, 7 skipped in 2.07s
```

## Additional Notes

Hardening pass in 4e49c802 after an adversarial self-review (edge cases + full-repo interaction audit; all 8 in-repo plugins verified unaffected):

- A bare **data directory** (`locale/`, `config/`, …) no longer shadows a same-named stdlib/third-party package — namespace-package specs are rejected when probing plugin-local names.
- The wrapping loader now forwards the full loader protocol (`get_data`, `get_resource_reader`, …), so `importlib.resources` / `pkgutil` work on namespaced plugin packages.
- A bare import of the entry file itself (`import main`) aliases the running entry module instead of executing it a second time.
- Namespace unregistration moved after the `sys.modules` sweeps (closes a hot-reload race), and `sys.path` cleanup now also removes plugin **sub**directories (nested-entry plugins insert `backend/`).

Known limitations, documented in the module docstring: only the `import` statement is redirected — `importlib.import_module("utils")` bypasses `__import__` and stays global (the existing `__file__`-based sweeps still clean those up on unload, and the cleanup test now exercises exactly that path); directories a plugin adds to `sys.path` itself (vendored `lib/`) are not covered; the plugin `__builtins__` is a snapshot dict; `plugin_<id>.…` pickles require the plugin loaded in the unpickling process.

The new meta-path finder is registered lazily (first plugin load) and matches only registered `plugin_<id>.` prefixes, so it adds zero overhead to non-plugin imports. Module `__name__` for bare-imported plugin files changes from e.g. `utils.env` to `plugin_qwenpaw_creator.utils.env`; in-repo plugins do not depend on the bare form.

### DIFF
--- a/src/qwenpaw/plugins/loader.py
+++ b/src/qwenpaw/plugins/loader.py
@@ -23,6 +23,11 @@ from packaging.requirements import Requirement
 
 from .architecture import PluginManifest, PluginRecord
 from .api import PluginApi
+from .module_isolation import (
+    build_plugin_builtins,
+    get_namespace_finder,
+    unregister_namespace,
+)
 from .registry import PluginRegistry
 
 logger = logging.getLogger(__name__)
@@ -509,11 +514,18 @@ class PluginLoader:
         """
         module_name = f"plugin_{plugin_id.replace('-', '_')}"
         plugin_dir_str = str(source_path)
+        # Plugins with a nested entry (e.g. ``backend/main.py``) resolve
+        # their bare imports against the entry file's directory, so it
+        # must be searchable alongside the plugin root.
+        entry_dir_str = str(backend_entry_file.parent)
+        search_paths = [plugin_dir_str]
+        if _norm_realpath(entry_dir_str) != _norm_realpath(plugin_dir_str):
+            search_paths.append(entry_dir_str)
 
         spec = importlib.util.spec_from_file_location(
             module_name,
             backend_entry_file,
-            submodule_search_locations=[plugin_dir_str],
+            submodule_search_locations=search_paths,
         )
         if spec is None or spec.loader is None:
             raise ImportError(
@@ -522,10 +534,17 @@ class PluginLoader:
 
         module = importlib.util.module_from_spec(spec)
 
+        # Redirect the plugin's bare absolute imports (``import utils``)
+        # into its private ``plugin_<id>`` namespace so plugins cannot
+        # collide with each other's top-level module names (#6683).
+        plugin_builtins = build_plugin_builtins(module_name, search_paths)
+        module.__dict__["__builtins__"] = plugin_builtins
+        get_namespace_finder().register(module_name, plugin_builtins)
+
         try:
             sys.modules[module_name] = module
             module.__package__ = module_name
-            module.__path__ = [plugin_dir_str]
+            module.__path__ = search_paths
             spec.loader.exec_module(module)
 
             plugin_def = getattr(module, "plugin", None)
@@ -601,6 +620,9 @@ class PluginLoader:
             "Cleaning up failed plugin load for '%s'",
             plugin_id,
         )
+
+        # 0. Import redirection for the plugin's private namespace
+        unregister_namespace(module_name)
 
         # 1. Registry (manifest, providers, hooks, middleware, routes, …)
         self.registry.unregister_plugin(plugin_id)
@@ -1264,6 +1286,7 @@ class PluginLoader:
         # Remove Python module and all sub-modules so the next import
         # gets a fresh copy (e.g. plugin_foo.utils must not be reused).
         module_name = f"plugin_{plugin_id.replace('-', '_')}"
+        unregister_namespace(module_name)
         prefix = module_name + "."
         stale = [
             k for k in sys.modules if k == module_name or k.startswith(prefix)
--- a/src/qwenpaw/plugins/loader.py
+++ b/src/qwenpaw/plugins/loader.py
@@ -27,6 +27,7 @@ from .module_isolation import (
     build_plugin_builtins,
     get_namespace_finder,
     strip_plugin_sys_path,
+    sweep_bare_tree_modules,
     unregister_namespace,
 )
 from .registry import PluginRegistry
@@ -614,6 +615,11 @@ class PluginLoader:
             # Shared dependency locations (plugin site dir) are
             # untouched — only paths under the plugin's own tree go.
             strip_plugin_sys_path(source_path)
+            # sys.modules is the other residue channel: a bypass import
+            # or a data-directory fallthrough during load can cache a
+            # bare name rooted in this plugin's tree, which would keep
+            # serving later plugins even with sys.path clean.
+            sweep_bare_tree_modules(source_path)
 
         return plugin_def
 
@@ -652,27 +658,17 @@ class PluginLoader:
         for k in stale:
             sys.modules.pop(k, None)
 
-        # 3. sys.modules — by __file__ path (catches bare imports that
-        #    bypassed the plugin_<id> namespace, e.g. ``import utils``
-        #    after the plugin inserted its dir into sys.path).
-        source_resolved = _norm_realpath(source_path)
-        if not source_resolved.endswith(os.sep):
-            source_resolved = source_resolved + os.sep
-        stale_by_file = [
-            k
-            for k, mod in list(sys.modules.items())
-            if (mod_file := getattr(mod, "__file__", None)) is not None
-            and _norm_realpath(mod_file).startswith(source_resolved)
-        ]
-        for k in stale_by_file:
-            sys.modules.pop(k, None)
-
-        # 4. Import redirection — after the sys.modules sweeps, so a
+        # 3. Import redirection — after the sys.modules sweep, so a
         #    concurrent lazy import cannot resolve a plugin submodule
         #    without the plugin builtins in the window between the two.
+        #    Bare (non-namespaced) residue is swept by the caller's
+        #    finally, AFTER strip_plugin_sys_path — sweeping while the
+        #    plugin's sys.path entries are still present could merge
+        #    plugin-tree portions into a shared namespace package's
+        #    __path__ recalculation and evict a host package.
         unregister_namespace(module_name)
 
-        # 5. sys.path — remove the plugin directory and its subdirs
+        # 4. sys.path — remove the plugin directory and its subdirs
         strip_plugin_sys_path(source_path)
 
     async def load_plugin(
@@ -1312,35 +1308,25 @@ class PluginLoader:
         for k in stale:
             sys.modules.pop(k, None)
 
-        # Plugins that manipulate ``sys.path`` (e.g. inserting their own
-        # directory) and use bare ``from sibling import …`` load sibling
-        # modules as top-level entries in ``sys.modules`` — the prefix
-        # cleanup above misses them.  Sweep any module whose ``__file__``
-        # lives inside the plugin directory so a reinstall always gets
-        # fresh code.  Use normcase so Windows drive/dir letter case
-        # differences do not leave stale modules behind.
-        source_resolved = _norm_realpath(record.source_path)
-        if not source_resolved.endswith(os.sep):
-            source_resolved = source_resolved + os.sep
-        stale_by_file = [
-            k
-            for k, mod in list(sys.modules.items())
-            if (mod_file := getattr(mod, "__file__", None)) is not None
-            and _norm_realpath(mod_file).startswith(source_resolved)
-        ]
-        for k in stale_by_file:
-            sys.modules.pop(k, None)
+        # Remove the plugin directory and its subdirectories from
+        # sys.path BEFORE the location-based sweep below: a namespace
+        # package's __path__ recalculation reads the live sys.path, and
+        # sweeping while the plugin's entries are still present could
+        # merge plugin-tree portions into a shared host package's
+        # portions and evict it.
+        strip_plugin_sys_path(record.source_path)
+
+        # Bypass imports (e.g. importlib.import_module after the plugin
+        # inserted its dir into sys.path) land as top-level entries in
+        # ``sys.modules`` — the prefix cleanup above misses them.
+        # Sweep by module location (including __file__-less namespace
+        # packages) so a reinstall always gets fresh code.
+        sweep_bare_tree_modules(record.source_path)
 
         # Drop the import redirection after the sys.modules sweeps, so
         # a concurrent lazy import cannot resolve a plugin submodule
         # without the plugin builtins in the window between the two.
         unregister_namespace(module_name)
-
-        # Remove the plugin directory and its subdirectories from
-        # sys.path (plugins add these at import time for sibling
-        # imports; leaving them leaks into later imports and prevents
-        # clean hot-reload).
-        strip_plugin_sys_path(record.source_path)
 
         # Remove tools from agents.tools + runtime registries while
         # ownership records still exist, then drop plugin registry state.
--- a/src/qwenpaw/plugins/loader.py
+++ b/src/qwenpaw/plugins/loader.py
@@ -26,6 +26,7 @@ from .api import PluginApi
 from .module_isolation import (
     build_plugin_builtins,
     get_namespace_finder,
+    strip_plugin_sys_path,
     unregister_namespace,
 )
 from .registry import PluginRegistry
@@ -659,16 +660,8 @@ class PluginLoader:
         #    without the plugin builtins in the window between the two.
         unregister_namespace(module_name)
 
-        # 5. sys.path — remove the plugin directory and any of its
-        #    subdirectories (nested-entry plugins insert e.g. backend/)
-        plugin_dir_real = _norm_realpath(source_path)
-        plugin_dir_prefix = plugin_dir_real + os.sep
-        sys.path[:] = [
-            p
-            for p in sys.path
-            if _norm_realpath(p) != plugin_dir_real
-            and not _norm_realpath(p).startswith(plugin_dir_prefix)
-        ]
+        # 5. sys.path — remove the plugin directory and its subdirs
+        strip_plugin_sys_path(source_path)
 
     async def load_plugin(
         self,
@@ -1331,19 +1324,11 @@ class PluginLoader:
         # without the plugin builtins in the window between the two.
         unregister_namespace(module_name)
 
-        # Remove the plugin directory — and any of its subdirectories,
-        # e.g. the backend/ dir nested-entry plugins insert — from
+        # Remove the plugin directory and its subdirectories from
         # sys.path (plugins add these at import time for sibling
         # imports; leaving them leaks into later imports and prevents
         # clean hot-reload).
-        plugin_dir_real = _norm_realpath(record.source_path)
-        plugin_dir_prefix = plugin_dir_real + os.sep
-        sys.path[:] = [
-            p
-            for p in sys.path
-            if _norm_realpath(p) != plugin_dir_real
-            and not _norm_realpath(p).startswith(plugin_dir_prefix)
-        ]
+        strip_plugin_sys_path(record.source_path)
 
         # Remove tools from agents.tools + runtime registries while
         # ownership records still exist, then drop plugin registry state.
--- a/src/qwenpaw/plugins/loader.py
+++ b/src/qwenpaw/plugins/loader.py
@@ -516,11 +516,14 @@ class PluginLoader:
         plugin_dir_str = str(source_path)
         # Plugins with a nested entry (e.g. ``backend/main.py``) resolve
         # their bare imports against the entry file's directory, so it
-        # must be searchable alongside the plugin root.
+        # must be searchable alongside the plugin root.  The entry
+        # directory comes first: nested-entry plugins put it at
+        # ``sys.path[0]``, so it must win over a same-named module in
+        # the plugin root.
         entry_dir_str = str(backend_entry_file.parent)
-        search_paths = [plugin_dir_str]
+        search_paths = [entry_dir_str]
         if _norm_realpath(entry_dir_str) != _norm_realpath(plugin_dir_str):
-            search_paths.append(entry_dir_str)
+            search_paths.append(plugin_dir_str)
 
         spec = importlib.util.spec_from_file_location(
             module_name,
--- a/src/qwenpaw/plugins/loader.py
+++ b/src/qwenpaw/plugins/loader.py
@@ -540,7 +540,11 @@ class PluginLoader:
         # Redirect the plugin's bare absolute imports (``import utils``)
         # into its private ``plugin_<id>`` namespace so plugins cannot
         # collide with each other's top-level module names (#6683).
-        plugin_builtins = build_plugin_builtins(module_name, search_paths)
+        plugin_builtins = build_plugin_builtins(
+            module_name,
+            search_paths,
+            entry_file=backend_entry_file,
+        )
         module.__dict__["__builtins__"] = plugin_builtins
         get_namespace_finder().register(module_name, plugin_builtins)
 
@@ -624,9 +628,6 @@ class PluginLoader:
             plugin_id,
         )
 
-        # 0. Import redirection for the plugin's private namespace
-        unregister_namespace(module_name)
-
         # 1. Registry (manifest, providers, hooks, middleware, routes, …)
         self.registry.unregister_plugin(plugin_id)
 
@@ -653,10 +654,20 @@ class PluginLoader:
         for k in stale_by_file:
             sys.modules.pop(k, None)
 
-        # 4. sys.path — remove the plugin directory if it was added
+        # 4. Import redirection — after the sys.modules sweeps, so a
+        #    concurrent lazy import cannot resolve a plugin submodule
+        #    without the plugin builtins in the window between the two.
+        unregister_namespace(module_name)
+
+        # 5. sys.path — remove the plugin directory and any of its
+        #    subdirectories (nested-entry plugins insert e.g. backend/)
         plugin_dir_real = _norm_realpath(source_path)
+        plugin_dir_prefix = plugin_dir_real + os.sep
         sys.path[:] = [
-            p for p in sys.path if _norm_realpath(p) != plugin_dir_real
+            p
+            for p in sys.path
+            if _norm_realpath(p) != plugin_dir_real
+            and not _norm_realpath(p).startswith(plugin_dir_prefix)
         ]
 
     async def load_plugin(
@@ -1289,7 +1300,6 @@ class PluginLoader:
         # Remove Python module and all sub-modules so the next import
         # gets a fresh copy (e.g. plugin_foo.utils must not be reused).
         module_name = f"plugin_{plugin_id.replace('-', '_')}"
-        unregister_namespace(module_name)
         prefix = module_name + "."
         stale = [
             k for k in sys.modules if k == module_name or k.startswith(prefix)
@@ -1316,12 +1326,23 @@ class PluginLoader:
         for k in stale_by_file:
             sys.modules.pop(k, None)
 
-        # Remove the plugin directory from sys.path (plugins add it at
-        # import time for sibling imports; leaving it leaks into later
-        # imports and prevents clean hot-reload).
+        # Drop the import redirection after the sys.modules sweeps, so
+        # a concurrent lazy import cannot resolve a plugin submodule
+        # without the plugin builtins in the window between the two.
+        unregister_namespace(module_name)
+
+        # Remove the plugin directory — and any of its subdirectories,
+        # e.g. the backend/ dir nested-entry plugins insert — from
+        # sys.path (plugins add these at import time for sibling
+        # imports; leaving them leaks into later imports and prevents
+        # clean hot-reload).
         plugin_dir_real = _norm_realpath(record.source_path)
+        plugin_dir_prefix = plugin_dir_real + os.sep
         sys.path[:] = [
-            p for p in sys.path if _norm_realpath(p) != plugin_dir_real
+            p
+            for p in sys.path
+            if _norm_realpath(p) != plugin_dir_real
+            and not _norm_realpath(p).startswith(plugin_dir_prefix)
         ]
 
         # Remove tools from agents.tools + runtime registries while
--- a/src/qwenpaw/plugins/loader.py
+++ b/src/qwenpaw/plugins/loader.py
@@ -602,6 +602,18 @@ class PluginLoader:
                 source_path,
             )
             raise
+        finally:
+            # A loaded plugin no longer needs the sys.path entries it
+            # inserted: its bare imports resolve through the private
+            # namespace (search_paths), not sys.path.  Sweeping here —
+            # on success, failure, AND BaseException (a cancelled
+            # startup) — keeps other plugins' non-local fallthrough
+            # imports from ever resolving into this plugin's source
+            # tree (data-dir fallthrough, uncached stdlib names, or
+            # plain bare imports would otherwise pick up the residue).
+            # Shared dependency locations (plugin site dir) are
+            # untouched — only paths under the plugin's own tree go.
+            strip_plugin_sys_path(source_path)
 
         return plugin_def
 
--- a/src/qwenpaw/plugins/module_isolation.py
+++ b/src/qwenpaw/plugins/module_isolation.py
@@ -148,6 +148,22 @@ def _norm(path: Any) -> str:
     return os.path.normcase(os.path.realpath(str(path)))
 
 
+_CODE_SUFFIXES = tuple(importlib.machinery.all_suffixes())
+
+
+def _has_importable_code(portions: List[str]) -> bool:
+    """True if any directory portion contains importable code.
+
+    Walks recursively (short-circuiting on the first hit) so a PEP 420
+    package whose code lives only in nested subpackages still counts.
+    """
+    for portion in portions:
+        for _dirpath, _dirnames, filenames in os.walk(portion):
+            if any(f.endswith(_CODE_SUFFIXES) for f in filenames):
+                return True
+    return False
+
+
 def build_plugin_builtins(
     module_name: str,
     search_paths: List[str],
@@ -174,15 +190,17 @@ def build_plugin_builtins(
                 search_paths,
             )
             # A namespace-package portion (loader is None) needs a
-            # tie-break mirroring the real machinery: a plain data
-            # directory (``locale/`` assets) loses to a concrete module
-            # elsewhere on sys.path (the stdlib ``locale`` must win),
-            # but a genuine PEP 420 namespace package the plugin ships
-            # stays plugin-local — sending it to the global namespace
-            # would reopen the cross-plugin collision (#6683).
+            # tie-break: a genuine PEP 420 package the plugin ships
+            # stays plugin-local, but a plain data directory
+            # (``locale/`` assets) must fall through so a concrete
+            # stdlib/third-party module can win.  Decide purely from
+            # the plugin's own files — probing sys.path here would
+            # couple the decision to sys.path entries left behind by
+            # earlier plugins, reopening the cross-plugin collision
+            # this module exists to fix (#6683).
             if spec is not None and spec.loader is None:
-                resolved = importlib.machinery.PathFinder.find_spec(top)
-                if resolved is None or resolved.loader is not None:
+                portions = list(spec.submodule_search_locations or [])
+                if not _has_importable_code(portions):
                     spec = None
             spec_cache[top] = spec
         return spec_cache[top]
--- a/src/qwenpaw/plugins/module_isolation.py
+++ b/src/qwenpaw/plugins/module_isolation.py
@@ -38,6 +38,11 @@ Known limitations (documented, not silently broken):
   not visible inside plugin modules.
 - Objects pickled with a ``plugin_<id>.…`` ``__module__`` are only
   unpicklable in a process where that plugin is loaded.
+- The local/non-local decision per bare top-level name is cached for
+  the plugin's lifetime; code generated into the plugin directory at
+  runtime after a name was first probed resolves globally.  Extension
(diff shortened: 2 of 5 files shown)


RerankerGuo:
Independent review at exact head `93289ff498bf7886ab9bf73d5659a49df219d035`.

The new isolation works for the covered collision cases, but I found a nested-entry precedence regression. For `backend/main.py`, both runtime loading and CLI validation construct:

```python
search_paths = [plugin_dir_str, entry_dir_str]
```

The plugin entry may explicitly put its own directory first with `sys.path.insert(0, os.path.dirname(__file__))` (the pattern this PR is intended to support). The custom importer now ignores that ordering: if the plugin root and `backend/` both contain the same bare module name, `PathFinder` resolves the root copy first. Before the custom redirection, the explicit `sys.path[0]` made the entry-directory copy win.

I added this boundary to the existing namespace suite:

```python
plugin_dir = tmp_path / "nested-priority"
backend = plugin_dir / "backend"
backend.mkdir(parents=True)
(plugin_dir / "helper.py").write_text("VALUE = 'plugin-root'\n")
(backend / "helper.py").write_text("VALUE = 'entry-dir'\n")
_write_manifest(plugin_dir, backend_entry="backend/main.py")
(backend / "main.py").write_text(
    "import os, sys\n"
    "sys.path.insert(0, os.path.dirname(__file__))\n"
    "from helper import VALUE\n" + _REGISTER_OK
)

await _load(loader, plugin_dir, backend_entry="backend/main.py")
assert sys.modules["plugin_nested_priority"].VALUE == "entry-dir"
```

Results:

```text
PR head, existing suite:
6 passed

PR head + boundary:
FAILED test_nested_entry_dir_precedes_plugin_root
E assert 'plugin-root' == 'entry-dir'
1 failed, 6 passed
```

A minimal fix is to preserve entry-directory precedence while keeping the root as fallback in both `loader.py` and `validation.py`:

```python
search_paths = [entry_dir_str]
if norm(entry_dir_str) != norm(plugin_dir_str):
    search_paths.append(plugin_dir_str)
```

With that ordering:

```text
focused namespace suite: 7 passed
complete tests/unit/plugins/: 164 passed, 7 skipped
```

This does not affect plugins whose entry is at the root or whose module names are unique. It matters when nested-entry plugins intentionally shadow a root-level support module, and it keeps validation behavior aligned with runtime behavior.

An-idd:
@RerankerGuo Good catch, thank you — confirmed the regression: your boundary test fails at `93289ff4` exactly as reported (`'plugin-root' == 'entry-dir'`). Nested-entry plugins put their entry directory at `sys.path[0]`, so it must take precedence over a same-named module in the plugin root.

Fixed in e212585c with your suggested ordering, applied to both `loader.py` and `validation.py`, and your boundary case added to the suite as `test_nested_entry_dir_precedes_plugin_root`:

```text
tests/unit/plugins/test_plugin_namespace_isolation.py: 7 passed
tests/unit/plugins/: 164 passed, 7 skipped
pre-commit on touched files: all hooks passed
```

One note on a related edge: this also reorders `module.__path__` for nested-entry plugins, so *relative* imports now prefer the entry directory too (pre-PR they only searched the plugin root). That keeps bare and relative resolution consistent with each other; a plugin relying on a root-level module shadowed by a same-named entry-dir module *via relative import* would see the entry-dir copy now, which matches the plugin's own `sys.path` intent.

RerankerGuo:
Revalidated the fix independently at exact head `e212585cf3588ad327db1119f504840b2685ab33`.

I confirmed that both runtime loading (`loader.py`) and CLI validation (`validation.py`) now use entry-directory-first ordering with the plugin root as fallback. The regression case from my earlier review is included as `test_nested_entry_dir_precedes_plugin_root`.

Results:

```text
.venv/bin/pytest tests/unit/plugins/test_plugin_namespace_isolation.py -q
7 passed

.venv/bin/pytest tests/unit/plugins/ -q
164 passed, 7 skipped

  src/qwenpaw/plugins/loader.py \
  src/qwenpaw/plugins/validation.py \
  tests/unit/plugins/test_plugin_namespace_isolation.py
all applicable hooks passed (including mypy, black, flake8, and pylint)
```

The nested-entry precedence finding is resolved at this head. The fallback to the plugin root remains covered by the existing nested-entry tests, and runtime/validation behavior stays aligned.

An-idd:
Thanks for the revalidation! One heads-up: after your re-check at `e212585c` I pushed one more commit, 4e49c802 — a hardening pass from an adversarial self-review, summarized in the updated **Additional Notes** section of the PR body (data-directory shadowing, loader-protocol forwarding for `importlib.resources`, entry-file aliasing, unload ordering/`sys.path` sweep, plus documented known limitations). Full plugin suite at the new head: 168 passed, 7 skipped; pre-commit all hooks passing. Happy to walk through any of it if useful.

RerankerGuo:
Independent review of the hardening update at exact head `4e49c802b97f950fc7fe6cf1f44800f91181110c`.

The new data-directory guard is too broad:

```python
if spec is not None and spec.loader is None:
    spec = None
```

`loader is None` also identifies legitimate PEP 420 namespace packages. Those packages now bypass the private plugin namespace and fall back to the process-wide top-level cache, reopening the cross-plugin collision that this PR fixes.

I reproduced this with two plugins that each ship a package without `__init__.py`:

```text
pep-a/helpers/value.py  -> VALUE = "A"
pep-b/helpers/value.py  -> VALUE = "B"
```

Each entry inserts its own directory at `sys.path[0]` and executes:

```python
from helpers.value import VALUE
```

After loading A then B:

```text
assert plugin_pep_a.VALUE == "A"  # passes
assert plugin_pep_b.VALUE == "B"  # fails: received "A"
```

Focused result on this head:

```text
2 passed, 1 failed
E assert 'A' == 'B'
```

The global modules are `helpers` / `helpers.value`, so B reuses A's cached submodule instead of resolving `plugin_pep_b.helpers.value`.

A narrow way to retain the new `wave/` data-directory protection without rejecting real namespace packages is to compare against the normal resolver's final result:

```python
if spec is not None and spec.loader is None:
    resolved = importlib.machinery.PathFinder.find_spec(top)
    if resolved is None or resolved.loader is not None:
        spec = None
```

For the tested plugin pattern, its directory is already at `sys.path[0]`:

- a data-only `wave/` plus stdlib `wave.py` resolves to the concrete stdlib loader, so it remains non-local;
- a real `helpers/` namespace package resolves as a namespace package, so it stays plugin-local.

With that distinction and the PEP 420 regression added:

```text
namespace isolation suite: 12 passed
complete tests/unit/plugins/: 169 passed, 7 skipped
applicable hooks: all passed
```

All temporary source/test changes were reverted after validation. The earlier nested-entry precedence fix remains covered and passing; this finding is limited to the new namespace-package hardening.

An-idd:
@RerankerGuo Confirmed again — your PEP 420 repro fails at `4e49c802` exactly as reported (plugin B received A's `helpers.value`). The `loader is None` check was indeed too broad: it conflated "data directory that should lose to a concrete module elsewhere" with "genuine namespace package the plugin ships".

Fixed in 591fca04 with your suggested tie-break — a loader-less portion is treated as non-local only when a bare sys.path resolution yields a concrete loader elsewhere; otherwise it stays plugin-local. Your two-plugin repro is added to the suite as `test_pep420_namespace_packages_stay_isolated` (fails before the fix, passes after).

```text
tests/unit/plugins/test_plugin_namespace_isolation.py: 12 passed
tests/unit/plugins/: 169 passed, 7 skipped
pre-commit on touched files: all hooks passed
```

The `wave/` data-directory case still passes with the narrower guard.

RerankerGuo:
Revalidated the PEP 420 fix independently at exact head `591fca04632c917b189d0fb0e7658103987bdac7`.

The new tie-break preserves both sides of the boundary:

- two plugins shipping the same loader-less `helpers/value.py` namespace package remain isolated (`A` and `B`, respectively);
- a plugin data directory named `wave/` still does not shadow the concrete stdlib `wave` module.

Results:

```text
.venv/bin/pytest tests/unit/plugins/test_plugin_namespace_isolation.py -q
12 passed

.venv/bin/pytest tests/unit/plugins/ -q
169 passed, 7 skipped

  src/qwenpaw/plugins/module_isolation.py \
  tests/unit/plugins/test_plugin_namespace_isolation.py
all applicable hooks passed

git diff --check 4e49c802..591fca04
passed
```

The PEP 420 finding is resolved at this head. The earlier nested-entry precedence regression remains covered and passing as part of the same namespace suite.

RerankerGuo (approved):
No blocking issues found at exact head `591fca04632c917b189d0fb0e7658103987bdac7`.

The implementation now preserves the intended import flow across the original collision, nested/lazy imports, nested-entry precedence, PEP 420 namespaces, resource loaders, entry-module aliasing, and failed-load/unload cleanup:

```mermaid
flowchart LR
    A[Plugin executes bare import] --> B{Top name exists in plugin paths?}
    B -->|No| C[Standard Python import]
    B -->|Yes| D[Resolve plugin_id.name]
    D --> E[Namespace finder wraps concrete loader]
    E --> F[Submodule inherits plugin builtins]
    F --> G[Lazy and nested imports remain isolated]
    H[Unload or failed load] --> I[Remove namespaced and file-owned modules]
    I --> J[Unregister namespace]
    style D fill:#c8e6c9,color:#1a5e20
    style E fill:#bbdefb,color:#0d47a1
    style I fill:#fff3e0,color:#e65100
```

I independently reproduced and revalidated both previously identified boundaries before this approval:

- nested entries prefer their explicit entry directory over the plugin root;
- genuine PEP 420 namespace packages remain plugin-local while data-only directories still lose to concrete stdlib modules.

Fresh verification used the current-main merge tree `f9d61836b9df472032acc77a889d1115a15b6036`:

```text
pytest tests/unit/plugins/ -q
172 passed, 7 skipped

pre-commit on all five changed files
all applicable hooks passed (AST, mypy, black, flake8, pylint, etc.)

git diff --check origin/main...HEAD
passed
```

Runtime loading and CLI validation use the same entry-directory-first search paths and the same namespace redirection. The documented dynamic-import and vendored-path limitations do not regress the issue's normal `import` statement path.

XiuShenAl (commented):
Thanks for the continued work on this — the isolation approach is in good shape overall, and the follow-up commits clearly improve entry-directory precedence, `importlib.resources` loader forwarding, cleanup ordering, and the PEP 420 / data-directory boundary handling.

I re-checked the current head (`591fca04`) and found **two related blocking issues** in the PEP 420 / data-directory tie-break. Both should be fixed (with regression tests) before merge.

### Blocking 1: `resolved is None` incorrectly rejects plugin-local PEP 420 packages

In `src/qwenpaw/plugins/module_isolation.py` (`_find_local`):

```python
if spec is not None and spec.loader is None:
    resolved = importlib.machinery.PathFinder.find_spec(top)
    if resolved is None or resolved.loader is not None:
        spec = None
```

This doesn’t match the intended rule from the commit / discussion:

> a loader-less portion should be treated as non-local **only when** a bare `sys.path` resolution yields a **concrete** loader elsewhere; otherwise it should stay plugin-local.

As written, `resolved is None` incorrectly forces the name out of the plugin namespace. Empirically, a PEP 420 package imported **without** `sys.path.insert` fails with `ModuleNotFoundError`, while an equivalent regular package (with `__init__.py`) succeeds.

The existing `test_pep420_namespace_packages_stay_isolated` inserts the plugin dir onto `sys.path` first, so it doesn’t catch this case.

### Blocking 2 (more serious): bare global `PathFinder` tie-break breaks under a common load order

Even if Blocking 1 is fixed by changing the condition to:

```python
if resolved is not None and resolved.loader is not None:
    spec = None
```

…isolation still fails in this realistic pattern:

1. Plugin A ships `helpers/` as a **regular** package (`__init__.py`) and does `sys.path.insert`
2. Plugin B ships `helpers/` as a **PEP 420** package (no `__init__.py`) and also does `sys.path.insert`

Because a successful load leaves the plugin directory on `sys.path`, the bare `PathFinder.find_spec(top)` probe can observe A’s concrete `helpers` and treat B’s local namespace portion as non-local. B then falls back to the process-global import and reads **A’s** `helpers.value` (and binds the global `helpers` module), instead of `plugin_b.helpers`.

I reproduced this at `591fca04`: both the current code and the “condition-only” fix yield `B.VALUE == "A"` with `helpers.__name__ == "helpers"` (not namespaced).

Current coverage only exercises **PEP 420 vs PEP 420**, not **regular package → later PEP 420**, so this hole is also untested.

### Recommended direction

Please avoid using a bare global `PathFinder.find_spec(top)` (full `sys.path` scan) to distinguish “data directory” from “real PEP 420 package”. That couples plugin-local resolution to leftover `sys.path` entries from earlier plugins, which is exactly what this PR is trying to eliminate.

A more robust local check is whether the plugin’s own portion contains importable code, for example:

- empty / data-only `wave/` → **not** plugin-local → fall through to stdlib / third-party  
- PEP 420 `helpers/` with `helpers/value.py` → **always** plugin-local, regardless of other plugins’ residual `sys.path` entries

### Requested follow-up before merge

1. Fix Blocking 1 (`resolved is None` mishandling), as part of redesigning the tie-break rather than only flipping `or` / `and`.
2. Fix Blocking 2 by replacing the bare global `PathFinder` tie-break with a plugin-local “has importable code” (or equivalent) check that is independent of other plugins’ `sys.path` residue.
3. Add regression tests for:
   - PEP 420 import **without** `sys.path.insert` (success + cross-plugin isolation)
   - **regular package → later PEP 420** (both with `sys.path.insert`; later plugin must see its own value under `plugin_<id>.*`, not the global module)
4. Keep the existing `wave/` data-directory vs stdlib coverage.

Thanks again for the careful iteration on this PR!

An-idd:
@XiuShenAl Thank you for the thorough review — both findings confirmed at `591fca04` before fixing:

- **Blocking 1**: a PEP 420 package imported *without* a `sys.path` insert failed with `ModuleNotFoundError` (while a regular package succeeded) — my `resolved is None` branch contradicted the rule stated in my own commit message.
- **Blocking 2**: with plugin A shipping a regular `helpers/` and plugin B a PEP 420 `helpers/` (both inserting `sys.path`), B received A's `helpers.value` — the bare global `PathFinder` probe observed A's residual `sys.path` entry, exactly the coupling you described.

Fixed in fe28b61b following your recommended direction — the global probe is gone entirely. A loader-less portion is now plugin-local iff the plugin's own portion contains importable code (recursive short-circuit scan over `importlib.machinery.all_suffixes()`), decided purely from the plugin's files:

- data-only `wave/` → falls through, concrete stdlib/third-party wins (existing test kept and passing);
- code-bearing PEP 420 `helpers/` → always plugin-local, independent of other plugins' `sys.path` residue.

Both requested regression tests added (`test_pep420_package_without_sys_path_insert` covering success + cross-plugin isolation, `test_regular_package_then_pep420_package` covering the A-then-B order with the module asserted under `plugin_<id>.*`); both fail at `591fca04` and pass at the new head.

```text
tests/unit/plugins/test_plugin_namespace_isolation.py: 14 passed
tests/unit/plugins/: 171 passed, 7 skipped
pre-commit on touched files: all hooks passed
```

XiuShenAl:
@An-idd Thanks for the quick follow-up in `fe28b61b` — the redesign looks good.

I re-checked the new plugin-local tie-break (`_has_importable_code`) and confirmed that the two earlier blockers are fixed:

- PEP 420 imports work without `sys.path.insert`, and stay isolated
- regular package → later same-named PEP 420 (both with `sys.path.insert`) no longer falls back to the earlier plugin’s global module
- the requested regression tests cover those cases, and `wave/` data directories still resolve to the stdlib in the clean case

One remaining **blocking** issue before merge:

### Blocking: `__pycache__` / `.pyc` makes data directories look “code-bearing”

`_has_importable_code` treats any file matching `importlib.machinery.all_suffixes()` as importable code. That includes bytecode (`.pyc`). So a data-only directory that happens to contain leftover `__pycache__/*.pyc` is incorrectly classified as a real PEP 420 package and captured into the plugin namespace — which can shadow a concrete stdlib / third-party module.

Repro at `fe28b61b`:

```text
plugin/
  wave/
    sample.bin
    __pycache__/x.cpython-312.pyc
```

```python
import wave
```

Observed: `wave.__name__ == "plugin_….wave"` and `hasattr(wave, "open") is False`  
Expected (per this PR’s data-dir rule): the real stdlib `wave`

This breaks the invariant that data-only directories should fall through so concrete modules can win.

### Requested follow-up

1. Tighten `_has_importable_code` so stale bytecode doesn’t count — for example skip `__pycache__` during the walk, and/or ignore bytecode suffixes and only treat source / extension modules as code.
2. Add a regression test: `wave/` with only `sample.bin` + `__pycache__/*.pyc` must still resolve to the stdlib `wave` (and must not create `plugin_*.wave`).

Thanks again!

An-idd:
Follow-up hardening in b2c80cb9, from a multi-angle self-review of the full branch diff (all findings verified with failing tests or reasoning before changes; the review also vetoed two of its own candidate fixes that would have degraded isolation):

- **`sys.path` cleanup no longer strips relative entries** (`''` — the CWD): unloading a plugin while the process CWD was inside it used to delete the CWD entry permanently. The sweep is now a shared helper (`strip_plugin_sys_path`) used by loader cleanup, unload, and CLI validation — validation previously leaked plugin-inserted `sys.path` entries into the CLI process.
- **Validation cleanup ordering aligned with the loader** (sweep `sys.modules`, then unregister the namespace).
- **`_NamespaceLoader.__getattr__` recursion guard** for instances created without `__init__` (e.g. `cls.__new__` during copy).
- **`_has_importable_code` counts only importable paths**: code reachable through non-identifier components (`locale/en_US.UTF-8/tool.py`) cannot be imported, so it no longer makes a data directory read as a package; the identifier pruning also bounds the walk on large asset trees.
- **Entry-file alias handles `fromlist`**: `from <entry> import <sibling_module>` now imports the sibling as a namespaced submodule instead of failing the attribute lookup.

```text
tests/unit/plugins/test_plugin_namespace_isolation.py: 16 passed
tests/unit/plugins/: 174 passed, 7 skipped
pre-commit on touched files: all hooks passed
```

An-idd:
@XiuShenAl Confirmed — reproduced before fixing: at the previous head a data-only `wave/` with a stray legacy `wave/x.pyc` was captured into the plugin namespace (`__pycache__`-tagged bytecode like `x.cpython-312.pyc` was already excluded by the identifier-stem check, but untagged legacy bytecode slipped through, and `__pycache__` itself was not pruned).

Fixed in e2207d00, both tightenings you suggested:

- `_CODE_SUFFIXES` is now `SOURCE_SUFFIXES + EXTENSION_SUFFIXES` — bytecode is never evidence of code. Documented trade-off: a PEP 420 portion shipping *only* sourceless bytecode reads as data now.
- The walk skips `__pycache__` entirely.

Regression test added as requested (`test_stale_bytecode_does_not_make_data_dir_a_package`): `wave/` holding `sample.bin`, `__pycache__/x.cpython-312.pyc`, a `__pycache__/x.py` planted to lock the cache-dir pruning specifically, and a legacy `x.pyc` — resolves to the stdlib `wave`, no `plugin_*.wave` created. Fails before the fix (via the legacy `.pyc`), passes after. The clean `wave/` data-dir test is unchanged and still passing.

```text
tests/unit/plugins/test_plugin_namespace_isolation.py: 18 passed
tests/unit/plugins/: 175 passed, 7 skipped
pre-commit on touched files: all hooks passed
```

XiuShenAl:
@An-idd Thanks again for the solid work so far — the earlier PEP 420 / data-directory / stale-bytecode issues look resolved at `e2207d00`, and the regression coverage there is in good shape.

After another pass, I still see **one remaining root cause** that should be fixed before merge. It shows up in a few related ways, so I’m listing them together so we can close the whole class in one go.

### Root cause

For names that are **not** plugin-local, the importer falls through to:

```python
builtins.__import__(name, ...)
```

That resolves against the full process `sys.path` / `sys.modules`, including source/entry directories left behind by **successfully loaded** plugins (`sys.path.insert` residue).  
Locality no longer uses a global `PathFinder` probe (good), but **fallthrough is still fully exposed to that residue**.

### Blocking manifestations (same root cause)

Please treat these as one fix with shared acceptance criteria — not three unrelated bugs.

**Blocking-1 — data directory fallthrough hits another plugin**

1. Plugin A ships a concrete package/module (e.g. `helpers/`) and does `sys.path.insert`
2. Plugin B ships only a same-named **data** directory and does `import helpers`

Observed: B gets A’s module and may bind the global top-level name.  
This breaks the invariant that data directories should fall through to a real stdlib/third-party module, not another plugin.

**Blocking-2 — uncached stdlib / host name gets polluted**

1. Plugin A ships a concrete module with a stdlib-like name (e.g. `wave.py`); A’s own import is namespaced, so the global name may still be clean
2. Plugin B has only a data directory `wave/` and imports `wave`
3. Important: the name is **not** already cached in `sys.modules` (this is the natural state after A’s namespaced import)

Observed: fallthrough loads A’s file into global `sys.modules['wave']`; later host `import wave` also gets the fake module.  
This is more severe than Blocking-1 because it pollutes process-level stdlib resolution. The existing single-plugin, already-cached `wave/` tests don’t catch it.

**Blocking-3 — no local name at all, same fallthrough path**

1. Plugin A leaves a concrete package/module on `sys.path`
2. Plugin B has **neither** a package nor a data directory for that name, but still `import <name>`

Observed: same mechanism as above — B loads A’s module into the global namespace.  
This is partly pre-existing relative to a full `sys.path` import world, but under this PR’s architecture it is the remaining pollution valve and must be closed by the **same** fallthrough fix. If we only special-case data directories and leave general non-local imports on the full `sys.path`, this case will slip through.

### Requested fix / acceptance

1. **All** non-local bare-name fallthroughs (not only data-dir cases) must not resolve into other **loaded plugins’ source/entry directories**.
2. Real stdlib / third-party modules must still win over pure data directories.
3. Please avoid breaking shared plugin dependency locations (e.g. plugin site-packages / deps install dirs) — only plugin source/entry trees should be excluded.
4. Please add regressions for Blocking-1, Blocking-2, and Blocking-3. For Blocking-2, assert the true stdlib module when the name is not pre-cached, and assert that host / `sys.modules` are not left with the wrong binding afterward.
5. Please keep the existing coverage for the #6683 main path, PEP 420 isolation, entry-directory precedence, clean `wave/`, and stale-bytecode cases.

### Out of scope for this round (not merge blockers)

- Documented `importlib.import_module` / `builtins.__import__` bypasses  
- Vendored dirs a plugin adds outside `search_paths`  
- Keeping plugin dirs on `sys.path` after a successful load itself (acceptable if fallthrough can no longer see other plugins’ source/entry paths)

Happy to take another look once this fallthrough class is closed. Thanks!

An-idd:
@XiuShenAl Confirmed all three manifestations at `e2207d00` before fixing — each reproduced as a failing test (B1: plugin B received A's `helpers`; B2: A's fake `wave.py` landed in the global `sys.modules['wave']`; B3: B bound A's module with no local candidate).

Fixed in ee139ea9 with one change for the class, per your acceptance criteria:

**Fix**: a plugin's inserted `sys.path` entries are swept in a `finally` as its load finishes (success, failure, or a cancelled startup). Its own imports never needed `sys.path` — they resolve through the private namespace (`search_paths`) — so once a plugin is loaded there is no residue for any later fallthrough to see. This closes B1/B2/B3 at the shared root rather than special-casing data directories.

Against your acceptance list:

1. ✅ All non-local fallthroughs (not only data-dir) can no longer resolve into loaded plugins' source/entry trees — the trees are simply gone from `sys.path`.
2. ✅ Real stdlib/third-party still wins over pure data directories (B2 test asserts the true stdlib `wave` when uncached; clean-`wave/` test unchanged).
3. ✅ The plugin site/deps dir lives outside plugin source trees and is untouched; the sweep only removes paths under the plugin's own tree, and never relative entries (`''`).
4. ✅ Regressions added for B1 (data-dir fallthrough must not bind A's tree), B2 (uncached stdlib name → real module, host `sys.modules` asserted clean), B3 (no local candidate → `ModuleNotFoundError`, no global binding).
5. ✅ All prior coverage intact: #6683 main path, PEP 420 isolation, entry-dir precedence, clean `wave/`, stale bytecode — 178 passed, 7 skipped.

One documented behavior shift worth noting (updated in the module docstring): dynamic `importlib.import_module("<bare-local-name>")` after load now raises `ModuleNotFoundError` instead of resolving globally, since the residue it used to ride on is gone — a loud failure in place of silent cross-plugin coupling. Repo plugins are unaffected (the only dynamic case, creator's `__import__("utils.logger", ...)`, goes through `__import__` and stays namespaced).

```text
tests/unit/plugins/test_plugin_namespace_isolation.py: 21 passed
tests/unit/plugins/: 178 passed, 7 skipped
pre-commit on touched files: all hooks passed
```

An-idd:
Addendum in 651645f0, from auditing the fallthrough class for remaining instances rather than stopping at the `sys.path` valve:

**`sys.modules` is the second residue channel.** During a plugin's load, a data-directory fallthrough caches a bare namespace package rooted in the plugin's tree (observable in the Blocking-1 repro), and a bypass import caches bare modules — both outlive the load and keep serving later plugins from the import cache even with `sys.path` clean. The failed-load/unload sweeps also missed namespace packages entirely (they matched by `__file__`, which namespace packages don't have).

Fix: a shared `sweep_bare_tree_modules(source_path)` (matches `__file__`, or `__path__` portions for namespace packages; registered `plugin_<id>` roots exempt) runs wherever `sys.path` residue is already stripped — load end on every exit path, unload, and CLI validation. Ordering matters and is now uniform: always **strip `sys.path` first, then sweep** — a namespace package's `__path__` recalculation reads the live `sys.path`, so sweeping first could merge plugin-tree portions into a shared host package (e.g. a site `google/` namespace) and evict it; the pre-strip sweep in the failed-load path is removed for that reason.

Two regression tests added: bare namespace residue is gone at load end (a later plugin importing the name gets `ModuleNotFoundError`, not the cached leftover), and failed-load residue is swept too. All prior coverage intact.

```text
tests/unit/plugins/test_plugin_namespace_isolation.py: 23 passed
tests/unit/plugins/: 180 passed, 7 skipped
pre-commit on touched files: all hooks passed
```

XiuShenAl (approved):
LGTM

PR #6692, opened 2026-08-04
fix: avoid logging conversation command arguments

niceIrene:
## Summary

- Stop logging raw conversation command arguments at INFO level.
- Keep the command name in logs for operational visibility.
- Add regression coverage for `/compact` hint redaction and one-shot behavior.

## Root Cause

The conversation command dispatcher logged the raw `args` value before command-specific processing. As a result, credentials included in a `/compact` hint could be written to `qwenpaw.log` in plaintext even though the hint was redacted before compaction and persistence.

## Changes

- Log only the conversation command name, excluding user-controlled arguments.
- Verify that a fake `sk-...` credential does not appear in captured logs.
- Verify that the hint passed to the compactor is redacted.
- Verify that a subsequent `/compact` without a hint does not reuse the previous hint.

## Test Plan

- [x] `tests/unit/agents/test_command_handler.py` — 22 passed
- [x] Pre-commit hooks passed, including mypy, Black, Flake8, and Pylint
- [x] `git diff --check` passed

## Related Test Case

`TC-CTX-15` — `/compact` hint one-shot behavior and secret redaction

### DIFF
--- a/src/qwenpaw/agents/command_handler.py
+++ b/src/qwenpaw/agents/command_handler.py
@@ -1262,7 +1262,11 @@ class CommandHandler(ConversationCommandHandlerMixin):
         parts = query.strip().lstrip("/").split(" ", maxsplit=1)
         command = parts[0]
         args = parts[1] if len(parts) > 1 else ""
-        logger.info(f"Processing command: {command}, args: {args}")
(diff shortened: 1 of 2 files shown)


cuiyuebing (approved):
LGTM

PR #6704, opened 2026-08-05
feat(chat): session fork — snapshot conversation context to new session

iluv7:
## Description

**Related Issue:** Relates to https://github.com/agentscope-ai/QwenPaw/issues/6560

Add session fork feature: users can right-click a session and fork it into a new independent session with the full conversation context copied. This is a checkpoint-style snapshot — the forked session starts with all history intact and evolves independently.

**Backend:**
- `POST /chats/{chat_id}/fork` (201)
- `SafeJSONSession.clone_session_state()` — copies the raw session JSON to a new file under path lock. If the source file was manually deleted, writes `{}` so the fork still succeeds with an empty session.
- `SafeJSONSession.delete_session_state()` — removes the session JSON file when ChatSpec creation fails, skips without error if already deleted.
- `ChatManager.fork_chat()` — creates a new `ChatSpec` with user/channel copied from the source, plus `forked_from_chat_id`, `forked_from_session_id`, and `forked_at` in meta so the lineage is traceable.
- Returns 409 if the source is still generating, since the persisted state may be incomplete.

**Frontend:**
- Added `HttpError` to `request.ts` so callers can check `err.status === 409`.
- `SessionItem` dropdown now has a Fork item (`GitFork` icon from lucide). Disabled for sessions that haven't sent a message yet (no backend UUID).
- Flow: call fork API → refresh the session list in the sidebar → navigate to the new chat. Refresh and navigation are independent — a failed refresh still lands you in the new chat.
- `useRef(Set)` tracks in-flight fork requests so double-clicking doesn't create duplicates.
- If `source_state` comes back `"empty"`, a warning toast lets the user know the original context wasn't available.

**Security Considerations:** Fork endpoint inherits existing chat API auth. No new secrets or permissions.

## Type of Change

- [x] New feature

## Component(s) Affected

- [x] Tests

## Checklist

- [x] Ready for review

## Testing

**Backend (17 tests):**

| Scope | Tests | What they verify |
|-------|-------|-----------------|
| `test_fork.py` | 10 | `ChatManager.fork_chat`: creates independent spec with lineage meta, raises on unknown source, deep-copies meta. `clone_session_state`: full state copy, raises on missing source, `allow_missing_source=True` writes `{}`, cloned session is independent. `delete_session_state`: removes file, returns false on missing, handles double-delete. |
| `test_chats_fork_router.py` | 7 | API layer: 201 with Fork name and `source_state`, custom name forwarding, 404 on unknown chat, 409 when running, 500 on clone failure (no orphan), 500 on ChatSpec failure (cleanup called), 500 on cleanup failure (still returns 500). |

**Frontend:**
- `npx tsc --noEmit` — 0 errors, all props/types aligned across SessionItem/Drawer/Sidebar.

## Evidence

All local checks and tests passed — see details below.

### Pre-commit

All hooks passed (check ast, mypy, black, flake8, pylint).

### Tests — 17 passed

What the 17 tests prove:
- Fork API returns 201 with correct lineage metadata on success
- 404 / 409 / 500 error paths all respond correctly
- Cloned session file is fully independent — mutations don't cross
- Missing source session file gracefully writes `{}` instead of crashing
- Orphan session file is cleaned up when ChatSpec creation fails
- Cleanup failure does not block the error response

### Console UI

04ef6491c989dacbafec4557371ba0b0

### DIFF
--- a/console/src/api/modules/chat.ts
+++ b/console/src/api/modules/chat.ts
@@ -7,6 +7,7 @@ import type {
   ChatDeleteResponse,
   ChatUpdateRequest,
   BatchArchiveResult,
+  ForkChatResponse,
   Session,
 } from "../types";
 
@@ -126,6 +127,15 @@ export const chatApi = {
     request<void>(`/console/chat/stop?chat_id=${encodeURIComponent(chatId)}`, {
       method: "POST",
     }),
+
+  forkChat: (chatId: string, name?: string) =>
+    request<ForkChatResponse>(
+      `/chats/${encodeURIComponent(chatId)}/fork`,
+      {
+        method: "POST",
+        body: name ? JSON.stringify({ name }) : undefined,
+      },
+    ),
 };
 
 export const sessionApi = {
--- a/console/src/api/request.ts
+++ b/console/src/api/request.ts
@@ -2,6 +2,16 @@ import { getApiUrl, clearAuthToken } from "./config";
 import { buildAuthHeaders } from "./authHeaders";
 import { getLoginHref, isLoginPath } from "../utils/navigationMode";
 
+/** HTTP error with status code, so callers can distinguish 409 vs 500 etc. */
+export class HttpError extends Error {
+  status: number;
+  constructor(status: number, message: string) {
+    super(message);
+    this.name = "HttpError";
+    this.status = status;
+  }
+}
+
 function getErrorMessageFromBody(
   text: string,
   contentType: string,
@@ -138,7 +148,7 @@ export async function request<T = unknown>(
           ? `${errorMessage} - ${text}`
           : `Request failed: ${response.status} ${response.statusText}`;
 
-        throw new Error(finalMessage);
+        throw new HttpError(response.status, finalMessage);
       }
 
       if (response.status === 204) {
--- a/console/src/api/types/chat.ts
+++ b/console/src/api/types/chat.ts
@@ -45,5 +45,13 @@ export interface BatchArchiveResult {
   }>;
 }
 
+export interface ForkChatResponse {
+  chat_id: string;
+  session_id: string;
+  name: string;
+  created_at: string;
+  source_state: "ok" | "empty";
+}
+
 // Legacy Session type alias for backward compatibility
 export type Session = ChatSpec;
--- a/console/src/components/SessionItem/index.tsx
+++ b/console/src/components/SessionItem/index.tsx
@@ -2,6 +2,7 @@ import React, { useCallback, useMemo, useRef, useState } from "react";
 import { Dropdown, Input } from "antd";
 import type { InputRef } from "antd";
 import { useTranslation } from "react-i18next";
+import { GitFork } from "lucide-react";
 import {
   SparkMoreLine,
   SparkDeleteLine,
@@ -41,6 +42,8 @@ export interface SessionItemProps {
   onDelete?: (sessionId: string) => void;
   onPin?: (sessionId: string) => void;
   onArchive?: (sessionId: string) => void;
+  onFork?: (sessionId: string) => void;
+  forkDisabled?: boolean;
   onEditChange?: (value: string) => void;
   onEditSubmit?: () => void;
   onEditCancel?: () => void;
@@ -66,6 +69,8 @@ const SessionItem: React.FC<SessionItemProps> = ({
   onDelete,
   onPin,
   onArchive,
+  onFork,
+  forkDisabled,
   onEditChange,
   onEditSubmit,
   onEditCancel,
@@ -128,6 +133,13 @@ const SessionItem: React.FC<SessionItemProps> = ({
           : t("sessions.archive.action", "Archive"),
         onClick: () => onArchive?.(sessionId),
       },
+      {
+        key: "fork",
+        icon: <GitFork size={14} />,
+        label: t("chat.contextMenu.fork", "Fork"),
+        disabled: forkDisabled,
+        onClick: () => onFork?.(sessionId),
+      },
       { type: "divider" as const },
       {
         key: "delete",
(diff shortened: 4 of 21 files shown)


RerankerGuo (changes requested):
I validated exact head `3358e26e` and a semantic current-main trial based on `bacac741`. The API happy/error paths are well covered, but cloning the raw session JSON does not create an independent full-context session when the production agent state and Scroll context manager are used.

The persisted `agent` module contains both:

1. `state.session_id`, AgentScope's durable internal session identity; and
2. `scroll` bookkeeping for messages already written to the shared `history.db`.

`clone_session_state()` copies both unchanged. On the fork's first build, `QwenPawAgent.load_state_dict()` restores the source `state.session_id`, and `MemoryMiddleware` explicitly prefers that value over the fork request context. Consequently, the source and fork continue using the same memory search/write scope instead of evolving independently.

The Scroll side loses history as well. I exercised this production sequence:

1. create a source `AgentState` and real `ScrollContextManager`;
2. write an old and a recent message to the real `HistoryStore`;
3. evict the old message from the live window;
4. persist the real `QwenPawAgent.state_dict()` via `SafeJSONSession`;
5. call this PR's `clone_session_state()`;
6. restore through a new `ScrollContextManager` keyed by the fork session and the real `QwenPawAgent.load_state_dict()`;
7. run the fork's first Scroll write-through.

Both exact and current-main trial produce:

```text
identity request='console:user:fork-12345678'
loaded='agentscope-source-identity'
memory_scope='agentscope-source-identity'

scroll_history source_rows=2
fork_rows_before_resume=0
fork_rows_after_resume=0
fork_live_messages=1
```

The evicted source row is absent from the fork because it lives only in `history.db`, not session JSON. Worse, the copied `persisted_ids` marks the one remaining live message as already durable, so the fork's write-through does not store even that message under the fork ID. It remains visible temporarily but has no fork-owned recall record once later compression evicts it.

Please make forking rebase every session-owned persistence layer rather than copying raw JSON unchanged:

- assign a destination-specific AgentScope/memory identity instead of retaining the source `AgentState.session_id`;
- copy the source session's durable Scroll rows into the destination scope without moving or sharing them;
- rebuild sequence-dependent Scroll bookkeeping/eviction index against the destination rows;
- keep source and fork mutations independent.

The regression should use the real `QwenPawAgent.state_dict()` / `load_state_dict()`, `ScrollContextManager`, `HistoryStore`, and `SafeJSONSession`, with at least one already-evicted message. Assert that memory routing resolves to a destination identity, both old and live messages are recallable under the fork, and subsequent writes change only the fork.

Verification:

- exact fork tests: `17 passed`;
- semantic current-main chat/session/Scroll/memory matrix: `178 passed`;
- Console `tsc --noEmit`: pass;
- exact `py_compile` and `git diff --check`: pass.

The branch also has a production merge conflict in `src/qwenpaw/app/chats/api.py`; the semantic trial retained both current main's project-directory routes and the new fork route.

iluv7:
Hi, @RerankerGuo 
Thank you for the detailed production-path reproduction. Your analysis was correct: copying only the session JSON retained the source `AgentState.session_id`, omitted compressed messages that existed only in `history.db`, and left Scroll checkpoint sequences pointing to the source session.

These issues are addressed in commits a0535f53 and e3c4edff: 

1. **Assign an independent AgentScope/memory identity to the fork.** Before writing the destination snapshot, `SafeJSONSession.clone_session_state()` now replaces `agent.state.session_id` with the new fork session ID. Future session-scoped memory operations are therefore recorded under the fork instead of the source session.

2. **Copy and isolate durable Scroll history.** `HistoryStore.clone_session_rows()` now copies every durable Scroll row owned by the source session into the destination session in a single SQLite transaction. The copied rows receive new globally unique `seq` values while preserving their content, structure, dedup keys, and FTS indexing. Source rows are neither moved nor modified, and subsequent source and fork writes go to their respective scopes.

3. **Rebuild all sequence-dependent Scroll state.** The history clone returns a `source seq -> destination seq` mapping. `ScrollContextManager.rebase_checkpoint()` uses it to rebase `seq_by_id`, `seq_by_tcid`, `model_turn_seq`, `leaf_by_id`, `persisted_ids`, `persisted_tcids`, the eviction index, and the continuation summary with its source ranges. If a checkpoint contains a stale pointer that cannot be mapped, the corresponding state is dropped rather than retaining a reference to a source-session row.

4. **Roll back partially created forks.** The API now performs the fork in this order: copy durable Scroll history, write the rebased session snapshot, then create the `ChatSpec`. If either of the latter steps fails, it best-effort removes the destination history rows and session file so a partial fork is not left behind.

The new regression test, `tests/unit/app/chats/test_session_fork_scroll.py`, exercises the production state path using the real `QwenPawAgent.state_dict()` / `load_state_dict()`, `ScrollContextManager`, `HistoryStore`, `SafeJSONSession`, and `MemoryMiddleware` / `MemorySpace`.

The test persists an old message and a live message, evicts the old message from the live window, forks the session, restores it, and runs the first Scroll write-through. It verifies that:

- Uses its own session and memory identity
- Retains both active messages and messages previously moved out of the live context
- Rebuilds its Scroll metadata against its own copied history
- Stores new messages independently without losing or duplicating inherited context or modifying the source session

The branch has also been rebased onto the current main you used for the semantic trial, `bacac741`. The conflict in `src/qwenpaw/app/chats/api.py` is resolved, with the current-main project-directory/checkpoint behavior retained alongside the fork route.

Latest verification:

- Backend chat/session/Scroll/memory matrix: `518 passed`
- Focused fork and cleanup tests: `46 passed`
- Console TypeScript check: passed
- `pre-commit run --all-files`: passed
- `git diff --check`: passed

iluv7:
Regarding long-term memory, QwenPaw currently scopes distilled ReMe memory and its semantic index at the agent level rather than the session level. This implementation follows that existing model: the fork gets an independent copy of its conversation state and Scroll history while continuing to use the agent-level knowledge base. We did not copy the ReMe artifacts or semantic index because doing so could introduce duplicate memories and search results and would require separate per-session storage, indexing, and lifecycle management. Please let me know if full per-session isolation of long-term memory is expected for this feature; we would be happy to discuss that broader design separately.

PR #6715, opened 2026-08-05
feat(onebot): localize inbound media before agent processing

GMsure:
# feat(onebot): localize inbound media before agent processing

## Summary

This PR aligns OneBot inbound image, audio, video, and file handling with the AgentScope 2.0 local `DataBlock` pipeline. OneBot-specific references are resolved and downloaded into managed local storage before agent processing, while the shared media layer continues to handle transcription, formatting, and model-specific behavior.

The branch now includes all maintainer/reviewer follow-ups: inbound/outbound limit separation, bounded streaming downloads, workspace isolation, source-segment-safe resolution, deterministic RIFF/MIME handling, per-session ordering, early ACL rejection, Console configuration, and removal of the superseded image downloader.

## What Problem This Solves

I reproduced a QQ voice message that arrived as a remote OneBot media URL but produced no reply because remote Pydantic `DataBlock`s are skipped by the common AgentScope 2.0 media path. Images, videos, and generic files had the same architectural gap or inconsistent channel-specific behavior.

The OneBot adapter now localizes all supported inbound media types first, including quoted media and OneBot `file_id` references. The resulting local paths become `file://` DataBlocks in the common pipeline, so audio reaches transcription/native delivery and other media types follow the existing shared behavior.

## Changes

- Localize inbound HTTP(S) image/audio/video/file media by default, independently of outbound `media_base64`.
- Resolve OneBot `file_id`, image, and record references through the appropriate OneBot API before downloading.
- Add `media_download_max_mb` (default `50`) as an independent inbound limit; keep `media_base64_max_mb` outbound-only.
- Reuse a channel-owned `aiohttp.ClientSession`, honor environment proxy settings, bound concurrent downloads, stream to a temporary file, enforce limits while streaming, and atomically finalize completed files.
- Reject clearly mismatched image/audio/video response MIME types and use deterministic MIME suffixes plus RIFF subtype detection for WAVE, WebP, and AVI.
- Match runtime media parts to their original segment by media kind and source reference, preserving the original segment index and preventing positional drift when an earlier segment is skipped.
- Forward `workspace_dir` through `from_config()`; default storage is `<workspace>/media/onebot`.
- Move local filesystem probing and file writes/finalization off the event loop; bound generated filename length.
- Preserve message order with one bounded queue/worker per session, reject dispatch while stopping, and reclaim idle workers.
- Run mention and ACL gates before quoted-message API calls or media downloads.
- Preserve the local path hint when audio transcription fails, matching the legacy fallback contract.
- Remove the dead `_inline_remote_images` / `_download_image_data_url` implementation.
- Expose `media_download_max_mb` and `media_dir` in the Console type/form with all seven locales, and document both fields in English and Chinese.

## Evidence

Real OneBot/QQ behavior verified with SnowLuma on this branch:

- A remote voice message reaches the localized local-file path and existing transcription/native-audio pipeline.
- With `require_mention` enabled, an unmentioned image-only group message is rejected before queue/model processing; the same image with `@bot` enters the normal path and receives a description.
- Quoted image/file messages are expanded through `get_msg`, retain their media type, and are localized for downstream access.

Automated validation after syncing the latest `upstream/main`:

- `python -m pytest tests/unit/channels/test_onebot_channel.py tests/unit/agents/utils/test_message_processing.py tests/unit/runtime/test_message_convert.py tests/unit/config/test_channel_shape_consistency.py -q` → `158 passed`.
- Targeted pre-commit checks pass for all changed Python files: AST, mypy, Black, Flake8, and Pylint.
- `npx tsc -b --noEmit` passes for the Console.
- Current-project Prettier checks pass for the changed Console type/form/locales and both website documentation files.
- The branch is synchronized with current `upstream/main` (`0` commits behind).

Regression coverage includes default-config localization, all four media types, quoted/current file IDs, skipped-segment source matching, RIFF WAVE/WebP/AVI detection, deterministic MIME suffixes, mismatched content-type rejection, download overflow cleanup, session reuse/concurrency, ACL-before-I/O ordering, per-session ordering/idle cleanup, Unicode/space paths, workspace config propagation, and audio localization-to-transcription integration.

## Scope

- Affects OneBot inbound media handling, its related configuration surface, tests, and channel documentation.
- Does not change OneBot outbound text/media semantics; `media_base64` remains outbound-only.
- Does not change other channel adapters.


hongxicheng:
Thank you for the contribution. After reviewing the code, I believe the root cause is broader than remote audio handling: OneBot's inbound file and media processing no longer matches the current AgentScope 2.0 DataBlock pipeline. Remote Pydantic DataBlocks are skipped by the common media-processing path, while the legacy dict-based path still supports downloading them.

Instead of adding separate patches for audio in the shared utility and images in the OneBot channel, I think a cleaner solution would be to introduce a unified inbound media-handling flow for OneBot, similar to the approach used by channels such as Feishu or DingTalk.

The OneBot channel should resolve channel-specific file IDs and URLs, download the supported image, audio, video, and generic file types into managed local storage, and then pass local file paths or file:// DataBlocks to the agent. The existing common processing layer can then handle transcription, media formatting, and model-specific inlining consistently.

This would address the underlying AgentScope 2.0 compatibility issue, avoid media-specific patches across different layers, and prevent video and generic files from remaining unsupported or behaving differently from images and audio. Could you please consider refactoring the PR in this direction?

GMsure:
Okay~ However, I think I need quoted-reply support before I can fully test this refactor(QQ messages' videos or files can only be sent separately and then referenced; they cannot be sent inline). I have submitted a new PR: [#6769](https://github.com/agentscope-ai/QwenPaw/pull/6769). Could you please review that PR first? Once quoted-message support is available, I will complete the end-to-end testing for this PR's refactor.

好的～不过我觉得，要完整测试这个重构，需要先解决引用回复的问题(QQ消息的视频和文件只能单独发送再引用，没办法内联)。我已经提交了一个新的 PR：[#6769](https://github.com/agentscope-ai/QwenPaw/pull/6769)。请先帮忙评审这个 PR；我会基于可用的消息引用能力，再完整测试本 PR 的重构。

GMsure:
## Update / 更新


This branch has been updated with the latest upstream `main`, including #6769, and the inbound-media implementation has been integrated with quoted-reply expansion.

- Quoted and current message segments are resolved independently, so each file keeps its own `file_id`.
- Remote image, audio, video, and file inputs are localized into the configured OneBot media directory with size limits and safe filenames.
- The downstream message pipeline now consumes localized `file://` DataBlocks.
- Real OneBot verification passed for quoted images and files: `get_msg` expanded the original segment, preserved its media type, added `[Quoted message]`, and localized the file for tool access.
- Targeted tests: **138 passed**.
- Commit-time checks for all changed files passed: AST, mypy, black, flake8, and pylint.



该分支已同步最新上游 `main`（包含 #6769），并完成了入站媒体本地化与引用消息展开的整合。

- 引用消息和当前消息分别使用各自的 segments 解析，确保每个文件保留自己的 `file_id`。
- 远程图片、语音、视频和文件会在大小限制与安全文件名约束下，本地化到配置的 OneBot 媒体目录。
- 下游消息处理统一消费本地化后的 `file://` DataBlock。
- 真实 OneBot 验证已覆盖引用图片和文件：`get_msg` 能展开原消息、保留媒体类型、添加 `[Quoted message]`，文件也已成功本地化并可由工具读取。
- 针对性测试：**138 passed**。
- 本次改动文件的提交检查全部通过：AST、mypy、black、flake8、pylint。

RerankerGuo:
Independent review at exact head `e9ce512070401e58bc83d6c816f342668935c76d`.

I found a default-configuration gap in the new unified inbound-media path. `OneBotConfig.media_base64` defaults to `False`, and `_localize_media_part()` returns the remote HTTP URL without calling `_download_remote_media()` in that state:

```python
if not self._media_base64:
    return ref
```

`_resolve_inbound_media()` then wraps that same remote URL back into `AudioContent`/`ImageContent`/etc. The AgentScope 2.0 common media path explicitly skips remote Pydantic `DataBlock`s and expects channel adapters to localize them. Consequently, the original remote-audio path can remain unfixed for users on the default OneBot configuration.

This flag also appears to have an outbound meaning: `_normalize_media_ref()` uses `media_base64` to decide whether a local outbound file should be serialized as `base64://`. It should not need to enable inbound downloading into the newly configured `media_dir`.

Minimal regression:

```python
channel = OneBotChannel.from_config(
    noop_process,
    OneBotConfig(enabled=True, media_dir=str(tmp_path)),
)
assert channel._media_base64 is False  # default

channel._download_remote_media = AsyncMock(
    return_value=str(tmp_path / "voice.mp3"),
)
remote_audio = AudioContent(
    type=ContentType.AUDIO,
    data="https://cdn.example.com/voice.amr",
)

resolved = await channel._resolve_inbound_media(
    [remote_audio],
    [{"type": "record", "data": {"url": remote_audio.data}}],
    "private",
    {},
)

assert resolved[0].data == str(tmp_path / "voice.mp3")
channel._download_remote_media.assert_awaited_once()
```

Result:

```text
FAILED test_default_config_localizes_remote_audio
E assert 'https://cdn.example.com/voice.amr' == '.../voice.mp3'
Expected _download_remote_media to have been awaited once.
Awaited 0 times.
```

The PR's complete stated target suite passes:

```bash
.venv/bin/pytest \
  tests/unit/channels/test_onebot_channel.py \
  tests/unit/agents/utils/test_message_processing.py \
  tests/unit/runtime/test_message_convert.py -q
# 138 passed
```

The current all-media localization test constructs the channel with `media_base64=True`, so it does not cover the default.

Suggested direction:

1. Localize inbound HTTP(S) media regardless of the outbound `media_base64` preference.
2. Keep `media_base64` only for deciding how local media is serialized to the OneBot client on outbound sends.
3. Add the default-config regression above, ideally continuing through message conversion/common processing to prove remote audio reaches transcription.

I did not run a live OneBot client in this environment; this finding is limited to the deterministic default-config code path and its downstream contract.

RerankerGuo (changes requested):
作者意图：统一 OneBot/AgentScope 2.0 入站 image/audio/video/file 本地化，让 downstream common media pipeline 只处理受控本地文件。

默认配置仍绕过该修复。`OneBotConfig.media_base64` 默认 `False`，但 `_localize_media_part()` 在远程 HTTP(S) media 上执行：

```python
if not self._media_base64:
    return ref
```

于是 `_resolve_inbound_media()` 把同一远程 URL 包回 `AudioContent`/其他 DataBlock。AgentScope 2.0 common processing 明确跳过 remote Pydantic DataBlock，原始“QQ 语音无响应”路径因此对默认用户仍然存在。

```mermaid
flowchart LR
    A[Remote inbound audio] --> B{media_base64 default false}
    B -->|false| C[Return HTTP URL unchanged]
    C --> D[Wrap as remote AudioContent]
    D --> E[Common media path skips remote block]
    E --> F[No transcription or native delivery]
```

确定性 default-config regression：

```text
expected localized path: .../voice.mp3
actual: https://cdn.example.com/voice.amr
_download_remote_media expected awaited once, actual 0
```

完整声明套件仍为 138 passed，因为 all-media localization test 显式设置了 `media_base64=True`。该 flag 在 `_normalize_media_ref()` 中本来控制的是出站本地文件是否序列化为 `base64://`，不应决定入站 remote media 是否下载进 `media_dir`。

请：

1. 对入站 HTTP(S) media 始终执行本地化，不依赖出站 `media_base64`；
2. 将 `media_base64` 仅用于 outbound reference serialization；
3. 增加默认 `OneBotConfig(enabled=True, media_dir=...)` regression；
4. 最好继续穿过 message conversion/common processing，证明默认配置的远程 audio 最终到达 transcription，其他 media types 也得到受控 local/file reference。

hongxicheng:
Thanks for moving OneBot inbound media handling toward the same channel-level
localization flow used by Feishu, DingTalk, WeCom, Telegram, and QQ. Resolving
the platform reference, downloading into managed local storage, preserving the
typed content, and then using the shared runtime/media hook is the right
direction.

I think the following issues should be addressed before merge:

1. **The outbound Base64 limit is incorrectly applied to inbound downloads.**
   `media_base64_max_mb` is documented in the Console as the limit for sending
   local media as Base64, but `_download_remote_media()` uses it
   unconditionally. With the default configuration, inbound files larger than
   10 MB fail even when `media_base64` is disabled. Please remove this
   dependency from the inbound path. If an inbound limit is required, it needs
   a separate name, configuration, default, and tests.

2. **Removing that limit alone would make the current memory behavior unsafe.**
   Each download is accumulated into a complete `bytearray` and copied again
   through `bytes(data)`, while the channel allows up to 500 concurrent event
   tasks. Please use bounded download concurrency, reuse a channel-owned
   `aiohttp.ClientSession`, and stream each response to a temporary file before
   an atomic rename. An unlimited business file size must not mean unlimited
   in-memory buffering.

3. **The production change in `message_processing.py` is unnecessary and
   regresses the audio fallback.** `_process_audio_block()` returns `False`
   when transcription fails so that the local file path hint remains visible
   to the agent. Ignoring that result removes the path hint and leaves only the
   placeholder. OneBot-specific remote localization should remain in the
   OneBot adapter; please revert this production change and keep the OneBot
   integration test in the OneBot test module.

4. **`workspace_dir` is not wired through `OneBotChannel.from_config()`.** The
   constructor accepts it, but `ChannelManager` filters arguments against the
   factory signature, so the active agent workspace is never passed and media
   falls back to global storage. Please accept and forward `workspace_dir`, as
   the reference channels do. The new `media_dir` setting also needs to be
   exposed in the OneBot Console type/form and documentation, or omitted in
   favor of the standard workspace media directory.

5. **There is still synchronous filesystem I/O in the event loop.**
   `_local_media_path()` calls `Path.is_file()` and `Path.resolve()` directly
   from the async receive path. These calls may block on Windows UNC paths,
   NAS mounts, macOS network volumes, or removable storage. Please move the
   complete local-path probe to `asyncio.to_thread()`. Generated filenames
   should also be length-bounded for common Windows/macOS/Linux filesystem
   component limits.

6. **Media downloads can reorder messages from the same conversation.** Every
   WebSocket event runs in an independent task before localization and
   enqueueing, so a later text message can overtake an earlier media message.
   Please preserve ordering per session without blocking the WebSocket reader,
   since it must continue receiving OneBot API echo responses. Access control
   should also reject blocked/pending senders before quoted-message lookups and
   media downloads, rather than after the payload has already been enqueued.

7. **The old image-inlining implementation is now dead production code.**
   `_resolve_inbound_media()` uses `_download_remote_media()`, while
   `_inline_remote_images()` and `_download_image_data_url()` are only called
   by their tests. Please remove the superseded path and move any relevant
   redirect/content-type coverage to the active downloader.

OneBot inbound events normally provide a URL, a cache/local `file` reference,
or a `file_id` resolved through a OneBot API. `base64://...` is primarily an
outbound transport representation; the new inbound resolver does not directly
decode Base64 either. The outbound Base64 policy should therefore stay fully
separate from inbound reception.

The PR title and description also need to be refreshed. The current title only
mentions voice and image, while the branch now handles image, audio, video, and
file content. I suggest:

`feat(onebot): localize inbound media before agent processing`

The description still documents the older design where generic
`message_processing.py` downloaded remote audio, claims a four-file/533-line
diff and 102 passing tests, and says the scope is limited to audio and image.
The current comparison contains six changed files with 1,027 insertions and
135 deletions. Please rewrite the Summary, Changes, Scope, test evidence, and
commit list after the issues above are fixed.

hongxicheng:
## Overall review

**Request changes**

The async HTTP download path looks reasonable, and moving the file write to
`asyncio.to_thread()` is a good choice. Before merging, please address the
comments below. The main concerns are workspace isolation, media resolution
correctness, event-loop safety for local filesystem operations, and completing
the `media_dir` integration across the built-in channel surfaces.

## Inline comments

### [P1] Forward `workspace_dir` in `from_config`

**File:** `src/qwenpaw/app/channels/onebot/channel.py:416-454`

`ChannelManager` passes the active agent workspace to channel factories, but
`OneBotChannel.from_config()` neither accepts nor forwards `workspace_dir`.
The default therefore falls back to the global media directory instead of the
current workspace. With multiple agents, inbound media can be written to a
shared directory and become mixed between workspaces. Please add
`workspace_dir` to this factory and pass it to the constructor, with a test
covering the `from_config(..., workspace_dir=...)` path.

### [P1] Separate the inbound download limit from the Base64 limit

**File:** `src/qwenpaw/app/channels/onebot/channel.py:1453-1474`

All inbound HTTP downloads are capped by `media_base64_max_mb`, even when
`media_base64` is disabled. The setting name and Console text describe the
outbound Base64 behavior, so this silently limits ordinary inbound media as
well. Please introduce a separate inbound-download limit, or explicitly
rename/document the setting and update the schema, UI, environment variable,
documentation, and tests to reflect the broader semantics.

### [P2] Keep local filesystem probes off the event loop

**File:** `src/qwenpaw/app/channels/onebot/channel.py:1280-1298`

`_local_media_path()` runs `Path.is_file()` and `Path.resolve()` directly from
the async message path. These calls can block on Windows network/UNC paths or
slow mounted volumes on macOS. Please run the complete local-path conversion
and existence check in `asyncio.to_thread()` so a slow filesystem cannot delay
the WebSocket reader and other incoming messages.

### [P2] Preserve the source segment when resolving media

**Files:** `src/qwenpaw/app/channels/onebot/channel.py:1195-1203` and
`1308-1337`

The code filters media segments and then pairs them with content parts using a
second positional counter. `_parse_message_segments()` can skip a media
segment when it has no URL/file (for example, a segment that only contains a
`file_id`). In a mixed message this shifts the mapping, so a later content part
may receive the wrong `file_id`, filename, or media type. Please preserve the
original segment index/source object during parsing and resolve by identity.
Add a regression test with an omitted media segment before another media
segment.

### [P2] Distinguish RIFF subtypes before choosing the suffix

**File:** `src/qwenpaw/app/channels/onebot/channel.py:1224-1241`

Every payload beginning with `RIFF` is currently saved as `.wav`, but RIFF is
also used by WebP and AVI. A WebP image or AVI video can therefore be written
with the wrong extension and downstream MIME handling may fail. Please inspect
the container subtype at bytes `data[8:12]` (`WAVE`, `WEBP`, `AVI `, etc.) or
use a deterministic MIME-aware mapping before selecting the suffix.

### [P2] Complete the `media_dir` configuration surface

**Files:** `src/qwenpaw/config/config.py:295-309`,
`console/src/api/types/channel.ts:180-187`,
`console/src/pages/Control/Channels/components/ChannelDrawer.tsx:1328-1418`,
`website/public/docs/channels.en.md:539-547`, and the corresponding Chinese
section.

The backend now accepts `OneBotConfig.media_dir`, but the Console type, the
OneBot form, and both OneBot documentation tables still omit it. Users cannot
discover or set the directory through the main configuration surface. Please
add the field, localized label/help text, config example, and documentation
entry across all layers, or keep the option internal until that integration is
complete.

### [P2] Remove or reconnect the unused Base64 image helpers

**File:** `src/qwenpaw/app/channels/onebot/channel.py:1524-1626`

There is no production caller for `_inline_remote_images()` or
`_download_image_data_url()`; the current inbound path only calls
`_resolve_inbound_media()`. This leaves a second, overlapping download
pipeline with different behavior and makes it unclear whether Base64 mode is
still intended to be active here. Please remove these helpers, or reconnect
them with a clear inbound/outbound boundary and an end-to-end test.

### [P3] Remove unrelated wording-only changes

**File:** `src/qwenpaw/agents/utils/message_processing.py:316-339, 493-510`

Please keep documentation/comment changes limited to the behavior changed by
this PR. The wording-only updates in this shared utility make the diff noisier
and obscure the actual OneBot change. If the new wording is required for the
audio behavior, please keep only the minimal relevant comment; otherwise,
restore the existing wording.

## General review notes

- Remote I/O is asynchronous, and the actual media write is already moved to
  `asyncio.to_thread()`. That part does not appear to block the event loop.
- `file_url_to_local_path()` already contains Windows drive and UNC handling,
  but the OneBot path should be tested with spaces, Unicode, drive-letter paths,
  and UNC paths. The synchronous existence/resolve check still needs to move to
  a worker thread.
- `mimetypes.guess_extension()` can produce different results across Windows,
  macOS, and Linux. Prefer a deterministic mapping for supported media types,
  then fall back to magic bytes.
- `trust_env=False` disables environment-configured HTTP proxies. This may
  break downloads in corporate Windows/macOS environments. Please document
  this choice or make proxy handling consistent with the other channels.
- Failure placeholders such as `[image: download failed]` are consistent with
  Feishu, WeCom, and QQ. Please also cover quoted messages and mixed
  text/media messages, not only one-media-per-kind cases.
- `onebot/channel.py` now contains WebSocket lifecycle, OneBot API calls,
  segment parsing, media URL resolution, MIME detection, filename sanitizing,
  storage, and content conversion. The current change can remain focused, but
  a follow-up split into `media.py`, `inbound.py`, and `channel.py` would make
  the design easier to extend and compare with other channels.

## Suggested overall PR comment

> Request changes. The async HTTP part and the `asyncio.to_thread()` file write
> are good, but there are still correctness/integration issues: `workspace_dir`
> is not forwarded by `from_config`, the Base64 size setting is reused for all
> inbound downloads, local filesystem probes run in the event-loop task, and
> media segments are matched by a filtered positional counter. Please also fix
> RIFF subtype detection, complete `media_dir` in Console/i18n/docs, remove or
> reconnect the unused Base64 image helpers, and clean up unrelated wording-only
> changes before merge.

GMsure:
Implemented the remaining review items and synchronized this branch with the latest `upstream/main`.

Review closure:

- `workspace_dir` is accepted and forwarded by `from_config()`, with coverage for the workspace default.
- Inbound downloads use independent `media_download_max_mb`; outbound `media_base64_max_mb` remains outbound-only.
- Downloads are concurrency-bounded, streamed to temp files, atomically finalized, and reuse a channel-owned session.
- Local filesystem probes and writes/finalization run through `asyncio.to_thread()`; filenames are length-bounded.
- Media parts now match their original source segment by kind + reference, retaining the original segment index; a skipped-segment regression test covers the prior positional drift.
- RIFF subtypes distinguish WAVE, WebP, and AVI; known MIME suffixes are deterministic, and mismatched image/audio/video content types are rejected.
- Message events are ordered per session without blocking the WebSocket reader; workers are reclaimed after idle timeout and no new worker is created while stopping.
- Mention/ACL gates run before `get_msg`, media API resolution, or downloads.
- `media_dir` and `media_download_max_mb` are exposed in the Console type/form, all seven locales, config examples, and EN/ZH docs.
- The old Base64 image downloader is removed.
- Shared `message_processing.py` changes are reduced to the required audio fallback behavior only; wording-only changes were restored.
- `trust_env=True` keeps media downloads consistent with environment-configured proxies.

Validation on the final branch:

- 158 focused tests passed, including channel shape consistency.
- All changed Python files pass AST, mypy, Black, Flake8, and Pylint.
- Console TypeScript compilation passes.
- Changed Console and website files pass the current project Prettier checks.
- Independent final code review found no merge-blocking issue.
- Branch is 0 commits behind current `upstream/main`.

hongxicheng:
Thank you for updating the implementation based on my previous review comments.

After reviewing the latest version, I felt that a few additional adjustments were still needed. I have completed these changes and pushed them directly to the current PR branch:

1. The channel module is currently being refactored, and the position where ACL checks are performed may change in the future. For now, I removed the OneBot-specific early ACL check so that its behavior remains consistent with the other channels.
2. I simplified and reorganized parts of the implementation to improve its structure and long-term maintainability.
3. I adjusted some configuration wording to align it with the conventions used by the other channels.
4. I removed several redundant and unrelated changes from the diff to keep the PR focused.

If convenient, you are very welcome to pull the updated branch and test it locally. Your feedback on the revised implementation would be highly valuable. Thanks again for your contribution!

lalaliat (approved):
LGTM

PR #6723, opened 2026-08-05
fix(provider): expire stale capability cache entries and clear on model switch

ningblue:
## What Problem This Solves

The learned-capability cache (`rejects_media`, `needs_reasoning_content`) never expired and was never cleared. A single transient upstream failure — e.g. an API gateway routing a multimodal model (`kimi2.6`) to a non-multimodal backend (`dsv4`) for a few minutes — permanently poisoned the cache. Every subsequent request then silently stripped user-uploaded images (`Proactively stripped 1 media block(s)`), and the model responded "I don't see any image" until the qwenpaw process was restarted.

Observed in production: after a gateway misroute returned `400 dsv4 is not a multimodal model`, qwenpaw called `get_capability_cache().learn(model_key, "rejects_media", True)`. From that point on, `_reasoning()` hit the proactive-strip branch on every turn — even after the gateway was fixed and the model could accept images again. The only recovery was deleting the pod.

Root cause: `ModelCapabilityCache` has a `clear()` method but nothing in the codebase ever calls it, and entries have no TTL.

## Fix

**1. Cache entries now expire after a TTL** — each entry records a monotonic timestamp; `get()` evicts entries older than `CAPABILITY_CACHE_TTL_SECONDS` (default 30 min, `0` disables expiry).

**2. `activate_model()` clears the cache** — switching or re-selecting a model discards findings that no longer apply.

## Evidence

Before the fix, a poisoned cache strips images permanently:

```
# gateway returned 400 once (transient misroute)
Learned capability for yunnet-kimi:kimi2.6: rejects_media=True

# every subsequent request, even after gateway was fixed:
WARNING | Proactively stripped 1 media block(s) before _reasoning (model lacks multimodal support).
# model response: "我没有看到图片"
```

After the fix, the same scenario self-heals:

```
# cache entry expires after TTL → get() returns default (False)
# → should_strip is False → image reaches the model
# model response: "红色"  (correctly identified the uploaded image)
```

Switching the model via `activate_model()` also clears the cache immediately, so an operator can recover without waiting for TTL or restarting the pod.

## Config

The TTL constant follows the existing `constant.py` convention (`_get_env` with `COPAW_` legacy fallback), consistent with `TOOL_GUARD_APPROVAL_TIMEOUT_SECONDS`.

| Env | Default | Effect |
|---|---|---|
| `QWENPAW_CAPABILITY_CACHE_TTL_SECONDS` | `1800` (30 min) | `0` disables expiry |

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Core / Backend (providers)

## Checklist

- [x] Ready for review

### DIFF
--- a/src/qwenpaw/providers/provider_manager.py
+++ b/src/qwenpaw/providers/provider_manager.py
@@ -2501,9 +2501,11 @@ class ProviderManager:  # pylint: disable=too-many-public-methods
                 if "extra_models" in saved_config:
                     provider_info.extra_models = [
                         ModelInfo.model_validate(
-                            model.model_dump()
-                            if isinstance(model, BaseModel)
-                            else model,
+                            (
+                                model.model_dump()
+                                if isinstance(model, BaseModel)
+                                else model
+                            ),
                         )
                         for model in saved_config["extra_models"]
                     ]
--- a/src/qwenpaw/providers/provider_manager.py
+++ b/src/qwenpaw/providers/provider_manager.py
@@ -1672,6 +1672,12 @@ class ProviderManager:  # pylint: disable=too-many-public-methods
         )
         self.save_active_model(self.active_model)
 
+        # Stale capability findings (e.g. rejects_media from a transient
+        # upstream failure) no longer apply after a model switch.
+        from .model_capability_cache import get_capability_cache
+
+        get_capability_cache().clear()
+
         self.maybe_probe_multimodal(provider_id, model_id)
 
     def maybe_probe_multimodal(self, provider_id: str, model_id: str) -> None:
--- a/src/qwenpaw/providers/provider_manager.py
+++ b/src/qwenpaw/providers/provider_manager.py
@@ -1672,11 +1672,16 @@ class ProviderManager:  # pylint: disable=too-many-public-methods
         )
         self.save_active_model(self.active_model)
 
-        # Stale capability findings (e.g. rejects_media from a transient
-        # upstream failure) no longer apply after a model switch.
+        # Drop a stale ``rejects_media`` entry for this model so that a
+        # transient upstream failure (e.g. a gateway misroute) does not
+        # keep stripping images after re-selection.  Other capabilities
+        # and other models are left untouched.
         from .model_capability_cache import get_capability_cache
 
-        get_capability_cache().clear()
+        get_capability_cache().forget(
+            f"{provider_id}:{model_id}",
+            "rejects_media",
+        )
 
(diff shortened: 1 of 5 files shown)


ningblue:
请关注，希望可以优化这个功能，因为这个对于企业环境是需要的。感谢

src/qwenpaw/providers/provider_manager.py
```diff
@@ -1672,6 +1672,12 @@ async def activate_model(self, provider_id: str, model_id: str):
         )
         self.save_active_model(self.active_model)
 
+        # Stale capability findings (e.g. rejects_media from a transient
+        # upstream failure) no longer apply after a model switch.
+        from .model_capability_cache import get_capability_cache
+
+        get_capability_cache().clear()
```
qbc2016: Please avoid clearing the entire process-wide capability cache here. Entries are already scoped by `provider_id:model_name`, and other agents may still be using different models, so a global clear makes unrelated models repeat previously learned failures.

If model re-selection is intended to provide immediate recovery, please invalidate only the selected model's transient `rejects_media` entry, while preserving `needs_reasoning_content` and entries for other models. The same targeted invalidation should be used by any agent-scoped model-selection path where this recovery behavior is expected.

Alternatively, removing switch-triggered invalidation and relying solely on the TTL would also avoid the cross-model regression. Please add a regression test confirming that activating one model does not evict capabilities learned for another model.

qbc2016 (commented):

ningblue:
Thanks for the review! 

Updated in `8378d7f2`:

- Added `ModelCapabilityCache.forget(model_key, capability)` that drops a single entry while preserving other capabilities (e.g. `needs_reasoning_content`) and all entries for other models.
- `activate_model()` now calls `forget("{provider_id}:{model_id}", "rejects_media")` instead of the global `clear()`.
- Added regression tests confirming that activating one model does not evict capabilities learned for another model (`test_activate_model_preserves_other_models_and_capabilities`).

The TTL mechanism is kept as-is for the self-healing case (transient failures auto-recover without any model switch).

qbc2016:
Please fix the unit tests and run `pre-commit run --all-files` to format the code.

ningblue:
Pushed a fix for the failing test and applied formatting (`7de9bd5d`)

qbc2016:
Please run `pip install pre-commit; pre-commit install; pre-commit run --all-files` to format the code.

qbc2016 (approved):
LGTM

PR #6743, opened 2026-08-06
fix(ci): export nightly coverage artifacts even when tests fail

yutai78786:
> **提交人**: 张衡·Appraiser@QPQAT

## Summary

The nightly workflow must export coverage even when tests fail. Today the `integrated-tests` coverage export/upload is chained to pytest exiting 0: on any night with ≥1 failing integration case, `bash -e` aborts the run block at the failing `pytest` line, and the upload step is then skipped (no `if: always()`), so the coverage artifact is silently dropped on the runner.

Result: **9 consecutive nights (2026-07-28 → 2026-08-05) with no integration coverage data** — the tests genuinely ran every night and runtime coverage was genuinely collected, yet nothing reached the artifacts.

## Root cause (evidence chain)

Nightly runs (`full-tests-nightly`) and their integrated-tests results, checked run by run:

| Date | run id | integrated-tests result | integration artifacts | `INTEGRATION` in Coverage Report log |
|---|---|---|---|---|
| 07-27 | 30295442263 | 4/4 success | all present | 36 |
| 07-28 | 30389112849 | 4/4 failure | none | n/a |
| 07-29 | 30480742681 | 3 fail / windows success | windows only | 0 (windows data has zero hits) |
| 07-30 | 30572037154 | 4/4 failure | none | n/a |
| 07-31 | 30656555365 | 3 fail / windows success | windows only | 0 |
| 08-01 | 30712925059 | 4/4 failure | none | n/a |
| 08-02 | 30761526176 | 4/4 failure | none | n/a |
| 08-03 | 30843414044 | 3 fail / windows success | windows only | 0 |
| 08-04 | 30940502146 | 4/4 failure | none | n/a |

Key evidence from the failing ubuntu job log of 07-29 (run 30480742681):

- `== 1 failed, 442 passed ==` (only `test_import_local_copies_source` failed)
- `[integration coverage] HTML report: /home/runner/.../htmlcov-integration/index.html` → runtime coverage **was** collected and combined by the app subprocesses
- the log ends right after the pytest summary — no `cp .integration_coverage/integration_subproc`, no `coverage xml`, no upload record → the export was aborted by `bash -e` and the upload step was skipped
- contrast: the succeeding windows job of the same run shows `Artifact coverage-data-integration-windows-latest-py3.11 ... successfully uploaded!`

Structural gap vs e2e: in `_e2e-job.yml`, both `Stop backend and collect coverage` and `Upload E2E coverage data` carry `if: always()`, so e2e coverage survives failures/cancellation. `integrated-tests` was missing exactly that.

## Changes

1. `integrated-tests` → "Run integration suite": capture the pytest exit code (`set +e` / `PYTEST_RC` / `set -e`), export coverage anyway (`cp` + `coverage xml`, tolerating the fail-under exit code 2 as before), then `exit "$PYTEST_RC"` — **a failing suite still fails the step**. Emits a `::warning` if the subprocess data file is missing despite coverage being enabled.
2. `integrated-tests` → "Upload integration coverage data": `if: always() && steps.cov_flag.outputs.enabled == '1'`.
3. `unit-tests` / `contract-tests` coverage upload steps: add `always()` to the existing matrix gate. Same latent defect (upload gated on the test step succeeding); it has not fired only because those suites are currently green.

Not changed: failure semantics — any failing suite still turns the run red. The `coverage-report` job already runs under `if: always()` and needs no change.

## Verification

After merge: trigger `workflow_dispatch` (or wait for the next nightly). On a run where integration cases fail, the Coverage Report log should show the integration data files being combined and `INTEGRATION=<non-zero>`, and `coverage-data-integration-*` artifacts should exist for the failed matrix entries.

Known remaining issue (separate track): windows subprocess coverage data currently combines to a zero-hit xml (verified from the 07-29/07-31/08-03 windows artifacts: 668 files / 91921 lines / 0 lines hit). With this PR, ubuntu/macos data backs `INTEGRATION`; the windows zero-hit needs its own investigation.

### DIFF
--- a/.github/workflows/full-tests-nightly.yml
+++ b/.github/workflows/full-tests-nightly.yml
@@ -108,6 +108,7 @@ jobs:
 
       - name: Upload unit coverage data
         if: |
+          always() &&
           matrix.os == 'ubuntu-latest' &&
           matrix.python-version == '3.13'
         uses: actions/upload-artifact@v4
@@ -187,6 +188,7 @@ jobs:
 
       - name: Upload contract coverage data
         if: |
+          always() &&
           matrix.os == 'ubuntu-latest' &&
           matrix.python-version == '3.13'
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


cuiyuebing (approved):
LGTM

PR #6750, opened 2026-08-06
fix(chat): session identity deadlock, early session save, oversized prompt collapse

lllyfff:
## Description

Fixes three confirmed front-end session-interaction issues (all root-caused in source before this change):

**A1 — Session identity desync: messages queued but never sent (Bug)**
- `chatLoading` is driven solely by the runtime SDK with no error-path reset; once stuck, Enter enqueues forever and nothing is sent (`index.tsx` queue guard).
- `getBackendSessionId` silently fell back to the raw library id, leaking stale chat UUIDs / unresolved local timestamp ids into the tool-call APIs. The backend validates `session_id` and returns 404 on mismatch (2200 `session_id mismatch` warnings observed in one session log).

Changes:
- `sessionApi`: new `getBackendSessionIdStrict` / `isKnownBackendSessionId` — return `""` instead of an unconfirmed id (the lenient variant is kept only for POST paths the backend ignores).
- `resolveBackendSessionId`: every branch now resolves strictly; callers already retry with backoff.
- `customFetch`: force-reset SDK loading via `RuntimeLoadingBridge` on non-2xx and network failures.
- `scheduleNextSend`: watchdog verifies backend chat status when `chatLoading` is stuck — resets and proceeds when the backend is idle, re-arms when genuinely running.
- Extracted `queryChatStatus`/`waitForChatIdle` into `pages/Chat/chatStatus.ts` (unit-testable).

**A2 — Session view lost on mid-turn refresh (robustness)**
The session file was only written at `POST_RESPONSE` (or cancel/error), so refreshing mid-turn showed the previous turn's state — the just-sent user message disappeared (it only survived same-tab via the sessionStorage patch). Added `SessionEarlySaveHook` (`PRE_EXECUTE`, priority 95, shared `_do_session_save` helper) that persists the final pre-execution state — including the current turn's user message — right before execution begins. The canonical save overwrites it later; `SESSION_SAVE_SUCCEEDED_KEY` remains reserved for the `POST_RESPONSE` save.

**A3 — ~25K mission controller prompt rendered as a plain user bubble (UX)**
`/mission` replaces the user message with the full controller prompt (`_rewrite_user_msg`), which is persisted and rendered verbatim on reload, drowning the original task text. Added `CollapsibleRequestCard`: user bubbles whose combined text exceeds 2000 chars collapse behind a 300-char summary with an expand toggle (vendor card lazy-mounted on expand). Length-based on text parts only, so attachment-only and short messages pass through untouched, and already-persisted histories need no metadata marker.

**Related Issue:** tracked in internal issue document (no upstream issue number)

**Security Considerations:** The `tool_calls.py` 404-on-mismatch behavior is an intentional cross-session protection and is deliberately unchanged; this PR stops invalid ids from being sent instead. The early session save reuses the existing atomic write + per-path lock. No auth, env, or config handling changes.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes (run on all changed files; see Evidence)
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks (no auto-fixes were needed)
- [x] Documentation updated (if needed) (only code comments/docstrings; no user-facing docs affected)
- [x] Ready for review

Not applicable — no channel runtime/contract behavior is changed (the console router is untouched); changes are in the Console web UI, session lifecycle hooks, and hook registration.

## Testing

Automated (all added in this PR):
- console: `npx vitest run src/utils/resolveBackendSessionId.test.ts src/pages/Chat/tests/strictSessionId.test.ts src/pages/Chat/tests/chatStatus.test.ts src/pages/Chat/components/CollapsibleRequestCard.test.tsx` — 32 tests covering strict id resolution, status probing, and collapse behavior.
- backend: `pytest tests/unit/agents/hooks/test_session_hook.py` — 8 tests incl. 5 new for `SessionEarlySaveHook` (phase/priority, ephemeral skip, success-key isolation, error swallowing).

Manual:
1. **A1**: start chat, stop the backend mid-conversation and send a message → loading clears and the next message sends; Ctrl+Enter ×3 to queue → all send in order; DevTools Network shows no more tool-calls 404 mismatches.
2. **A2**: send a message and refresh immediately (turn still running) → the user message bubble is still visible.
3. **A3**: run `/mission <task>` → the user bubble shows a ~300-char summary with "Show full message"; expand/collapse works; short and attachment-only messages render normally.

## Evidence

Local test-run transcripts (Windows dev environment):

```bash
$ cd console && npx vitest run src/utils/resolveBackendSessionId.test.ts \
    src/pages/Chat/tests/strictSessionId.test.ts \
    src/pages/Chat/tests/chatStatus.test.ts \
    src/pages/Chat/components/CollapsibleRequestCard.test.tsx
 Test Files  4 passed (4)
      Tests  32 passed (32)

$ cd console && npx tsc -b --noEmit
(exit code 0 — no type errors)

$ cd console && npm run test:run
 Test Files  1 failed | 207 passed (208)
      Tests  1 failed | 1591 passed (1592)
# the single failure is replaySdkIntegration.test.ts, which fails identically
# on the clean tree (verified via git stash) — pre-existing, unrelated.

$ pytest tests/unit/agents/hooks/test_session_hook.py -v
tests/unit/agents/hooks/test_session_hook.py::test_ephemeral_request_skips_session_load_and_save PASSED
tests/unit/agents/hooks/test_session_hook.py::test_normal_request_loads_and_saves_session_state PASSED
tests/unit/agents/hooks/test_session_hook.py::test_failed_session_save_does_not_mark_turn_as_persisted PASSED
tests/unit/agents/hooks/test_session_hook.py::test_early_save_hook_runs_last_in_pre_execute PASSED
tests/unit/agents/hooks/test_session_hook.py::test_early_save_hook_saves_state_at_turn_start PASSED
tests/unit/agents/hooks/test_session_hook.py::test_early_save_hook_skips_ephemeral_request PASSED
tests/unit/agents/hooks/test_session_hook.py::test_early_save_hook_does_not_publish_success_key PASSED
tests/unit/agents/hooks/test_session_hook.py::test_early_save_hook_swallows_save_errors PASSED
8 passed in 0.71s

$ pytest tests/unit -q
8 failed, 6209 passed, 69 skipped in 1612.43s
# all 8 failures are pre-existing environment issues (Windows symlink/path
# handling in checkpoints/cli/governance/skill_scanner); 5 of them were
# re-verified to fail identically on the clean tree via git stash.

$ python -m pre_commit run --files <all 19 changed files>
# yaml/xml/toml/prettier hooks: no matching files -> Skipped
# actionlint skipped: Go module proxy unreachable in this env; no workflow files touched
```

## Additional Notes

- Three commits map 1:1 to the issues: `fix(chat):` (A1), `fix(session):` (A2), `feat(console):` (A3).
- Branch was merged with upstream `origin/fix/frontend` before pushing (clean merge, no conflicts; only locale-file additions overlapped and auto-merged — all 7 JSONs validated).
- `POST /console/chat` ignores the request-body `session_id` (channel resolves it from the sender), so the lenient id fallback was kept only on non-validating paths; the strict resolvers are used for every validating endpoint (tool-call APIs).


RerankerGuo:
I found a queue-ordering issue in the loading watchdog at exact head `d8d77b92c3f101c5636b84c1d5ce80baa57e44ab`.

When `chatLoading` is still true and the one-shot status request has a transient network failure, `queryChatStatus()` returns `null`. `scheduleNextSend()` currently treats every value other than `"running"` as proof that the backend is idle, calls `setLoading(false)`, and submits the next queued message. A temporary inability to determine status can therefore send the next turn while the previous turn may still be running.

I reproduced this with a component-level Vitest harness around `ChatPage`:

- initialize the runtime loading bridge with `loading = true`
- enqueue `{ text: "queued" }` for the active session
- make `GET /chats/{id}` reject with a network error
- wait for the exact watchdog status request
- assert that the queue item is retained and `input.submit` is not called

Command:

```bash
npm run test:run -- src/pages/Chat/ChatPage.test.tsx \
  -t "does not send a queued message when the busy-chat status is unknown"
```

Result on this head:

```text
FAIL does not send a queued message when the busy-chat status is unknown
expected mockSubmit not to be called
received 1 call: { query: "queued", fileList: undefined }
```

Environment: macOS, Node `v24.7.0`, Vitest `4.1.6`. The existing `ChatPage.test.tsx` exclusion was temporarily lifted and its stale mocks were completed so the real page state machine could run; all harness/config edits were reverted after validation.

A minimal behavior change that distinguishes unknown from idle fixes the same test:

```ts
if (status === "running" || status === null) {
  setTimeout(() => scheduleNextSend(), 3000);
  return;
}
```

With that change, the exact test passes and the queued item remains pending. Only an explicit non-running backend status should clear the SDK loading state and permit the next send.

lllyfff:
@RerankerGuo fixed in `38be322c` (pushed to `fix/frontend`).

  Root cause was conflating 404 with network failure: both returned `null`, so the watchdog treated an undetermined
  status as idle and cleared loading. The probe now returns a tri-state — `"running"` / `"idle"` / `"unknown"` — and
  `shouldResetStuckLoading(status)` is true only for a confirmed `"idle"`:

  ```ts
  if (!shouldResetStuckLoading(status)) {
    setTimeout(() => scheduleNextSend(), 3000);
    return;
  }

  Network failures and 5xx now map to "unknown" → the watchdog re-arms instead of sending, so an out-of-order send can't
  happen. I deliberately kept 404 as "idle" (a non-existent chat can't be running; treating it as unknown would break
  the stale-id recovery path).

  Verified: shouldResetStuckLoading("unknown") === false regression test + new tri-state probe tests in
  chatStatus.test.ts; 36/36 pass, tsc clean. (The ChatPage harness stays out of CI since ChatPage.test.tsx is excluded
  in vitest — the same decision path is unit-covered.)

RerankerGuo:
Revalidated exact head `38be322c2359fd0cd6e7ecacbe8580c85fb7cc4b`.

The tri-state probe fixes the reported foreground watchdog path:

```text
npm run test:run -- src/pages/Chat/tests/chatStatus.test.ts
14 passed

npx tsc --noEmit
passed
```

There is still a second queue sender that collapses `"unknown"` into idle. `startBackgroundQueue()` calls `waitForChatIdle()` immediately before its direct chat POST:

```ts
const idle = await waitForChatIdle(...);
if (!idle) break;
// set sending state, then fetch("/api/chats/...", { method: "POST", ... })
```

At this head, `waitForChatIdle()` retains the old condition:

```ts
if (status !== "running") return true;
```

Consequently, a transient network failure maps to `"unknown"`, returns `true`, and releases the background worker to send the queued item without proof that the prior turn is idle. This is the same ordering invariant through a different production caller; `shouldResetStuckLoading()` is not used on this path.

I added a bounded regression around the real helper:

```ts
const waiting = waitForChatIdle("chat-1", ctrl.signal);
const outcome = await Promise.race([
  waiting.then(() => "released"),
  new Promise<string>((resolve) =>
    setTimeout(() => resolve("still-waiting"), 20),
  ),
]);
expect(outcome).toBe("still-waiting");
```

Exact-head result:

```text
FAIL does not release the background queue while status is unknown
Expected: "still-waiting"
Received: "released"
```

Changing the helper to release only on confirmed idle makes this new regression pass:

```ts
if (status === "idle") return true;
```

The existing test named `returns true when the status probe fails (does not block the queue)` then needs its contract updated; that assertion encodes the unsafe behavior. Avoiding a permanent wedge does not require sending on unknown: the worker can retain the item and retry with the existing one-second backoff until status is confirmed, the request succeeds again, or the AbortSignal stops the worker.

The foreground finding is fixed, but queue ordering is not yet enforced across both foreground and background send paths at this head. All temporary source/test changes were reverted after the red/green check.

RerankerGuo:
Revalidated the background-queue correction independently at exact head `b05e36b9e4706f73ed5ca94090b03df1abd63946` on macOS.

The second send path now has the same safe tri-state contract as the foreground watchdog: `waitForChatIdle()` releases only on confirmed `"idle"`, keeps polling on `"unknown"`, and still releases a 404 because `queryChatStatus()` maps that case to idle.

```text
cd console
npm exec vitest -- run src/pages/Chat/tests/chatStatus.test.ts
1 file passed, 15 tests passed

npm exec vitest -- run src/pages/Chat/tests
9 files passed, 91 tests passed

npx tsc --noEmit
passed

npx prettier --check \
  src/pages/Chat/chatStatus.ts \
  src/pages/Chat/tests/chatStatus.test.ts
All matched files use Prettier code style

git diff --check 38be322c..b05e36b9
passed
```

The changed-file pre-commit checks also pass with `actionlint` skipped because no workflow file changed. A separate isolated-`HOME` attempt could not install `actionlint` because the host's system Go is older than that tool's dependencies; this is unrelated to the TypeScript changes and all applicable checks above are green.

The foreground and background unknown-status queue-ordering findings are both resolved at this head.

RerankerGuo (changes requested):
The queue-ordering fixes are resolved at `b05e36b9`, and the current-main merge is green in the exercised Console scope. However, A2 (mid-turn refresh persistence) does not persist the current user message.

`Runtime.run()` executes every `PRE_EXECUTE` hook before it calls `AgentExecutor.run(ctx.input_msgs)`. AgentScope only appends those inputs to `agent.state.context` inside `Agent._handle_incoming_messages()`, which is entered by `reply_stream()` after the hook has returned. Therefore `SessionEarlySaveHook` snapshots the previous turn's state, not the current input it claims to make durable.

The new test at `tests/unit/agents/hooks/test_session_hook.py:110-118` does not cover the acceptance boundary; it explicitly asserts that the early-saved `context` is empty.

```mermaid
sequenceDiagram
    participant Runtime
    participant EarlySave as SessionEarlySaveHook
    participant Session
    participant Executor as AgentExecutor / AgentScope
    Runtime->>EarlySave: PRE_EXECUTE
    EarlySave->>Session: save agent.state.context
    Note over Session: Current user message is absent
    Runtime->>Executor: run(ctx.input_msgs)
    Executor->>Executor: _handle_incoming_messages appends current turn
    Note over Executor: Too late for early snapshot
```

I reproduced this against the actual PR hook and AgentScope input handler on the current-main merge tree `5b066b2a8acf32412b2563b57b2f9a3a166baae2`:

```text
pre_execute_snapshot= []
after_executor_input= ['CURRENT TURN']
```

Fresh merge-tree checks otherwise pass:

```text
session hooks: 8 passed
four focused Console files: 37 passed
complete Chat tests: 9 files / 91 tests passed
tsc -b --noEmit: passed
```

Please move or redesign the early-save boundary so the persisted snapshot includes `ctx.input_msgs` after the same normalization/rewrite that execution consumes, without appending the input twice to the live agent. Add a production-path regression that pauses a turn before model completion, reloads the persisted session, and asserts the current user message is present. The existing empty-context assertion should not be used as proof of A2.

lllyfff:
@RerankerGuo Fixes verified in 59de6f0c — thanks for the sharp diagnosis.

  You were right: PRE_EXECUTE runs before executor.run(ctx.input_msgs), and AgentScope only commits inputs to the
  context inside reply_stream() — so the old early save snapshotted the previous turn.

  Fix: SessionEarlySaveHook now projects ctx.input_msgs into a deep copy of the serialized state
  (Msg.model_dump(mode="json"), identical to state_dict()'s serialization) instead of mutating the live agent — no
  double-append, POST_RESPONSE save untouched.

  Tests as requested:
  - Removed the empty-context assertion as proof; replaced with a test asserting CURRENT TURN is present in the
  PRE_EXECUTE snapshot.
  - Added test_early_save_survives_mid_turn_refresh: real Runtime + SafeJSONSession + AgentScope input handling, pauses
  the turn pre-completion, reloads the persisted session, asserts the current user message is there.
  - Pinned no-double-append (live_context == [] after the hook).

  Result: all 12 tests in test_session_hook.py pass.

PR #6754, opened 2026-08-06
feat(acp): support portable runtime sessions

rayrayraykk:
## Summary

- add an ephemeral OpenAI-compatible provider for headless ACP runtimes
- use the ACP session `cwd` as the runtime project context
- connect client-provided stdio, HTTP, and SSE MCP servers with session-scoped Driver lifecycles
- isolate transient Driver discovery and invocation by request scope
- expose typed tagged tool calls and complete ACP tool results
- make cancellation wait for active prompts before session resources are released

## Why this belongs in core

These are protocol and runtime capabilities that apply to any ACP client, not one frontend or launcher.

The ACP session setup contract defines `cwd` as the primary filesystem context and requires it to be used independently of the directory where the Agent subprocess was spawned. The same contract allows clients to provide MCP servers on session creation, load, and resume. Stdio MCP is required for all Agents, while HTTP and SSE support is advertised through `mcpCapabilities`.

Other ACP implementations follow the same architecture:

- Gemini CLI creates per-session configuration from `cwd`, merges client-provided MCP servers, and binds stdio MCP processes to the session directory.
- Codex ACP keeps the working directory, MCP servers, model, and active prompt lifecycle in session state.
- GitHub Copilot CLI documents the working directory and MCP servers as core `session/new` inputs and supports process-level BYOK configuration for headless ACP runtimes.

This change makes QwenPaw behave consistently with those clients while reusing the existing Provider, Workspace, Driver, policy, and approval infrastructure.

References:

- https://agentclientprotocol.com/protocol/v1/session-setup
- https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/acp/acpSessionManager.ts
- https://github.com/agentclientprotocol/codex-acp/blob/main/src/CodexAcpServer.ts
- https://docs.github.com/en/copilot/reference/copilot-cli-reference/acp-server

### Ephemeral runtime provider

`qwenpaw acp --runtime-provider openai-env` reads `OPENAI_BASE_URL`, `OPENAI_API_KEY`, and `OPENAI_MODEL`.

The provider exists only in memory for the ACP process:

- credentials are not written to provider configuration or secret storage
- every runtime-backed session receives an explicit model override
- ACP model state exposes only the runtime model, matching actual execution
- the previous process model state is restored during shutdown

### Session project context

The ACP `cwd` becomes the default `project_dir` request context. A controlled explicit project override still takes precedence.

This gives prompts, tools, file operations, and runtime context one session directory without modifying the Agent default.

### Session-scoped MCP

ACP MCP definitions are normalized into transient Driver cards:

- stdio, HTTP, and SSE transports are supported
- transient cards and credentials are never persisted
- every session receives a distinct Driver scope
- persistent Drivers remain available to all sessions
- transient Drivers are discoverable and invokable only from their owning scope
- replacement is build-before-publish and rolls back on initialization failure
- close removes and shuts down the session MCP handlers

The Agent now advertises `mcpCapabilities.http` and `mcpCapabilities.sse`, so compliant clients can use the implemented remote transports.

### Tool-call compatibility

Tagged tool calls recovered from OpenAI-compatible model text are emitted as typed AgentScope `ToolCallBlock` values with stable unique IDs. ACP tool result updates also preserve raw output and map unsuccessful states to `failed`.

## Merge blockers addressed

This version resolves the issues identified during review of the earlier follow-up:

- remote MCP transports are now declared during ACP initialization
- runtime model state matches the forced runtime model and does not advertise unusable alternatives
- transient and persistent Driver names cannot overwrite each other, including disabled persistent cards
- `session/close` waits for the active prompt to stop before MCP resources are removed
- session MCP replacement and metadata updates roll back together on failure
- direct unit and integration coverage has been restored and expanded

## Security

- client-provided MCP tools retain the default `ask` policy
- request scope checks apply to capability listing, direct Driver listing, and invocation
- MCP environment values and HTTP headers remain in memory and are not written to Driver storage
- runtime provider credentials remain in memory and are removed on shutdown
- unexpected prompt failures return a generic ACP error unless local diagnostics are explicitly enabled

## Validation

```text
Focused new runtime/session tests:
41 passed

ACP, Driver, provider compatibility, and Driver integration tests:
247 passed, 2 skipped

pre-commit on all changed Python files:
AST, encoding, trailing commas, mypy, black, flake8, and pylint passed

git diff --check:
passed
```

## Scope

This PR intentionally does not add persistent runtime-provider configuration or a new compatibility layer. It uses the existing Provider and Driver contracts and keeps all client-supplied runtime state ephemeral.

### DIFF
--- a/src/qwenpaw/agents/acp/meta.py
+++ b/src/qwenpaw/agents/acp/meta.py
@@ -5,12 +5,12 @@ This module is intentionally lightweight so CLI code can import constants
 without importing the ACP server implementation.
 """
 
-ACP_CODING_PROJECT_META_KEY = "qwenpaw.coding_project_dir"
+ACP_PROJECT_DIR_META_KEY = "qwenpaw.project_dir"
 ACP_EPHEMERAL_META_KEY = "qwenpaw.ephemeral"
 ACP_APPROVAL_EXPIRES_AT_META_KEY = "qwenpaw.approval_expires_at"
 
 __all__ = [
     "ACP_APPROVAL_EXPIRES_AT_META_KEY",
-    "ACP_CODING_PROJECT_META_KEY",
     "ACP_EPHEMERAL_META_KEY",
+    "ACP_PROJECT_DIR_META_KEY",
 ]
--- a/src/qwenpaw/agents/acp/runtime_provider.py
+++ b/src/qwenpaw/agents/acp/runtime_provider.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+"""Ephemeral model providers for headless ACP runtimes."""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+from typing import Mapping
+from urllib.parse import urlparse
+
+from ...config.config import ModelSlotConfig
+from ...providers.openai_provider import OpenAIProvider
+from ...providers.provider import ModelInfo
+
+RUNTIME_OPENAI_PROVIDER_ID = "runtime-openai"
+
+
+@dataclass(frozen=True)
+class OpenAIRuntimeProviderConfig:
+    """One process-scoped OpenAI-compatible model connection."""
+
+    base_url: str
+    api_key: str
+    model: str
+
+    @classmethod
+    def from_env(
+        cls,
+        environ: Mapping[str, str] | None = None,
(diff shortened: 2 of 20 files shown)


cuiyuebing (approved):
LGTM

PR #6504, opened 2026-07-27
feat: unify project directories and harden file workspace

rayrayraykk:
## Description

This PR unifies project-directory resolution and the file workspace across normal and coding-enabled agent sessions.

### Project directory model

- Decouples the project directory from the Coding tools switch and makes it part of the Agent runtime context.
- Adds a controlled Session-level project-directory override without mutating the Agent default.
- Resolves one effective directory for prompts, tools, file APIs, Git operations, and session execution.
- Keeps the Agent Files page and the Chat workspace as separate tab/workspace owners.
- Refreshes the active Session workspace immediately when the runtime project changes.

### Unified file workspace

- Adds a shared file navigator, preview, editor, diff, upload, download, and workspace watcher contract.
- Keeps project and Agent configuration roots explicit while using one compact directory selector.
- Adds independent Agent-level file tabs and Session-level preview/workspace tabs.
- Adds a Chat header action for opening and closing the workspace.
- Supports editable text/CSV files, media previews, transactional diff actions, and file references.
- Replaces the textarea overlay with a Lexical rich composer whose file and code references are atomic editor nodes while preserving the original plain-text request value.

### UI and interaction

- Aligns the Agent Files workspace and Chat workspace styling, colors, icon sizing, spacing, and responsive behavior.
- Uses a compact recent-project/browser picker with selected-state feedback and removable project-path pills.
- Adds bounded, left/upward popover placement to prevent viewport overflow.
- Adds smoother workspace drawer transitions without changing workspace ownership semantics.
- Keeps all new labels internationalized and uses the existing Lucide icon system.

### Correctness, concurrency, and cross-platform hardening

- Carries ETags through file load, editor tabs, normal saves, and diff keep/undo actions.
- Rejects stale saves when a file is externally modified or deleted, and refreshes the ETag after a successful save.
- Ensures multi-chunk reads use one file version and retry when the file changes between chunks.
- Allows paths returned by the directory API to address existing macOS NFD filenames.
- Detects upload collisions using the target filesystem's case and Unicode-normalization behavior.
- Atomically reserves upload destinations and serializes competing versioned saves.
- Makes legacy project-directory migration a pure field move instead of resolving foreign Windows/UNC paths on the current platform.
- Replaces the backtracking editor-reference regex with a linear parser supporting paths with spaces, UNC paths, and colon-containing filenames.
- Moves newly introduced configuration and filesystem work out of async request loops.

## Why

[...]

PR #6764, opened 2026-08-06
feat(ci): gate main mergeability on tests (required checks ready)

yutai78786:

> **Plain-language summary**: This PR fixes the "no gate on `main`" problem. In short: previously `main` had no required status checks, so code could be merged even when tests were all red — this already happened (#6418 was merged with three failing test jobs). This PR: (1) adds required-check rules to `main` (tests, frontend tests, and code-style checks must all be green before merging); (2) resolves the trap where path filtering causes checks to stay pending forever (docs-only PRs will not be stuck); (3) since the PR tier is the only reliable execution layer (push-triggered runs are never actually approved in practice), the PR stage runs the full integration test suite. Note: merging this PR does **not** automatically enable the gate — a repository admin must manually import the ruleset configuration file.

> **Merging this PR does not enable the gate.** The ruleset must be imported by a repository admin, and there is a prerequisite to clear first — see [Before importing the ruleset](#before-importing-the-ruleset).

## Problem

Two independent gaps mean a red test run does not stop anything today.

**1. `main` has no `required_status_checks`.** Tests, frontend vitest and pre-commit all turn red without blocking a merge. Concretely: #6418 was merged with three failing test jobs.

**2. The push tier never actually runs.** The tiered plan assumed post-merge push runs provide deeper coverage (PR=p0, push=p0+p1, nightly=full). But push-triggered runs wait on the maintainer-approved environment and are never approved in practice — **the last 40 push runs on `main` all expired in `waiting`**. The PR gate is the only layer that reliably executes.

## Changes

**1. Required checks that cannot deadlock a PR.** `tests.yml` and `frontend-tests.yml` drop the `pull_request` `paths` filter and replace it with an in-workflow `changes` job (dorny/paths-filter, same path list). A path-filtered workflow that never triggers leaves its required check pending forever, which blocks the PR permanently — the classic trap. With the filter moved inside, a docs-only PR reports the checks as *skipped*, which satisfies a required check. Push triggers are unchanged, and docs-only PRs also stop pinging the maintainer-approved environment.

**2. Readable check contexts.** `pre-commit.yml`'s bare `run` job is named `Pre-commit`.

**3. Importable ruleset.** `.github/rulesets/main-required-checks.json` requires Test Summary, Pre-commit, Vitest Unit Tests and both npm format checks (GitHub Actions app pinned), with repository admins as bypass actors for emergencies.

**4. PR gate runs the full integration suite** instead of p0 only, compensating for the dead push tier. The `workflow_dispatch` marker input remains as a manual override to narrow a run.

**5. macOS integration HTTP timeout floor.** The 120s floor was applied to Windows only; extended to every non-Linux runner. Nightly evidence: `test_plugins_install_real_via_agentscoped` hit `httpx.ReadTimeout` on macOS 2 nights in 10.

## Fail-closed revision (head 7fa21070)

Addresses the review finding that the required-check design fails open when a path-detection job fails:

- `Test Summary` (tests.yml) and the new `Frontend Summary` (frontend-tests.yml) now run with `if: always()` and make a three-state decision: detection not `success` → fail; docs-only → explicit pass; tests required → every tier must be strictly `success` (`failure`/`cancelled`/`skipped` all rejected).
- Ruleset required check `Vitest Unit Tests` → `Frontend Summary`.
- Verified with 7 canary scenarios against the real Actions state machine on the fork (docs-only, all-green, detection failed, detection cancelled, test failed, test cancelled mid-run, test unexpectedly skipped): only the first two are mergeable; every failure mode is blocked.

## Verification

Verified on a fork via a real `pull_request` event (not `workflow_dispatch`), so the actual gate path was exercised — `changes` job filtering, full integration selection, and the job names used as check contexts.

- Rebased onto latest `main` with no conflicts; confirmed the `playwright install` step added by #6678 survives in both workflows, and the PR-gate marker resolves to `integration`.
- `check yaml`, `check json` and `actionlint` all pass.
- Earlier rounds used two canary branches to close the loop both ways: a deliberately failing unit test (red is actually blocked) and a docs-only change (docs-only PRs are not left hanging).

## Before importing the ruleset

Because the PR gate now runs the full suite, **any pre-existing failure on `main` will block every PR once the ruleset is active.** As of this PR there is one, and it must be fixed first:

`tests/integration/test_fork_and_coding_mode.py::test_coding_mode_get_default_disabled` — fails on all four platforms.

It is a test-side adaptation gap, not a regression: #6504 intentionally removed `project_dir` from the `GET /api/coding-mode` response (`- "project_dir": cm.project_dir,` in `src/qwenpaw/app/routers/coding_mode.py`), while the test still asserts `"project_dir" in body`. Response is now `{'enabled': False, 'agent_id': 'default'}`. Reported separately for the integration test owner.

Two notes on things that look alarming in this PR's own CI but are not caused by it:

- The same `test_coding_mode_get_default_disabled` failure appears here. It arrived with the rebase: `#6504` is in this branch's base but not in the 8/05 nightly base, which is why nightly showed a different failure set.
- `Coverage Report` fails with `Couldn't combine from non-existent path '.coverage.integration'` — a cascade of the integration job failing, so the artifact was never uploaded. #6743 is already addressing artifact export on failure.

Recommended sequence: merge this PR, land the `test_coding_mode_get_default_disabled` fix, confirm a green run on `main`, then import the ruleset.

### DIFF
--- a/.github/rulesets/main-required-checks.json
+++ b/.github/rulesets/main-required-checks.json
@@ -1,0 +1,34 @@
+{
+  "name": "Required status checks (main)",
+  "target": "branch",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "include": ["~DEFAULT_BRANCH"],
+      "exclude": []
+    }
+  },
+  "rules": [
+    {
+      "type": "required_status_checks",
+      "parameters": {
+        "strict_required_status_checks_policy": false,
+        "do_not_enforce_on_create": false,
+        "required_status_checks": [
+          { "context": "Test Summary", "integration_id": 15368 },
+          { "context": "Pre-commit", "integration_id": 15368 },
+          { "context": "Vitest Unit Tests", "integration_id": 15368 },
+          { "context": "console format check", "integration_id": 15368 },
+          { "context": "website format check", "integration_id": 15368 }
+        ]
+      }
+    }
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 5,
+      "actor_type": "RepositoryRole",
+      "bypass_mode": "always"
+    }
+  ]
+}
--- a/.github/rulesets/main-required-checks.json
+++ b/.github/rulesets/main-required-checks.json
@@ -17,7 +17,7 @@
         "required_status_checks": [
           { "context": "Test Summary", "integration_id": 15368 },
           { "context": "Pre-commit", "integration_id": 15368 },
-          { "context": "Vitest Unit Tests", "integration_id": 15368 },
+          { "context": "Frontend Summary", "integration_id": 15368 },
           { "context": "console format check", "integration_id": 15368 },
           { "context": "website format check", "integration_id": 15368 }
         ]
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -61,10 +61,10 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Run unit tests
-        run: npm run test:run
-
-      - name: Coverage ratchet (blocks PR on regression)
+      # A single vitest pass: runs the full suite AND enforces the coverage
+      # ratchet (fails on test failures or coverage regression). The previous
+      # separate `npm run test:run` pass executed the identical suite twice.
+      - name: Run unit tests with coverage (ratchet blocks PR on regression)
         run: npm run test:coverage
 
       - name: Upload coverage report
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -8,9 +8,6 @@ on:
       - '.github/workflows/frontend-tests.yml'
   pull_request:
     branches: [main, master, dev, develop]
-    paths:
-      - 'console/**'
-      - '.github/workflows/frontend-tests.yml'
 
 jobs:
   spam-gate:
@@ -20,11 +17,32 @@ jobs:
     with:
       author: ${{ github.event.pull_request.user.login }}
 
+  # Replaces the former `on.pull_request.paths` filter: the workflow now runs
+  # (and reports a status) on every PR so `Vitest Unit Tests` can be a
+  # required check; PRs not touching the console skip the actual test run.
+  changes:
+    name: Detect console changes
+    runs-on: ubuntu-latest
+    outputs:
+      code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@v4
+        if: github.event_name == 'pull_request'
+      - uses: dorny/paths-filter@v3
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            code:
+              - 'console/**'
+              - '.github/workflows/frontend-tests.yml'
+
   vitest:
     name: Vitest Unit Tests
-    needs: [spam-gate]
+    needs: [spam-gate, changes]
     if: |
       always() &&
+      needs.changes.outputs.code == 'true' &&
       (needs.spam-gate.result == 'skipped' || needs.spam-gate.outputs.blocked != 'true')
     runs-on: ubuntu-latest
     defaults:
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -18,8 +18,9 @@ jobs:
       author: ${{ github.event.pull_request.user.login }}
 
   # Replaces the former `on.pull_request.paths` filter: the workflow now runs
-  # (and reports a status) on every PR so `Vitest Unit Tests` can be a
-  # required check; PRs not touching the console skip the actual test run.
+  # (and reports a status) on every PR so `Frontend Summary` can be a
+  # required check; PRs not touching the console skip the actual test run
+  # and the summary reports an explicit success instead of a skip.
   changes:
     name: Detect console changes
     runs-on: ubuntu-latest
@@ -40,8 +41,13 @@ jobs:
   vitest:
     name: Vitest Unit Tests
     needs: [spam-gate, changes]
+    # Also guard on the detection job's RESULT: if detection failed/was
+    # cancelled its `code` output is empty and this job would silently skip,
+    # which a ruleset cannot distinguish from a bypass. The frontend-summary
+    # job below turns that detection failure into a hard red.
     if: |
       always() &&
+      needs.changes.result == 'success' &&
       needs.changes.outputs.code == 'true' &&
       (needs.spam-gate.result == 'skipped' || needs.spam-gate.outputs.blocked != 'true')
     runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


RerankerGuo:
Independent review at exact head `4b8feeba7c36b2dc9a862e0b7ff5c078f676dd10`.

The required-check design fails open when either path-detection job fails.

The PR intentionally relies on a skipped required job being accepted for an irrelevant/docs-only change. However, both required jobs use the detector's output without checking whether the detector itself succeeded:

```yaml
# tests.yml
test-summary:
  if: always() && needs.changes.outputs.code == 'true'

# frontend-tests.yml
vitest:
  if: |
    always() &&
    needs.changes.outputs.code == 'true' &&
    ...
```

If checkout, `dorny/paths-filter`, or the changed-files API fails:

1. `changes` concludes `failure`;
2. `steps.filter.outputs.code` is unavailable/empty;
3. `Test Summary` or `Vitest Unit Tests` is skipped by the job-level `if`;
4. the ruleset sees the same skipped required context that this PR intentionally uses to satisfy docs-only PRs.

The ruleset does not require `Detect code changes` or `Detect console changes`, so the detector failure itself is not a required context:

```json
[
  "Test Summary",
  "Pre-commit",
  "Vitest Unit Tests",
  "console format check",
  "website format check"
]
```

I encoded the fail-closed invariant as a temporary exact-head contract:

```python
@pytest.mark.parametrize(
    ("workflow", "required_job"),
    [
        (".github/workflows/tests.yml", "test-summary"),
        (".github/workflows/frontend-tests.yml", "vitest"),
    ],
)
def test_required_gate_fails_closed_when_change_detection_fails(
    workflow,
    required_job,
):
    data = yaml.safe_load(Path(workflow).read_text())
    job = data["jobs"][required_job]
    guard_and_steps = str(job.get("if", "")) + "\n" + "\n".join(
        str(step.get("run", ""))
        for step in job.get("steps", [])
        if isinstance(step, dict)
    )
    assert "needs.changes.result" in guard_and_steps
```

Exact-head result:

```text
2 failed

tests.yml:test-summary does not distinguish a successful
docs-only detection from a failed changes job

frontend-tests.yml:vitest does not distinguish a successful
docs-only detection from a failed changes job
```

The normal PR runs show both detector jobs succeeding, so they do not exercise this failure path:

```text
Detect code changes      success
Detect console changes   success
```

`check-yaml`, ruleset JSON parsing, and the PR's remote Pre-commit check pass; this is a gate-state issue rather than malformed syntax.

Two viable corrections:

1. Add both detector contexts to `required_status_checks`. Then a detector failure blocks while its success still allows the downstream test job to be skipped for irrelevant changes.
2. Prefer always-running summary contexts. The summary should fail when `needs.changes.result != 'success'`, return success explicitly when detection succeeded with `code != 'true'`, and otherwise require every gated job to conclude exactly `success`. The frontend workflow would need an equivalent always-running summary context instead of requiring the conditionally skipped Vitest job directly.

The same fail-closed check should also reject `cancelled`/unexpectedly `skipped` test dependencies rather than checking only the literal `failure` result.

RerankerGuo (changes requested):
作者意图：让 main 的测试结果真正约束 merge，同时避免 docs-only PR 因 required workflow 未触发而永久 pending。

当前设计在 path detector 失败时会 fail open。`Test Summary` 与 `Vitest Unit Tests` 都只读取 `needs.changes.outputs.code`，没有检查 `needs.changes.result`；checkout、changed-files API 或 `dorny/paths-filter` 失败时 output 为空，于是 required job 被 `if` 跳过。ruleset 又没有要求两个 detector context，因此该故障与“成功检测为 docs-only”具有相同的 merge-gate 表象。

```mermaid
flowchart LR
    A[Detect changes job] --> B{Result}
    B -->|success, code false| C[Required test job skipped]
    B -->|failure, output empty| C
    C --> D[Ruleset accepts skipped context]
    D --> E[Merge gate opens]
```

Exact-head contracts 对两条 required path 均失败：

```text
tests.yml / Test Summary:
  no needs.changes.result fail-closed handling

frontend-tests.yml / Vitest Unit Tests:
  no needs.changes.result fail-closed handling

result: 2 failed
```

正常 PR evidence 中 detector 都是 `success`，所以未覆盖该状态。YAML、ruleset JSON 与远端 Pre-commit 均通过；这是 gate state-machine 问题，不是语法问题。

请采用以下任一完整方案：

1. 将 `Detect code changes` 和 `Detect console changes` 也加入 required status checks；成功 docs-only 仍允许 downstream job skip，而 detector failure 会直接阻止 merge；或
2. 使用 always-running summary contexts：detector 非 `success` 时失败，成功且 `code != true` 时显式成功，需要测试时则只接受所有依赖均为 `success`。frontend 也应要求该 summary，而非条件性 Vitest job。

回归测试应覆盖 detector 的 `success/code=false`、`failure`、`cancelled`，以及 test dependency 的 `failure`、`cancelled`、unexpected `skipped`，证明只有真正 docs-only 才能绕过测试执行。

yutai78786:
@RerankerGuo Thanks for the careful review and the exact-head reproduction — the fail-open chain you described is real, and we agree with the substance of every point. Commit 7fa21070 (now the PR head) addresses all of them; details below.

**1. Fail-open when the detection jobs fail — confirmed, fixed.**
Verified line by line: if `Detect code changes` / `Detect console changes` do not reach `success`, their outputs go empty, the downstream test jobs are skipped by their `if:` guards, and the ruleset treats skipped required checks as satisfied — a failed detection was indistinguishable from "detection succeeded, docs-only change". This is now closed.

**2. Why not option 1 (add the two detection jobs to the required checks).**
It blocks detection failures, but leaves a residual hole: rulesets treat a *skipped* required check as satisfied, so if the summary job itself is unexpectedly skipped, the gate silently passes. We want every gate state to resolve to an explicit red or an explicit green — never a silent pass.

**3. Adopted option 2 — always-running summary gates with three-state decisions.**
- `Test Summary` (tests.yml) now runs with `if: always()`; a new `Frontend Summary` (frontend-tests.yml) does the same for the frontend workflow. The conditional `vitest` job keeps its original behavior plus a `needs.changes.result == 'success'` guard.
- Decision logic in each summary:
  1. any detection job not `success` (failure / cancelled / skipped) → the summary **fails**;
  2. detection succeeded and no code/console changes (docs-only) → the summary **explicitly succeeds** — an explicit green instead of a skip standing in for satisfaction;
  3. tests required → the approval gate and every test tier must be **strictly `success`**; `failure`, `cancelled` and `skipped` are all rejected.
- `rulesets/main-required-checks.json`: required check `Vitest Unit Tests` → `Frontend Summary`.
- `coverage-report` stays advisory (not a required check): a coverage-tooling outage should not block PRs.

Self-check for new silent-skip paths: docs-only PRs now show an explicit green summary instead of skipped-as-satisfied; any non-green test tier is an explicit red. Every path terminates in red or green.

**4. Rejecting `cancelled` and unexpected `skipped` — agreed; this exposed a second, pre-existing hole.**
The old `test-summary` script passed anything that was not literally `failure` — cancelled or unexpectedly skipped tiers were reported as passing. The fix inverts the logic entirely: only strict `success` passes (state 3 above), rather than patching in two extra conditions.

**5. Regression evidence — 7 canary scenarios on the real Actions state machine, not paper analysis.**
Setup: a disposable target branch plus a disposable ruleset (required checks = `Test Summary` + `Frontend Summary`) on the fork, one canary PR per scenario, everything cleaned up afterward (canary PRs yutai78786/QwenPaw#41–#45 closed, branches and ruleset deleted, each deletion verified).

| # | Scenario | Setup | Expected | Actual |
|---|---|---|---|---|
| A | Docs-only | one-line README change | both summaries explicitly green, mergeable | both `success`, `mergeable=true` |
| B | Code change, all green | one-line code comment | green, mergeable | green and mergeable (one rerun of a pre-existing flaky Windows unit test unrelated to this change) |
| C | Detection failure | `exit 1` injected into `Detect code changes` | blocked | `Test Summary=failure`, `mergeable_state=blocked` |
| D | Detection cancelled | whole run cancelled while queued | blocked | the always-running summary still executed, saw `detection=cancelled` and reported `failure` itself → blocked (stronger than expected) |
| E | Test tier failure | injected an always-failing unit test | blocked | unit tests failed on all four platforms → summary `failure`, per-tier verdict logged, blocked |
| F | Test tier cancelled mid-run | run cancelled while unit tests in progress | blocked | summary judged `failure` first; run concluded `cancelled`; blocked |
| G | Test tier unexpectedly skipped | forced the unit-test tier's `if:` to `false` | blocked (skipped ≠ success) | `Unit Tests=skipped` → summary `failure`, blocked |

Scenarios D/F also surfaced a platform behavior worth noting: cancelling a *queued* run takes effect immediately, while cancelling an *in-progress* run lagged by tens of minutes in our runs; the always-running summary keeps the gate safe under both.

The summary decision log prints every tier's status, so a red gate always states its reason, e.g. `Unit tests: skipped → Every test tier must be success (failure/cancelled/skipped are all rejected) → exit 1`.

**6. The exact-head contract test — good proof of the hole; we would not commit it as-is.**
It asserts on literal implementation text (that `needs.changes.result` must appear in specific scripts), so it will false-positive the moment the guard is written differently. The 7 canary scenarios exercise the actual GitHub Actions state machine, which we consider stronger evidence than text assertions. If an in-repo regression line is still wanted, we would suggest a narrower actionlint-level check in pre-commit — happy to discuss.

**Note on the remaining red on this PR:** the integration tier currently fails on a known pre-existing test (`test_coding_mode_get_default_disabled`, broken by the `project_dir` removal; fix in progress separately). That failure is independent of this gate change — and the new gate correctly reports it instead of letting it slip by.

Head updated to 7fa21070. Requesting re-review.

cuiyuebing (approved):
LGTM

cuiyuebing (approved):
LGTM

PR #6767, opened 2026-08-06
fix(config): harden agent persistence on shared filesystems

rayrayraykk:
## Summary

- use the existing atomic JSON writer for agent and ACL migrations, and retain legacy source fields when destination persistence fails
- replace the mtime-only agent cache key with device, inode, size, and nanosecond mtime while preserving the existing shared-object `Workspace.config` contract
- reject stale saves with a loaded-content digest check
- move per-agent `last_dispatch` runtime data to `state/last_dispatch.json` and migrate legacy values publication-first
- offload watcher, heartbeat, and reply-dispatch persistence I/O from the event loop
- update configuration and heartbeat documentation

## Why

This addresses three shared-filesystem failure modes: truncated or trailing migration writes, stale same-mtime cache hits that can overwrite newer configuration, and high-frequency runtime state rewrites of `agent.json`.

The implementation intentionally keeps:

```python
@property
def config(self):
    self._config = load_agent_config(self.agent_id)
    return self._config
```

No deep-copy compatibility layer is introduced.

## Validation

- `conda run -n Codex-QwenPaw pytest -q tests/unit`: 6474 passed, 15 skipped
- config and heartbeat integration selection: 46 passed
- channel compatibility selection: 404 passed, 1 skipped
- pre-commit: all hooks passed, including mypy, black, flake8, and pylint

## Filesystem boundary

QwenPaw now consistently uses its same-directory atomic replacement helper. Actual cross-client rename guarantees still depend on the mounted FUSE or OSSFS implementation; stale-save validation protects updates visible before the save transaction.


RerankerGuo:
Independent A/B review of exact head `e2e682da8262dd488cde469de137c42ba1b2f545`.

The new private source digest breaks the existing **Copy Agent** endpoint.

`load_agent_config(source_id)` now records `_source_digest` on the loaded `AgentProfileConfig`. The copy path derives the new identity with:

```python
agent_config = source_config.model_copy(deep=True)
agent_config.id = new_id
...
save_agent_config(new_id, agent_config)
```

Pydantic copies private attributes in `model_copy(deep=True)`, so the new agent inherits the **source agent's** digest. Its new workspace does not have `agent.json` yet. `save_agent_config()` sees a non-`None` source digest plus a missing target file and raises:

```text
ConfigurationException:
Agent 'copied' changed on disk; reload it before saving
```

I reproduced this with one identical test on current main and the PR head:

```python
source = load_agent_config("source")
copied = _build_copied_agent_config(
    source_config=source,
    new_id="copied",
    new_name="Source Copy",
    workspace_dir=copied_workspace,
)
save_agent_config("copied", copied)

assert json.loads(
    (copied_workspace / "agent.json").read_text()
)["id"] == "copied"
```

Results:

```text
current main bacac741:
1 passed

PR head e2e682da:
1 failed
qwenpaw.exceptions.ConfigurationException:
Agent 'copied' changed on disk; reload it before saving
```

This also leaves partial durable state. `copy_agent()` calls `save_config(config)` after registering the new profile, then calls the failing `save_agent_config()`. The request returns 500 with a new agent registered in root config but no corresponding `agent.json`.

The committed suites remain green:

```text
new persistence tests + agent router tests: 44 passed
all config tests: 25 passed
remote checks: all completed successfully
```

The copy router tests mock both `load_agent_config()` (returning a model with no recorded digest) and `save_agent_config()`, so the private-attribute interaction is not exercised.

Suggested correction:

1. Treat the digest as identity/version metadata, not copyable profile data. Build a copied agent from `source_config.model_dump(...)` / `AgentProfileConfig.model_validate(...)`, or provide an explicit helper that clears the source version when deriving a new identity.
2. Add a disk-backed regression that loads the source through real `load_agent_config()`, invokes the copy path, and verifies the new `agent.json`.
3. Consider rollback/order handling so a failed target save cannot leave a root profile pointing to a missing configuration file.

The stale-save guard should remain intact for edits to the same agent; only identity-changing derivations need an unversioned new model.

RerankerGuo (changes requested):
作者意图：通过 fingerprint cache、content digest stale-save guard 和独立 runtime state，强化共享文件系统上的 Agent 配置持久化。

核心 guard 对同一 identity 的编辑是合理的，但私有 `_source_digest` 被 Pydantic `model_copy(deep=True)` 复制，破坏现有 Copy Agent 路径。该路径从已加载的源配置派生新 identity：

```python
source = load_agent_config(source_id)  # records source digest
copied = source.model_copy(deep=True)  # copies private digest
copied.id = new_id
save_agent_config(new_id, copied)
```

目标 workspace 尚无 `agent.json`，但 `copied.source_digest()` 非空，所以保存被判定为“目标在磁盘上已变更”并抛出 `ConfigurationException`。

```mermaid
flowchart LR
    A[Load source agent] --> B[Attach source file digest]
    B --> C[Deep-copy profile for new identity]
    C --> D[Private digest copied too]
    D --> E[Target agent.json does not exist]
    E --> F[Stale-save guard rejects first save]
```

相同 disk-backed A/B regression：

```text
current main bacac741:
  Copy Agent first save: passed

exact head e2e682da:
  Copy Agent first save: failed
  ConfigurationException: Agent 'copied' changed on disk
```

影响不只是 500。`copy_agent()` 先通过 `save_config(config)` 注册新 profile，再执行失败的 `save_agent_config()`，因此会留下 root config 中存在新 agent、目标 workspace 却没有 `agent.json` 的部分持久化状态。现有 44 项 persistence/router selection 与 25 项 config tests 全绿，是因为 router tests mock 了 load/save，未经过 private-attribute interaction。

请：

1. 将 digest 明确作为 identity/version metadata，而非可复制的 profile 数据；新 identity 应从 `model_dump()` + `model_validate()` 构造，或使用显式 helper 清除 source version；
2. 保留同一 agent 编辑时的 stale-save guard；
3. 添加真实 disk-backed Copy Agent regression，源配置必须经 `load_agent_config()` 加载，并验证目标 `agent.json`；
4. 调整写入顺序或 rollback，使目标配置保存失败时不会留下悬空 root profile。

qbc2016 (approved):
LGTM.

PR #6748, opened 2026-08-06
fix(shell): preserve multiline commands and avoid background pipe hangs

qbc2016:
## Description

Fixes #6565.

Preserves Unix multiline shell syntax and prevents background processes from blocking command completion by redirecting stdout/stderr to temporary files and waiting only for the direct shell process.

Trade-off: large outputs may consume temporary-disk space, and files held open by background descendants may not release that space until those processes close their descriptors.

**Security Considerations:** [If applicable, e.g. channel auth, env/config handling]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [ ] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

[How to test these changes]

## Evidence

Examples of valid evidence:
- Terminal transcript of the test run (e.g. `pytest tests/unit/app/chats/ -q` output)
- Screenshot of the Console UI showing the fix
- CI artifact link
- `pre-commit run --all-files` summary

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]

PR #6799, opened 2026-08-07
fix(shell): stop temp output file leakage and cap captured output

lllyfff:
## Description

On Windows, every `execute_shell_command` call redirects the child's stdout/stderr to temp files (`qwenpaw_out_*` / `qwenpaw_err_*`). Three defects combined into a real-world **26 GB orphan file** in `%TEMP%` that could not be deleted and was completely invisible (no log, no metric):

1. **Silent delete failure** — cleanup runs `os.unlink` inside `except OSError: pass`. On Windows, `unlink` fails with `PermissionError` whenever any process still holds the file open (POSIX allows deleting open files, which masked this on macOS/Linux). The failure was swallowed, leaking the file permanently and invisibly.
2. **No output size limit** — the temp files had no cap anywhere. A runaway command (infinite loop, endless log, detached background writer) could grow them without bound, and `_read_temp_file` then read the entire file into memory (`f.read()`), an OOM risk.
3. **Unreliable process-tree kill** — timeout/cancel relies on `taskkill /F /T`, which can miss descendants that detached from the PID tree; those survivors keep holding the temp-file handles (the precondition for the leak), and the kill helper swallowed all its own errors.

This PR fixes all three plus two adjacent gaps:

- **Unlink failures are now visible and recoverable**: failure logs a WARNING (path + error), registers the file for an `atexit` retry, and a once-per-process startup sweep deletes stale orphans (older than 1 h and no longer locked) left by previous runs.
- **Captured output is capped** at `QWENPAW_SHELL_MAX_OUTPUT_BYTES` (default 10 MB): the poll loop checks the combined temp-file size and terminates the process tree when exceeded; `_read_temp_file` reads at most the cap (plus a truncation marker), so memory stays bounded. Downstream `ToolResultPruningMiddleware` already caps what the LLM sees, so the cap only protects disk/memory.
- **Windows job object for reliable kills**: the child is placed in a job object right after `Popen` (descendants inherit membership automatically), so timeout/cancel terminates *every* member via `TerminateJobObject` regardless of tree topology; `taskkill` remains as fallback. Kill failures are logged instead of swallowed. Job creation/assignment failures degrade gracefully to the previous taskkill-only behavior.
- **Background processes are still allowed to survive normal exit** (the job is only terminated on timeout/cancel), preserving the documented design where commands like `Start-Process` return immediately while their children keep running.
- **Unexpected post-`Popen` errors now kill the child** too (previously an exception after spawn left the child running with the temp-file handles open).

**Related Issue:** Fixes #(issue_number) 

**Security Considerations:** None for auth/env/config handling. Process termination is strictly limited to the process tree/job of commands the shell tool itself spawned; job-object creation or assignment failures fall back to the previous behavior. No new privileges required (`OpenProcess` with `PROCESS_SET_QUOTA | PROCESS_TERMINATE` on our own child).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

N/A — not a channel change.

## Testing

**Unit tests** (added/updated in `tests/unit/agents/tools/test_shell.py`):

```bash
pytest tests/unit/agents/tools/test_shell.py -q
```

New coverage:

- `TestShellOutputLimit` — a runaway writer is terminated at the cap with `-1` + explanatory stderr; read-back is capped; normal commands under the limit are unaffected.
- `TestShellTempFileCleanup` — temp files are gone after a normal command; a simulated locked file (unlink raises `PermissionError`) is logged + registered instead of swallowed; the exit-time retry deletes released files.
- `TestStaleTempFileSweep` — sweep deletes stale (`>1h`) unlocked files with our prefixes only, skips fresh/foreign/directory entries, and runs once per process.
- `TestKillProcessTreeHelpers` — kill helper never raises on a missing PID; job-object lifecycle on Windows / graceful unavailability off Windows.
- `TestReadTempFile` — capped reads with truncation marker.

**Manual reproduction of the original incident** (Windows): run a command whose detached child keeps the stdout handle open after the direct child exits, e.g.

```python
from qwenpaw.agents.tools.shell import _execute_subprocess_sync
code, out, err = _execute_subprocess_sync(
    'start /B "" python -c "import time; time.sleep(30)"', ".", 10.0)
```

Before: the temp file silently stays in `%TEMP%` forever. After: a WARNING is logged with the exact path, the file is registered, and it is deleted by the atexit retry once the holder exits (verified end-to-end below).

## Evidence

The fix is validated by pre-commit (all applicable hooks green on the changed files), the unit suites (101 passed / 15
skipped for test_shell.py, 266 passed / 15 skipped for tests/unit/agents/tools/ on Windows), and a real-machine
reproduction that reclaimed the original 26 GB orphan file (details below).

### 1. pre-commit

**1.1 On the 3 changed files — all hooks pass:**

```text
$ pre-commit run --files src/qwenpaw/agents/tools/shell.py src/qwenpaw/constant.py tests/unit/agents/tools/test_shell.py
```

(black's 2 reformats and the pylint fixes were committed; rerun is fully green as above.)

**1.2 `pre-commit run --all-files` (full repo):**

```text
mypy: Success: no issues found in 469 source files
pylint...................................................................Failed (exit 4)
  tests\unit\plugins\computer_use\test_stop_semantics.py:61:8:
  W0613: Unused argument 'handler' (unused-argument)
```

The single pylint finding is **pre-existing and outside this PR's diff** — `test_stop_semantics.py` was introduced by #6424/#6671 and is untouched here; all files changed by this PR are rated 10.00/10.

### 2. pytest

```text
$ pytest tests/unit/agents/tools/test_shell.py -q
96 passed, 5 skipped in 17.80s          # 5 skips = pre-existing POSIX-only paths

$ pytest tests/unit/agents/tools/ -q
261 passed, 5 skipped in 20.99s

$ pytest tests/unit -q                  # full unit suite
6441 passed, 72 skipped, 6 failed in 1077.09s
```

The 6 full-suite failures are **pre-existing environment issues unrelated to this change** (Windows symlink privileges etc.) — verified by `git stash` + rerunning the same 6 tests on the unmodified base: identical failures (`test_cli_task` overlay ×2, `governance/test_policy` absolute path, `skill_scanner` symlink ×2 / yaml header).

### 3. Real-machine proof (the original 26 GB incident, Windows)

First run of the new startup sweep on the machine that produced the incident — it deleted the **exact 26,229,984,973-byte orphan file from the bug report** plus ~60 previously leaked files (excerpt):

```text
INFO shell.py:376 | Removed stale shell temp file: C:\Users\22812\AppData\Local\Temp\qwenpaw_out_1i_8m44n   ← the 26 GB file
INFO shell.py:376 | Removed stale shell temp file: C:\Users\22812\AppData\Local\Temp\qwenpaw_err_00hrsott
... (~60 more previously leaked qwenpaw_out_*/qwenpaw_err_* files removed)
```

Detached-child scenario (direct child exits, grandchild still holds the handle) — previously silent, now visible and self-healing:

```text
WARNING shell.py:401 | Could not remove shell temp file C:\Users\22812\AppData\Local\Temp\qwenpaw_out_omhpq9hh
    ([WinError 32] The process cannot access the file because it is being used by another process).
    A process likely still holds it open; QwenPaw will retry at exit and sweep the file on a future start once it is released.
rc=0 elapsed=0.25s
registered leaked files: ['...qwenpaw_out_omhpq9hh', '...qwenpaw_err_rbpn2phu']
```

After the holder exited, the `atexit` retry deleted both files — verified they no longer exist on disk.

The leak precondition was also confirmed live: orphan processes from past sessions were still holding handles days later (6 × `python -m http.server 8080` since 2026-08-04, 4 × orphaned `cmd` from today) — exactly the "`taskkill /T` missed them" scenario the job-object kill now closes.

## Additional Notes

**Known boundary (by design):** if a command *exits normally* but deliberately leaves a detached background process holding stdout (e.g. a user-intended daemon), QwenPaw does not kill it — that would break legitimate `start ...` usage. Such a file is now visible in logs and reclaimed once the holder exits (atexit retry / next-start sweep); growth while the daemon runs is the user's own process writing, bounded only when the command is under monitoring (timeout/cancel paths are fully covered by cap + job-object kill).

**Config:** `QWENPAW_SHELL_MAX_OUTPUT_BYTES` (default `10485760`, min `1024`) lets operators tune the cap.

**Files changed:** `src/qwenpaw/agents/tools/shell.py`, `src/qwenpaw/constant.py`, `tests/unit/agents/tools/test_shell.py` (+639/−30).

### DIFF
--- a/src/qwenpaw/agents/tools/shell.py
+++ b/src/qwenpaw/agents/tools/shell.py
@@ -5,6 +5,7 @@
 
 import asyncio
 import locale
+import logging
 import os
 import re
 import signal
@@ -26,31 +27,180 @@ from ...config.context import (
     get_current_shell_command_timeout,
     get_current_workspace_dir,
 )
-from ...constant import WORKING_DIR
+from ...constant import SHELL_MAX_OUTPUT_BYTES, WORKING_DIR
 from ...runtime.tool_registry import tool_descriptor
 from ...sandbox import ExecutionResult
 from ...sandbox.config import SandboxConfig
 from ...utils.io_utils import run_sync_io
 
+logger = logging.getLogger(__name__)
+
 _SHELL_OUTPUT_MAX_BYTES = 1024 * 1024
 _SHELL_OUTPUT_DRAIN_GRACE_SECS = 10.0
 
+# Prefixes of the temp files this module creates for shell stdout/stderr.
+_TEMP_FILE_PREFIXES = ("qwenpaw_out_", "qwenpaw_err_")
 
-def _kill_process_tree_win32(pid: int) -> None:
-    """Kill a process and all its descendants on Windows via taskkill.
+# A file with one of our prefixes that still exists but is not locked can
+# only be an orphan: current code either deletes temp files right after each
+# command or marks them for automatic deletion on last-handle close (Windows
+# ``O_TEMPORARY``).  Remaining files were leaked by older versions or by a
+# crashed run, so stale ones may be swept once per process.
+_STALE_TEMP_FILE_AGE_SECS = 3600
 
-    Uses ``taskkill /F /T`` which forcefully terminates the entire process
-    tree, including grandchild processes that ``Popen.kill()`` would miss.
-    """
+_sweep_done = False
+_sweep_lock = threading.Lock()
+
+
+# ---------------------------------------------------------------------------
+# Windows job-object helpers.
+#
+# ``taskkill /F /T`` walks the parent/child PID chain, so descendants that
+# detached from the tree (independent session, dead intermediate parent)
+# can survive a kill and keep holding the stdout/stderr temp-file handles —
+# which keeps the output files alive and growing on disk.  Placing the
+# command in a job object lets us terminate *every* member on
+# timeout/cancel regardless of tree topology.  All helpers degrade
+# gracefully: on any failure they return None/False and callers fall back
+# to the taskkill-only behaviour.
+# ---------------------------------------------------------------------------
+
+_PROCESS_TERMINATE = 0x0001
+_PROCESS_SET_QUOTA = 0x0100
+
+_win_kernel32 = None
+
+
+def _get_kernel32():
+    """Return a ctypes handle to ``kernel32`` with prototypes configured."""
+    global _win_kernel32
+    if _win_kernel32 is None:
+        import ctypes
+
+        kernel32 = ctypes.WinDLL("kernel32", use_last_error=True)
+        kernel32.CreateJobObjectW.restype = ctypes.c_void_p
+        kernel32.CreateJobObjectW.argtypes = [
+            ctypes.c_void_p,
+            ctypes.c_wchar_p,
+        ]
+        kernel32.OpenProcess.restype = ctypes.c_void_p
+        kernel32.OpenProcess.argtypes = [
+            ctypes.c_uint32,
+            ctypes.c_int,
+            ctypes.c_uint32,
+        ]
+        kernel32.AssignProcessToJobObject.restype = ctypes.c_int
+        kernel32.AssignProcessToJobObject.argtypes = [
+            ctypes.c_void_p,
+            ctypes.c_void_p,
+        ]
+        kernel32.TerminateJobObject.restype = ctypes.c_int
+        kernel32.TerminateJobObject.argtypes = [
+            ctypes.c_void_p,
+            ctypes.c_uint32,
+        ]
+        kernel32.CloseHandle.restype = ctypes.c_int
+        kernel32.CloseHandle.argtypes = [ctypes.c_void_p]
+        _win_kernel32 = kernel32
+    return _win_kernel32
+
+
+def _create_job_object_win32():
+    """Create a Windows job object; return its handle or ``None``."""
+    if sys.platform != "win32":
+        return None
     try:
-        subprocess.call(
+        import ctypes
+
+        handle = _get_kernel32().CreateJobObjectW(None, None)
+        if not handle:
+            logger.debug(
+                "CreateJobObjectW failed (error %s)",
+                ctypes.get_last_error(),
+            )
+            return None
+        return handle
+    except Exception as exc:  # pragma: no cover - defensive
+        logger.debug("Job object creation unavailable: %s", exc)
+        return None
+
+
+def _assign_process_to_job_win32(job_handle, pid: int) -> bool:
+    """Assign *pid* to the job object.  Returns ``True`` on success."""
+    try:
+        import ctypes
+
+        kernel32 = _get_kernel32()
+        proc_handle = kernel32.OpenProcess(
+            _PROCESS_TERMINATE | _PROCESS_SET_QUOTA,
+            False,
+            pid,
+        )
+        if not proc_handle:
+            logger.debug("OpenProcess(%s) failed", pid)
+            return False
+        try:
+            ok = kernel32.AssignProcessToJobObject(job_handle, proc_handle)
+        finally:
+            kernel32.CloseHandle(proc_handle)
+        if not ok:
+            logger.debug(
+                "AssignProcessToJobObject failed for PID %s (error %s); "
+                "falling back to taskkill-only kill",
+                pid,
+                ctypes.get_last_error(),
+            )
+            return False
+        return True
+    except Exception as exc:  # pragma: no cover - defensive
+        logger.debug("Job assignment unavailable: %s", exc)
+        return False
+
+
+def _terminate_job_object_win32(job_handle) -> None:
+    """Terminate every process that is a member of the job object."""
+    try:
+        _get_kernel32().TerminateJobObject(job_handle, 1)
+    except Exception as exc:  # pragma: no cover - defensive
+        logger.debug("TerminateJobObject failed: %s", exc)
+
+
+def _close_handle_win32(handle) -> None:
+    """Close a Windows handle, ignoring errors."""
+    try:
+        _get_kernel32().CloseHandle(handle)
+    except Exception:  # pragma: no cover - defensive
+        pass
+
+
+def _kill_process_tree_win32(pid: int, job_handle=None) -> None:
+    """Kill a process and all its descendants on Windows.
+
+    When *job_handle* is given, the job object is terminated first: this
+    catches descendants that detached from the PID tree and that
+    ``taskkill /F /T`` may miss.  ``taskkill /F /T`` still runs afterwards
+    as a fallback for processes created before the job assignment.
+    Failures are logged rather than silently swallowed.
+    """
+    if job_handle is not None:
+        _terminate_job_object_win32(job_handle)
+    try:
+        returncode = subprocess.call(
             ["taskkill", "/F", "/T", "/PID", str(pid)],
             stdout=subprocess.DEVNULL,
             stderr=subprocess.DEVNULL,
             timeout=10,
         )
-    except Exception:
-        pass
+        if returncode != 0:
+            # Non-zero usually means the process was already gone (killed
+            # via the job object or exited on its own), so stay at debug.
+            logger.debug(
+                "taskkill /F /T /PID %s exited with code %s",
+                pid,
+                returncode,
+            )
+    except Exception as exc:
+        logger.warning("taskkill for PID %s failed: %s", pid, exc)
 
 
 def _windows_shell_creationflags() -> int:
@@ -210,8 +360,8 @@ class _PosixTempOutputs:
             if path is not None:
                 try:
                     os.unlink(path)
-                except OSError:
-                    pass
+                except OSError as exc:
+                    _log_unlink_failure(path, exc)
                 setattr(self, attr, None)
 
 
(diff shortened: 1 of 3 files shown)


RerankerGuo:
Independent validation on commit `b727559e82ea402e2656e88eaa6503e2ffe63868`.

Environment:
- macOS
- repository virtualenv

Commands and results:

```text
.venv/bin/pytest tests/unit/agents/tools/test_shell.py -q
100 passed, 1 skipped in 2.92s

.venv/bin/pytest tests/unit/agents/tools/ -q
265 passed, 1 skipped in 4.92s
```

All Python hooks applicable to the three changed files also pass:
`check-ast`, `check-docstring-first`, `fix-encoding-pragma`,
`detect-private-key`, `trailing-whitespace`, `add-trailing-comma`, `mypy`,
`black`, `flake8`, and `pylint`.

I could not independently verify the Windows job-object and locked-handle
behavior from macOS, so the Windows real-machine evidence in the PR remains
the authoritative proof for that path.

Current merge blockers I observed:

1. The branch is conflicting with current `main` and needs a rebase.
2. The PR body still contains `Fixes #(issue_number)`.
3. The Real behavior proof run reports a missing `## Evidence` section. The
   current body now contains that section, so this appears to be a stale run
   from before the body update; editing/re-triggering after the rebase should
   let the gate evaluate the current body.

The focused implementation tests are green at the stated SHA.

lllyfff:
@RerankerGuo Thanks for the independent macOS validation and for catching the stale-gate issue — appreciated.

  Rebase / conflict — resolved
  The branch was conflicting with current main because upstream #6748 landed a large shell rewrite (temp files moved to
  O_TEMPORARY auto-deletion + bounded snapshots). Instead of a mechanical merge, I re-ported the fix onto that new
  architecture and force-pushed. The branch is now cleanly on top of upstream main (bacac741) as a single commit:

  c02b3c92 fix(shell): cap captured output on disk and kill full process tree

  What the PR now adds on top of #6748: a disk cap (QWENPAW_SHELL_MAX_OUTPUT_BYTES, default 10 MB) that terminates
  runaway writers instead of letting them fill the disk; a Windows job object so timeout/cancel kills descendants that
  taskkill /T misses; child kill on unexpected post-Popen errors; WARNING logging for unlink failures; and a
  once-per-process sweep that reclaims stale orphaned qwenpaw_out_*/qwenpaw_err_* files.

  Tests re-run on the new commit (Windows)
  pytest tests/unit/agents/tools/test_shell.py -q   → 101 passed, 14 skipped
  pytest tests/unit/agents/tools/ -q                → 266 passed, 14 skipped
  pre-commit run --files <3 changed files>          → all hooks green (mypy, black, flake8, pylint)
  Your macOS numbers were on the previous SHA; the same logic is preserved, and the Windows-only paths (job object /
  O_TEMPORARY) remain covered by the Windows results here.

  Evidence gate — stale run
  The "missing ## Evidence" result predates the body update. The current body contains the verbatim ## Evidence section
  (test transcripts + the real-machine log of the 26 GB file being reclaimed). The rebase push should re-trigger the
  check against the current body — please re-run it if it doesn't pick up automatically.

RerankerGuo:
Independent revalidation at exact head `c02b3c92ff609586d2d894d5808770bd044e7225`.

The re-port is clean against current `main`, and its existing suites are green:

```text
.venv/bin/pytest tests/unit/agents/tools/test_shell.py -q
113 passed, 2 skipped

.venv/bin/pytest tests/unit/agents/tools/ -q
278 passed, 2 skipped

  src/qwenpaw/agents/tools/shell.py \
  src/qwenpaw/constant.py \
  tests/unit/agents/tools/test_shell.py
all applicable hooks passed
```

I found one cross-platform gap in the disk-cap claim. `SHELL_MAX_OUTPUT_BYTES` is read and polled only in `_execute_subprocess_sync()`. `execute_shell_command()` routes Linux/macOS host commands through the new `_execute_posix_host()`, which never reads that setting. Its `_SHELL_OUTPUT_MAX_BYTES` bound limits only the snapshot copied back into memory after execution; it does not stop the temporary output files from growing while the command runs.

I reproduced this with `SHELL_MAX_OUTPUT_BYTES = 64 * 1024` and a POSIX child writing 4096 bytes every 5 ms. The command timeout was 0.8 s:

```python
shell.SHELL_MAX_OUTPUT_BYTES = 64 * 1024
returncode, stdout, stderr = await shell._execute_posix_host(
    command,
    cwd,
    0.8,
    os.environ.copy(),
    "/bin/sh",
)
```

Result:

```text
configured_disk_cap=65536
elapsed=0.815s timeout=0.800s
returncode=-1
captured_bytes=466944
disk_cap_error=False
timeout_error=True
```

The writer exceeded the configured cap by more than 7x and continued until the ordinary timeout. A capped path should terminate on the output threshold and report the limit, before the timeout.

This also affects the background-writer boundary documented in the new POSIX test. If the direct shell exits while a descendant retains the regular-file descriptors, unlinking the paths hides the files but does not release their allocated storage; the descendant can continue growing the unlinked inodes after `_execute_posix_host()` returns. The test avoids that production leak by explicitly killing the saved background PID in `finally`.

Suggested direction:

1. Monitor the POSIX temp-file sizes concurrently with `proc.wait()` and cancellation, and terminate the process group with an explicit output-limit result when the combined size exceeds `SHELL_MAX_OUTPUT_BYTES`.
2. Define and enforce the normal-exit behavior for descendants that retain captured stdout/stderr. Requiring explicit redirection is useful guidance, but it does not make the configured disk cap effective; either detach those descriptors safely or keep the process group under management.
3. Add a routed POSIX regression that sets a small cap and proves a slow writer is stopped by the cap rather than by timeout. Keep the existing bounded-snapshot assertion as a separate memory-safety test.

The current Real behavior proof failure is still the pre-body-update run (`31187762414`, started 14:28 UTC); its log reports the old missing `Evidence` section. The current body has that section, although `Fixes #(issue_number)` remains unfilled.

lllyfff:
@RerankerGuo Thanks for the precise repro — the cap was Windows-only. Fixed in 94e87bca fix(shell): enforce the output disk cap on
  the POSIX host path.

  What changed: _execute_posix_host() now runs a monitor alongside proc.wait() that polls the temp-file sizes every 0.2
  s and, when they exceed SHELL_MAX_OUTPUT_BYTES, terminates the process group and returns -1 with ⚠️ Command output
  exceeded the limit of ... bytes — before the ordinary timeout.

  Regression test: test_execute_posix_host_output_cap_stops_runaway_writer sets a 64 KB cap with a slow writer and a 30
  s timeout, asserting the output-limit error fires (not TimeoutError).

  One boundary, kept explicit: if the direct shell exits normally, detached descendants are deliberately left running
  (that's why output is redirected to files in the first place), so a descendant retaining the descriptors can keep
  growing the unlinked file after the call returns. The cap protects the monitored execution window; killing post-exit
  descendants would break legitimate nohup/& usage. This is noted in the PR's Additional Notes.

  Verification (Windows): shell tests 101 passed / 15 skipped, tools 266 passed / 15 skipped, pre-commit green. The new
  POSIX test is skipped on Windows and runs under macOS/Linux CI.

  The "missing ## Evidence" CI result is from a stale pre-body-update run; the current body contains the section, so a
  re-run should pick it up.

RerankerGuo:
Rechecked exact head `94e87bcaf45f16960c63255b03f2e6045aeb40b1` on macOS.

The POSIX monitor implementation works with a valid paced writer, but the newly committed regression test does not currently execute that writer. Its generated script contains an unescaped newline inside a Python string literal:

```python
"    sys.stdout.write('x' * 4096 + '\n')\n"
```

`write_text()` therefore produces:

```python
sys.stdout.write('x' * 4096 + '
')
```

Exact-head result:

```text
.venv/bin/pytest tests/unit/agents/tools/test_shell.py -q \
  -k output_cap_stops_runaway_writer

FAILED
E assert 1 == -1

stderr:
SyntaxError: unterminated string literal (detected at line 3)
```

The complete shell file consequently reports:

```text
1 failed, 113 passed, 2 skipped
```

Escaping the generated newline is sufficient:

```python
"    sys.stdout.write('x' * 4096 + '\\n')\n"
```

With only that temporary test correction, the identical focused command passes:

```text
1 passed, 115 deselected
```

I also independently ran the implementation with a 64 KiB cap and a paced 4096-byte writer. It now returns `-1` with `Command output exceeded the limit`, not `TimeoutError`, in about 0.22 s. So the monitored-execution fix is effective; the remaining issue at this head is that the committed regression is red and does not prove it. All temporary test/probe changes were reverted after validation.

lllyfff:
@RerankerGuo The committed regression test had an unescaped newline in the generated writer     
  script, so it crashed with SyntaxError and never exercised the monitor. Fixed in 709804ac fix(test): escape newline in
  POSIX output-cap writer script.                                                                                       
                                                                                                                        
  What changed: the write_text(...) literal now uses '\\n', so the generated writer.py contains a real \n escape inside 
  the string literal instead of a literal line break. Both writer literals (this one and the shared _WRITER_CODE) now   
  compile to valid Python — verified by parsing the AST of the committed file and compiling the generated content.      
                                                                                                                        
  Verified at 709804ac:                                                                                                 
  pytest tests/unit/agents/tools/test_shell.py -q   → 101 passed, 15 skipped                                            
  pre-commit run --files <changed files>            → all hooks green                                                   
                                                                                                                        
  Your independent run with a paced writer confirms the implementation (monitor kills at the cap in ~0.22 s with Command
  output exceeded the limit, not TimeoutError) — the fix was only to the test that proves it. On macOS/Linux the        
  focused test should now pass; the POSIX tests are skipped on Windows, so please confirm on your side.

RerankerGuo:
Revalidated the test correction independently at exact head `709804ac25d7dd48d12d5e08efe8a57089e8a0ae` on macOS.

The generated writer now contains a valid `\n` escape, so it reaches the POSIX output monitor and proves the intended terminal result:

```text
.venv/bin/pytest tests/unit/agents/tools/test_shell.py -q \
  -k output_cap_stops_runaway_writer
1 passed, 115 deselected

HOME=/tmp/qwenpaw-pr6799-home \
  .venv/bin/pytest tests/unit/agents/tools/test_shell.py -q
114 passed, 2 skipped

HOME=/tmp/qwenpaw-pr6799-home \
    src/qwenpaw/agents/tools/shell.py \
    src/qwenpaw/constant.py \
    tests/unit/agents/tools/test_shell.py
all applicable hooks passed

git diff --check 94e87bca..709804ac
passed
```

The isolated `HOME` avoids an unrelated local sandbox denial on `~/.qwenpaw/.qwenpaw_restore.lock`. The POSIX monitored-execution finding and its regression-test gap are resolved at this head.

lllyfff:
@RerankerGuo Thanks for the thorough validation — the escape fix at 709804ac and the POSIX monitored-execution cap are confirmed on your side. All raised points are resolved at this head; only the CI Evidence gate (PR body formatting, now fixed)
  remains to re-trigger.

RerankerGuo (approved):
Intent: bound shell capture disk growth and reliably terminate the spawned process group on output-limit, timeout, and cancellation paths while retaining bounded in-memory snapshots.

Approved at unchanged exact head `709804ac25d7dd48d12d5e08efe8a57089e8a0ae`. The earlier POSIX routing gap and invalid generated-writer regression are resolved. Independent exact-head evidence passes the focused cap regression (`1 passed`), the complete shell suite (`114 passed, 2 skipped`), all applicable changed-file hooks, and `git diff --check`. The current branch cleanly merges with `main` (`bacac741`; merge tree `18235b5e`), and the Real behavior proof check is green. I found no remaining blocker in the affected paths.

```mermaid
flowchart LR
    A[Shell command starts] --> B[Capture output to temp files]
    B --> C{Output exceeds configured cap?}
    C -->|No| D[Continue or complete normally]
    C -->|Yes| E[Terminate process group or Windows job]
    E --> F[Return explicit output-limit result]
    D --> G[Return bounded snapshot]
    style C fill:#bbdefb,color:#0d47a1
    style F fill:#c8e6c9,color:#1a5e20
```

PR #6809, opened 2026-08-07
fix(providers): sanitize Chat Completions content for strict providers

### LINKED ISSUE #6803: [Bug]: OpenAI-compatible chat requests carry Responses-API `input_text` content type + raw streaming fields, rejected by strict providers (StepFun: 400 Unrecognized chat message)
## QwenPaw Version

2.0.1 (Python 3.12.3, Ubuntu 24.04 aarch64, installed on a self-hosted VPS, channel: Telegram)

## Description

When calling an OpenAI-compatible Chat Completions endpoint that **strictly validates message content structure** (e.g. StepFun's `step-3.5-flash`), requests fail with `400 Unrecognized chat message`. The same provider works fine with `curl` using standard payloads, so the problem is on QwenPaw's request serialization side.

Two serialization defects were captured from the actual outgoing request:

**1. Content items leak raw streaming/runtime fields.**

QwenPaw's error dump (`/tmp/qwenpaw_query_error_h32qvkij.json`) shows a plain user message serialized as:

```json
{
  "type": "text",
  "delta": "False",
  "index": "None",
  "status": "created",
  "object": "content",
  "msg_id": "None",
  "text": "测试"
}
```

Standard OpenAI Chat Completions content item should only contain `{"type": "text", "text": "..."}`. The extra keys (`delta`, `index`, `status`, `object`, `msg_id`) look like internal runtime/streaming fields that were not stripped during serialization. StepFun rejects any content item with unknown fields (`request_params_invalid`).

**2. System message uses Responses-API content type `input_text` in a Chat Completions request.**

Captured from the gateway's request log (9router `requestDetails`, field `providerRequest` — i.e. the request as received from QwenPaw before any proxy transformation):

```json
"content": [
  {"type": "text", "text": "You are helpful."},
  {"type": "input_text", "text": "Respond like terse caveman..."}
]
```

Chat Completions only defines `text` for content items; `input_text` is a Responses API type. Strict Chat Completions endpoints reject it. (An upstream gateway-level patch that rewrites `input_text` → `text` for *injected* prompts does not help because this `input_text` is emitted by QwenPaw itself and merely passed through the gateway.)

## Environment

- **QwenPaw version:** 2.0.1
- **OS:** Ubuntu 24.04 (aarch64)
- **Python version:** 3.12.3
- **Install method:** pip (venv, self-hosted)
- **Channel:** Telegram
- **Provider:** StepFun `step-3.5-flash` via OpenAI-compatible base URL (tested both direct and through a local gateway)

## Steps to Reproduce

1. Add an OpenAI-compatible provider pointing to StepFun (`https://api.stepfun.com/v1`), model `step-3.5-flash`.
2. Start a chat session (Telegram channel) and send any message.
3. Observe `400` error: `{"error":{"message":"Unrecognized chat message.","type":"request_params_invalid"}}`.

Control checks that point to QwenPaw serialization:

- Same provider/model via `curl` with a hand-written standard payload → HTTP 200.
- The gateway (`9router`) passes the request through unchanged; its captured `providerRequest` already contains the `input_text` item before any gateway logic runs.

## Actual vs Expected

- **Actual:** Chat Completions requests built by QwenPaw contain non-standard content items (`input_text` type; leaked `delta/index/status/object/msg_id` keys), which strict providers reject with `400 Unrecognized chat message`.
- **Expected:** In Chat Completions mode, all message content items should be serialized as plain `{"type": "text", "text": "..."}` (strip runtime fields; do not use Responses-API types like `input_text`), so requests work with any strict OpenAI-compatible endpoint.

## Additional Notes

- This likely affects **any** OpenAI-compatible provider that validates content item schema strictly; more lenient providers (OpenAI, Gemini) silently accept it, which is why the bug goes unnoticed there.
- Temporary workarounds used on our side: clearing polluted history reduces occurrences; a gateway-level rewrite patch mitigates the `input_text` case but does not fix the root cause.
- Disclosure per contribution policy (#4333): this issue was reproduced in a real self-hosted environment by the reporter; the report was compiled with AI assistance from the captured request dumps/logs above.

axelray-dev:
## Description

Strict OpenAI-compatible Chat Completions providers (for example StepFun) reject QwenPaw requests when message content parts include internal runtime envelope fields or Responses API text types.

Observed on the wire:

1. Text content parts can carry runtime fields such as delta, index, status, object, and msg_id from the streaming envelope.
2. Chat Completions requests can include input_text (Responses API) instead of text.

This PR sanitizes Chat Completions message content in OpenAIChatModelCompat._call_api before the request is sent:

- rewrite input_text / output_text to text
- keep only type and text for text parts
- strip known runtime envelope keys from non-text multimodal parts
- leave plain string content unchanged
- do not mutate the original messages object

Responses API providers and formatters are intentionally unchanged.

**Related Issue:** Fixes #6803

**Security Considerations:** None. Wire-format cleanup only; no auth, secrets, or channel changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran black (line-length 79) and flake8 on touched files locally and they pass
- [x] I ran targeted unit tests locally and they pass
- [ ] Ready for review

## Testing

Focused unit tests cover polluted text parts, input_text/output_text rewrite, mixed content lists, image_url preservation, plain string content, immutability of the original messages object, and that _call_api forwards sanitized messages to the base model.

## Evidence

Local unit proof for the Chat Completions sanitizer path:

- test file: tests/unit/providers/test_openai_chat_content_sanitize.py
- result: 8 passed
- black --line-length=79 on touched files: clean
- flake8 --extend-ignore=E203 on touched files: clean

Commands used:

pytest tests/unit/providers/test_openai_chat_content_sanitize.py -q
black --line-length=79 src/qwenpaw/providers/openai_chat_model_compat.py tests/unit/providers/test_openai_chat_content_sanitize.py
flake8 --extend-ignore=E203 src/qwenpaw/providers/openai_chat_model_compat.py tests/unit/providers/test_openai_chat_content_sanitize.py

## Additional Notes

Defensive sanitization on the Chat Completions wire path only. Does not change Responses API content types.

### DIFF
--- a/src/qwenpaw/providers/openai_chat_model_compat.py
+++ b/src/qwenpaw/providers/openai_chat_model_compat.py
@@ -86,7 +86,7 @@ class _ChatCompletionFormatterWrapper:
         self._inner_formatter = formatter
 
     async def format(self, *args: Any, **kwargs: Any) -> list[Any]:
-        """Format messages and remove fields unsupported by Chat Completions."""
+        """Format messages and strip Chat Completions-unsupported fields."""
         formatted_messages = await self._inner_formatter.format(
             *args,
             **kwargs,
@@ -732,8 +732,10 @@ class OpenAIChatModelCompat(OpenAIChatModel):
         self._extra_generate_kwargs = extra_generate_kwargs or {}
         self._output_token_param = output_token_param
         super().__init__(**kwargs)
-        if not isinstance(self.formatter, _ChatCompletionFormatterWrapper):
-            self.formatter = _ChatCompletionFormatterWrapper(self.formatter)
+        # AgentScope sets self.formatter in OpenAIChatModel.__init__.
+        current_formatter = getattr(self, "formatter", None)
+        if not isinstance(current_formatter, _ChatCompletionFormatterWrapper):
+            self.formatter = _ChatCompletionFormatterWrapper(current_formatter)
         credential_id = str(getattr(self.credential, "id", "") or "")
         credential_provider_id = credential_id.removeprefix("qwenpaw-")
         if (
--- a/src/qwenpaw/providers/openai_chat_model_compat.py
+++ b/src/qwenpaw/providers/openai_chat_model_compat.py
@@ -79,6 +79,24 @@ def _sanitize_chat_completion_messages(messages: Any) -> list[Any]:
     return sanitized_messages
 
 
+class _ChatCompletionFormatterWrapper:
+    """Sanitize the wire messages returned by an OpenAI formatter."""
+
+    def __init__(self, formatter: Any) -> None:
+        self._inner_formatter = formatter
+
+    async def format(self, *args: Any, **kwargs: Any) -> list[Any]:
+        """Format messages and remove fields unsupported by Chat Completions."""
+        formatted_messages = await self._inner_formatter.format(
+            *args,
+            **kwargs,
+        )
+        return _sanitize_chat_completion_messages(formatted_messages)
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._inner_formatter, name)
+
+
 def _battr(block: Any, key: str, default: Any = None) -> Any:
     """Read an attribute from a dict *or* Pydantic block."""
     if isinstance(block, dict):
@@ -714,6 +732,8 @@ class OpenAIChatModelCompat(OpenAIChatModel):
         self._extra_generate_kwargs = extra_generate_kwargs or {}
         self._output_token_param = output_token_param
         super().__init__(**kwargs)
+        if not isinstance(self.formatter, _ChatCompletionFormatterWrapper):
+            self.formatter = _ChatCompletionFormatterWrapper(self.formatter)
         credential_id = str(getattr(self.credential, "id", "") or "")
         credential_provider_id = credential_id.removeprefix("qwenpaw-")
         if (
@@ -807,7 +827,6 @@ class OpenAIChatModelCompat(OpenAIChatModel):
         tool_choice: Any | None = None,
         **generate_kwargs: Any,
     ) -> Any:
-        messages = _sanitize_chat_completion_messages(messages)
         merged = {**self._extra_generate_kwargs, **generate_kwargs}
         self._consume_disable_thinking(merged)
         if self._output_token_param != "max_tokens":
--- a/src/qwenpaw/providers/openai_chat_model_compat.py
+++ b/src/qwenpaw/providers/openai_chat_model_compat.py
@@ -27,6 +27,58 @@ from qwenpaw.utils.tool_call_extra import (
 logger = logging.getLogger(__name__)
 
 
+_CHAT_COMPLETION_RUNTIME_KEYS = frozenset(
+    {"delta", "index", "status", "object", "msg_id"},
+)
+
+
+def _sanitize_chat_completion_messages(messages: Any) -> list[Any]:
+    """Remove non-Chat-Completions fields from message content parts."""
+    sanitized_messages: list[Any] = []
+    for message in messages:
+        if not isinstance(message, dict):
+            sanitized_messages.append(message)
+            continue
+
+        sanitized_message = dict(message)
+        content = message.get("content")
+        if isinstance(content, list):
+            sanitized_content = []
+            for item in content:
+                if not isinstance(item, dict):
+                    sanitized_content.append(item)
+                    continue
+
+                item_type = item.get("type")
+                if item_type in ("input_text", "output_text"):
+                    sanitized_item = {"type": "text"}
+                    if "text" in item:
+                        sanitized_item["text"] = item["text"]
+                    sanitized_content.append(sanitized_item)
+                elif item_type == "text":
+                    sanitized_content.append(
+                        {
+                            key: value
+                            for key, value in item.items()
+                            if key in ("type", "text")
+                        },
+                    )
+                else:
+                    sanitized_content.append(
+                        {
+                            key: value
+                            for key, value in item.items()
+                            if key not in _CHAT_COMPLETION_RUNTIME_KEYS
+                        },
+                    )
+
+            sanitized_message["content"] = sanitized_content
+
+        sanitized_messages.append(sanitized_message)
+
+    return sanitized_messages
+
+
 def _battr(block: Any, key: str, default: Any = None) -> Any:
     """Read an attribute from a dict *or* Pydantic block."""
     if isinstance(block, dict):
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Independent review at exact head `96b70753a19718f12296eea59a18cd76bc630891`.

I found that the sanitizer currently runs on the wrong side of the AgentScope formatter, so it does not sanitize the real Chat Completions wire payload described in #6803.

`OpenAIChatModelCompat._call_api()` receives `list[Msg]` in normal model execution. The PR calls:

```python
messages = _sanitize_chat_completion_messages(messages)
return await super()._call_api(...)
```

Since each input is a `Msg`, the helper's `if not isinstance(message, dict)` branch leaves it unchanged. AgentScope's `OpenAIChatModel._call_api()` then runs:

```python
formatted_messages = await self.formatter.format(messages)
response = await client.chat.completions.create(
    messages=formatted_messages,
    ...
)
```

That formatting step is where the wire dictionaries are produced, after this PR's sanitizer has already returned.

I reproduced the issue with a formatter that emits the exact two polluted shapes reported in #6803, then allowed the call to pass through the real OpenAI 2.33.0 SDK into an `httpx.MockTransport` (no network). The test asserts that the final `/v1/chat/completions` JSON is clean.

```text
PR head 96b70753:
HOME=/tmp/qwenpaw-pr6809-wire-home \
  .venv/bin/pytest -q /tmp/test_qwenpaw_pr6809_wire_sanitize.py

FAILED
actual system part:
  {"type":"input_text","text":"system prompt","status":"completed"}
expected:
  {"type":"text","text":"system prompt"}

The captured user part also still contains:
  delta, index, status, object, msg_id
```

The identical wire test fails on current `main` (`bacac741`) with the same payload. This means the PR head has not changed the actual request boundary. In contrast, the committed tests pass:

```text
HOME=/tmp/qwenpaw-pr6809-home \
  .venv/bin/pytest -q \
  tests/unit/providers/test_openai_chat_content_sanitize.py
8 passed in 0.60s
```

The coverage gap is in `test_call_api_passes_sanitized_messages_to_base`: it passes dictionaries directly to `_call_api` and monkeypatches the base `_call_api`, so it bypasses both the real `Msg` input contract and `self.formatter.format()`.

Suggested correction:

1. Apply `_sanitize_chat_completion_messages()` to the result of the configured formatter, not to its `Msg` inputs. A small formatter wrapper/decorator would preserve AgentScope's request and response handling without duplicating the base `_call_api`.
2. Keep the helper's current conversion tests.
3. Replace or supplement the base-method mock with an HTTP transport regression that uses a real `Msg`, a formatter emitting the #6803 payload, and asserts the final `/v1/chat/completions` JSON.
4. Retain the original formatter instance and its provider-specific behavior; only sanitize its returned ephemeral wire dictionaries.

This finding is limited to sanitizer placement. The field filtering and non-mutation behavior of the helper itself look consistent with the stated scope.

RerankerGuo (changes requested):
Intent: strip Responses/runtime-only fields from the final OpenAI Chat Completions request without mutating the source messages.

The sanitizer logic itself handles the reported dictionary shapes, but it runs before AgentScope creates those dictionaries, so the production wire payload remains unchanged.

| No. | Blocking issue | Evidence | Required fix |
|---|---|---|---|
| 1 | Sanitizer executes before formatter output exists | `src/qwenpaw/providers/openai_chat_model_compat.py:807` applies `_sanitize_chat_completion_messages()` to the normal `list[Msg]` input. Every `Msg` takes the helper non-dict branch; `OpenAIChatModel._call_api()` formats those messages only afterward. An exact-head test using a real `Msg`, formatter, OpenAI 2.33.0 client, and `httpx.MockTransport` captures unchanged `input_text`, `status`, `delta`, `index`, `object`, and `msg_id` in `/v1/chat/completions`. | Sanitize the configured formatter result, for example through a formatter wrapper that preserves provider-specific behavior. |
| 2 | Committed integration test bypasses the failing boundary | `tests/unit/providers/test_openai_chat_content_sanitize.py:149-165` supplies dictionaries directly and monkeypatches the base `_call_api`, bypassing both the real `Msg` contract and `self.formatter.format()`. The committed suite passes 8 tests while the real wire regression still fails on both main and this head. | Add a transport-level regression using a real `Msg` and formatter-emitted polluted parts, then assert the final request JSON is sanitized. |

```mermaid
flowchart LR
    A[Real Msg inputs] --> B[Current sanitizer sees non-dicts]
    B --> C[AgentScope formatter creates wire dictionaries]
    C --> D[Polluted Chat Completions request]
    A --> E[Required formatter wrapper]
    E --> F[Sanitize formatted dictionaries]
    F --> G[Clean wire request]
    style D fill:#ffcdd2,color:#b71c1c
    style F fill:#bbdefb,color:#0d47a1
```

qbc2016:
The linked issue, #6803, has now been closed as invalid.

The reporter confirmed that:

- `"Respond like terse caveman..."` was injected by 9router.
- The `input_text` content type came from the gateway's prompt-injection serialization.
- When QwenPaw was connected directly to StepFun, the request succeeded without `input_text` or any streaming-field leak.
- QwenPaw's Chat Completions formatter was working correctly.

Therefore, the behavior this PR is intended to fix is not produced by QwenPaw.

The runtime fields shown in `qwenpaw_query_error_*.json` are part of QwenPaw's internal pre-formatter message representation. They are removed when AgentScope's Chat Completions formatter reconstructs the provider request, so their presence in the error dump is not evidence of a wire-format leak.

The tests added by this PR use an artificial formatter that explicitly returns polluted content. They demonstrate that the sanitizer can clean a hypothetical invalid formatter output, but they do not reproduce a real QwenPaw production path.

In addition, the current formatter wrapper introduces regressions:

- It replaces a `FormatterBase` instance with `_ChatCompletionFormatterWrapper`, which is not a `FormatterBase`.
- QwenPaw's `_create_formatter_instance()` consequently raises a `TypeError` on the normal model-factory path.
- Existing provider tests expecting the configured OpenAI capping formatter fail.
- Attribute writes used by reasoning and media fallback logic are not forwarded to the wrapped formatter.

Since the underlying issue is not a QwenPaw bug and this PR introduces production-path regressions, I recommend closing this PR rather than merging or continuing to revise it.

If a real Chat Completions serialization problem is found in the future, it should be reported separately with a post-formatter HTTP request capture and a production-path reproduction.

PR #6817, opened 2026-08-08
feat: integrate AnySearch web search (SearchProvider + MCP)

anysearch-ai:
## Description

Integrate **AnySearch** (https://anysearch.com) into QwenPaw as the built-in web search capability, replacing Tavily:

1. **Fix MCP env-ref header binding defects** (`fix(mcp): handle empty env-ref header bindings safely`)
   - Fix: when `Authorization: Bearer ${VAR}` is configured but the environment variable is empty, the empty value is formatted into an illegal HTTP header (`Bearer ` trailing space) which h11 rejects, breaking the MCP connection at startup
   - Fix: Console no longer shows `${VAR}`-style env-backed headers (displayed blank)
   - Fix: editing and saving a client no longer silently deletes these env-backed headers
2. **Replace the built-in MCP client** (`feat(mcp): replace tavily with AnySearch as the built-in MCP client`)
   - Remove the `tavily_search` MCP template and ship AnySearch MCP instead (`https://api.anysearch.com/mcp`, streamable HTTP)
   - Disabled by default (`enabled=False`), follows the default ask policy — no blanket allow
3. **Abstract web_search behind a pluggable SearchProvider** (`feat(tools): abstract web_search behind a pluggable SearchProvider`)
   - Built-in `web_search` now uses AnySearch REST API by default (`https://api.anysearch.com/v1/search`, anonymous)
   - Setting `ANYSEARCH_API_KEY` automatically sends `Authorization: Bearer`
   - TavilyProvider kept as fallback (switch via `QWENPAW_WEBSEARCH_PROVIDER=tavily`)

**Related Issue:** None (partner integration)

**Security Considerations:** MCP client is disabled by default (`enabled=False`), follows the default ask policy, no blanket allow; header secrets are referenced via `${VAR}` env vars, never persisted in plaintext.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes（pre-commit not installed locally; CI will run it. black/flake8/pylint checked locally on changed files and pass）
- [x] I ran tests locally (`pytest` or as relevant) and they pass（19 related tests pass）
- [ ] Documentation updated (if needed)（README changes prepared separately, will be submitted in a follow-up PR）
- [x] Ready for review

## Testing

```bash
pytest tests/unit/config/test_mcp_config_defaults.py \
       tests/unit/drivers/adapters/test_credentials_bindings.py \
       tests/unit/agents/tools/test_websearch_providers.py \
       tests/integration/test_driver_mcp_http_flow.py \
       tests/integration/test_driver_mcp_env_ref_flow.py --no-cov
```

## Evidence

```bash
# 19 related unit/integration tests pass
$ pytest tests/unit/config/test_mcp_config_defaults.py tests/unit/drivers/adapters/test_credentials_bindings.py tests/unit/agents/tools/test_websearch_providers.py tests/integration/test_driver_mcp_http_flow.py tests/integration/test_driver_mcp_env_ref_flow.py --no-cov
19 passed in 0.51s
```

```text
# End-to-end against the real AnySearch endpoint (anonymous web_search call)
web_search 匿名调用(默认 AnySearchProvider): success
返回样例: [1] GitHub - agentscope-ai/QwenPaw: Your Personal AI Assistant; easy ... URL
```

```text
# MCP migration verified against the real legacy migration path
默认 clients: ['anysearch']   # tavily removed
anysearch: enabled=False, transport=streamable_http, url=https://api.anysearch.com/mcp
迁移后: enabled=False, policy_rules[0].effect=ask   # default ask policy
```

```text
# Code style on changed files (black --check)
All done! ✨ 🍰 ✨
2 files would be left unchanged.
```

## Additional Notes

- The PR surfaced a latent generic defect in QwenPaw (commit 1): any HTTP MCP client using a `${VAR}` header with an unset environment variable would fail to connect. It was never triggered before because no built-in client combined "optional auth" with "env-ref header" (Tavily uses stdio and never goes through the header binding path). AnySearch's "anonymous by default, optional key" configuration is the first to expose it.
- README changes (AnySearch integration docs, badge, news entry) are prepared but intentionally excluded from this PR; they will be submitted separately.

### DIFF
--- a/src/qwenpaw/agents/tools/websearch/__init__.py
+++ b/src/qwenpaw/agents/tools/websearch/__init__.py
@@ -6,8 +6,8 @@
 ``SearchProvider`` is the abstraction that lets QwenPaw swap the search
 backend without touching the tool surface.  The active provider is chosen
 via the ``QWENPAW_WEBSEARCH_PROVIDER`` environment variable
-(``anysearch`` by default, ``tavily`` to fall back to the legacy
-keyless backend).
+(``tavily`` by default for backward compatibility, ``anysearch`` as an
+explicit opt-in).
 """
 
 from __future__ import annotations
--- a/src/qwenpaw/agents/tools/websearch/__init__.py
+++ b/src/qwenpaw/agents/tools/websearch/__init__.py
@@ -1,0 +1,148 @@
+# -*- coding: utf-8 -*-
+# flake8: noqa: E501
+# pylint: disable=line-too-long
+"""Pluggable web search providers for the ``web_search`` tool.
+
+``SearchProvider`` is the abstraction that lets QwenPaw swap the search
+backend without touching the tool surface.  The active provider is chosen
+via the ``QWENPAW_WEBSEARCH_PROVIDER`` environment variable
+(``anysearch`` by default, ``tavily`` to fall back to the legacy
+keyless backend).
+"""
+
+from __future__ import annotations
+
+import os
+
+from abc import ABC, abstractmethod
+
+import httpx
+
+_TIMEOUT = 30
+
+
+async def _post(
+    url: str,
+    headers: dict,
+    payload: dict,
+) -> dict:
+    """Async HTTP POST with certificate validation always on."""
+    async with httpx.AsyncClient(timeout=_TIMEOUT) as client:
+        resp = await client.post(url, headers=headers, json=payload)
+    resp.raise_for_status()
+    return resp.json()
+
+
+def format_search_results(results: list[dict]) -> str:
+    """Format search results into readable text."""
+    if not results:
+        return "No results found."
+    lines: list[str] = []
+    for i, r in enumerate(results, 1):
+        title = r.get("title", "")
+        url = r.get("url", "")
+        content = r.get("content", "")
+        lines.append(f"[{i}] {title}")
+        lines.append(f"    URL: {url}")
+        if content:
+            lines.append(f"    {content}")
+        lines.append("")
+    return "\n".join(lines).rstrip()
+
+
+class SearchProvider(ABC):
+    """Abstract web search backend."""
+
+    name = ""
+
+    @abstractmethod
+    async def search(
+        self,
+        query: str,
+        max_results: int = 5,
+    ) -> list[dict]:
+        """Return a list of ``{title, url, snippet, content}`` dicts."""
+        raise NotImplementedError
+
+
+class TavilyProvider(SearchProvider):
+    """Legacy Tavily keyless search backend."""
+
+    name = "tavily"
+
+    _SEARCH_URL = "https://api.tavily.com/search"
+
+    async def search(
+        self,
+        query: str,
+        max_results: int = 5,
+    ) -> list[dict]:
+        payload = {
+            "query": query,
+            "max_results": max_results,
+            "search_depth": "basic",
+        }
+        data = await _post(
+            self._SEARCH_URL,
+            headers={
+                "Content-Type": "application/json",
+                "X-Tavily-Access-Mode": "keyless",
+            },
+            payload=payload,
+        )
+        return list(data.get("results") or [])
+
+
+class AnySearchProvider(SearchProvider):
+    """AnySearch REST search backend (https://api.anysearch.com)."""
+
+    name = "anysearch"
+
+    _SEARCH_URL = "https://api.anysearch.com/v1/search"
+
+    async def search(
+        self,
+        query: str,
+        max_results: int = 5,
+    ) -> list[dict]:
+        headers = {"Content-Type": "application/json"}
+        api_key = os.environ.get("ANYSEARCH_API_KEY", "")
+        if api_key:
+            headers["Authorization"] = f"Bearer {api_key}"
+        payload = {
+            "query": query,
+            "max_results": max_results,
+        }
+        data = await _post(
+            self._SEARCH_URL,
+            headers=headers,
+            payload=payload,
+        )
+        return list((data.get("data") or {}).get("results") or [])
+
+
+def get_search_provider() -> SearchProvider:
+    """Return the active search provider selected by env var.
+
+    Defaults to ``tavily`` (the legacy keyless backend). Set
+    ``QWENPAW_WEBSEARCH_PROVIDER=anysearch`` to use AnySearch.
+    Unknown values raise ``ValueError`` instead of silently routing.
+    """
+    choice = os.environ.get("QWENPAW_WEBSEARCH_PROVIDER", "").strip().lower()
+    if choice in {"", "tavily"}:
+        return TavilyProvider()
+    if choice == "anysearch":
+        return AnySearchProvider()
+    raise ValueError(
+        f"Unknown QWENPAW_WEBSEARCH_PROVIDER: {choice!r} "
+        "(expected 'tavily' or 'anysearch')",
+    )
+
+
+__all__ = [
+    "SearchProvider",
+    "TavilyProvider",
+    "AnySearchProvider",
+    "get_search_provider",
+    "format_search_results",
+]
--- a/src/qwenpaw/drivers/adapters/mcp_card_builder.py
+++ b/src/qwenpaw/drivers/adapters/mcp_card_builder.py
@@ -12,10 +12,13 @@ from .mcp_binding import (
     binding_to_response,
     classify_mcp_binding,
     normalize_secret_key,
+    parse_env_template,
+    plan_env_ref_bindings,
     restore_masked_value,
     source_binding_from_split,
     split_mcp_binding,
 )
+from .env_ref import env_ref
 from ..constants import (
     CAPABILITY_KIND_TOOL,
     CREDENTIAL_ALIAS_OAUTH,
@@ -50,7 +53,12 @@ def build_mcp_credential_record(
     *,
     existing: CredentialRecord | None = None,
 ) -> CredentialRecord:
-    """Build the static MCP credential record from console request data."""
+    """Build the static MCP credential record from console request data.
+
+    Values that embed a single ``${VAR}`` reference are env-backed and
+    never stored as static secrets — they resolve from the process
+    environment at runtime (see ``plan_env_ref_bindings``).
+    """
     ref = mcp_credential_ref(client_key)
     incoming_env = dict(_read_field(client, "env", {}) or {})
     incoming_headers = dict(_read_field(client, "headers", {}) or {})
@@ -62,6 +70,9 @@ def build_mcp_credential_record(
             classify_mcp_binding(section="env", key=str(key), value=str(value))
             == "public"
         ):
+            continue
+        template = parse_env_template(str(value))
+        if template is not None and template.is_single:
             continue
         secret_key = str(key)
         secrets[secret_key] = restore_masked_value(
@@ -79,6 +90,9 @@ def build_mcp_credential_record(
             )
             == "public"
         ):
+            continue
+        template = parse_env_template(str(value))
+        if template is not None and template.is_single:
             continue
         secret_key = normalize_secret_key(str(header), used)
         used.add(secret_key)
@@ -115,40 +129,51 @@ def build_mcp_driver_card(
     if transport == "stdio":
         env = dict(data.get("env") or {})
         public_env, secret_env = split_mcp_binding("env", env)
+        env_plan = plan_env_ref_bindings(secret_env)
+        env_binding = source_binding_from_split(
+            public_env,
+            {str(key): str(key) for key in env_plan.plain_secrets},
+            STATIC_CREDENTIAL_ALIAS,
+        )
+        env_binding.update(env_plan.env_bindings)
         endpoint: dict[str, Any] = {
             "transport": "stdio",
             "command": str(data.get("command") or ""),
             "args": list(data.get("args") or []),
-            "env": source_binding_from_split(
-                public_env,
-                {str(key): str(key) for key in secret_env},
-                STATIC_CREDENTIAL_ALIAS,
-            ),
+            "env": env_binding,
         }
         cwd = str(data.get("cwd") or "")
         if cwd:
             endpoint["cwd"] = cwd
+        env_aliases = env_plan.env_aliases
     else:
         headers = dict(data.get("headers") or {})
         public_headers, secret_headers = split_mcp_binding("headers", headers)
+        header_plan = plan_env_ref_bindings(secret_headers)
         used: set[str] = set()
         secret_refs: dict[str, str] = {}
-        for header in secret_headers:
+        for header in header_plan.plain_secrets:
             secret_key = normalize_secret_key(str(header), used)
             used.add(secret_key)
             secret_refs[str(header)] = secret_key
+        header_binding = source_binding_from_split(
+            public_headers,
+            secret_refs,
+            STATIC_CREDENTIAL_ALIAS,
+        )
+        header_binding.update(header_plan.env_bindings)
         endpoint = {
             "transport": transport,
             "url": str(data.get("url") or ""),
-            "headers": source_binding_from_split(
-                public_headers,
-                secret_refs,
-                STATIC_CREDENTIAL_ALIAS,
-            ),
+            "headers": header_binding,
         }
         _preserve_oauth_authorization_binding(existing, endpoint)
+        env_aliases = header_plan.env_aliases
 
     credentials = _credential_refs_from_existing(existing)
+    for alias, ref in list(credentials.items()):
+        if ref.ref.startswith("env:"):
+            credentials.pop(alias, None)
     if secrets:
         credentials[STATIC_CREDENTIAL_ALIAS] = CredentialRef(
             kind=CREDENTIAL_KIND_STATIC,
@@ -156,6 +181,11 @@ def build_mcp_driver_card(
         )
     else:
         credentials.pop(STATIC_CREDENTIAL_ALIAS, None)
+    for alias, var in env_aliases.items():
+        credentials[alias] = CredentialRef(
+            kind=CREDENTIAL_KIND_STATIC,
+            ref=env_ref(var),
+        )
     # Generic DriverPolicy defaults to deny as the low-level safe fallback.
     # Console-created MCP clients default to ask so a newly added external
     # server requires human approval instead of failing silently before the
@@ -180,6 +210,21 @@ def build_mcp_driver_card(
     )
 
 
+def _env_aliases_from_credentials(
+    credentials: dict[str, CredentialRef],
+) -> dict[str, str]:
+    """Return alias -> var name for ``env:`` credential refs.
+
+    Preserves the original variable name (case-sensitive) recorded on the
+    card, instead of reverse-engineering it from the lowercased alias.
+    """
+    aliases: dict[str, str] = {}
+    for alias, ref in (credentials or {}).items():
+        if ref.ref.startswith("env:"):
+            aliases[str(alias)] = ref.ref[len("env:") :]
+    return aliases
+
+
 def build_mcp_client_info_payload(
     card: DriverCard,
     credential: CredentialRecord | None,
@@ -188,15 +233,18 @@ def build_mcp_client_info_payload(
     """Return the MCPClientInfo-compatible API response payload."""
     endpoint = card.endpoint
     transport = str(endpoint.get("transport") or "stdio")
+    env_aliases = _env_aliases_from_credentials(card.credentials)
     env = binding_to_response(
         endpoint.get("env") or {},
         credential,
         credential_alias=STATIC_CREDENTIAL_ALIAS,
+        env_aliases=env_aliases,
     )
     headers = binding_to_response(
         endpoint.get("headers") or {},
         credential,
         credential_alias=STATIC_CREDENTIAL_ALIAS,
+        env_aliases=env_aliases,
     )
     return {
         "key": card.name,
@@ -310,6 +358,7 @@ def _card_to_client_data(card: DriverCard | None) -> dict[str, Any]:
     if card is None:
         return {}
     endpoint = card.endpoint
+    env_aliases = _env_aliases_from_credentials(card.credentials)
     return {
         "name": card.config.get("display_name") or card.name,
         "description": card.config.get("description") or "",
@@ -319,12 +368,14 @@ def _card_to_client_data(card: DriverCard | None) -> dict[str, Any]:
         "headers": binding_plain_keys(
             endpoint.get("headers") or {},
             credential_alias=STATIC_CREDENTIAL_ALIAS,
+            env_aliases=env_aliases,
         ),
         "command": endpoint.get("command") or "",
         "args": list(endpoint.get("args") or []),
         "env": binding_plain_keys(
             endpoint.get("env") or {},
             credential_alias=STATIC_CREDENTIAL_ALIAS,
+            env_aliases=env_aliases,
         ),
         "cwd": endpoint.get("cwd") or "",
     }
--- a/src/qwenpaw/agents/tools/web_search.py
+++ b/src/qwenpaw/agents/tools/web_search.py
@@ -3,7 +3,8 @@
 # pylint: disable=line-too-long
 """Web search and fetch tools.
 
-web_search uses Tavily keyless API.
+web_search delegates to a pluggable search provider (AnySearch by
+default, Tavily when QWENPAW_WEBSEARCH_PROVIDER=tavily).
 web_fetch uses direct HTTP GET + html2text.
 """
 
@@ -21,12 +22,11 @@ from agentscope.tool import ToolChunk
 from agentscope.message import ToolResultState
 
 from ...runtime.tool_registry import tool_descriptor
+from .websearch import format_search_results, get_search_provider
 
 logger = logging.getLogger(__name__)
 
-_TAVILY_SEARCH_URL = "https://api.tavily.com/search"
 _DEFAULT_TIMEOUT = 30
-_DEFAULT_MAX_RESULTS = 5
 
 _SEARCH_FALLBACK_HINT = (
     "This tool uses a free API with rate limits. "
@@ -64,57 +64,6 @@ def _is_ssl_error(exc: BaseException) -> bool:
             return True
         cur = cur.__cause__
     return False
-
-
-async def _post(
-    url: str,
-    headers: dict,
-    payload: dict,
-) -> dict:
-    """Async HTTP POST with SSL-verification fallback."""
-    try:
-        async with httpx.AsyncClient(
-            timeout=_DEFAULT_TIMEOUT,
-        ) as client:
-            resp = await client.post(
-                url,
-                headers=headers,
-                json=payload,
-            )
-    except Exception as first_exc:
-        if not _is_ssl_error(first_exc):
-            raise
-        logger.warning(
-            f"SSL verify failed for {url}, retrying",
-        )
-        async with httpx.AsyncClient(
-            timeout=_DEFAULT_TIMEOUT,
-            verify=False,
-        ) as client:
-            resp = await client.post(
-                url,
-                headers=headers,
-                json=payload,
-            )
-    resp.raise_for_status()
-    return resp.json()
-
-
-def _format_search_results(results: list[dict]) -> str:
-    """Format Tavily search results into readable text."""
-    if not results:
-        return "No results found."
-    lines: list[str] = []
-    for i, r in enumerate(results, 1):
-        title = r.get("title", "")
-        url = r.get("url", "")
-        content = r.get("content", "")
-        lines.append(f"[{i}] {title}")
-        lines.append(f"    URL: {url}")
-        if content:
-            lines.append(f"    {content}")
-        lines.append("")
-    return "\n".join(lines).rstrip()
 
 
 async def _fetch_html(url: str) -> str:
@@ -225,20 +174,12 @@ async def web_search(search_term: str) -> ToolChunk:
         )
 
     try:
-        data = await _post(
-            _TAVILY_SEARCH_URL,
-            headers={
-                "Content-Type": "application/json",
-                "X-Tavily-Access-Mode": "keyless",
-            },
-            payload={
-                "query": query,
-                "max_results": _DEFAULT_MAX_RESULTS,
-                "search_depth": "basic",
-            },
+        provider = get_search_provider()
+        results = await provider.search(
+            query,
+            max_results=5,
         )
-        results = data.get("results", [])
-        text = _format_search_results(results)
+        text = format_search_results(results)
         if not text:
             text = "No content searched."
         return ToolChunk(
--- a/src/qwenpaw/config/config.py
+++ b/src/qwenpaw/config/config.py
@@ -2002,17 +2002,20 @@ class MCPConfig(BaseModel):
     """MCP clients configuration.
 
     Uses a dict to allow dynamic client definitions.
-    Default tavily_search client is created and auto-enabled if API key exists.
+    Default anysearch client is provided but disabled by default; enable it
+    in the Console to use AnySearch through MCP. Access follows the default
+    ask policy (no blanket allow).
     """
 
     clients: Dict[str, MCPClientConfig] = Field(
         default_factory=lambda: {
-            "tavily_search": MCPClientConfig(
-                name="tavily_mcp",
+            "anysearch": MCPClientConfig(
+                name="anysearch_mcp",
                 enabled=False,
-                command="npx",
-                args=["-y", "tavily-mcp@latest"],
-                env={"TAVILY_API_KEY": ""},
+                transport="streamable_http",
+                url="https://api.anysearch.com/mcp",
+                headers={"Authorization": "Bearer ${ANYSEARCH_API_KEY}"},
+                description="AnySearch web search via MCP",
             ),
         },
     )
--- a/src/qwenpaw/drivers/adapters/mcp_binding.py
+++ b/src/qwenpaw/drivers/adapters/mcp_binding.py
@@ -144,6 +144,7 @@ def plan_env_ref_bindings(secrets: dict[str, str]) -> EnvRefPlan:
     migration can warn about them.
     """
     plan = EnvRefPlan()
+    used_aliases: set[str] = set()
     for raw_key, raw_value in dict(secrets or {}).items():
         key = str(raw_key)
         value = str(raw_value)
@@ -151,6 +152,12 @@ def plan_env_ref_bindings(secrets: dict[str, str]) -> EnvRefPlan:
         if template is not None and template.is_single:
             var = template.var_names[0]
             alias = env_alias(var)
+            if alias in used_aliases:
+                index = 2
+                while f"{alias}_{index}" in used_aliases:
+                    index += 1
+                alias = f"{alias}_{index}"
+            used_aliases.add(alias)
             plan.env_aliases[alias] = var
             spec: dict[str, str] = {
                 "source": "credential",
@@ -167,11 +174,47 @@ def plan_env_ref_bindings(secrets: dict[str, str]) -> EnvRefPlan:
     return plan
 
 
+def _env_ref_template(
+    spec: dict[str, Any],
+    env_aliases: dict[str, str] | None = None,
+) -> str | None:
+    """Render an env-backed credential binding as its ${VAR} template.
+
+    Returns e.g. ``Bearer ${ANYSEARCH_API_KEY}`` for a binding whose
+    ``credential`` alias is ``env_<var>`` with a ``format`` of
+    ``Bearer {value}``.  ``None`` when the binding is not env-backed.
+
+    The original variable name (preserving case) is taken from
+    ``env_aliases`` (alias -> var name, derived from the card's
+    ``env:`` credential refs); when absent the alias is uppercased as a
+    fallback.
+    """
+    credential = str(spec.get("credential") or "")
+    if not credential.startswith("env_"):
+        return None
+    var_name = ""
+    if env_aliases:
+        var_name = env_aliases.get(credential, "")
+    if not var_name:
+        var_name = credential[len("env_") :].upper()
+    fmt = str(spec.get("format") or "{value}")
+    return fmt.replace("{value}", "${" + var_name + "}")
+
+
+def _is_env_ref_spec(spec: Any) -> bool:
+    return (
+        isinstance(spec, dict)
+        and spec.get("source") == "credential"
+        and str(spec.get("credential") or "").startswith("env_")
+    )
+
+
 def binding_to_response(
     binding: Any,
     credential: CredentialRecord | None,
     *,
     credential_alias: str,
+    env_aliases: dict[str, str] | None = None,
 ) -> dict[str, str]:
     """Return masked Console response values from a Driver endpoint binding."""
     if not isinstance(binding, dict):
@@ -182,6 +225,10 @@ def binding_to_response(
         for key, spec in binding.items():
             if isinstance(spec, dict) and spec.get("source") == "literal":
                 result[str(key)] = str(spec.get("value") or "")
+            elif _is_env_ref_spec(spec):
+                template = _env_ref_template(spec, env_aliases)
+                if template:
+                    result[str(key)] = template
             elif (
                 isinstance(spec, dict)
                 and spec.get("source") == "credential"
@@ -209,6 +256,7 @@ def binding_plain_keys(
     binding: Any,
     *,
     credential_alias: str,
+    env_aliases: dict[str, str] | None = None,
 ) -> dict[str, str]:
     """Return unmasked public values and blank placeholders for secret keys."""
     if not isinstance(binding, dict):
@@ -218,6 +266,10 @@ def binding_plain_keys(
         for key, spec in binding.items():
             if isinstance(spec, dict) and spec.get("source") == "literal":
                 result[str(key)] = str(spec.get("value") or "")
+            elif _is_env_ref_spec(spec):
+                template = _env_ref_template(spec, env_aliases)
+                if template:
+                    result[str(key)] = template
             elif (
                 isinstance(spec, dict)
                 and spec.get("source") == "credential"
--- a/src/qwenpaw/drivers/adapters/mcp_binding.py
+++ b/src/qwenpaw/drivers/adapters/mcp_binding.py
@@ -201,12 +201,18 @@ def _env_ref_template(
     return fmt.replace("{value}", "${" + var_name + "}")
 
 
-def _is_env_ref_spec(spec: Any) -> bool:
-    return (
-        isinstance(spec, dict)
-        and spec.get("source") == "credential"
-        and str(spec.get("credential") or "").startswith("env_")
-    )
+def _is_env_ref_spec(
+    spec: Any,
+    env_aliases: dict[str, str] | None = None,
+) -> bool:
+    """Whether a binding is env-backed, decided by the env: credential
+    refs (via env_aliases) rather than the alias naming convention — a
+    static alias like ``env_custom`` must not be misclassified."""
+    if not (isinstance(spec, dict) and spec.get("source") == "credential"):
+        return False
+    if not env_aliases:
+        return False
+    return str(spec.get("credential") or "") in env_aliases
 
 
 def binding_to_response(
@@ -225,7 +231,7 @@ def binding_to_response(
         for key, spec in binding.items():
             if isinstance(spec, dict) and spec.get("source") == "literal":
                 result[str(key)] = str(spec.get("value") or "")
-            elif _is_env_ref_spec(spec):
+            elif _is_env_ref_spec(spec, env_aliases):
                 template = _env_ref_template(spec, env_aliases)
                 if template:
                     result[str(key)] = template
@@ -266,7 +272,7 @@ def binding_plain_keys(
         for key, spec in binding.items():
             if isinstance(spec, dict) and spec.get("source") == "literal":
                 result[str(key)] = str(spec.get("value") or "")
-            elif _is_env_ref_spec(spec):
+            elif _is_env_ref_spec(spec, env_aliases):
                 template = _env_ref_template(spec, env_aliases)
                 if template:
                     result[str(key)] = template
--- a/src/qwenpaw/drivers/credentials/bindings.py
+++ b/src/qwenpaw/drivers/credentials/bindings.py
@@ -134,6 +134,7 @@ def lookup_credential_value(
     return None
 
 
+# pylint: disable-next=too-many-return-statements
 def _resolve_value_source(
     spec: Any,
     credentials: dict[str, ResolvedCredential],
@@ -154,11 +155,14 @@ def _resolve_value_source(
         if field
         else None
     )
-    if value is None:
+    if value is None or value == "":
         return None
 
     text = str(value)
     fmt = spec.get("format")
-    if isinstance(fmt, str) and fmt:
-        return fmt.replace("{value}", text)
-    return text
+    if not (isinstance(fmt, str) and fmt):
+        return text
+    resolved = fmt.replace("{value}", text)
+    if not resolved:
+        return None
+    return resolved
--- a/tests/integration/test_driver_mcp_http_flow.py
+++ b/tests/integration/test_driver_mcp_http_flow.py
@@ -71,3 +71,70 @@ async def test_driver_mcp_http_header_secret_flow(
         FakeHttpClient.instances[0].kwargs["headers"]
         == result.value["headers"]
     )
+
+
+@pytest.mark.asyncio
+@pytest.mark.integration
+@pytest.mark.p1
+async def test_driver_mcp_anysearch_default_config_shape(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Test purpose:
+    - Verify the default MCP config ships an anysearch streamable_http
+      template that migrates to a buildable DriverCard.
+
+    Test flow:
+    1. Build MCPConfig() and grab the anysearch client entry.
+    2. Convert it to a DriverCard via the legacy migration helper.
+    3. Register the card and build the driver handler (HTTP faked).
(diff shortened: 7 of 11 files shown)


RerankerGuo:
Independent review at exact head `d8cbc4d995e824f167ba33cdf5ff1edf556830c3`.

The env-ref header fix is only one-way. The GET response now displays an env-backed binding as `Bearer ${ANYSEARCH_API_KEY}`, but the Console save path converts that template into a **static literal secret**. Editing any unrelated client property therefore breaks the new default AnySearch MCP authentication.

Production flow:

1. `legacy_mcp_client_to_driver()` correctly converts the default header to:

```python
{
    "source": "credential",
    "credential": "env_anysearch_api_key",
    "field": "value",
    "format": "Bearer {value}",
}
```

2. `build_mcp_client_info_payload()` returns:

```python
{"Authorization": "Bearer ${ANYSEARCH_API_KEY}"}
```

3. When the Console submits that unchanged payload, `build_mcp_credential_record()` stores the whole template under the static credential. `build_mcp_driver_card()` then rebuilds the header with `credential="static"` rather than preserving/reconstructing the env binding.

4. Runtime resolution sends the literal string:

```text
Authorization: Bearer ${ANYSEARCH_API_KEY}
```

It no longer reads the process environment.

I reproduced this with a temporary exact-head test using only production helpers:

```python
client = MCPConfig().clients["anysearch"]
original, _ = legacy_mcp_client_to_driver("anysearch", client)
payload = build_mcp_client_info_payload(original, None)

static_record = build_mcp_credential_record("anysearch", payload)
saved = build_mcp_driver_card(
    "anysearch",
    payload,
    mcp_credential_ref("anysearch"),
    credential_record=static_record,
    existing=original,
)

headers = resolve_binding(
    saved.endpoint["headers"],
    {
        "static": ResolvedCredential(
            kind="static",
            secrets=static_record.secrets,
        ),
        "env_anysearch_api_key": ResolvedCredential(
            kind="env",
            secrets={"value": "live-token"},
        ),
    },
)
assert headers == {"Authorization": "Bearer live-token"}
```

Exact-head result:

```text
Expected: Bearer live-token
Actual:   Bearer ${ANYSEARCH_API_KEY}
```

The committed tests remain green:

```text
PR focused suite: 19 passed
all driver adapter tests: 34 passed
```

The two new `binding_to_response()` / `binding_plain_keys()` tests only prove that the template is visible. They do not exercise the subsequent save and runtime resolution.

Suggested correction:

1. During save, either preserve an unchanged existing env-backed binding or parse `${VAR}` values with the same `plan_env_ref_bindings()` semantics used by legacy migration.
2. Exclude env-ref values from the static credential record.
3. Retain/create the corresponding `CredentialRef(ref="env:ANYSEARCH_API_KEY")` and source binding.
4. Add the migration -> API payload -> unchanged Console save -> runtime resolve regression above for both headers and stdio env.

There is also a generic case-preservation issue in `_env_ref_template()`: it reconstructs the variable name by uppercasing the lowercased credential alias. `${ApiKey}` therefore comes back as `${APIKEY}`, which points to a different variable on case-sensitive systems. The original variable is available in the card's credential ref (`env:ApiKey`), so response serialization should use that source of truth rather than reverse-engineering the alias.

RerankerGuo (changes requested):
Intent: add AnySearch while preserving optional environment-backed MCP authentication across migration, Console display/edit/save, and runtime resolution.

The anonymous search path works, but the credential fix is only one-way: GET reconstructs `${VAR}` for display, while an unchanged Console save stores that template as a static secret.

| No. | Blocking issue | Evidence | Required fix |
|---|---|---|---|
| 1 | Unchanged env-backed headers become static literals on save | `build_mcp_client_info_payload()` returns `Bearer ${ANYSEARCH_API_KEY}`, then `build_mcp_credential_record()` at `src/qwenpaw/drivers/adapters/mcp_card_builder.py:73-86` stores the whole string as a static secret and `build_mcp_driver_card()` at lines 132-147 binds it to `static`. An exact production-helper round trip resolves `Bearer ${ANYSEARCH_API_KEY}` instead of the supplied environment value `Bearer live-token`. | Preserve unchanged env-backed bindings or parse `${VAR}` through `plan_env_ref_bindings()`, exclude them from the static record, retain the `env:` CredentialRef, and add migration -> API -> unchanged save -> runtime tests for headers and stdio env. |
| 2 | Response serialization changes case-sensitive variable names | `_env_ref_template()` at `src/qwenpaw/drivers/adapters/mcp_binding.py:177-182` reconstructs the variable by uppercasing a lowercased alias. `${ApiKey}` therefore becomes `${APIKEY}` and targets a different variable on case-sensitive systems. | Render from the original `env:ApiKey` credential reference rather than reverse-engineering the alias; add a mixed-case regression. |

The committed 19 focused and all 34 adapter tests pass because they stop before the save/runtime boundary. The exact head remains `d8cbc4d995e824f167ba33cdf5ff1edf556830c3` and cleanly merges with current main via tree `9acc62e4`.

```mermaid
flowchart LR
    A[env ANYSEARCH_API_KEY] --> B[Migrated env credential binding]
    B --> C[Console displays template]
    C --> D[Unchanged save]
    D -->|current| E[Static literal template]
    E --> F[Broken Authorization header]
    D -->|required| G[Preserved env credential ref]
    G --> H[Runtime resolves live token]
    style G fill:#bbdefb,color:#0d47a1
    style H fill:#c8e6c9,color:#1a5e20
```

anysearch-ai:
> Independent review at exact head `d8cbc4d995e824f167ba33cdf5ff1edf556830c3`.
> 
> The env-ref header fix is only one-way. The GET response now displays an env-backed binding as `Bearer ${ANYSEARCH_API_KEY}`, but the Console save path converts that template into a **static literal secret**. Editing any unrelated client property therefore breaks the new default AnySearch MCP authentication.
> 
> Production flow:
> 
> 1. `legacy_mcp_client_to_driver()` correctly converts the default header to:
> 
> ```python
> {
>     "source": "credential",
>     "credential": "env_anysearch_api_key",
>     "field": "value",
>     "format": "Bearer {value}",
> }
> ```
> 
> 2. `build_mcp_client_info_payload()` returns:
> 
> ```python
> {"Authorization": "Bearer ${ANYSEARCH_API_KEY}"}
> ```
> 
> 3. When the Console submits that unchanged payload, `build_mcp_credential_record()` stores the whole template under the static credential. `build_mcp_driver_card()` then rebuilds the header with `credential="static"` rather than preserving/reconstructing the env binding.
> 4. Runtime resolution sends the literal string:
> 
> ```
> Authorization: Bearer ${ANYSEARCH_API_KEY}
> ```
> 
> It no longer reads the process environment.
> 
> I reproduced this with a temporary exact-head test using only production helpers:
> 
> ```python
> client = MCPConfig().clients["anysearch"]
> original, _ = legacy_mcp_client_to_driver("anysearch", client)
> payload = build_mcp_client_info_payload(original, None)
> 
> static_record = build_mcp_credential_record("anysearch", payload)
> saved = build_mcp_driver_card(
>     "anysearch",
>     payload,
>     mcp_credential_ref("anysearch"),
>     credential_record=static_record,
>     existing=original,
> )
> 
> headers = resolve_binding(
>     saved.endpoint["headers"],
>     {
>         "static": ResolvedCredential(
>             kind="static",
>             secrets=static_record.secrets,
>         ),
>         "env_anysearch_api_key": ResolvedCredential(
>             kind="env",
>             secrets={"value": "live-token"},
>         ),
>     },
> )
> assert headers == {"Authorization": "Bearer live-token"}
> ```
> 
> Exact-head result:
> 
> ```
> Expected: Bearer live-token
> Actual:   Bearer ${ANYSEARCH_API_KEY}
> ```
> 
> The committed tests remain green:
> 
> ```
> PR focused suite: 19 passed
> all driver adapter tests: 34 passed
> ```
> 
> The two new `binding_to_response()` / `binding_plain_keys()` tests only prove that the template is visible. They do not exercise the subsequent save and runtime resolution.
> 
> Suggested correction:
> 
> 1. During save, either preserve an unchanged existing env-backed binding or parse `${VAR}` values with the same `plan_env_ref_bindings()` semantics used by legacy migration.
> 2. Exclude env-ref values from the static credential record.
> 3. Retain/create the corresponding `CredentialRef(ref="env:ANYSEARCH_API_KEY")` and source binding.
> 4. Add the migration -> API payload -> unchanged Console save -> runtime resolve regression above for both headers and stdio env.
> 
> There is also a generic case-preservation issue in `_env_ref_template()`: it reconstructs the variable name by uppercasing the lowercased credential alias. `${ApiKey}` therefore comes back as `${APIKEY}`, which points to a different variable on case-sensitive systems. The original variable is available in the card's credential ref (`env:ApiKey`), so response serialization should use that source of truth rather than reverse-engineering the alias.

---
Thank you for the thorough review and the reproduction script — it was spot on. I've fixed the one-way env-ref handling and added regression tests covering the full round trip.

**Fix summary:**

1. **Console save path now preserves env-backed bindings**
   - `build_mcp_credential_record()` skips values embedding a single `${VAR}` — they are never stored as static secrets.
   - `build_mcp_driver_card()` routes secret env/header values through `plan_env_ref_bindings()` (the same semantics as `legacy_mcp_client_to_driver()`), so an unchanged save keeps the `env:` credential ref and the `source: credential` binding instead of degrading to a static literal.

2. **Stale env refs are cleaned up**
   - When a user switches a header from `${VAR}` to a static literal, the old `env_*` credential ref is removed from the card.

3. **Env var case is preserved**
   - `binding_to_response()` / `binding_plain_keys()` now take `env_aliases` derived from the card's `env:` credential refs, so `${ApiKey}` round-trips as `${ApiKey}` (not the uppercased alias `${APIKEY}`).

**Regression tests added** (`tests/unit/drivers/adapters/test_mcp_legacy_env_ref.py`):
- migration → API payload → unchanged Console save → runtime resolve (header env ref)
- same round trip for stdio env refs
- static secrets still stored/resolved correctly
- env var case preservation
- env → static switch drops the stale env ref

**Verification against your exact reproduction:**

```python
# Before fix: headers == {"Authorization": "Bearer ${ANYSEARCH_API_KEY}"}
# After fix:
headers == {"Authorization": "Bearer live-token"}   # ✅ resolves from process env
Full run: 52 related tests pass (test_mcp_config_defaults, test_credentials_bindings, test_websearch_providers, test_driver_mcp_http_flow, test_driver_mcp_env_ref_flow, test_mcp_legacy_env_ref). black / flake8 / pylint clean on changed files.

File: src/qwenpaw/drivers/adapters/mcp_card_builder.py [hunk 1]
```diff
@@ -63,6 +71,8 @@ def build_mcp_credential_record(
             == "public"
         ):
             continue
+        if parse_env_template(str(value)) is not None:
```
cuiyuebing: 结合`line 93`与 `src/qwenpaw/drivers/adapters/mcp_binding.py:150`一起看，这几段代码可以正确处理只包含一个环境变量的配置，例如`Authorization: Bearer ${API_KEY}`，但处理不了两个环境变量拼在一起的配置，例如：`Authorization: ${USER}:${PASS}`

保存第二种配置时，代码一方面认为它“包含环境变量”，所以不把值保存到凭证库；另一方面又认为它不能按单个环境变量处理，于是生成一个指向静态凭证的配置。结果是配置指向的凭证根本不存在。

修复建议：只有代码确实支持的单变量格式，例如 `Bearer ${API_KEY}`，才跳过静态保存。`${USER}:${PASS}` 这种多变量格式如果暂时不支持，应选择下面一种明确行为：

1. 按现有规则把整段内容作为静态值保存；或
2. 保存时直接报错，告诉用户当前不支持这种格式。

不能继续接受保存请求后悄悄把值丢掉。

src/qwenpaw/agents/tools/websearch/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+
+    try:
+        async with httpx.AsyncClient(timeout=_TIMEOUT) as client:
+            resp = await client.post(url, headers=headers, json=payload)
+    except Exception as first_exc:
+        if not _is_ssl_error(first_exc):
+            raise
+        logger.warning("SSL verify failed for %s, retrying", url)
+        async with httpx.AsyncClient(timeout=_TIMEOUT, verify=False) as client:
+            resp = await client.post(url, headers=headers, json=payload)
```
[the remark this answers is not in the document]
anysearch-ai: This is a pre-existing pattern, not introduced by this PR. The
`verify=False` retry fallback was already present in the original
`web_search.py` (Tavily backend) — the same SSL-fallback pattern
exists in `_fetch_html`. The migration to `websearch/__init__.py`
kept the project's established behavior.
It only applies when the initial request fails with an SSL error
(the fallback path), and is needed to support environments with
self-signed certificates or intercepting proxies. We've kept it
consistent with the existing codebase behavior.

src/qwenpaw/drivers/adapters/mcp_card_builder.py
anysearch-ai: 已修复，按建议 1 处理：`${USER}:${PASS}` 这类多变量模板不再被静默丢弃，
保存时按现有规则把整段内容作为静态值整体保存，运行时原样解析。

新增回归测试 `test_console_save_treats_multi_env_ref_as_static` 覆盖该场景。

cuiyuebing (approved):
LGTM

rayrayraykk (changes requested):
# PR #6817 — Request Changes

- PR: [agentscope-ai/QwenPaw#6817](https://github.com/agentscope-ai/QwenPaw/pull/6817)
- Title: `feat: integrate AnySearch web search (SearchProvider + MCP)`
- Base: `main@ca40a47c024e11b591ca320fe3b6564e3aa26762`
- Reviewed head: `ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed`
- Scope: 11 files, `+884/-90`
- Review date: 2026-08-10
- Verdict: **Request changes.** This design should not merge as it stands.
  QwenPaw already has an enabled built-in `web_search` tool, and
  this PR makes that tool use AnySearch while also adding a second AnySearch
  entry through the default MCP configuration. That creates two overlapping
  search surfaces with different configuration, credential, tool-discovery,
  and approval semantics. The PR also leaves two reproducible generic
  credential-binding regressions.

## Checklist

- [x] Read the latest PR diff, commits, discussions, and CI state
- [x] Check synchronous I/O and event-loop boundaries
- [x] Check Windows and macOS compatibility
- [x] Check functional behavior and regression boundaries
- [x] Check code clarity, responsibility boundaries, and modularity
- [x] Check overlap between built-in tools and default MCP capabilities
- [x] Check provider/API-key configuration and secret-storage boundaries
- [x] Run the focused tests against the exact reviewed head
- [x] Reproduce the remaining edge cases with production helpers
- [x] Recheck the report after the PR was force-pushed during review

## Findings

### [Blocking design] Keep one canonical web-search surface

Locations:

- [`web_search.py:136-182`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/agents/tools/web_search.py#L136-L182)
- [`websearch/__init__.py:96-129`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/agents/tools/websearch/__init__.py#L96-L129)
- [`config.py:1956-1976`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/config/config.py#L1956-L1976)
- [`tools.py:104-177`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/app/routers/tools.py#L104-L177)
- [`tools.py:304-444`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/app/routers/tools.py#L304-L444)

The AnySearch MCP client is created by default but disabled by default. It is
not “enabled by default,” but that distinction does not resolve the design
conflict: the built-in `web_search` tool is already enabled, this PR changes
its default backend to AnySearch, and the same PR places another AnySearch
search surface in every user's MCP configuration.

Once a user enables that MCP entry, the model can see two ways to perform the
same basic operation:

```text
built-in web_search  -> AnySearch REST API
default AnySearch MCP -> AnySearch MCP search tools
```

These are not two interchangeable implementations behind one interface. They
behave differently at every product boundary:

| Concern | Built-in `web_search` | AnySearch MCP |
|---|---|---|
| Provider selection | Process-wide `QWENPAW_WEBSEARCH_PROVIDER` | Fixed MCP endpoint |
| API key | Process-wide `ANYSEARCH_API_KEY` | MCP env credential binding |
| Configuration UI | No provider/key fields | MCP client editor |
| Tool surface | One `web_search` tool | MCP-advertised search tools |
| Default governance | `allow` | `ask` |
| Scope | Process environment | Agent/workspace driver configuration |

The governance mismatch is especially problematic. The same external search
can be automatically allowed through the built-in tool but require approval
through MCP. Which path is used is left to model tool selection, not to a
single user-visible policy decision. Credentials can also disagree: the
built-in path may read one process environment key while the MCP card resolves
a different per-workspace binding.

Required direction before merge: remove the default AnySearch MCP entry from
this PR and keep the existing built-in `web_search` as the canonical product
surface. Extend that tool so the user can configure it directly in the
Console:

1. Add a provider selector with explicit supported values, initially
   `anysearch` and `tavily`. Reject unknown values instead of silently routing
   them to AnySearch.
2. Add an optional provider API-key field. Anonymous AnySearch access should
   continue to work when the field is empty.
3. Resolve provider selection per agent/workspace from the tool configuration,
   rather than using a process-wide environment variable as the primary
   runtime configuration.
4. Store the API key in the encrypted credential store and return only a
   masked value to the Console. Do **not** write it directly to
   `BuiltinToolConfig.config`: the current `PluginRegistry.set_tool_config()`
   persists that dictionary into the agent configuration, so using it for a
   raw API key would store the secret in agent JSON.
5. Make the built-in descriptor/config API expose typed `select` and
   `password` fields generically. The current tool response only derives
   `config_fields` from plugin manifests (plus a browser special case), so the
   existing Console form cannot configure this built-in cleanly.

This gives users one search tool, one configuration card, one credential
source, and one governance decision. Users who deliberately want AnySearch
MCP can still add it as a normal custom MCP client; it should not be injected
as a second default representation of a capability QwenPaw already owns.

### [P2] Preserve valid `0` and `False` credential values

Location:

- [`bindings.py:151-168`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/credentials/bindings.py#L151-L168)

The change from `if value is None` to `if not value` is broader than the
intended AnySearch fix. `ResolvedCredential.public` and `.secrets` are typed
as `dict[str, Any]`, and the resolver previously converted any present value
to a string. The new truthiness check drops valid numeric zero and boolean
false values before string conversion.

Exact-head reproduction:

```text
credential public value 0     -> resolved binding {}
credential public value False -> resolved binding {}
```

Before this PR, the same bindings resolved to `{"X-Value": "0"}` and
`{"X-Value": "False"}`. This can silently remove a credential-backed MCP or
harness header/environment entry whose legitimate value is zero or false.
The committed tests cover an empty string from an unset environment variable,
but not other falsey values.

Suggested minimal fix:

```python
if value is None or value == "":
    return None
```

Add regression cases for `0`, `False`, `""`, and `None`. This keeps the
desired behavior of omitting `Authorization: Bearer ` when
`ANYSEARCH_API_KEY` is unset without changing the generic binding contract.

### [P2] Do not lowercase case-sensitive environment-variable identities

Locations:

- [`env_ref.py:51-57`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/adapters/env_ref.py#L51-L57)
- [`mcp_binding.py:146-162`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/adapters/mcp_binding.py#L146-L162)
- [`mcp_card_builder.py:173-188`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/adapters/mcp_card_builder.py#L173-L188)

The latest patch correctly preserves the original variable name in each
`env:` credential reference, but the credential alias is still generated by
lowercasing that name. Two distinct POSIX environment variables therefore
collapse to one alias:

```text
X-First:  ${ApiKey} -> env_apikey
X-Second: ${APIKEY} -> env_apikey
```

`EnvRefPlan.env_aliases` is a dictionary, so the second entry overwrites the
first. The resulting card contains only:

```text
env_apikey -> env:APIKEY
```

Both endpoint bindings then resolve from `APIKEY`; `ApiKey` is never read.
Environment-variable names are case-sensitive on macOS and Linux, so this is
a silent cross-platform correctness problem. Windows treats environment names
case-insensitively, which is why a Windows-only check would not expose it.

The same prefix-based namespace also makes aliases beginning with `env_`
implicitly reserved even though that restriction is not represented by the
DriverCard contract. The safest design is to derive a collision-free alias
from the exact variable identity and to determine whether a binding is
environment-backed from its `CredentialRef.ref == "env:..."`, not from the
alias prefix alone.

Add a round-trip/runtime regression containing both `${ApiKey}` and
`${APIKEY}`. On macOS/Linux it must resolve each header from its corresponding
environment variable; on Windows the test should account for the OS's
case-insensitive environment semantics.

## Synchronous I/O and Event-Loop Review

**No material synchronous network or file I/O is introduced on the normal
event-loop path.**

- [`_post()`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/agents/tools/websearch/__init__.py#L26-L37)
  uses `httpx.AsyncClient` and awaits the request.
- MCP environment credential lookup is implemented by the existing
  `AsyncCredentialStore.get()` path, whose synchronous YAML/environment work
  runs through `asyncio.to_thread`; this PR does not move it onto the loop.
- Provider selection, environment-template parsing, card construction, and
  result formatting are bounded in-memory operations.
- `response.json()` and text formatting execute on the event-loop thread, but
  the observed five-result AnySearch response was about 21 KB; no CPU hotspot
  was reproduced.

There is one non-blocking hygiene issue: `_post()` imports `httpx` inside the
coroutine even though project conventions require imports at module scope. In
the normal `web_search.py` path `httpx` is already imported, so the lookup is
cached and did not cause measurable blocking. A direct consumer of the public
`AnySearchProvider`, however, could perform a cold synchronous module import
inside the event loop. Moving the import to the file header is simpler and
keeps the async boundary unambiguous.

## Windows and macOS Compatibility

The new HTTP-based default is otherwise more portable than the removed
`npx tavily-mcp@latest` default: it introduces no shell quoting, executable
lookup, path separator, permission, signal, or subprocess behavior.

- The exact reviewed code uses platform-neutral `httpx`, `os.environ`, regex,
  and Python mappings.
- The default MCP transport is native Streamable HTTP, matching
  [AnySearch's official configuration](https://www.anysearch.com/docs).
- Anonymous access works because an empty environment credential is omitted,
  rather than sending an empty `Authorization` header. The official API also
  documents anonymous operation without that header.
- The case-collision finding above is the only PR-attributable macOS/Linux
  compatibility defect found.

The exact-head cross-platform matrix is currently running. `CodeQL` for Python
and TypeScript, the aggregate `CodeQL` check, Console/website formatting,
maintainer approval, and `Real behavior proof` have passed. Ubuntu, macOS, and
Windows unit/contract/integration jobs were still in progress when I finalized
this review. I am not treating the previous head's Windows failures as caused
by this PR because they were in unrelated shell/Codex cancellation tests, but
the exact-head matrix still needs to finish before merge.

## Functional Review

Apart from the blocking product design and the two binding regressions, the
individual implementation paths work as intended:

- An anonymous live request to `POST https://api.anysearch.com/v1/search`
  returned `code: 0` and the response shape consumed by
  `AnySearchProvider.search()`.
- `ANYSEARCH_API_KEY`, when present, is sent as a Bearer token; when absent,
  the request remains anonymous as documented by AnySearch.
- `QWENPAW_WEBSEARCH_PROVIDER=tavily` retains the legacy keyless provider.
- The default MCP client uses the documented Streamable HTTP endpoint and is
  disabled by default with the existing `ask` policy.
- The Console migration/display/save/runtime round trip now retains a single
  `${VAR}` reference, preserves its original case, avoids persisting the
  secret, handles stdio environment bindings, removes stale references when
  switching to a static secret, and treats multi-reference templates as
  static values rather than silently dropping them.
- The latest commit removes the former `verify=False` retry. This is important
  because the new provider may carry `ANYSEARCH_API_KEY`; certificate
  validation is now always enforced in the direct search path.

Those facts show that the provider implementation itself is viable. They do
not justify exposing the same capability twice. I want the REST provider kept
behind `web_search`, with user-visible provider/key configuration, and the
default MCP duplicate removed.

## Code Quality and Modularity

The `SearchProvider` boundary is useful and appropriately small: the tool
surface depends on one async method, provider-specific payload parsing is
isolated, and the old Tavily implementation remains selectable without
branching in `web_search()`. This is exactly why the built-in tool should own
provider selection instead of adding a second default MCP surface.

The package layout is less coherent than the abstraction itself. A new
`websearch/` package contains all transport code, formatting, the interface,
both implementations, and the factory directly in `__init__.py`, while the
caller remains `web_search.py`. The nearly identical names make navigation
harder, and the package currently provides no module separation. With only
two compact providers, a single `websearch.py` module would be simpler; if the
package is retained, move implementation into focused modules and keep
`__init__.py` as exports. This is a maintainability recommendation, not a
reason to expand the current patch into a large refactor.

The credential work is appropriately reused between legacy migration and the
Console save path through `plan_env_ref_bindings()`. However, the alias is
being used simultaneously as a display convention, a dictionary key, and an
identity for a case-sensitive external resource. The second finding is the
result of those concerns not being separated. Preserve exact identity in the
credential reference and use an independently collision-free internal key.

The built-in tool configuration boundary also needs one clean design instead
of another special case. `BuiltinToolConfig` already owns non-secret per-agent
tool configuration, while `AsyncCredentialStore` owns encrypted secrets. The
Console tool API should compose those two responsibilities and expose typed
descriptor metadata for built-ins. It should not reuse plugin-manifest lookup
as an implicit schema for built-in settings, and it should not persist API
keys in the generic `config` dictionary.

## Verification Record

Executed in a detached temporary worktree at exact head `ecb1377d`:

```text
git diff --check upstream/main...refs/review/pr-6817: passed

pytest:
  tests/unit/agents/tools/test_websearch_providers.py
  tests/unit/config/test_mcp_config_defaults.py
  tests/unit/drivers/adapters/test_credentials_bindings.py
  tests/unit/drivers/adapters/test_mcp_legacy_env_ref.py
  tests/integration/test_driver_mcp_http_flow.py
  tests/integration/test_driver_mcp_env_ref_flow.py

result: 32 passed in 0.63s
```

Additional production-helper probes reproduced both findings:

```text
falsey=0:     {}
falsey=False: {}

credential_refs={
  'env_apikey': CredentialRef(kind='static', ref='env:APIKEY')
}
bindings={
  'X-First':  {'credential': 'env_apikey', ...},
  'X-Second': {'credential': 'env_apikey', ...}
}
```

A live anonymous AnySearch probe returned five results with the expected
`title`, `url`, `snippet`, and `content` fields. No API key or user data was
used in that probe.

## Recommended Fix Order

1. Remove the default AnySearch MCP entry and make built-in `web_search` the
   single user-facing search surface.
2. Add first-class, per-agent provider selection and encrypted/masked API-key
   configuration to the built-in tool card and runtime.
3. Narrow the empty-value check so `0` and `False` retain their previous
   string-conversion behavior.
4. Make environment credential aliases collision-free for case-sensitive
   variable names and add a two-variable runtime round-trip test.
5. Move the lazy `httpx` import to module scope, reject unknown provider names,
   and rerun the full Ubuntu, macOS, and Windows matrices on exact head.

I am requesting changes until the duplicate default capability is consolidated,
the two P2 regressions are fixed, and the exact-head cross-platform CI passes.

src/qwenpaw/agents/tools/websearch/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+        return list((data.get("data") or {}).get("results") or [])
+
+
+def get_search_provider() -> SearchProvider:
+    """Return the active search provider selected by env var."""
+    choice = os.environ.get("QWENPAW_WEBSEARCH_PROVIDER", "").strip().lower()
+    if choice == "tavily":
+        return TavilyProvider()
+    return AnySearchProvider()
```
cuiyuebing: Could we keep Tavily as the default provider and make AnySearch an explicit opt-in, for example via `QWENPAW_WEBSEARCH_PROVIDER=anysearch`?

This change switches the search backend—and therefore query routing and user-visible search behavior—for every existing QwenPaw installation that does not set this environment variable. Adding AnySearch as an optional provider is useful, but changing the default should be supported by representative QwenPaw-specific evaluation data.

At minimum, I would expect a comparison covering search relevance, Chinese and English queries, latency, reliability, rate limits, and cost on typical QwenPaw workloads (see https://github.com/agentscope-ai/PawBench). Until such evidence is available, please preserve Tavily as the default for backward compatibility:

```python
if choice == "anysearch":
   return AnySearchProvider()
return TavilyProvider()
```

This still allows users to switch to AnySearch explicitly without changing the behavior of existing installations.

cuiyuebing (changes requested):

src/qwenpaw/agents/tools/websearch/__init__.py
anysearch-ai: Done — Tavily is the default again; AnySearch is an explicit opt-in via
`QWENPAW_WEBSEARCH_PROVIDER=anysearch`:

```python
if choice in {"", "tavily"}:
    return TavilyProvider()      # default, backward compatible
if choice == "anysearch":
    return AnySearchProvider()   # explicit opt-in
raise ValueError(...)            # unknown values rejected
Existing installs keep Tavily behavior unless the env var is set. The
updated branch is pushed
```

anysearch-ai:
> # PR #6817 — Request Changes
> * PR: [agentscope-ai/QwenPaw#6817](https://github.com/agentscope-ai/QwenPaw/pull/6817)
> * Title: `feat: integrate AnySearch web search (SearchProvider + MCP)`
> * Base: `main@ca40a47c024e11b591ca320fe3b6564e3aa26762`
> * Reviewed head: `ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed`
> * Scope: 11 files, `+884/-90`
> * Review date: 2026-08-10
> * Verdict: **Request changes.** This design should not merge as it stands.
>   QwenPaw already has an enabled built-in `web_search` tool, and
>   this PR makes that tool use AnySearch while also adding a second AnySearch
>   entry through the default MCP configuration. That creates two overlapping
>   search surfaces with different configuration, credential, tool-discovery,
>   and approval semantics. The PR also leaves two reproducible generic
>   credential-binding regressions.
> 
> ## Checklist
> * [x]  Read the latest PR diff, commits, discussions, and CI state
> * [x]  Check synchronous I/O and event-loop boundaries
> * [x]  Check Windows and macOS compatibility
> * [x]  Check functional behavior and regression boundaries
> * [x]  Check code clarity, responsibility boundaries, and modularity
> * [x]  Check overlap between built-in tools and default MCP capabilities
> * [x]  Check provider/API-key configuration and secret-storage boundaries
> * [x]  Run the focused tests against the exact reviewed head
> * [x]  Reproduce the remaining edge cases with production helpers
> * [x]  Recheck the report after the PR was force-pushed during review
> 
> ## Findings
> ### [Blocking design] Keep one canonical web-search surface
> Locations:
> 
> * [`web_search.py:136-182`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/agents/tools/web_search.py#L136-L182)
> * [`websearch/__init__.py:96-129`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/agents/tools/websearch/__init__.py#L96-L129)
> * [`config.py:1956-1976`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/config/config.py#L1956-L1976)
> * [`tools.py:104-177`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/app/routers/tools.py#L104-L177)
> * [`tools.py:304-444`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/app/routers/tools.py#L304-L444)
> 
> The AnySearch MCP client is created by default but disabled by default. It is not “enabled by default,” but that distinction does not resolve the design conflict: the built-in `web_search` tool is already enabled, this PR changes its default backend to AnySearch, and the same PR places another AnySearch search surface in every user's MCP configuration.
> 
> Once a user enables that MCP entry, the model can see two ways to perform the same basic operation:
> 
> ```
> built-in web_search  -> AnySearch REST API
> default AnySearch MCP -> AnySearch MCP search tools
> ```
> 
> These are not two interchangeable implementations behind one interface. They behave differently at every product boundary:
> 
> Concern	Built-in `web_search`	AnySearch MCP
> Provider selection	Process-wide `QWENPAW_WEBSEARCH_PROVIDER`	Fixed MCP endpoint
> API key	Process-wide `ANYSEARCH_API_KEY`	MCP env credential binding
> Configuration UI	No provider/key fields	MCP client editor
> Tool surface	One `web_search` tool	MCP-advertised search tools
> Default governance	`allow`	`ask`
> Scope	Process environment	Agent/workspace driver configuration
> The governance mismatch is especially problematic. The same external search can be automatically allowed through the built-in tool but require approval through MCP. Which path is used is left to model tool selection, not to a single user-visible policy decision. Credentials can also disagree: the built-in path may read one process environment key while the MCP card resolves a different per-workspace binding.
> 
> Required direction before merge: remove the default AnySearch MCP entry from this PR and keep the existing built-in `web_search` as the canonical product surface. Extend that tool so the user can configure it directly in the Console:
> 
> 1. Add a provider selector with explicit supported values, initially
>    `anysearch` and `tavily`. Reject unknown values instead of silently routing
>    them to AnySearch.
> 2. Add an optional provider API-key field. Anonymous AnySearch access should
>    continue to work when the field is empty.
> 3. Resolve provider selection per agent/workspace from the tool configuration,
>    rather than using a process-wide environment variable as the primary
>    runtime configuration.
> 4. Store the API key in the encrypted credential store and return only a
>    masked value to the Console. Do **not** write it directly to
>    `BuiltinToolConfig.config`: the current `PluginRegistry.set_tool_config()`
>    persists that dictionary into the agent configuration, so using it for a
>    raw API key would store the secret in agent JSON.
> 5. Make the built-in descriptor/config API expose typed `select` and
>    `password` fields generically. The current tool response only derives
>    `config_fields` from plugin manifests (plus a browser special case), so the
>    existing Console form cannot configure this built-in cleanly.
> 
> This gives users one search tool, one configuration card, one credential source, and one governance decision. Users who deliberately want AnySearch MCP can still add it as a normal custom MCP client; it should not be injected as a second default representation of a capability QwenPaw already owns.
> 
> ### [P2] Preserve valid `0` and `False` credential values
> Location:
> 
> * [`bindings.py:151-168`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/credentials/bindings.py#L151-L168)
> 
> The change from `if value is None` to `if not value` is broader than the intended AnySearch fix. `ResolvedCredential.public` and `.secrets` are typed as `dict[str, Any]`, and the resolver previously converted any present value to a string. The new truthiness check drops valid numeric zero and boolean false values before string conversion.
> 
> Exact-head reproduction:
> 
> ```
> credential public value 0     -> resolved binding {}
> credential public value False -> resolved binding {}
> ```
> 
> Before this PR, the same bindings resolved to `{"X-Value": "0"}` and `{"X-Value": "False"}`. This can silently remove a credential-backed MCP or harness header/environment entry whose legitimate value is zero or false. The committed tests cover an empty string from an unset environment variable, but not other falsey values.
> 
> Suggested minimal fix:
> 
> ```python
> if value is None or value == "":
>     return None
> ```
> 
> Add regression cases for `0`, `False`, `""`, and `None`. This keeps the desired behavior of omitting `Authorization: Bearer ` when `ANYSEARCH_API_KEY` is unset without changing the generic binding contract.
> 
> ### [P2] Do not lowercase case-sensitive environment-variable identities
> Locations:
> 
> * [`env_ref.py:51-57`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/adapters/env_ref.py#L51-L57)
> * [`mcp_binding.py:146-162`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/adapters/mcp_binding.py#L146-L162)
> * [`mcp_card_builder.py:173-188`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/drivers/adapters/mcp_card_builder.py#L173-L188)
> 
> The latest patch correctly preserves the original variable name in each `env:` credential reference, but the credential alias is still generated by lowercasing that name. Two distinct POSIX environment variables therefore collapse to one alias:
> 
> ```
> X-First:  ${ApiKey} -> env_apikey
> X-Second: ${APIKEY} -> env_apikey
> ```
> 
> `EnvRefPlan.env_aliases` is a dictionary, so the second entry overwrites the first. The resulting card contains only:
> 
> ```
> env_apikey -> env:APIKEY
> ```
> 
> Both endpoint bindings then resolve from `APIKEY`; `ApiKey` is never read. Environment-variable names are case-sensitive on macOS and Linux, so this is a silent cross-platform correctness problem. Windows treats environment names case-insensitively, which is why a Windows-only check would not expose it.
> 
> The same prefix-based namespace also makes aliases beginning with `env_` implicitly reserved even though that restriction is not represented by the DriverCard contract. The safest design is to derive a collision-free alias from the exact variable identity and to determine whether a binding is environment-backed from its `CredentialRef.ref == "env:..."`, not from the alias prefix alone.
> 
> Add a round-trip/runtime regression containing both `${ApiKey}` and `${APIKEY}`. On macOS/Linux it must resolve each header from its corresponding environment variable; on Windows the test should account for the OS's case-insensitive environment semantics.
> 
> ## Synchronous I/O and Event-Loop Review
> **No material synchronous network or file I/O is introduced on the normal event-loop path.**
> 
> * [`_post()`](https://github.com/agentscope-ai/QwenPaw/blob/ecb1377d1c8a9ed6acfd539ff9d910d08b9739ed/src/qwenpaw/agents/tools/websearch/__init__.py#L26-L37)
>   uses `httpx.AsyncClient` and awaits the request.
> * MCP environment credential lookup is implemented by the existing
>   `AsyncCredentialStore.get()` path, whose synchronous YAML/environment work
>   runs through `asyncio.to_thread`; this PR does not move it onto the loop.
> * Provider selection, environment-template parsing, card construction, and
>   result formatting are bounded in-memory operations.
> * `response.json()` and text formatting execute on the event-loop thread, but
>   the observed five-result AnySearch response was about 21 KB; no CPU hotspot
>   was reproduced.
> 
> There is one non-blocking hygiene issue: `_post()` imports `httpx` inside the coroutine even though project conventions require imports at module scope. In the normal `web_search.py` path `httpx` is already imported, so the lookup is cached and did not cause measurable blocking. A direct consumer of the public `AnySearchProvider`, however, could perform a cold synchronous module import inside the event loop. Moving the import to the file header is simpler and keeps the async boundary unambiguous.
> 
> ## Windows and macOS Compatibility
> The new HTTP-based default is otherwise more portable than the removed `npx tavily-mcp@latest` default: it introduces no shell quoting, executable lookup, path separator, permission, signal, or subprocess behavior.
> 
> * The exact reviewed code uses platform-neutral `httpx`, `os.environ`, regex,
>   and Python mappings.
> * The default MCP transport is native Streamable HTTP, matching
>   [AnySearch's official configuration](https://www.anysearch.com/docs).
> * Anonymous access works because an empty environment credential is omitted,
>   rather than sending an empty `Authorization` header. The official API also
>   documents anonymous operation without that header.
> * The case-collision finding above is the only PR-attributable macOS/Linux
>   compatibility defect found.
> 
> The exact-head cross-platform matrix is currently running. `CodeQL` for Python and TypeScript, the aggregate `CodeQL` check, Console/website formatting, maintainer approval, and `Real behavior proof` have passed. Ubuntu, macOS, and Windows unit/contract/integration jobs were still in progress when I finalized this review. I am not treating the previous head's Windows failures as caused by this PR because they were in unrelated shell/Codex cancellation tests, but the exact-head matrix still needs to finish before merge.
> 
> ## Functional Review
> Apart from the blocking product design and the two binding regressions, the individual implementation paths work as intended:
> 
> * An anonymous live request to `POST https://api.anysearch.com/v1/search`
>   returned `code: 0` and the response shape consumed by
>   `AnySearchProvider.search()`.
> * `ANYSEARCH_API_KEY`, when present, is sent as a Bearer token; when absent,
>   the request remains anonymous as documented by AnySearch.
> * `QWENPAW_WEBSEARCH_PROVIDER=tavily` retains the legacy keyless provider.
> * The default MCP client uses the documented Streamable HTTP endpoint and is
>   disabled by default with the existing `ask` policy.
> * The Console migration/display/save/runtime round trip now retains a single
>   `${VAR}` reference, preserves its original case, avoids persisting the
>   secret, handles stdio environment bindings, removes stale references when
>   switching to a static secret, and treats multi-reference templates as
>   static values rather than silently dropping them.
> * The latest commit removes the former `verify=False` retry. This is important
>   because the new provider may carry `ANYSEARCH_API_KEY`; certificate
>   validation is now always enforced in the direct search path.
> 
> Those facts show that the provider implementation itself is viable. They do not justify exposing the same capability twice. I want the REST provider kept behind `web_search`, with user-visible provider/key configuration, and the default MCP duplicate removed.
> 
> ## Code Quality and Modularity
> The `SearchProvider` boundary is useful and appropriately small: the tool surface depends on one async method, provider-specific payload parsing is isolated, and the old Tavily implementation remains selectable without branching in `web_search()`. This is exactly why the built-in tool should own provider selection instead of adding a second default MCP surface.
> 
> The package layout is less coherent than the abstraction itself. A new `websearch/` package contains all transport code, formatting, the interface, both implementations, and the factory directly in `__init__.py`, while the caller remains `web_search.py`. The nearly identical names make navigation harder, and the package currently provides no module separation. With only two compact providers, a single `websearch.py` module would be simpler; if the package is retained, move implementation into focused modules and keep `__init__.py` as exports. This is a maintainability recommendation, not a reason to expand the current patch into a large refactor.
> 
> The credential work is appropriately reused between legacy migration and the Console save path through `plan_env_ref_bindings()`. However, the alias is being used simultaneously as a display convention, a dictionary key, and an identity for a case-sensitive external resource. The second finding is the result of those concerns not being separated. Preserve exact identity in the credential reference and use an independently collision-free internal key.
> 
> The built-in tool configuration boundary also needs one clean design instead of another special case. `BuiltinToolConfig` already owns non-secret per-agent tool configuration, while `AsyncCredentialStore` owns encrypted secrets. The Console tool API should compose those two responsibilities and expose typed descriptor metadata for built-ins. It should not reuse plugin-manifest lookup as an implicit schema for built-in settings, and it should not persist API keys in the generic `config` dictionary.
> 
> ## Verification Record
> Executed in a detached temporary worktree at exact head `ecb1377d`:
> 
> ```
> git diff --check upstream/main...refs/review/pr-6817: passed
> 
> pytest:
>   tests/unit/agents/tools/test_websearch_providers.py
>   tests/unit/config/test_mcp_config_defaults.py
>   tests/unit/drivers/adapters/test_credentials_bindings.py
>   tests/unit/drivers/adapters/test_mcp_legacy_env_ref.py
>   tests/integration/test_driver_mcp_http_flow.py
>   tests/integration/test_driver_mcp_env_ref_flow.py
> 
> result: 32 passed in 0.63s
> ```
> 
> Additional production-helper probes reproduced both findings:
> 
> ```
> falsey=0:     {}
> falsey=False: {}
> 
> credential_refs={
>   'env_apikey': CredentialRef(kind='static', ref='env:APIKEY')
> }
> bindings={
>   'X-First':  {'credential': 'env_apikey', ...},
>   'X-Second': {'credential': 'env_apikey', ...}
> }
> ```
> 
> A live anonymous AnySearch probe returned five results with the expected `title`, `url`, `snippet`, and `content` fields. No API key or user data was used in that probe.
> 
> ## Recommended Fix Order
> 1. Remove the default AnySearch MCP entry and make built-in `web_search` the
>    single user-facing search surface.
> 2. Add first-class, per-agent provider selection and encrypted/masked API-key
>    configuration to the built-in tool card and runtime.
> 3. Narrow the empty-value check so `0` and `False` retain their previous
>    string-conversion behavior.
> 4. Make environment credential aliases collision-free for case-sensitive
>    variable names and add a two-variable runtime round-trip test.
> 5. Move the lazy `httpx` import to module scope, reject unknown provider names,
>    and rerun the full Ubuntu, macOS, and Windows matrices on exact head.
> 
> I am requesting changes until the duplicate default capability is consolidated, the two P2 regressions are fixed, and the exact-head cross-platform CI passes.

---

All findings addressed in the latest head (`9b9ab8a4`). Note: after
discussion with the team, the resolution for the blocking item differs
slightly from the original ask — the default AnySearch **MCP** entry is
kept (disabled by default, `ask` policy) rather than removed, because
the built-in `web_search` tool now defaults back to **Tavily**, so
AnySearch is no longer exposed twice by default:

- **Blocking**: `web_search` defaults back to Tavily (backward
  compatible); AnySearch is an explicit opt-in via
  `QWENPAW_WEBSEARCH_PROVIDER=anysearch`. The MCP client stays disabled
  by default with `ask` policy — one search surface per configuration.
- **P2 falsy values**: guard narrowed to `None`/`""` — `0`/`False`
  preserved, with regression tests.
- **P2 case-sensitive aliases**: collision-free aliases for `${ApiKey}`
  vs `${APIKEY}`, with round-trip regression.
- **httpx import** moved to module scope; **unknown provider values**
  now raise instead of silently routing.
- Per-agent Console provider config + API-key field: planned as a
  follow-up PR (landing the abstraction first).
- `websearch/__init__.py` monolith: kept for now (non-blocking).

58 focused tests pass; full-repo pre-commit green.

rayrayraykk:
Thanks for the update. Most of the concrete implementation issues from my previous review have been addressed, including falsey credential values, case-insensitive alias collisions, import placement, and unknown-provider handling. I also confirmed that the focused test suite passes.

One functional issue remains: environment-backed credentials are still identified by checking whether the credential alias starts with `env_`. A valid static credential alias such as `env_custom` is therefore incorrectly serialized as `${CUSTOM}`. Environment references should be identified from the credential reference/type, not from the alias naming convention.

The search-provider change also only partially addresses the architecture concern. The built-in `web_search` now defaults to Tavily and supports an environment-variable-based provider switch, which avoids two active AnySearch paths by default. However, provider and API-key configuration are still unavailable per agent in the Console, and enabling the bundled AnySearch MCP still exposes two semantically overlapping search tools. I understand that part is being deferred, but it means the product-level concern raised in the previous review is not fully resolved in this PR.

Finally, the `websearch` module docstrings still state that AnySearch is the default, while the implementation now defaults to Tavily.

I’m keeping the Request Changes review until the `env_` alias misclassification is fixed. The remaining search configuration question should also be explicitly resolved or tracked as an agreed follow-up.

anysearch-ai:
> Thanks for the update. Most of the concrete implementation issues from my previous review have been addressed, including falsey credential values, case-insensitive alias collisions, import placement, and unknown-provider handling. I also confirmed that the focused test suite passes.
> 
> One functional issue remains: environment-backed credentials are still identified by checking whether the credential alias starts with `env_`. A valid static credential alias such as `env_custom` is therefore incorrectly serialized as `${CUSTOM}`. Environment references should be identified from the credential reference/type, not from the alias naming convention.
> 
> The search-provider change also only partially addresses the architecture concern. The built-in `web_search` now defaults to Tavily and supports an environment-variable-based provider switch, which avoids two active AnySearch paths by default. However, provider and API-key configuration are still unavailable per agent in the Console, and enabling the bundled AnySearch MCP still exposes two semantically overlapping search tools. I understand that part is being deferred, but it means the product-level concern raised in the previous review is not fully resolved in this PR.
> 
> Finally, the `websearch` module docstrings still state that AnySearch is the default, while the implementation now defaults to Tavily.
> 
> I’m keeping the Request Changes review until the `env_` alias misclassification is fixed. The remaining search configuration question should also be explicitly resolved or tracked as an agreed follow-up.

---


Thanks for the follow-up. Both issues are addressed in the latest head (`39003c4c`):

**1. env-backed identification from credential refs (fixed)**
You're right that checking the `env_` alias prefix misclassifies a static alias like `env_custom`. To be clear, this is a **pre-existing limitation of QwenPaw's env-ref alias convention** — the `env_` prefix is generated by `env_alias()`, and the identification logic relied on that prefix convention, so any static alias starting with `env_` was already at risk. The env-ref work in this PR surfaced the issue but did not introduce it.

The fix identifies env-backed bindings from the `env:` credential refs (`CredentialRef.ref == "env:..."`, via `env_aliases`) instead of the alias prefix. A static `env_custom` alias now serializes as a plain static value, not `${CUSTOM}`. Regression test added (`test_static_alias_starting_with_env_is_not_misclassified`).

**2. Stale docstring (fixed)**
The `websearch` module docstring now correctly states Tavily is the default provider with AnySearch as an explicit opt-in.

PR #6818, opened 2026-08-08
fix(summary): honor disable_thinking and interruption (#6811)

### LINKED ISSUE #6811: [Bug]: OpenAI Responses continuation summary ignores `disable_thinking` and misreports 60-second cancellation as malformed output
## QwenPaw Version

QwenPaw `2.1.0b2`

## Description

When Scroll evicts old conversation turns, QwenPaw synchronously asks the active main model to generate a continuation summary. With an OpenAI Responses provider and a reasoning-enabled model, this summary call blocks the main conversation for approximately 53–60 seconds. Most observed calls then fail with:

```text
_SummaryCandidateError: empty or malformed plain Markdown summary
```

The summary implementation explicitly calls the model with `disable_thinking=True`:

```python
response = await agent.model(
    messages=messages,
    tools=None,
    max_tokens=max_tokens,
    disable_thinking=True,
)
```

However, `OpenAIResponseModelCompat._call_api()` currently removes the flag without disabling inherited reasoning:

```python
generate_kwargs.pop("disable_thinking", None)
max_tokens = generate_kwargs.pop("max_tokens", None)
if max_tokens is not None and "max_output_tokens" not in generate_kwargs:
    generate_kwargs["max_output_tokens"] = max_tokens
merged = {**self._extra_generate_kwargs, **generate_kwargs}
```

In the affected configuration, the provider-level effective generation parameters contain:

```python
{
    "reasoning": {"effort": "xhigh"},
    "max_tokens": 100000,
}
```

After applying the continuation-summary call overrides, the effective Responses request still contains:

```python
{
    "model": "gpt-5.6-sol",
    "stream": True,
    "max_output_tokens": 4000,
    "reasoning": {"effort": "xhigh"},
    "input": "<two continuation-summary messages>",
}
```

Therefore, the internal summary call is not actually running with thinking disabled. Since the Responses API counts reasoning tokens toward `max_output_tokens`, an `xhigh` request can spend much of the 4,000-token output budget and most of the 60-second deadline on reasoning before producing the strict Markdown body required by Scroll.

There is also a cancellation-reporting issue in the streaming path. Scroll wraps summary generation in a 60-second timeout, but AgentScope's `ChatModelBase` catches `asyncio.CancelledError` while consuming the stream and converts it into an empty `FinishedReason.INTERRUPTED` response:

```python
    acc_res.finished_reason = FinishedReason.INTERRUPTED
    yield_acc_res = True

if yield_acc_res:
    yield acc_res
```

As a result, a timeout cancellation can become an empty final response. Scroll then parses an empty string and reports `empty or malformed plain Markdown summary` instead of a timeout. This matches the observed failures that consistently finish at approximately 60 seconds.

The continuation-summary implementation may also make a second model request after the first candidate fails local format or quality validation. Both attempts share the same 60-second deadline. This can further amplify the delay when the first response is incomplete or does not exactly match the required Markdown headings.

Expected behavior:

1. `disable_thinking=True` should remove or override provider/model-level reasoning configuration for OpenAI Responses summary calls.
2. A cancelled or timed-out stream should remain distinguishable from a completed empty response.
3. The log should report the real timeout/interruption rather than a misleading malformed-summary error.
4. A normal no-thinking continuation summary should not routinely block the main conversation for approximately one minute.

**Related PR(s):** None

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** `2.1.0b2`
- **Protocol:** OpenAI Responses API
- **Model:** `gpt-5.6-sol`
- **Provider type:** Custom OpenAI-compatible Responses provider (`openai-custom`)
- **Effective reasoning setting:** `reasoning.effort=xhigh`
- **OS:** Windows 11
- **Install method:** pip
- **Python version:** `3.12.10`
- **AgentScope version:** `2.0.4.post1`
- **OpenAI SDK version:** `2.33.0`

## Steps to Reproduce

1. Configure a custom OpenAI-compatible provider that uses the Responses API.
2. Configure a reasoning-capable model with a provider/model-level generation setting such as:

   ```json
   {
     "reasoning": {
       "effort": "xhigh"
     }
   }
   ```

3. Start a sufficiently long conversation or tool-heavy task so that Scroll must evict older turns and generate a continuation summary.
4. Observe that the main conversation waits while `ScrollManager._update_continuation_summary()` synchronously invokes the current model.
5. Inspect the effective call parameters through the provider compatibility path. `max_tokens=4000` is translated to `max_output_tokens=4000`, but `disable_thinking=True` is removed without removing the inherited `reasoning={"effort":"xhigh"}` value.
6. Wait for the summary call to reach the approximately 60-second deadline.
7. Observe that the log may report `empty or malformed plain Markdown summary` instead of a timeout because stream cancellation was converted into an empty interrupted response.

## Actual vs Expected

- **Actual:**
  - The continuation-summary request still carries `reasoning.effort=xhigh` despite being invoked with `disable_thinking=True`.
  - Scroll blocks the main conversation synchronously for approximately 53–60 seconds.
  - Four of five observed long summary calls failed after approximately 60 seconds.
  - The failures were logged as `empty or malformed plain Markdown summary`, even though their duration closely matches the hardcoded 60-second timeout.
  - The previous valid summary is preserved, so conversation history is not lost, but the delay is paid even when the newly generated summary is discarded.

- **Expected:**
  - `disable_thinking=True` should produce a Responses request without inherited reasoning, or with a provider-supported explicit no-reasoning setting.
  - The internal summary model should spend its output budget on the requested Markdown summary rather than hidden `xhigh` reasoning.
  - Cancellation and timeout should propagate as cancellation/timeout, not as a completed empty model response.
  - Logs should distinguish timeout, interruption, empty model output, malformed Markdown, and local quality-validation failure.

## Logs / Screenshots

Five observed eviction-time continuation-summary calls:

```text
2026-08-07 15:46:26 | scroll: compact timing outcome=done total=60521.3ms ... continuation_summary=60460.7ms ...
2026-08-07 17:56:44 | scroll: compact timing outcome=done total=53039.5ms ... continuation_summary=52985.5ms ...
2026-08-07 19:36:46 | scroll: compact timing outcome=done total=60605.1ms ... continuation_summary=60538.1ms ...
2026-08-07 20:25:12 | scroll: compact timing outcome=done total=60564.6ms ... continuation_summary=60504.3ms ...
2026-08-07 22:10:18 | scroll: compact timing outcome=done total=60609.5ms ... continuation_summary=60545.6ms ...
```

Four of those calls logged this failure; one call completed without the warning:

```text
2026-08-07 15:46:26 | WARNING | scroll: continuation summary update failed; preserving the previous valid summary
  File ".../qwenpaw/agents/context/scroll/manager.py", line 1323, in _update_continuation_summary_inner
    updated = await self._validated_summary_attempt(...)
  File ".../qwenpaw/agents/context/scroll/manager.py", line 1206, in _validated_summary_attempt
    raise _SummaryCandidateError(...)
qwenpaw.agents.context.scroll.manager._SummaryCandidateError: empty or malformed plain Markdown summary
```

The non-summary compression stages took only milliseconds in the same event:

```text
persist=6.8ms
prepare_input=20.1ms
count_tokens=2.0ms
pre_fold_tool_results=0.7ms
split=14.4ms
offload_dialog=0.2ms
continuation_summary=60460.7ms
index_evicted=0.1ms
rebuild_context=0.3ms
live_tokens=16.1ms
```

Relevant constants and call path:

```python
_SUMMARY_UPDATE_TIMEOUT_SECONDS = 60.0

async with asyncio.timeout(_SUMMARY_UPDATE_TIMEOUT_SECONDS):
    await self._update_continuation_summary_inner(...)
```

```python
for attempt in range(2):
    # Build prompt, call model, parse Markdown, validate quality.
    # A candidate-format failure may cause one repair attempt.
```

## Additional Notes

- Plain tool-result folding is not the source of this one-minute delay. In the same environment, pre-fold-only operations generally complete in roughly 0.5 seconds.
- The summary call correctly uses `tools=None`; it does not include the agent's normal tool schemas.
- `max_tokens=4000` is correctly mapped to `max_output_tokens=4000`. The problematic part is that inherited Responses reasoning remains active after `disable_thinking` is discarded.
- The provider-level request also retains provider `instructions` and `prompt_cache_key`. These were not proven to cause the delay, but they show that the summary request is not fully isolated from ordinary main-model generation settings.
- Possible areas to evaluate in a fix:
  1. Make `OpenAIResponseModelCompat` translate `disable_thinking=True` by removing inherited `reasoning`, or setting a supported no-reasoning value where applicable.
  2. Preserve `CancelledError` through the stream wrappers, or explicitly detect `FinishedReason.INTERRUPTED` before treating the result as candidate text.
  3. Log first-attempt/repair-attempt latency, first output-text latency, reasoning/output token counts, and the final interruption/validation category separately.
  4. Add a regression test where `_extra_generate_kwargs` contains `reasoning={"effort":"xhigh"}` and a call-level `disable_thinking=True` must prevent it from reaching `client.responses.create()`.
  5. Add a timeout regression test proving that a cancelled stream is reported as timeout/interruption rather than malformed Markdown.

RerankerGuo:
## Description

Continuation summaries sent through the OpenAI Responses provider could still inherit provider-level `reasoning` settings despite passing `disable_thinking=True`. Separately, AgentScope converts a cancelled model stream into a final `ChatResponse` with `FinishedReason.INTERRUPTED`; Scroll discarded that status, treated the empty text as malformed Markdown, and could start a repair attempt instead of preserving timeout semantics.

This change:

- merges provider and call-level generation kwargs before consuming `disable_thinking`, then explicitly sets `reasoning.effort=none` only for the documented GPT-5.5 alias/snapshot and GPT-5.6 alias/Sol/Terra/Luna models that support it; Pro, Codex, unknown, and custom model families instead have inherited reasoning removed;
- preserves `reasoning` for ordinary Responses calls where thinking remains enabled;
- detects both streaming and non-streaming interrupted summary responses before text parsing and re-raises cancellation;
- reads `finished_reason` from dict-like `ChatResponse` objects, where the mapping value contains the actual runtime status;
- adds deterministic regressions for both provider kwargs and AgentScope cancellation conversion.

**Related Issue:** Fixes #6811

**Security Considerations:** None. This only changes summary request adaptation and cancellation handling.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran all applicable changed-file pre-commit hooks locally and they pass
- [x] Documentation updated (not needed; behavior is covered by regression tests)
- [x] Ready for review

## Testing

- Provider regressions configure inherited `reasoning={"effort": "xhigh"}` and verify `disable_thinking=True` emits explicit `none` for `gpt-5.5`, its documented `gpt-5.5-2026-04-23` snapshot, the official `gpt-5.6` alias, and the GPT-5.6 Sol/Terra/Luna family. They also verify inherited reasoning is removed for Pro, Codex, future, unknown, and custom models while preserving the per-call output limit.
- A companion provider test verifies reasoning remains enabled without the neutral flag.
- Scroll regression uses the real AgentScope `ChatModelBase` stream wrapper around a no-network hanging generator. The wrapper converts cancellation to an interrupted response; the test verifies QwenPaw restores `TimeoutError`.
- A non-stream regression verifies an interrupted `ChatResponse` propagates cancellation before Markdown parsing.

## Evidence

Before the fix on `bacac7410cf2ac122dd892152beff8b03f38729f`:

```text
FAILED test_disable_thinking_removes_inherited_responses_reasoning
AssertionError: assert 'reasoning' not in {
    'reasoning': {'effort': 'xhigh'},
    'max_output_tokens': 4000,
}

FAILED test_cancelled_summary_stream_remains_a_timeout
Failed: expected timeout from cancelled summary stream,
got completed result ''
```

After the fix:

```text
$ HOME=/tmp/qwenpaw-6811-home .venv/bin/pytest -q \
    tests/unit/providers tests/unit/agents/context
777 passed in 7.48s

$ HOME=/tmp/qwenpaw-6811-home SKIP=actionlint \
    src/qwenpaw/providers/openai_response_provider.py \
    src/qwenpaw/agents/context/scroll/manager.py \
    tests/unit/providers/test_openai_response_provider.py \
    tests/unit/agents/context/test_scroll_manager.py
```

`actionlint` was skipped because this PR changes no workflow files; all applicable hooks passed.

The post-review capability matrix verifies that `gpt-5.5`, its documented
`gpt-5.5-2026-04-23` snapshot, the official `gpt-5.6` alias, and the
`gpt-5.6-sol` / `gpt-5.6-terra` / `gpt-5.6-luna` family receive
`reasoning={"effort":"none"}`. Pro/Codex variants, including dated forms,
plus future, unknown, and custom models do not.

## Additional Notes

The two fixes are kept together because they are the two independently reproduced causes in #6811's continuation-summary path. AI-assisted analysis was independently verified against the exact default-branch commit with deterministic no-network tests and review of every changed line.

### DIFF
--- a/src/qwenpaw/agents/context/scroll/manager.py
+++ b/src/qwenpaw/agents/context/scroll/manager.py
@@ -26,6 +26,7 @@ from datetime import datetime, timedelta, timezone
 from typing import Any
 
 from agentscope.message import Msg, SystemMsg, TextBlock, UserMsg
+from agentscope.model import FinishedReason
 
 from ....constant import (
     QWENPAW_MESSAGE_TAG_KEY,
@@ -1004,16 +1005,29 @@ class ScrollContextManager:
             disable_thinking=True,
         )
         if not inspect.isasyncgen(response):
+            self._raise_if_summary_interrupted(response)
             return self._response_text(response)
         deltas: list[str] = []
         final = ""
         async for chunk in response:
+            self._raise_if_summary_interrupted(chunk)
             text = self._response_text(chunk)
             if getattr(chunk, "is_last", False):
                 final = text
             elif text:
                 deltas.append(text)
         return final or "".join(deltas).strip()
+
+    @staticmethod
+    def _raise_if_summary_interrupted(response: Any) -> None:
+        """Preserve cancellation converted into an AgentScope response."""
+        finished_reason = (
+            response.get("finished_reason")
+            if isinstance(response, dict)
+            else getattr(response, "finished_reason", None)
+        )
+        if finished_reason == FinishedReason.INTERRUPTED:
+            raise asyncio.CancelledError()
 
     @staticmethod
     def _summary_messages(prompt: str, language: str) -> list[Msg]:
--- a/src/qwenpaw/providers/openai_response_provider.py
+++ b/src/qwenpaw/providers/openai_response_provider.py
@@ -15,6 +15,16 @@ from .openai_provider import OpenAIProvider
 logger = logging.getLogger(__name__)
 
 
+def _supports_none_reasoning_effort(model_name: str) -> bool:
+    """Whether an official GPT model accepts ``reasoning.effort=none``."""
+    canonical_name = model_name.rsplit("/", 1)[-1].lower()
+    prefix = "gpt-5."
+    if not canonical_name.startswith(prefix):
+        return False
+    minor_version = canonical_name[len(prefix) :].split("-", 1)[0]
+    return minor_version.isdigit() and int(minor_version) >= 2
+
+
 def _extract_response_text(response: Any) -> str:
     """Extract aggregated output text from a Responses API
     response object.
@@ -96,7 +106,10 @@ class OpenAIResponseModelCompat(OpenAIResponseModel):
         ):
             merged["max_output_tokens"] = inherited_max_tokens
         if disable_thinking:
-            merged.pop("reasoning", None)
+            if _supports_none_reasoning_effort(model_name):
+                merged["reasoning"] = {"effort": "none"}
+            else:
+                merged.pop("reasoning", None)
         return await super()._call_api(
             model_name,
             messages,
--- a/src/qwenpaw/providers/openai_response_provider.py
+++ b/src/qwenpaw/providers/openai_response_provider.py
@@ -14,15 +14,20 @@ from .openai_provider import OpenAIProvider
 
 logger = logging.getLogger(__name__)
 
+_NONE_REASONING_EFFORT_MODELS = frozenset(
+    {
+        "gpt-5.5",
+        "gpt-5.6-luna",
+        "gpt-5.6-sol",
+        "gpt-5.6-terra",
+    },
+)
+
 
 def _supports_none_reasoning_effort(model_name: str) -> bool:
-    """Whether an official GPT model accepts ``reasoning.effort=none``."""
+    """Whether a known model accepts ``reasoning.effort=none``."""
     canonical_name = model_name.rsplit("/", 1)[-1].lower()
-    prefix = "gpt-5."
-    if not canonical_name.startswith(prefix):
-        return False
-    minor_version = canonical_name[len(prefix) :].split("-", 1)[0]
-    return minor_version.isdigit() and int(minor_version) >= 2
+    return canonical_name in _NONE_REASONING_EFFORT_MODELS
 
 
 def _extract_response_text(response: Any) -> str:
--- a/src/qwenpaw/providers/openai_response_provider.py
+++ b/src/qwenpaw/providers/openai_response_provider.py
@@ -17,6 +17,8 @@ logger = logging.getLogger(__name__)
 _NONE_REASONING_EFFORT_MODELS = frozenset(
     {
         "gpt-5.5",
+        "gpt-5.5-2026-04-23",
+        "gpt-5.6",
         "gpt-5.6-luna",
         "gpt-5.6-sol",
         "gpt-5.6-terra",
@@ -25,7 +27,7 @@ _NONE_REASONING_EFFORT_MODELS = frozenset(
 
 
 def _supports_none_reasoning_effort(model_name: str) -> bool:
-    """Whether a known model accepts ``reasoning.effort=none``."""
+    """Whether a documented model accepts ``reasoning.effort=none``."""
     canonical_name = model_name.rsplit("/", 1)[-1].lower()
     return canonical_name in _NONE_REASONING_EFFORT_MODELS
 
@@ -111,10 +113,9 @@ class OpenAIResponseModelCompat(OpenAIResponseModel):
         ):
             merged["max_output_tokens"] = inherited_max_tokens
         if disable_thinking:
+            merged.pop("reasoning", None)
             if _supports_none_reasoning_effort(model_name):
                 merged["reasoning"] = {"effort": "none"}
-            else:
-                merged.pop("reasoning", None)
         return await super()._call_api(
             model_name,
             messages,
--- a/src/qwenpaw/providers/openai_response_provider.py
+++ b/src/qwenpaw/providers/openai_response_provider.py
@@ -81,8 +81,6 @@ class OpenAIResponseModelCompat(OpenAIResponseModel):
         tool_choice: Any | None = None,
         **generate_kwargs: Any,
     ) -> Any:
-        # Pop the neutral ``disable_thinking`` flag
-        generate_kwargs.pop("disable_thinking", None)
         max_tokens = generate_kwargs.pop("max_tokens", None)
         if (
             max_tokens is not None
@@ -90,6 +88,15 @@ class OpenAIResponseModelCompat(OpenAIResponseModel):
         ):
             generate_kwargs["max_output_tokens"] = max_tokens
         merged = {**self._extra_generate_kwargs, **generate_kwargs}
+        disable_thinking = merged.pop("disable_thinking", False)
+        inherited_max_tokens = merged.pop("max_tokens", None)
+        if (
+            inherited_max_tokens is not None
+            and "max_output_tokens" not in merged
+        ):
+            merged["max_output_tokens"] = inherited_max_tokens
+        if disable_thinking:
+            merged.pop("reasoning", None)
         return await super()._call_api(
             model_name,
             messages,
--- a/tests/unit/agents/context/test_scroll_manager.py
+++ b/tests/unit/agents/context/test_scroll_manager.py
@@ -21,7 +21,7 @@ from agentscope.message import (
     ToolCallBlock,
     ToolResultBlock,
 )
-from agentscope.model import ChatResponse
+from agentscope.model import ChatModelBase, ChatResponse, FinishedReason
 
 from qwenpaw.agents.context.base import ContextManager
 from qwenpaw.agents.context.scroll import manager as scroll_manager_module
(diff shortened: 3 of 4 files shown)


RerankerGuo:
Requesting review of current head `8392faa9`. The branch is mergeable with no failing executable gate and includes the reporter-requested compatibility refinement: GPT-5.2+ receives explicit `effort=none`, while older or unknown models retain the fallback behavior. Verification passes 765 related tests, OpenAI 2.33.0 wire serialization probes, applicable hooks, interruption/token-precedence regressions, and all refreshed non-approval CI.

qbc2016 (commented):
## `reasoning.effort="none"` capability detection is too broad

`_supports_none_reasoning_effort()` currently classifies every `gpt-5.<minor>-*` model with a minor version greater than or equal to `2` as supporting `reasoning.effort="none"`.

This incorrectly includes models that do not support `none`, for example:

- `gpt-5.2-pro`: supports `medium`, `high`, and `xhigh`
- `gpt-5.4-pro`: supports `medium`, `high`, and `xhigh`
- `gpt-5.3-codex`: supports `low`, `medium`, `high`, and `xhigh`

As a result, continuation-summary calls using these models will send an invalid parameter and fail at the API boundary.

### Suggested fix

Please replace the version-based check with an explicit capability check.

For this PR, a safe minimal approach would be:

1. Always remove inherited `reasoning` when `disable_thinking=True`.
2. Explicitly set `reasoning={"effort": "none"}` only for known models that support it and require an explicit override, such as GPT-5.5 and the GPT-5.6 Sol/Terra/Luna family.
3. Do not send `none` to Pro, Codex, unknown, or custom model families.
4. Add regression cases for `gpt-5.2-pro`, `gpt-5.3-codex`, `gpt-5.4-pro`, and `gpt-5.5-pro`.

Longer term, this should use `ModelInfo.reasoning_effort_options` instead of inferring capabilities from the model name.

The interruption handling and the issue-specific `gpt-5.6-sol` fix otherwise look correct.

RerankerGuo:
@qbc2016 Addressed the capability concern in `6df7a2a8`.

- Replaced the `gpt-5.<minor> >= 2` inference with an exact capability whitelist for `gpt-5.5` and the GPT-5.6 Sol/Terra/Luna family.
- `disable_thinking=True` now always removes inherited `reasoning` first, then adds `{"effort": "none"}` only for those known models.
- Added the requested regressions for `gpt-5.2-pro`, `gpt-5.3-codex`, `gpt-5.4-pro`, and `gpt-5.5-pro`.
- Also covered base `gpt-5.6`, future, unknown, and custom models as unsupported, plus provider-prefixed `openai/gpt-5.6-luna` as supported.

Verification:

```text
pytest -q tests/unit/providers/test_openai_response_provider.py
23 passed

pytest -q tests/unit/providers tests/unit/agents/context
774 passed

SKIP=actionlint pre-commit run --files <the four PR files>
all applicable hooks passed
```

The PR description now reflects the explicit capability matrix. Re-review of the current head would be appreciated.

Remote executable checks on `6df7a2a8` are now green, including Python/TypeScript CodeQL, pre-commit, Real behavior proof, frontend formatting, and the PR Spam Gate.

qbc2016:
## Official aliases and snapshots are incorrectly treated as not supporting `reasoning.effort="none"`

The updated allowlist fixes the Pro/Codex false positives, but it now excludes the official `gpt-5.6` alias. The regression test even explicitly expects inherited reasoning to be removed without sending `none` for `gpt-5.6`.

According to the official OpenAI documentation, `gpt-5.6` routes to `gpt-5.6-sol`, supports `reasoning.effort="none"`, and defaults to `medium` when the effort is omitted. Therefore, merely removing the inherited `reasoning` field does not disable reasoning for this alias; the request falls back to `medium`, so the original summary latency and timeout behavior can still occur.

The exact-name allowlist also excludes documented snapshots such as `gpt-5.5-2026-04-23`. That snapshot has the same supported efforts and `medium` default as `gpt-5.5`, but `_supports_none_reasoning_effort()` currently returns `False` for it.

Please include the official `gpt-5.6` alias and support documented dated snapshots while continuing to exclude Pro and Codex variants. The current negative `gpt-5.6` test should become a positive `effort="none"` case, and a GPT-5.5 snapshot regression should be added.

The previously reported Pro/Codex issue is otherwise fixed, and the related test directories pass locally.

RerankerGuo:
@qbc2016 Thank you for the precise follow-up. You were right that the previous exact allowlist fixed the Pro/Codex false positives but missed the official `gpt-5.6` alias and the documented GPT-5.5 snapshot.

Addressed in `48134c11`:

- added `gpt-5.6`, which the official model page documents as routing to GPT-5.6 Sol;
- added the documented `gpt-5.5-2026-04-23` snapshot;
- kept the capability check as an explicit allowlist, so Pro/Codex variants, including dated forms, remain excluded;
- made the `disable_thinking=True` path remove inherited reasoning before applying the explicit `{"effort": "none"}` override.

Verification:

```text
uv run --frozen pytest -q tests/unit/providers/test_openai_response_provider.py
27 passed

uv run --frozen pytest -q tests/unit/providers tests/unit/agents/context
777 passed

SKIP=actionlint uv run --frozen pre-commit run --files \
  src/qwenpaw/providers/openai_response_provider.py \
  tests/unit/providers/test_openai_response_provider.py
all applicable hooks passed
```

I also updated the PR description and capability matrix to match the corrected alias/snapshot behavior. Thanks again for catching this distinction.

qbc2016 (approved):
LGTM.

PR #6824, opened 2026-08-08
fix(scroll): fix of recall correctness issues from aone

niceIrene:
## Description

Fix two Scroll recall correctness issues:

1. CJK substring searches could return no results when FTS5 was available.
   SQLite's `porter unicode61` tokenizer indexes a continuous CJK sentence as
   a single token, so a query such as `紫水晶河马` could not match
   `紫水晶河马在周二跳舞`.

   Queries containing CJK characters now use the existing bounded,
   parameterized LIKE substring-search path. English and other tokenizable
   queries continue to use FTS5, BM25 ranking, and Porter stemming.

2. `recall_history(op="expand")` accepted reversed sequence ranges such as
   `lo=999999, hi=999990`. The empty SQL result was incorrectly reported as a
   successful, genuine absence of history.

   Expand arguments are now normalized and checked for `lo <= hi`. Reversed
   ranges return an actionable error such as:

   ```text
   RECALL FAILED — invalid op="expand" seq span
   (ValueError: lo (...) must be less than or equal to hi (...);
   swap lo and hi and retry).

### DIFF
--- a/src/qwenpaw/agents/context/scroll/memoryspace.py
+++ b/src/qwenpaw/agents/context/scroll/memoryspace.py
@@ -68,6 +68,21 @@ _SAVED_TOOL_FILE_RE = re.compile(
 )
 
 _FTS_TOKEN_RE = re.compile(r"\w+", re.UNICODE)
+# ``unicode61`` does not perform word segmentation for CJK scripts. A
+# punctuation-delimited run such as ``紫水晶河马在周二跳舞`` is indexed as one
+# token, so a natural keyword query like ``紫水晶河马`` cannot MATCH it. Route
+# queries containing these scripts through the bounded literal-substring path
+# instead. This keeps Porter stemming/BM25 for languages it tokenizes well.
+_CJK_QUERY_RE = re.compile(
+    "["
+    "\u3040-\u30ff"  # Hiragana and Katakana
+    "\u3400-\u4dbf"  # CJK Extension A
+    "\u4e00-\u9fff"  # CJK Unified Ideographs
+    "\uac00-\ud7af"  # Hangul syllables
+    "\uf900-\ufaff"  # CJK Compatibility Ideographs
+    "\U00020000-\U0002fa1f"  # Supplementary CJK ideographs
+    "]",
+)
 # FTS5's boolean operators are UPPERCASE-only; we pass these through bare so a
 # query like ``tank OR aquarium`` casts a wide net, while every other token is
 # quoted as a literal phrase. A lowercase ``or`` stays a search term.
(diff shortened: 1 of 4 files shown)


RerankerGuo:
Early independent check of the draft at exact head
`150940d8a8d06bddd1ce985dbcf262777b2abf91` on macOS arm64 /
Python 3.11.15.

The two target fixes are reproducible and the exercised scope is stable:

```text
PR's 7 new target cases, unchanged:
current main bacac741: 6 failed, 1 passed
exact head 150940d8: 7 passed

complete tests/unit/agents/context:
402 passed

Black 23.3.0, Flake8, git diff --check:
passed
```

The negative control confirms that current main misses the CJK substring and
accepts all four integer/string variants of a descending span; the PR fixes
both.

I found one remaining correctness gap in the new CJK route. Every query that
contains CJK is sent to `_search_like_rows()`, but that function binds the
*entire* query as one pattern:

```python
params = [f"%{query}%", ...]
```

That fixes a single continuous substring, but a natural multi-keyword CJK
query still produces a false empty-history conclusion. This production-path
contract stores `项目的截止日期是周二` and calls the real `recall_history`
tool:

```text
query='项目'
1 row(s): 项目的截止日期是周二

query='项目 截止日期'
0 rows ... the history genuinely holds nothing there
```

Both keywords are present in the same row. The FTS path's documented
plain-query behavior is implicit AND across words, while the CJK LIKE route
requires the literal space and adjacency. Also, because FTS5 *is* available,
the existing LIKE-degraded notice that says to query one term at a time is not
inserted, so the model receives a confident false absence rather than a
recoverable hint.

I suggest preserving the single-substring case while splitting
whitespace-delimited CJK/mixed queries into terms and AND-combining bounded
LIKE predicates (with SQL LIKE metacharacters handled according to the
documented plain-text semantics). A tool-level regression should assert that
`项目 截止日期` finds `项目的截止日期是周二`; keeping the existing continuous
`紫水晶河马` case proves the original fix remains intact.

I did not find an active Issue or PR covering this multi-keyword boundary.

RerankerGuo (dismissed):
作者意图：修复 Scroll 对连续 CJK substring 的召回，以及 reversed expand span 的输入校验。

七个目标 case 已通过，但所有含 CJK 的查询都被送入 `_search_like_rows()`，它将完整 query 绑定为单个 `%query%`。因此自然的空格分隔多关键词查询要求字面空格和相邻顺序，明明同一行包含全部词却返回“历史中不存在”。

```mermaid
flowchart LR
 A[Query: 项目 截止日期] --> B[CJK LIKE route]
 B --> C[Pattern: %项目 截止日期%]
 C --> D[Stored row: 项目的截止日期是周二]
 D --> E[No literal match]
 E --> F[Confident false empty-history result]
```

真实 `recall_history` tool contract：`项目` 能找到该行，`项目 截止日期` 返回 0 rows；FTS5 实际可用，因此也不会附加 degraded LIKE 提示。

请保留连续 substring 行为，同时将 whitespace-delimited CJK/mixed query 拆为 terms，使用有界 AND-combined LIKE predicates，并按 plain-text contract 正确转义 `%`/`_` 等 LIKE metacharacters。提交 tool-level regression：`项目 截止日期` 必须找到 `项目的截止日期是周二`，并保留现有 `紫水晶河马` 连续字符串 case。

rayrayraykk (approved):

rayrayraykk (approved):

rayrayraykk (approved):

PR #6873, opened 2026-08-10
fix: normalize legacy local-path media sources when loading sessions

### LINKED ISSUE #6872: Legacy sessions with local-path media sources fail to load (Internal error)
## Summary

Legacy (pre-2.0) sessions that contain media blocks whose `source.url` is a **local filesystem path** fail to load on every message. The user sees a generic `Internal error` reply and the session can never be used again.

## Actual error

```
channel _stream_with_tracker failed: 1 validation error for URLSource
url
  Input should be a valid URL, relative URL without a base [type=url_parsing, input_value='/app/working/media/wecom_...png', input_type=str]
```

Traceback path:

```
runtime/builder.py build -> react_agent.load_state_dict
  -> app/chats/utils.py parse_legacy_memory_state
    -> _compat/message.py msg_from_dict -> _coerce_block
      -> _coerce_source -> URLSource(url=<local path>)  # pydantic ValidationError
```

## Root cause

The agentscope 2.0 migration introduced `URLSource` with strict `AnyUrl` validation and added `_ensure_url_scheme` to convert local paths to `file://` URIs — but only wired it into the **new-input direction** (`runtime/message_convert.py::_request_input_to_msgs`). The **legacy session-load direction** (`_compat/message.py::_coerce_source`) still builds `URLSource` from the raw stored value, so any legacy session containing a local-path media source (e.g. a `view_image` tool result whose output image block stored `url: "/app/working/media/xxx.png"`) crashes during load, and the channel fallback replies `Internal error`.

## Steps to reproduce

```python
from qwenpaw._compat.message import msg_from_dict

msg_from_dict({
    "id": "m1", "name": "user", "role": "user",
    "timestamp": "2026-01-01 00:00:00.000",
    "content": [{
        "type": "image",
        "source": {"type": "url", "url": "/app/working/media/x.png"},
    }],
})
# -> pydantic_core.ValidationError: Input should be a valid URL, relative URL without a base
```

## Expected behavior

Legacy local-path media sources should load like they did before the 2.0 migration (the media is a local file, resolvable on disk). `file://` conversion is already the established pattern in `_request_input_to_msgs` and the providers already resolve `file://` media back to disk and inline them as base64.

## Suggested fix

Apply `_ensure_url_scheme` in `_coerce_source` before constructing `URLSource`, and move the helper to `_compat/message.py` so both directions share one implementation. See the linked PR.

xingliu228:
## Description

Legacy (pre-2.0) sessions can contain media blocks whose `source.url` is a **local filesystem path** (e.g. WeCom images saved under the working dir). When such a session is loaded, `_coerce_source` builds `URLSource` directly from the raw path and pydantic rejects it (`Input should be a valid URL, relative URL without a base`), so every message in that session fails with a generic `Internal error`.

The agentscope 2.0 migration added `_ensure_url_scheme` (local path → `file://`) but only wired it into the new-input direction (`_request_input_to_msgs`), not the legacy session-load direction. This PR moves the helper to `_compat/message.py` so both directions share one implementation, applies it in `_coerce_source`, and extends it to Windows UNC paths.

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core
- [x] Tests

## Checklist

- [x] Run and pass relevant tests (`pytest` or as applicable)

## Testing

- Legacy image block with a platform-native absolute local path (`tmp_path`) loads successfully and produces a `file://` `URLSource`
- Windows UNC path normalization (`\\server\share\image.png` → `file://server/share/image.png`)
- Percent-encoded non-ASCII local paths are unquoted before normalization

```bash
pre-commit run --files src/qwenpaw/_compat/message.py src/qwenpaw/runtime/message_convert.py tests/unit/_compat/test_message.py
# check python ast ....... Passed
# check docstring is first  Passed
# mypy .................... Passed
# black ................... Passed
# flake8 .................. Passed
# pylint .................. Passed

pytest tests/unit/_compat/test_message.py tests/unit/runtime/test_message_convert.py -q
# 6 passed in 0.17s
```

Fixes #6872

### DIFF
--- a/src/qwenpaw/_compat/message.py
+++ b/src/qwenpaw/_compat/message.py
@@ -16,7 +16,9 @@ from __future__ import annotations
 
 import json
 import mimetypes
+from pathlib import Path
 from typing import Any, Mapping
+from urllib.parse import unquote
 
 from agentscope.message import (
     Base64Source,
@@ -31,6 +33,36 @@ _MODALITY_DEFAULT_MIME = {
     "audio": "audio/*",
     "video": "video/*",
 }
+
+
+def _ensure_url_scheme(url: str) -> str:
+    """Prepend ``file://`` when *url* is an absolute local path.
+
+    Handles Unix paths (``/``, ``~``), Windows drive paths
+    (e.g. ``C:\\`` or ``C:/``) and Windows UNC paths
+    (e.g. ``\\\\server\\share\\x.png`` → ``file://server/share/x.png``).
+
+    Always ``unquote()`` first so percent-encoded non-ASCII characters
+    (e.g. ``%E6%B5%8B%E8%AF%95`` → ``测试``) resolve to the real
+    filename on disk.  Then uses ``file://`` + raw path (not
+    ``Path.as_uri()``) to avoid re-encoding.
+    """
+    if url.startswith(("/", "~")):
+        resolved = str(Path(unquote(url)).expanduser().resolve())
+    elif len(url) >= 3 and url[1] == ":" and url[2] in ("/", "\\"):
+        resolved = str(Path(unquote(url)).resolve())
+    elif url.startswith("\\\\"):
+        resolved = unquote(url)
+    else:
+        return url
+
+    resolved = resolved.replace("\\", "/")
+    if resolved.startswith("//"):
+        # UNC path: \\server\share\... -> file://server/share/...
+        return "file://" + resolved.lstrip("/")
+    if not resolved.startswith("/"):
+        resolved = "/" + resolved
+    return "file://" + resolved
 
 
 def _coerce_source(
@@ -53,7 +85,12 @@ def _coerce_source(
 
     src_type = source.get("type")
     if src_type == "url":
-        url = source["url"]
+        # Legacy sessions (pre-2.0) sometimes stored local filesystem
+        # paths in ``url`` (e.g. WeCom images under the working dir).
+        # ``URLSource`` requires an absolute RFC 3986 URI, so normalize
+        # local paths to ``file://`` URIs here; the provider layer
+        # resolves those back to on-disk files when building requests.
+        url = _ensure_url_scheme(str(source["url"]))
         media_type = source.get("media_type")
         if not media_type:
             guessed, _ = mimetypes.guess_type(str(url))
--- a/tests/unit/_compat/test_message.py
+++ b/tests/unit/_compat/test_message.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+"""Regression tests for legacy session message compatibility shims."""
+
+from __future__ import annotations
+
+from agentscope.message import DataBlock, URLSource
+
+from qwenpaw._compat.message import _ensure_url_scheme, msg_from_dict
+
+
+def test_ensure_url_scheme_unc_path():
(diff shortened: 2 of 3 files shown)


tests/unit/_compat/test_message.py
```diff
@@ -0,0 +1,100 @@
...
+
+from urllib.parse import unquote
+
+from agentscope.message import Base64Source, DataBlock, URLSource
+
+from qwenpaw._compat.message import _ensure_url_scheme, msg_from_dict
+
+
+def test_ensure_url_scheme_absolute_unix_path():
+    assert _ensure_url_scheme("/app/working/media/x.png") == (
```
qbc2016, line 14: This hard-coded POSIX expectation will fail on Windows, where `Path("/app/...").resolve()` includes the current drive letter.

Please keep a focused regression test for the reported failure, but use
`tmp_path` to construct a platform-native absolute path. This will make the
test portable across the repository's Linux, macOS, and Windows CI matrix.

The unrelated HTTPS and base64 pass-through tests can be removed if we want
to keep this fix minimal.

src/qwenpaw/_compat/message.py
```diff
@@ -33,6 +35,30 @@
...
+    (e.g. ``C:\\`` or ``C:/``).
+
+    Always ``unquote()`` first so percent-encoded non-ASCII characters
+    (e.g. ``%E6%B5%8B%E8%AF%95`` → ``测试``) resolve to the real
+    filename on disk.  Then uses ``file://`` + raw path (not
+    ``Path.as_uri()``) to avoid re-encoding.
+    """
+    if url.startswith(("/", "~")):
+        resolved = str(Path(unquote(url)).expanduser().resolve())
+    elif len(url) >= 3 and url[1] == ":" and url[2] in ("/", "\\"):
```
qbc2016, line 51: A Windows UNC path such as `\\server\share\image.png` does not match this
branch, so it is passed unchanged to `URLSource` and still fails validation.

Could we normalize UNC paths as well, or explicitly narrow the documented
Windows-path support?

qbc2016 (commented):
Please see the inline comments and run `pip install pre-commit; pre-commit install; pre-commit run --all-files` to format the code.

tests/unit/_compat/test_message.py
xingliu228: Fixed. The test now builds the path from pytest `tmp_path` (platform-native) and asserts only the `file://` prefix, so it passes on Linux, macOS and Windows. I also removed the unrelated HTTPS/base64 pass-through tests to keep the PR focused on the reported regression; the remaining cases cover a local-path legacy image block, UNC normalization and percent-encoded paths.

src/qwenpaw/_compat/message.py
xingliu228: Fixed. `_ensure_url_scheme` now also normalizes Windows UNC paths: `\\server\share\image.png` → `file://server/share/image.png` (backslashes converted, leading slashes stripped before the authority). Added a regression test for it. The docstring was updated to reflect the supported forms.

xingliu228:
iShot_2026-08-12_08 47 16

qbc2016 (approved):
LGTM.

PR #6874, opened 2026-08-10
feat(mcp): add configurable tool call timeout

### LINKED ISSUE #6724: [Feature]: Configurable MCP tool-call timeout (per-client config + call-level guard)
## Summary

MCP tool calls currently have **no configurable timeout**: `MCPClientConfig` has no `timeout` field (silently dropped by Pydantic), and `call_tool` awaits the MCP session with no upper bound — so a slow or hung MCP server can stall a turn indefinitely. Add a per-client configurable tool-call timeout with an `asyncio.wait_for` guard in `call_tool`, and surface timeout errors to the caller (the LLM) instead of hanging.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Problem / Motivation

Two independent gaps make MCP timeouts effectively unmanageable:

1. **No configuration surface.** `MCPClientConfig` (`src/qwenpaw/config/config.py`) has no `timeout` field; adding one to config is silently discarded (already reported in #3997, still open). The driver layer (`drivers/handlers/mcp.py::_setup`) constructs `HttpStatefulClient` without passing any timeout values, so even the constructor defaults (`timeout=30`, `sse_read_timeout=300`) are never reachable from configuration.

2. **No call-level guard, and the transport-level timeout can be swallowed.** `call_tool` does `await self.session.call_tool(...)` with no `asyncio.wait_for` and no SDK `read_timeout_seconds`, so the MCP SDK waits forever by default. The HTTP transport read timeout (300 s) does fire, but the exception occurs inside `streamable_http_client`'s `post_writer` background task and is silently swallowed while the stream is closed — the awaiting caller never learns about it (this is even acknowledged in the `_handle_transport_error` docstring). Practical impact: a slow or wedged MCP server hangs the agent turn for minutes (observed with a remote MCP endpoint), with no way for the user to bound it.

## Proposed Solution

1. Add a `timeout` field (e.g. `timeout: float = 120.0`, or `None` = no limit) to `MCPClientConfig`, mirror it in the MCP create/update schemas and the UI form.
2. Pass it through the driver layer into `HttpStatefulClient` / `StdIOStatefulClient`.
3. Wrap `call_tool` in `asyncio.wait_for(timeout)`; on timeout, raise `TimeoutError` upward so the tool layer reports it to the LLM, which can decide to retry, switch tools, or tell the user — **no forced reconnect**. This is safe with the MCP SDK: `send_request` cleans up the request stream on cancellation, a late response is ignored by `request_id`, and a genuinely dead transport already triggers the existing `_handle_transport_error` reconnect on the next call.
4. (Optional but valuable) Make transport read timeouts propagate to the waiting caller instead of being swallowed by `post_writer`.

## Alternatives Considered

- **Reuse the existing global shell/tool timeout** (e.g. `shell_command_timeout`): works as a stopgap but conflates unrelated concerns and cannot be tuned per MCP server.
- **Lower `sse_read_timeout`**: ineffective for the hang, because the 300 s error is swallowed by `post_writer` anyway.
- **Leave it unbounded (status quo)**: unacceptable for remote MCP servers that can wedge for minutes.

## Additional Context

A local prototype (call_tool wrapped in `asyncio.wait_for`, config-driven, fallback 120 s, no forced reconnect) was verified end-to-end: a slow `tavily_extract` batch was cut at exactly the configured 5 s with `MCP tool call ... timed out after 5s`; the connection stayed healthy and the next call succeeded; with 180 s configured the same batch completed fully. Related: #3997 (config timeout silently dropped), #4033 (execution_timeout incorrectly bound to the 30 s HTTP connect timeout, fixed by #4061 to use sse_read_timeout instead), #4227 (calls blocking until timeout on HTTP error statuses).

## Willing to Contribute

- [ ] I am willing to open a PR for this feature (after discussion).

AaronZ345:
## Description

Add a positive per-client MCP tool-call deadline named `tool_call_timeout`, defaulting to 300 seconds. HTTP/SSE read budgets are raised to at least the configured deadline so values above 300 seconds are honored. The old `timeout` key is accepted only for legacy stdio configurations, where it is migrated to `tool_call_timeout`.

Closes #6724. Related to #3997.

## What Problem This Solves

A slow or wedged MCP server can currently block an agent turn indefinitely because per-client tool-call timeout values are discarded. This setting bounds one complete tool invocation, including implicit `tools/list`, header-refresh retries, and `tools/call`. It does not change HTTP connect, write, or pool timeouts.

## Type of Change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend
- [ ] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Summary

- validate finite tool-call deadlines in `(0, 86400]`, with a shared schema field and 300-second default
- apply one deadline across modern Streamable-HTTP discovery, header refresh, retry, and tool invocation
- raise HTTP/SSE read budgets to at least `tool_call_timeout`
- map direct and grouped MCP 408 / httpx timeout failures to the contextual `TimeoutError` without reconnecting
- preserve mixed exception groups so non-timeout failures are not hidden
- migrate legacy `timeout` only for stdio configurations; HTTP transport timeout values are not reinterpreted
- treat `tool_call_timeout=None` as unset during legacy migration

## Testing

- 145 focused MCP stateful, dual-protocol, and legacy-upgrade tests passed
- pre-commit checks passed for all changed files, including mypy, black, flake8, pylint, AST, trailing whitespace, and private-key detection
- regression coverage includes direct/grouped/mixed timeout exceptions, HTTP read-budget propagation, implicit `tools/list`, header-refresh retry, alias boundaries, schema descriptions, and `None` migration

## Risk & Scope

- No real multi-minute MCP server was used; timeout behavior is covered through SDK and transport-level test doubles.
- Existing HTTP connect, write, and pool timeout semantics are unchanged.

## Checklist

- [x] Run and pass relevant pre-commit hooks
- [x] Add/update regression coverage
- [x] Document the config field semantics

### DIFF
--- a/src/qwenpaw/app/mcp/schemas.py
+++ b/src/qwenpaw/app/mcp/schemas.py
@@ -6,7 +6,9 @@ from __future__ import annotations
 from datetime import datetime
 from typing import Any, Dict, List, Literal, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
+
+from ...mcp_timeout import DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS
 
 
 class MCPClientOAuthStatus(BaseModel):
@@ -60,10 +62,13 @@ class MCPClientInfo(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
-    timeout: float = Field(
-        default=120.0,
+    tool_call_timeout: float = Field(
+        default=DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
         gt=0,
-        description="Maximum duration of one MCP tool call in seconds",
+        description=(
+            "Maximum duration of one MCP tool call in seconds. This does not "
+            "change HTTP transport timeouts."
+        ),
     )
     tools: Optional[List[str]] = Field(
         default=None,
@@ -117,16 +122,28 @@ class MCPClientCreateRequest(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
-    timeout: float = Field(
-        default=120.0,
+    tool_call_timeout: float = Field(
+        default=DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
         gt=0,
-        description="Maximum duration of one MCP tool call in seconds",
+        description=(
+            "Maximum duration of one MCP tool call in seconds. This does not "
+            "change HTTP transport timeouts."
+        ),
     )
     tools: Optional[List[str]] = Field(
         default=None,
         description="Tool whitelist. Only listed tools will be loaded. "
         "None means load all tools.",
     )
+
+    @model_validator(mode="before")
+    @classmethod
+    def _normalize_legacy_timeout(cls, data):
+        if isinstance(data, dict) and "timeout" in data:
+            payload = dict(data)
+            payload.setdefault("tool_call_timeout", payload["timeout"])
+            return payload
+        return data
 
 
 class MCPClientUpdateRequest(BaseModel):
@@ -166,16 +183,28 @@ class MCPClientUpdateRequest(BaseModel):
         None,
         description="Working directory for stdio MCP command",
     )
-    timeout: Optional[float] = Field(
+    tool_call_timeout: Optional[float] = Field(
         None,
         gt=0,
-        description="Maximum duration of one MCP tool call in seconds",
+        description=(
+            "Maximum duration of one MCP tool call in seconds. This does not "
+            "change HTTP transport timeouts."
+        ),
     )
     tools: Optional[List[str]] = Field(
         None,
         description="Tool whitelist (omit to leave unchanged). "
         "Set to null to remove the whitelist.",
     )
+
+    @model_validator(mode="before")
+    @classmethod
+    def _normalize_legacy_timeout(cls, data):
+        if isinstance(data, dict) and "timeout" in data:
+            payload = dict(data)
+            payload.setdefault("tool_call_timeout", payload["timeout"])
+            return payload
+        return data
 
 
 class MCPAccessRule(BaseModel):
--- a/src/qwenpaw/app/mcp/schemas.py
+++ b/src/qwenpaw/app/mcp/schemas.py
@@ -60,6 +60,11 @@ class MCPClientInfo(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
+    timeout: float = Field(
+        default=120.0,
+        gt=0,
+        description="Maximum duration of one MCP tool call in seconds",
+    )
     tools: Optional[List[str]] = Field(
         default=None,
         description="Tool whitelist. Only listed tools will be loaded. "
@@ -112,6 +117,11 @@ class MCPClientCreateRequest(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
+    timeout: float = Field(
+        default=120.0,
+        gt=0,
+        description="Maximum duration of one MCP tool call in seconds",
+    )
     tools: Optional[List[str]] = Field(
         default=None,
         description="Tool whitelist. Only listed tools will be loaded. "
@@ -155,6 +165,11 @@ class MCPClientUpdateRequest(BaseModel):
     cwd: Optional[str] = Field(
         None,
         description="Working directory for stdio MCP command",
+    )
+    timeout: Optional[float] = Field(
+        None,
+        gt=0,
+        description="Maximum duration of one MCP tool call in seconds",
     )
     tools: Optional[List[str]] = Field(
         None,
--- a/src/qwenpaw/app/mcp/schemas.py
+++ b/src/qwenpaw/app/mcp/schemas.py
@@ -8,7 +8,10 @@ from typing import Any, Dict, List, Literal, Optional
 
 from pydantic import BaseModel, Field, model_validator
 
-from ...mcp_timeout import DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS
+from ...mcp_timeout import (
+    DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+    MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+)
 
 
 class MCPClientOAuthStatus(BaseModel):
@@ -65,9 +68,12 @@ class MCPClientInfo(BaseModel):
     tool_call_timeout: float = Field(
         default=DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
         gt=0,
+        le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+        allow_inf_nan=False,
         description=(
             "Maximum duration of one MCP tool call in seconds. This does not "
-            "change HTTP transport timeouts."
+            "change HTTP connect/write/pool transport timeouts or SSE read "
+            "timeouts."
         ),
     )
     tools: Optional[List[str]] = Field(
@@ -125,9 +131,12 @@ class MCPClientCreateRequest(BaseModel):
     tool_call_timeout: float = Field(
         default=DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
         gt=0,
+        le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+        allow_inf_nan=False,
         description=(
             "Maximum duration of one MCP tool call in seconds. This does not "
-            "change HTTP transport timeouts."
+            "change HTTP connect/write/pool transport timeouts or SSE read "
+            "timeouts."
         ),
     )
     tools: Optional[List[str]] = Field(
@@ -186,9 +195,12 @@ class MCPClientUpdateRequest(BaseModel):
     tool_call_timeout: Optional[float] = Field(
         None,
         gt=0,
+        le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+        allow_inf_nan=False,
         description=(
             "Maximum duration of one MCP tool call in seconds. This does not "
-            "change HTTP transport timeouts."
+            "change HTTP connect/write/pool transport timeouts or SSE read "
+            "timeouts."
         ),
     )
     tools: Optional[List[str]] = Field(
--- a/src/qwenpaw/app/mcp/schemas.py
+++ b/src/qwenpaw/app/mcp/schemas.py
@@ -10,6 +10,7 @@ from pydantic import BaseModel, Field
 
 from ...mcp_timeout import (
     DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+    MCPToolCallTimeout,
     mcp_tool_call_timeout_field,
 )
 
@@ -65,7 +66,7 @@ class MCPClientInfo(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
-    tool_call_timeout: float = mcp_tool_call_timeout_field(
+    tool_call_timeout: MCPToolCallTimeout = mcp_tool_call_timeout_field(
         DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
     )
     tools: Optional[List[str]] = Field(
@@ -120,7 +121,7 @@ class MCPClientCreateRequest(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
-    tool_call_timeout: float = mcp_tool_call_timeout_field(
+    tool_call_timeout: MCPToolCallTimeout = mcp_tool_call_timeout_field(
         DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
     )
     tools: Optional[List[str]] = Field(
@@ -167,7 +168,9 @@ class MCPClientUpdateRequest(BaseModel):
         None,
         description="Working directory for stdio MCP command",
     )
-    tool_call_timeout: Optional[float] = mcp_tool_call_timeout_field(None)
+    tool_call_timeout: Optional[
+        MCPToolCallTimeout
+    ] = mcp_tool_call_timeout_field(None)
     tools: Optional[List[str]] = Field(
         None,
         description="Tool whitelist (omit to leave unchanged). "
--- a/src/qwenpaw/app/mcp/schemas.py
+++ b/src/qwenpaw/app/mcp/schemas.py
@@ -10,8 +10,7 @@ from pydantic import BaseModel, Field
 
 from ...mcp_timeout import (
     DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
-    MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
-    MCP_TOOL_CALL_TIMEOUT_DESCRIPTION,
+    mcp_tool_call_timeout_field,
 )
 
 
@@ -66,12 +65,8 @@ class MCPClientInfo(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
-    tool_call_timeout: float = Field(
-        default=DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
-        gt=0,
-        le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
-        allow_inf_nan=False,
-        description=MCP_TOOL_CALL_TIMEOUT_DESCRIPTION,
+    tool_call_timeout: float = mcp_tool_call_timeout_field(
+        DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
     )
     tools: Optional[List[str]] = Field(
         default=None,
@@ -125,12 +120,8 @@ class MCPClientCreateRequest(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
-    tool_call_timeout: float = Field(
-        default=DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
-        gt=0,
-        le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
-        allow_inf_nan=False,
-        description=MCP_TOOL_CALL_TIMEOUT_DESCRIPTION,
+    tool_call_timeout: float = mcp_tool_call_timeout_field(
+        DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
     )
     tools: Optional[List[str]] = Field(
         default=None,
@@ -176,13 +167,7 @@ class MCPClientUpdateRequest(BaseModel):
         None,
         description="Working directory for stdio MCP command",
     )
-    tool_call_timeout: Optional[float] = Field(
-        None,
-        gt=0,
-        le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
-        allow_inf_nan=False,
-        description=MCP_TOOL_CALL_TIMEOUT_DESCRIPTION,
-    )
+    tool_call_timeout: Optional[float] = mcp_tool_call_timeout_field(None)
     tools: Optional[List[str]] = Field(
         None,
         description="Tool whitelist (omit to leave unchanged). "
--- a/src/qwenpaw/app/mcp/schemas.py
+++ b/src/qwenpaw/app/mcp/schemas.py
@@ -6,11 +6,12 @@ from __future__ import annotations
 from datetime import datetime
 from typing import Any, Dict, List, Literal, Optional
 
-from pydantic import BaseModel, Field, model_validator
+from pydantic import BaseModel, Field
 
 from ...mcp_timeout import (
     DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
     MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+    MCP_TOOL_CALL_TIMEOUT_DESCRIPTION,
 )
 
 
@@ -70,11 +71,7 @@ class MCPClientInfo(BaseModel):
         gt=0,
         le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
         allow_inf_nan=False,
-        description=(
-            "Maximum duration of one MCP tool call in seconds. This does not "
-            "change HTTP connect/write/pool transport timeouts or SSE read "
-            "timeouts."
-        ),
+        description=MCP_TOOL_CALL_TIMEOUT_DESCRIPTION,
     )
     tools: Optional[List[str]] = Field(
         default=None,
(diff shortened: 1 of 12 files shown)


src/qwenpaw/config/config.py
```diff
@@ -2168,6 +2168,11 @@ class MCPClientConfig(BaseModel):
     args: List[str] = Field(default_factory=list)
     env: Dict[str, str] = Field(default_factory=dict)
     cwd: str = ""
+    timeout: float = Field(
+        default=120.0,
+        gt=0,
+        description="Maximum duration of one MCP tool call in seconds.",
+    )
```
yuanxs21, line 2377: This implemented default is a silent behavior change, and `gt=0` is not enough validation.

Currently, stdio `call_tool` has no call-level deadline, and HTTP MCP uses `sse_read_timeout=300`. In this implementation, each client without an explicit value fails at 120s. Long-running HTTP MCP tools may regress.

src/qwenpaw/drivers/adapters/mcp_card_builder.py
```diff
@@ -170,6 +170,7 @@ def build_mcp_driver_card(
         _preserve_oauth_authorization_binding(existing, endpoint)
         env_aliases = header_plan.env_aliases
 
+    endpoint["timeout"] = float(data.get("timeout") or 120.0)
```
yuanxs21: Please do not coerce `120.0` to the fallback.

- `0` / `0.0` / `""` become 120 instead of failing `gt=0`.
- Negatives are truthy, so `timeout: -1` is stored and later passed to `wait_for`.
- Non-numeric values (`"fast"`, a list) raise `ValueError` here or in the handler, not `DriverCardError`.

Missing should default, and present-but-invalid should reject. Same for L262 and L383.

src/qwenpaw/app/mcp/schemas.py
```diff
@@ -60,6 +60,11 @@ class MCPClientInfo(BaseModel):
         default="",
         description="Working directory for stdio MCP command",
     )
+    timeout: float = Field(
+        default=120.0,
+        gt=0,
+        description="Maximum duration of one MCP tool call in seconds",
+    )
```
yuanxs21, line 71: If keep the name `timeout` on the public API, Console types/UI will also need it later; otherwise, callers cannot configure the value that this schema now persists.

Please:
- Set `allow_inf_nan=False` and a finite upper bound.
- Make the description explicit: this bounds one `tools/call`, not HTTP connect / SSE read (`HttpStatefulClient.timeout=30`, `sse_read_timeout=300`).
- Revisit `default=120.0` (see comment on `MCPClientConfig`). 120s is below the existing HTTP read budget.

Same for L120 and L169.

src/qwenpaw/drivers/adapters/mcp_legacy_config.py
```diff
@@ -260,6 +260,7 @@ def legacy_mcp_client_to_driver(
             "headers": header_binding,
         }
 
+    endpoint["timeout"] = float(getattr(config, "timeout", 120.0) or 120.0)
```
yuanxs21: The default is written twice (`getattr(..., 120.0)` and `or 120.0`), and `or` still turns `0` into 120.

`MCPClientConfig.timeout` already defaults to 120 and rejects non-positive values, so the `or 120.0` is either redundant (valid configs) or hiding invalid cards. Please read `config.timeout` directly after validation, or share the helper used in `mcp_card_builder`.

src/qwenpaw/drivers/handlers/mcp.py
```diff
@@ -64,6 +64,7 @@ async def _setup(self) -> None:
         """Create and connect StdIOStatefulClient or HttpStatefulClient."""
         endpoint = self._card.endpoint
         transport = str(endpoint.get("transport") or "stdio")
+        tool_call_timeout = float(endpoint.get("timeout") or 120.0)
```
yuanxs21: Same as the former fault timeout setting, and this is on the hot path: a hand-edited DriverCard with `timeout: "120s"` or `[]` raises `ValueError` from `_setup` instead of `DriverCardError`.

`validate_mcp_endpoint()` was not updated for the new field. Please validate type / `> 0` / finite there, then read the already-checked value here.

Also: this value is passed as `tool_call_timeout` only. `HttpStatefulClient.timeout` and `sse_read_timeout` stay at constructor defaults. A configured `timeout=600` still cannot keep an HTTP tool alive past the 300s read timeout — and if `post_writer` swallows that error, the caller waits until `tool_call_timeout` anyway. Worth calling out, or wiring the call deadline and the HTTP read budget together.

src/qwenpaw/drivers/handlers/mcp_stateful_client.py
```diff
@@ -427,11 +428,20 @@ async def call_tool(self, name: str, arguments: dict | None = None):
...
-            return await self.session.call_tool(name, arguments or {})
+            return await asyncio.wait_for(
+                self.session.call_tool(name, arguments or {}),
+                timeout=self.tool_call_timeout,
+            )
+        except asyncio.TimeoutError as exc:
+            raise TimeoutError(
+                f"MCP tool call '{name}' on client '{self.name}' timed out "
+                f"after {self.tool_call_timeout:g}s",
+            ) from exc
```
yuanxs21: Please do not wrap `session.call_tool` in `asyncio.wait_for`.

In the current version, `wait_for` is `async with asyncio.timeout()` and cancels the *current* task. `mcp.ClientSession.send_request` then runs its `finally` (`await response_stream.aclose()`) while the task is still cancelled, so cleanup can raise `CancelledError` and the caller may not get this wrapped `TimeoutError`.

This file exists to keep anyio CancelScopes enter/exit on the same task. Cancelling an anyio-backed session from `asyncio.timeout` works against that.

Same for L93 below.

yuanxs21 (changes requested):
Thanks for taking this fix for #6724.

Here are the requesting changes before merge:

1. Defaulting every MCP client to 120s is a silent behavior change. Today stdio `call_tool` is unbounded and HTTP read is 300s (`sse_read_timeout`). Tools that currently finish in 2–5 minutes will start failing. Please treat this as a behavior change: raise the default to at least the existing HTTP read budget, or use `None` = unlimited, and call that out in the PR.

2. `asyncio.wait_for` around `session.call_tool` is the wrong timeout primitive on this path. On 3.11+ it uses `asyncio.timeout` and cancels the *current* task, which can interrupt `ClientSession.send_request`'s `finally` (`await stream.aclose()`). This module exists specifically to keep anyio cancel scopes well-behaved. The SDK already has `call_tool(..., read_timeout_seconds=timedelta(...))` via `anyio.fail_after`. Please use that, and map the SDK 408 `McpError` to the current timeout message.

3. `float(x or 120.0)` is used in ~8 places. `0` / `""` silently become 120; negatives pass through; non-numeric DriverCard values raise `ValueError` in `_setup` instead of `DriverCardError`. Please parse missing vs invalid vs non-positive separately, share one constant, and validate in `validate_mcp_endpoint`.

4. Config field name `timeout` collides with `HttpStatefulClient.timeout` (HTTP connect/write/pool, default 30s). This PR does not change transport timeouts, so `#3997` configs like `"timeout": 1200` will not do what those users asked for. Please document that this is a tools/call deadline only, or rename to `tool_call_timeout`.

The config round-trip itself looks fine (`exclude_unset` + `_card_to_client_data`). The issues are default, timeout implementation, and parsing — not persistence.

AaronZ345:
Addressed the requested changes:\n\n- Raised the default tool-call timeout to the existing HTTP SSE read budget (300s), avoiding the earlier 120s regression.\n- Renamed the persisted/public config field to tool_call_timeout while keeping legacy timeout input normalization.\n- Added finite validation with allow_inf_nan=false and an upper bound, and clarified the schema description that this limits one tools/call rather than HTTP connect/write/pool or SSE read timeouts.\n- Removed asyncio.wait_for around session.call_tool and now passes the deadline through the MCP SDK read_timeout_seconds path.\n\nValidation: uv run python -m pytest tests/unit/drivers/handlers/test_mcp_stateful_client.py -q -> 44 passed.

src/qwenpaw/drivers/adapters/mcp_legacy_config.py
```diff
@@ -260,6 +261,9 @@ def legacy_mcp_client_to_driver(
             "headers": header_binding,
         }
 
+    endpoint["tool_call_timeout"] = get_mcp_tool_call_timeout(
+        config.model_dump(),
+    )
```
yuanxs21, line 271: The rest of this function uses `getattr(config, ...)`. Callers and unit tests pass `types.SimpleNamespace`, which has no `model_dump()` — this will raise `AttributeError` before the default timeout can apply.

Production migration happens to pass `MCPClientConfig`, so it may look fine, but the helper is not Pydantic-only. Please dump the same way `_model_dump` does in `mcp_card_builder`, and only put a timeout key in the dict when the attribute is actually present. Putting `timeout: None` would hit `parse_mcp_tool_call_timeout`'s "must be provided when present" path.

yuanxs21 (changes requested):

yuanxs21:
Please resolve the conflicts.

yuanxs21 (approved):
LGTM.

src/qwenpaw/drivers/handlers/mcp_streamable_http.py
```diff
@@ -656,13 +659,25 @@ async def _call() -> Any:
...
+
         try:
-            result = await _call()
+            result = await _call_with_timeout()
         except _JsonRpcError as exc:
             if exc.code != _JSONRPC_HEADER_MISMATCH:
                 raise
             await self.list_tools()
-            result = await _call()
+            result = await _call_with_timeout()
```
yuanxs21: This timeout wrapper is incomplete.

`call_tool` still does an unbounded `await self.list_tools()` when `_tools_listed` is false (just above this hunk), and again on the header-mismatch retry at L679. A hung `tools/list` never enters `wait_for`, so the first invocation — the case this PR is meant to bound — can still stall the turn.

`wait_for` on 3.11+ is `asyncio.timeout` and cancels the *current* task. That is the same issue called out on the stateful path. After the timeout, `is_connected` stays True; a half-closed httpx stream can poison later calls on this client.

Please:
- include the implicit `list_tools` in the same deadline, or stop listing from the `call_tool` hot path;
- bound the tool POST with a per-request httpx timeout (or a dedicated task) instead of `wait_for` on the agent-turn task;
- map `httpx.ReadTimeout` / `TimeoutException` to the same `TimeoutError` message used below, without leaving the client in a dirty connected state.

src/qwenpaw/drivers/handlers/mcp_stateful_client.py
```diff
@@ -588,6 +606,11 @@ async def call_tool(self, name: str, arguments: dict | None = None):
                 raise RuntimeError(
                     f"MCP client '{self.name}' request was aborted.",
                 ) from exc
+            if _is_mcp_request_timeout(exc):
+                raise TimeoutError(
+                    f"MCP tool call '{name}' on client '{self.name}' timed "
+                    f"out after {self.tool_call_timeout:g}s",
+                ) from exc
             self._handle_transport_error(exc, session)
```
yuanxs21, line 622: This only maps the SDK `McpError(code=408)`. It does not cover the HTTP layer.

`HttpStatefulClient._setup_transport` still builds httpx with `read=sse_read_timeout` (default 300s) and SSE with the same budget. Those lines are unchanged in this PR. If the server sends no bytes, `httpx.ReadTimeout` is a `TransportError`, so `_handle_transport_error` tears down the session and sets `_reload_event`.

At the default 300/300, that is a race. If the user sets `tool_call_timeout` above 300, HTTP always wins: the configured deadline never becomes a 408, and the call reconnects instead of raising the `TimeoutError` this block promises.

`test_call_tool_maps_sdk_timeout_without_reconnect` only injects `McpError(408)`, so this path is untested.

Please treat `httpx.ReadTimeout` / `TimeoutException` on `tools/call` as a call deadline (same `TimeoutError`, no reconnect), and make the HTTP read budget at least `tool_call_timeout` for that request. Otherwise, HTTP/SSE does not actually honor this setting.

src/qwenpaw/drivers/handlers/mcp_stateful_client.py
```diff
@@ -981,6 +1007,7 @@ def __init__(
         headers: dict[str, str] | None = None,
         timeout: float = 30,
         sse_read_timeout: float = 60 * 5,
+        tool_call_timeout: float = DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
```
yuanxs21, line 1018: `StdIOStatefulClient.__init__` still stores read_timeout_seconds, but nothing reads it. `ClientSession` is constructed without it; `call_tool` uses `tool_call_timeout` only. The HTTP client already dropped this field.

Do not keep a second timeout attribute with no runtime effect. Drop the assignment (keep the param only if `agentscope` compatibility requires it), or pass `timedelta(seconds=tool_call_timeout)` into `ClientSession()` so `list_tools` / `initialize` are bounded too.

src/qwenpaw/app/mcp/schemas.py
```diff
@@ -112,12 +128,32 @@ class MCPClientCreateRequest(BaseModel):
...
+        default=DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+        gt=0,
+        le=MAX_MCP_TOOL_CALL_TIMEOUT_SECONDS,
+        allow_inf_nan=False,
+        description=(
+            "Maximum duration of one MCP tool call in seconds. This does not "
+            "change HTTP connect/write/pool transport timeouts or SSE read "
+            "timeouts."
+        ),
+    )
```
yuanxs21, line 126: This description is accurate and is also a bug.

If this field “does not change … SSE read timeouts”, then any HTTP/SSE value above the default 300s `sse_read_timeout` cannot work. The schema still accepts up to 86400. Users will think they configured a longer tool deadline; httpx will cut the socket at 300s, and the stateful client will reconnect.

Either:
- raise the HTTP/SSE read budget to at least this value (and say so here), or
- reject / clamp `tool_call_timeout` so it cannot exceed the active transport read timeout, and document `effective = min(tool_call_timeout, sse_read_timeout)`.

The same Field text is copied in `MCPClientInfo`, `MCPClientUpdateRequest`, and `MCPClientConfig`. Please fix the behavior once and share the Field.

src/qwenpaw/config/config.py
```diff
@@ -2392,6 +2407,9 @@ def _normalize_legacy_fields(cls, data):
         if "type" in payload and "transport" not in payload:
             payload["transport"] = payload["type"]
 
+        if "timeout" in payload and "tool_call_timeout" not in payload:
+            payload["tool_call_timeout"] = payload["timeout"]
+
```
yuanxs21: This still treats legacy `timeout` as `tool_call_timeout`. That is the #3997 collision this rename was meant to avoid.

A config like `"timeout": 1200` now persists as a 1200s *tool-call* deadline. HTTP connect/write/pool stay at 30s, SSE read stays at 300s, and on HTTP the 1200s value is then clipped by that 300s read budget.

Please do not silently alias `timeout` → `tool_call_timeout` for HTTP clients, or accept it only with a warning and without claiming it changes transport timeouts. If you keep the alias for stdio / old agent.json, say in the schema that `timeout` is a deprecated alias for `tool_call_timeout` and is not `HttpStatefulClient.timeout`.

src/qwenpaw/drivers/adapters/mcp_legacy_config.py
```diff
@@ -260,6 +261,12 @@ def legacy_mcp_client_to_driver(
             "headers": header_binding,
         }
 
+    timeout_config: dict[str, Any] = {}
+    if hasattr(config, "tool_call_timeout"):
+        timeout_config["tool_call_timeout"] = config.tool_call_timeout
+    elif hasattr(config, "timeout"):
+        timeout_config["timeout"] = config.timeout
+    endpoint["tool_call_timeout"] = get_mcp_tool_call_timeout(timeout_config)
```
yuanxs21, line 272: `hasattr` is true when the attribute exists and is `None` (dataclass / SimpleNamespace with `tool_call_timeout: float | None = None`). You then call `parse_mcp_tool_call_timeout(None)` → `ValueError("must be provided when present")` and never fall through to `timeout` or the 300s default. `_migrate_one_client` does not catch that per client.

Please read the value and treat `None` as unset:

```python
raw = getattr(config, "tool_call_timeout", None)
if raw is None:
    raw = getattr(config, "timeout", None)
timeout_config = {"tool_call_timeout": raw} if raw is not None else {}

File: src/qwenpaw/mcp_timeout.py [hunk 14]
```diff
@@ -0,0 +1,53 @@
...
+        return parse_mcp_tool_call_timeout(
+            endpoint[LEGACY_MCP_TOOL_CALL_TIMEOUT_FIELD],
+        )
+    return DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS
+
+
+def parse_mcp_tool_call_timeout(value: Any = _MISSING) -> float:
+    """Parse a configured MCP tool-call timeout in seconds."""
+    if value is _MISSING:
+        return DEFAULT_MCP_TOOL_CALL_TIMEOUT_SECONDS
```
yuanxs21: `get_mcp_tool_call_timeout` never calls `parse_mcp_tool_call_timeout()` without a value; a missing key returns the default above. The `_MISSING` default is unused in production. Drop it, or make `parse` require an argument.

yuanxs21 (changes requested):
Request extra changes.

The stdio path looks right: SDK `call_tool(..., read_timeout_seconds=timedelta(...))`, 408 → TimeoutError, no reconnect, and the parser rejects 0 / inf / values above 24h.

The HTTP/SSE path does not match the feature as described:

1. `tool_call_timeout` is capped in practice by the unchanged 300s `sse_read_timeout` / httpx read budget. Values above 300 are accepted by the schema and then silently ignored.
2. When the HTTP layer times out first, the failure is `httpx.ReadTimeout` (a transport error) → reconnect, not the mapped `TimeoutError`. That is the opposite of “keep the session and fail the call”.
3. On the modern Streamable-HTTP client, `call_tool` still runs an unbounded `list_tools()` before (and on header-mismatch retry after) `wait_for`. A hung `tools/list` can still stall the turn.

Please fix those three before merge. Nits on the `timeout` alias, `hasattr(None)`, and the copied Field/validator are secondary.

src/qwenpaw/mcp_timeout.py
```diff
@@ -0,0 +1,81 @@
...
+    if (
+        getattr(exc, "code", None) == MCP_REQUEST_TIMEOUT_CODE
+        or getattr(error, "code", None) == MCP_REQUEST_TIMEOUT_CODE
+    ):
+        return True
+    sub_excs = getattr(exc, "exceptions", None)
+    return bool(sub_excs) and all(
+        is_mcp_request_timeout(item) for item in sub_excs
+    )
+
```
yuanxs21, line 63: This classifier is too broad.

httpx.TimeoutException includes ConnectTimeout, WriteTimeout, and PoolTimeout. Those are transport failures, not a tools/call deadline. Mapping them here skips _handle_transport_error, so a dead TCP/pool
stays is_connected=True and the next turn waits another full deadline.

getattr(exc, "code", None) == 408 also matches any object with a .code attribute, not just MCP SDK McpError / _JsonRpcError.

Please treat as a call timeout only:
- MCP / JSON-RPC 408 (exc.error.code or _JsonRpcError.code)
- httpx.ReadTimeout

Let connect/write/pool timeouts keep going through reconnect.

src/qwenpaw/mcp_timeout.py
```diff
@@ -0,0 +1,81 @@
...
+            raise ValueError("must be a positive number")
+        raw_value = value.strip()
+    else:
+        raw_value = value
+    try:
+        timeout = float(raw_value)
+    except (TypeError, ValueError) as exc:
+        raise ValueError("must be a positive number") from exc
+    if not math.isfinite(timeout) or timeout <= 0:
+        raise ValueError("must be a positive number")
```
yuanxs21, line 101: bool is a subclass of int, so float(True) == 1.0 and passes gt=0. A hand-edited card with "tool_call_timeout": true becomes a 1s deadline instead of a validation error.

Reject isinstance(value, bool) before the float() coercion. Same for the shared Pydantic field if it currently accepts True.

src/qwenpaw/drivers/handlers/mcp.py
```diff
@@ -262,6 +266,13 @@ async def _guarded_execute(
 def validate_mcp_endpoint(card: DriverCard) -> None:
     """Validate MCP endpoint shape beyond generic DriverCard checks."""
     endpoint = card.endpoint
+    try:
+        get_mcp_tool_call_timeout(endpoint)
+    except ValueError as exc:
+        raise DriverCardError(
+            f"DriverCard {card.name} endpoint.tool_call_timeout " f"{exc}",
```
yuanxs21: Accidental adjacent f-strings. Merge into one message.

src/qwenpaw/drivers/handlers/mcp_stateful_client.py
```diff
@@ -1018,8 +1041,17 @@ def __init__(
...
+            sse_read_timeout.total_seconds()
+            if isinstance(sse_read_timeout, timedelta)
+            else float(sse_read_timeout)
+        )
+        self.sse_read_timeout = max(
+            sse_read_timeout_seconds,
+            float(tool_call_timeout),
+        )
+        self.read_timeout_seconds = self.sse_read_timeout
+        self.tool_call_timeout = tool_call_timeout
```
yuanxs21, line 1066: self.read_timeout_seconds is written here and never read. ClientSession is still constructed without it; call_tool uses tool_call_timeout; HTTP/SSE use sse_read_timeout.

Do not persist a second timeout field that has no runtime effect. Either pass it into ClientSession(read_stream, write_stream, read_timeout_seconds=...), or drop the assignment.

src/qwenpaw/drivers/handlers/mcp_streamable_http.py
```diff
@@ -656,13 +663,43 @@ async def _call() -> Any:
...
+            task.cancel()
+            await asyncio.gather(task, return_exceptions=True)
+            raise
+        if not done:
+            task.cancel()
+            await asyncio.gather(task, return_exceptions=True)
+            raise TimeoutError(
+                f"MCP tool call '{name}' on client '{self.name}' timed "
+                f"out after {self.tool_call_timeout:g}s",
+            )
```
yuanxs21, line 718: Cancelling the in-flight POST and leaving is_connected=True can poison the shared httpx.AsyncClient (half-closed stream / HTTP/1.1 reuse). The next call_tool then fails for a reason that is no longer a deadline.

After this cancel, check that _http is still usable. If the stream is dead, close/reconnect; do not keep a dirty client marked connected. Mapping the timeout to TimeoutError is correct; keeping a broken transport is not.

yuanxs21 (changes requested):

src/qwenpaw/drivers/handlers/mcp_streamable_http.py
```diff
@@ -656,13 +687,47 @@ async def _call() -> Any:
...
+            await asyncio.gather(task, return_exceptions=True)
+            raise
+        if not done:
+            task.cancel()
+            await asyncio.gather(task, return_exceptions=True)
+            await self._reset_http_client_after_deadline()
+            raise TimeoutError(
+                f"MCP tool call '{name}' on client '{self.name}' timed "
+                f"out after {self.tool_call_timeout:g}s",
+            )
```
yuanxs21, line 718: The connection pool is only replaced on the `asyncio.wait` deadline. The other two paths that abandon or end an in-flight request leave `self._http` untouched:

1. The mapped-timeout path: when the task itself fails with `httpx.ReadTimeout` (or JSON-RPC 408) it is mapped to `TimeoutError` without reset. At the default 300/300 the wait deadline and the HTTP read budget fire at the same time, so this path is reachable on a normal timeout; the next call then reuses a half-closed connection, and the failure is no longer a deadline error.

2. The `except BaseException` branch: parent-task cancellation cancels the in-flight POST and re-raises without reset — the same dirty-pool case.

Please call `_reset_http_client_after_deadline()` on every path that abandons an in-flight request: before raising the mapped `TimeoutError`, and in the `except BaseException` branch before `raise`. Add regression coverage for both (mapped `ReadTimeout` replaces the client; caller cancellation replaces the client).

yuanxs21 (changes requested):

PR #6884, opened 2026-08-10
fix: make Auto-Dream integration resilient

### LINKED ISSUE #6841: [Feature]: Auto-Dream: single unit integration failure (LLM returned empty schema) marks the whole task as error — suggest retry + tolerance
Description:

One of my agents hit a nightly Auto-Dream failure. The task summary said status=error, but looking at the logs, the actual outcome was mostly successful — one integration unit failed on schema validation after its file had already been written to disk.

Steps to reproduce (observed, nightly run):

Auto-Dream runs nightly, scans 2 days of daily notes, extracts 5 units.
Unit 2/5 (bucket=wiki, paths=2) — topic content is domain-specific and irrelevant here:
[WriteStep] wrote path=digest/wiki/<some-topic>.md bytes=2613 — file written successfully
[DreamIntegrateStep] unit 2/5 failed: ValidationError: 2 validation errors for IntegrateOutcome — action and target_path missing, input_value={} (empty dict)
Other units (3/5, 4/5, 5/5) all succeeded.
Result: integrate success=False integrated=4 failed=1 → finish success=False failed_units=1 → the whole dream task pushed to inbox as status=error, and cron logged ERROR Failed to execute dream task.
Root cause: The integrate step asks the LLM to return a structured IntegrateOutcome (action, target_path, ...) validated via Pydantic. In this run the LLM returned an empty object {} (likely output truncation or schema drift), which fails validation — even though the content itself had already been written to disk. One failed unit flips the entire nightly task to error.

Suggestions:

Retry on schema validation failure — one automatic retry with the previous LLM response as context would likely fix most transient empty/truncated outputs.
Don't fail the whole task for one unit — a unit whose file was already written could be recovered from the on-disk content (parse/regenerate the metadata), or at least be reported as partial instead of error. failed_units is already tracked; surfacing it as a warning (not error) would reduce noise.
Validate before write — ask the LLM to return the metadata first, or include a fallback that derives action/target_path from the written file when the structured response is empty.
Impact: Cosmetic-ish (files are fine, next run re-digests), but status=error on every ~1-in-N nights is misleading and hides real failures.

Image

huiyiyichen:
## Description

This PR makes Auto-Dream integration tolerant of malformed structured output from the LLM.

## What Problem This Solves

A single empty or invalid integration schema currently turns the whole Auto-Dream task into an error. Successful units should remain successful, failed units should be reported as partial progress, and completed digest files must not be written again during recovery.

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Documentation only
- [ ] Refactoring only

## Component(s) Affected

- [x] Core
- [x] Tests
- [ ] Console
- [ ] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation
- [ ] CI/CD
- [ ] Scripts

## Changes

- Added a single retry for empty or invalid structured integration output.
- Restrict the retry to read-only tools when the first attempt already changed a digest file.
- Preserve successful units and expose `success`, `partial`, or `error` Auto-Dream status in the final response and inbox event.
- Make integration result recording idempotent so completed units and target nodes are not duplicated.
- Added focused tests for retry recovery, partial results, full failure, idempotency, and status propagation.

Closes #6841

## Checklist

- [x] Ran and passed `pre-commit run --all-files`
- [x] Ran and passed the relevant focused tests
- [x] Documentation update is not needed for this internal reliability fix

## Testing

- `PYTHONPATH=src python -m pytest -q tests/unit/agents/memory` — **93 passed**
- `PYTHONPATH=src python -m pytest -q tests/unit/agents/memory/test_reme_dream.py tests/unit/agents/memory/test_reme_config.py` — **21 passed**

- `pre-commit run --all-files --verbose` — **Passed**, exit code 0
  - Runtime: Python 3.12.12 / pre-commit 4.6.1
  - check-ast, YAML/XML/TOML/JSON checks, trailing-whitespace, add-trailing-comma: Passed
  - mypy: Passed
  - black: Passed
  - flake8: Passed
  - pylint: Passed
  - prettier: Passed
  - actionlint: Passed
- Focused pre-commit on the changed files: mypy, Black, Flake8, Pylint, and structural hooks passed.

## Evidence

- Invalid schema recovers after exactly one retry.
- A unit that wrote a digest file is retried with read-only tools.
- Mixed unit outcomes produce `partial` status without discarding successful integrations.
- A second integration pass skips an already recorded unit and does not duplicate target nodes.

### DIFF
--- a/src/qwenpaw/agents/memory/reme_config.py
+++ b/src/qwenpaw/agents/memory/reme_config.py
@@ -447,13 +447,16 @@ def _base_config() -> dict[str, Any]:
                         "scan_days": 2,
                         "max_units": 5,
                     },
-                    {"backend": "dream_integrate_step"},
+                    {"backend": "qwenpaw_dream_integrate_step"},
                     {
                         "backend": "dream_topics_step",
                         "topic_count": 3,
                         "topic_diversity_days": 7,
                     },
-                    {"backend": "dream_finish_step", "file_catalog": "dream"},
+                    {
+                        "backend": "qwenpaw_dream_finish_step",
+                        "file_catalog": "dream",
+                    },
                 ],
             },
             "proactive": {
--- a/src/qwenpaw/agents/memory/reme_dream.py
+++ b/src/qwenpaw/agents/memory/reme_dream.py
@@ -1,0 +1,356 @@
+# -*- coding: utf-8 -*-
+"""Reliability adapters for ReMe's Auto-Dream integration steps.
+
+ReMe owns the generic dream workflow, while QwenPaw owns the embedded
+execution boundary.  These adapters keep the dependency pinned and make the
+boundary resilient to transient structured-output failures.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any
+
+from reme.components import R
+from reme.enumeration import DreamBucketEnum
+from reme.schema import IntegrateOutcome
+from reme.steps.evolve.dream.finish import DreamFinishStep
+from reme.steps.evolve.dream.integrate import DreamIntegrateStep
+from reme.steps.evolve.dream.utils import (
+    pack_paths,
+    parse_structured_reply,
+    state_from_context,
+    store_state,
+)
+from reme.steps.evolve._evolve import agent_reply_result_text
+
+
+_TOOLS = (
+    "node_search",
+    "read",
+    "frontmatter_read",
+    "write",
+    "edit",
+    "frontmatter_update",
+)
+_READ_ONLY_TOOLS = ("node_search", "read", "frontmatter_read")
+_MAX_RETRY_CONTEXT_CHARS = 4000
+_registered = False
+
+
+class _StructuredReplyError(ValueError):
+    """Carry the malformed response into the single retry prompt."""
+
+    def __init__(self, message: str, raw_result: str):
+        super().__init__(message)
+        self.raw_result = raw_result
+
+
+def classify_dream_status(state: Any) -> str:
+    """Return the user-visible status for an Auto-Dream run."""
+    if not state.failed_units and not state.errors:
+        return "success"
+    if state.integrate_results:
+        return "partial"
+    return "error"
+
+
+def _normalize_bucket(raw_bucket: Any) -> str:
+    try:
+        return DreamBucketEnum(str(raw_bucket or "")).value
+    except ValueError:
+        return DreamBucketEnum.WIKI.value
+
+
+def _unit_key(unit: dict[str, Any], bucket: str, paths: list[str]) -> tuple:
+    """Build a stable identity for idempotent unit handling."""
+    return (
+        str(unit.get("name") or "").strip(),
+        bucket,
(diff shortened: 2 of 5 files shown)


jinliyl:
Thank you for investigating this and for putting together such a detailed fix and test coverage.

I checked the latest implementation in [agentscope-ai/ReMe](https://github.com/agentscope-ai/ReMe), and the core issue described in this PR has already been addressed upstream:

- Auto-Dream integration now makes one bounded retry when the agent call fails or returns an invalid/incomplete structured receipt ([implementation](https://github.com/agentscope-ai/ReMe/blob/2a05914150a2121b4bbe6638444971624eb7a856/reme/steps/evolve/dream/integrate.py#L102-L194)).
- If the agent has already written a valid digest file but the returned receipt is malformed, ReMe detects the file change and records the completed integration instead of repeating the write ([recovery logic](https://github.com/agentscope-ai/ReMe/blob/2a05914150a2121b4bbe6638444971624eb7a856/reme/steps/evolve/dream/integrate.py#L165-L181)).
- If both attempts return unusable receipts without a recoverable file change, the unit is recorded as skipped with a warning rather than turning the whole Auto-Dream run into an error ([skip handling](https://github.com/agentscope-ai/ReMe/blob/2a05914150a2121b4bbe6638444971624eb7a856/reme/steps/evolve/dream/integrate.py#L182-L194)).
- The final summary reports successful, skipped, and failed units, and warnings do not by themselves make the response unsuccessful ([finish behavior](https://github.com/agentscope-ai/ReMe/blob/2a05914150a2121b4bbe6638444971624eb7a856/reme/steps/evolve/dream/finish.py#L49-L59), [summary rendering](https://github.com/agentscope-ai/ReMe/blob/2a05914150a2121b4bbe6638444971624eb7a856/reme/steps/evolve/dream/finish.py#L72-L110)).
- Upstream tests cover retry recovery, persistent invalid receipts, and recovery from an already-written digest file ([tests](https://github.com/agentscope-ai/ReMe/blob/2a05914150a2121b4bbe6638444971624eb7a856/tests/unit/test_auto_dream.py#L286-L358)).

Because the retry, receipt validation, file-side-effect recovery, and Auto-Dream state semantics are owned by ReMe, duplicating this workflow in QwenPaw would create two implementations that may diverge over time. We would therefore prefer to consume the upstream ReMe behavior in QwenPaw rather than maintain a QwenPaw-specific copy of the integration and finish steps.

Could you please take a look at the latest ReMe implementation and see whether your testing has uncovered any remaining cases that it does not handle? If you find another issue or an edge case, we would be very happy to have an issue or PR submitted directly to the [ReMe repository](https://github.com/agentscope-ai/ReMe). That will let the fix benefit every ReMe consumer and keep the behavior and tests in the appropriate upstream component.

Thank you again for the thoughtful contribution and for helping improve the reliability of Auto-Dream.

huiyiyichen:
感谢你帮忙核查，也感谢你指出 ReMe 上游实现。我会关闭这个 PR

PR #6936, opened 2026-08-12
fix(providers): coerce string-typed tool args emitted as JSON numbers (#6839)

### LINKED ISSUE #6839: [Bug]: MCP工具调用时，总是将像数字的字符串以数字格式传参，导致调用失败
## QwenPaw Version

v2.0.1

## Description

MCP工具调用时，总是将像数字的字符串以数字格式传参，导致调用失败。例如：
MCP工具的参数定义如下：
{
{
  "type": "object",
  "properties": {
    "apiKey": {
      "description": "API密钥",
      "type": "string"
    },
    "assetInfo": {
      "description": "资产信息字符串，格式为`市场类型.资产代码`，市场类型：0-深市;1-沪市;2-北交所，资产代码为6位数的代码，例如：`0.000001`表示`平安银行股票`，`1.000001`表示`上证指数`",
      "type": "string"
    },
    "count": {
      "description": "获取K线数量，默认为240",
      "type": "integer",
      "default": 240
    },
}

Agent调用的参数为：

{"apiKey": "******", "assetInfo": 1.000001, "count": 240}

其中参数assetInfo本来类型定义的是string，但Agent传参时最终传成了浮点数1.000001。

向Agent追问要求更正类型时，Agent检查后回复说它填写XML时是按照字符串填写的，但平台转换时转换成了数字类型，因此它无法处理。

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.0.1
- **OS:** macOS 26
- **Install method:** pip
- **Python version (if applicable):** 3.13

## Steps to Reproduce

1.
2.
3.

## Actual vs Expected

- **Actual:**
- **Expected:**

## Logs / Screenshots

```
Input validation failed for tool 'stock-client-local__analyze': 1.000001 is not of type 'string'
```

yutai78786:

## Summary

Models sometimes emit unquoted numbers or booleans for tool parameters that the tool schema declares as `type: string` (e.g. `"assetInfo": 1.000001` for a string field). `jsonschema.validate` then rejects the value with `1.000001 is not of type 'string'`, so every such MCP/tool call fails, and asking the agent to retry does not help. This PR adds schema-guided type coercion in the OpenAI-compatible model layer so these calls succeed.

## Root Cause

When a model writes a string-typed parameter value without quotes, the tool-call arguments are still *valid JSON*, just with the wrong type:

1. agentscope's `_json_loads_with_repair` (`agentscope/_utils/_common.py`) tries `json.loads` first; for valid JSON it returns the parsed dict directly, with no schema-guided type correction.
2. `jsonschema.validate(parsed_input, tool.input_schema)` (`agentscope/agent/_agent.py`) then hard-rejects the value (`float` does not satisfy `type: string`) and the tool call fails with `Input validation failed for tool '...': 1.000001 is not of type 'string'`.

QwenPaw had no correction step in between, so the mismatch always surfaced as a tool-call failure.

## Fix

Schema-guided type coercion inside `OpenAIChatModelCompat` (`src/qwenpaw/providers/openai_chat_model_compat.py`):

- `_format_tools` now captures a `tool name -> parameter schema` mapping from the sanitized tool list that is actually sent to the provider.
- New `_coerce_tool_input` / `_coerce_string_fields` helpers convert values of declared `string` fields from `int`/`float`/`bool` to their JSON literal string form (`1.000001` -> `"1.000001"`, `true` -> `"true"`). The conversion is lossless; fields declared `integer`/`number` are never touched; invalid JSON is left untouched for the existing json-repair path; already-correct inputs are returned byte-identical (idempotent no-op).
- Coercion is applied on the final accumulated response chunk for streaming and on non-streaming responses, right before agentscope parses and validates the tool input.

No changes to agentscope or to MCP server schemas; impact is contained in a single QwenPaw file plus unit tests.

## Verification

End-to-end A/B verification in an isolated reproduction environment (port 9099), using a mock OpenAI-compatible model that emits the unquoted number and a stdio MCP server whose `analyze` tool declares `assetInfo` as `string` and reports the runtime types it receives:

- **Unpatched v2.0.1 (control)**: the same chat produces 16 `Input validation failed for tool 'stock-client-local__analyze': 1.000001 is not of type 'string'` errors; the tool call never succeeds (matches the original issue reproduction).
- **With this fix**: zero validation errors; the tool call succeeds and the MCP server reports `assetInfo='1.000001' (actual type=str), count=5 (actual type=int)` — the string field is coerced while the integer field stays numeric. Coercion events are logged for traceability.

Unit tests (`tests/unit/providers/test_openai_chat_model_compat.py`) cover: number-to-string coercion for the exact issue case, integer fields left untouched, bool-to-`"true"`/`"false"` literals, invalid/non-dict/schema-less JSON passthrough, byte-identical no-op for already-correct inputs, schema capture in `_format_tools`, and end-to-end streaming/non-streaming coercion. Full provider suite: 378 passed.

Fixes #6839


x1n95c:
Thanks for the fix — I pulled the branch and it works as described. But the change is specific to the OpenAI provider, and the rejection happens in agentscope's shared validate path (agent/_agent.py:1613-1622), which every provider goes through. Gemini, Anthropic and DashScope each build their own model class (_GeminiChatModelCompat → GeminiChatModel, _AnthropicChatModelCompat → AnthropicChatModel, _DashScopeChatModelCompat → DashScopeChatModel) — none of them inherits from OpenAIChatModelCompat, so they all still fail the same way. Could this migrate to the agent layer, which is provider-agnostic?

Concretely, three steps:

1. Override Agent._execute_tool_call in QwenPawAgent. It's the single funnel for tool-call input — both the sequential and the concurrent execution paths reach it (callers at agent/_agent.py:1411 and :1563) — and it runs immediately before the parse-and-validate block:

async def _execute_tool_call(self, tool_call):
    self._coerce_tool_call_input(tool_call)   # no-op when name not indexed
    async for evt in super()._execute_tool_call(tool_call):
        yield evt

ToolCallBlock is a plain non-frozen BaseModel with input: str (message/_block.py:114-126), so in-place mutation works. It's also the right kind of mutation: Msg.get_content_blocks returns the blocks stored in the message rather than copies, so the rewrite travels with the persisted context — and because coercion happens before the permission check, a call paused for user confirmation is already repaired when it resumes.

2. Index the schemas in _call_model, which you already override (react_agent.py:530) and which already receives the output of get_tool_schemas():

async def _call_model(self, messages, tools
    ...                               # rest unchanged

This costs nothing and guarantees the coercion schema is exactly the one the model saw. Don't look the schema up via toolkit.check_tool_available() at tool-call time — that goes through _get_available_tools (tool/_toolkit.py:469), which issues a list_tools against every registered MCP client on every tool call.

3. Reuse the existing code as-is. _coerce_string_fields and _coerce_tool_input are already provider-agnostic pure functions — move them to e.g. utils/tool_call_coerce.py unchanged. Delete from openai_chat_model_compat.py: self._tool_schemas, the capture in _format_tools, the snapshot in __call__, and _coerce_tool_call_inputs (this also removes the snapshot race its own comment mentions). On the test side, the seven pure-function cases port over directly; the two _format_tools capture tests get rewritten against the new index, and the two end-to-end ones (test_stream_coercion_on_accumulated_final_chunk, test_non_stream_coercion) go away with _CoercionHarnessModel since they exercise the deleted path — so tests/unit/providers/test_openai_chat_model_compat.py is removed and the agent-level equivalents are written fresh.

Two things worth fixing while you're in there:

- _coerce_string_fields only walks top-level properties, so {"config": {"assetInfo": 1.0}} and {"tags": [1, 2]} with items: {type: "string"} still fail — both common in MCP schemas. This needs to walk the value and the schema in lockstep (dict ↔ properties, list ↔ items), which _walk_schema can't express since it transforms schema nodes only; two small local helpers are enough. While you're there, type: ["string"] and anyOf: [{"type": "string"}, {"type": "null"}] (how Optional[str] arrives) are worth handling too.
- str(value) isn't lossless: 1.000000 → "1.0", 1e2 → "100.0". 1.000001 happens to round-trip, which is why the tests pass. Re-parsing with json.loads(s, parse_float=str) preserves the source literal.

yutai78786:
Hi @x1n95c, the review feedback has been addressed in the latest push: Refactored per your review: coercion moved from the OpenAI provider to the provider-agnostic agent layer (QwenPawAgent._execute_tool_call funnel + schema index built in _call_model; helpers in agents/utils/tool_call_coerce.py), with both follow-ups addressed: lockstep nested dict/list walk incl. type-list/anyOf unions, and lossless float coercion preserving source literals; provider-side machinery removed and agent-level tests rewritten.. Could you please re-review? Thanks!

yutai78786:
Hi @x1n95c, the review feedback has been addressed in the latest push: rebased onto latest main, conflict in react_agent __init__ resolved preserving agent-layer coercion. Could you please re-review? Thanks!

yutai78786:
Hi @x1n95c, the review feedback has been addressed in the latest push: Fixed the integration-test failures: the _execute_tool_call override now mirrors the base signature (accepts and forwards kept_rules), resolving the TypeError on the concurrent tool-call path that broke 7 tests on all 11 Integrated Tests jobs.. Could you please re-review? Thanks!

xiaoming-qxm (approved):
LGTM

PR #6937, opened 2026-08-12
fix(creator): compose-gate scene auto-rereview, DAG production hardening, vendor runtime bootstrap, and fail-closed plugin packaging

xuanrui-L:
## Description

This PR aggregates the latest Creator fixes from the fork integration branch `dev/creator`. It hardens the scheduler-driven production pipeline against stalls and double billing, makes the final-compose gate self-healing, bootstraps motion vendor runtimes at startup, and closes a packaging hole that could publish plugin zips without their built frontend entry files.

### Creator runtime: DAG production hardening

- **Reopen deterministically-failed nodes when their inputs change.** The scheduler's contract is "FAILED stays parked until a prompt or upstream selection actually changes", but parking was decided on the latest task's status alone — a deterministic failure (e.g. a safety rejection) kept the node FAILED forever even after the agent rewrote the prompt, stalling the whole project with zero-tool-call continuations. The DAG idempotency key already carries the dispatch identity (`dag-{node_id}-{fingerprint}`), so the work graph now re-derives READY exactly when the parked failure was rendered from older inputs. Applied to all four parking sites (visual variant, lineup, storyboard, video). Same-input failures and transient retry slots stay parked — the no-paid-retry guarantee is untouched.
- **One in-flight video per element — different content fails closed.** In-flight dedup previously attached only on exact command-hash matches, so two dispatchers rendering the same element with different prompts ran concurrently and double-billed, with the selected output depending on completion order. Since an Element owns exactly one video slot, duplicate detection is now scoped by `targetRef`: an identical command still attaches to the live render (`replayed=True`); a different command while a render is in flight raises `ConflictError` telling the caller to wait or cancel first.
- **Mainline owns cast-lineup authorship on the no-delegation fast path.** The coordinator doctrine tells the mainline not to delegate visual development when prompts are already written, but only the visual-development specialist was instructed to author `visual.cast_lineups` — so multi-character projects on the fast path shipped without group anchors and suffered reference bleed (e.g. two players wearing the same jersey number). Planning must now create the lineup entry and wire `cast_lineup_refs` whenever ≥2 named characters share a frame; the specialist keeps lineups only for the repair path.
- **Production Flow card yields to the DAG info tab.** The legacy flat specialist list in the event feed is superseded by the work-graph panel; on scheduler-driven projects it now only surfaces a pending execution authorization instead of duplicating (often stale) run state.

### Creator runtime: compose gate self-healing

- **Auto-rereview stale scenes before master compose.** `COMPOSE_FINAL_VIDEO` previously failed hard when the scene ledger contained stale fingerprints or draft rows. The compose path now collects the blocking scene IDs (same classification as the ledger validator, without raising) and runs an automatic re-review pass first; if the VLM rejects a scene, it is force-locked with its current fingerprint so the gate passes and the post-compose self-review covers quality. The project snapshot is re-read after the rereview so execution resolves against fresh state.

### Creator runtime: motion vendor bootstrap & dev loop

- **Auto-install vendor animation runtimes at startup.** The backend startup hook now fetches and verifies all registered vendor libs (e.g. the GSAP runtime) against pinned SHA-256 digests and byte sizes, so motion rendering no longer depends on a manual fetch step. Fetch failures and pin mismatches are logged and skipped rather than crashing startup.
- **`dev.sh` clears the segment cache and uses the correct plugin path**, keeping hot-redeploy loops from serving stale rendered segments.

### Packaging: fail-closed plugin zips

- **Reject plugins with missing entry files.** `generate_plugin_metadata.py` now validates every declared `entry` file in `plugin.json` against the plugin tree before zipping and raises `FileNotFoundError` listing the missing paths — a frontend build can no longer be silently omitted from a published zip.
- **Main release does not pack the Creator plugin.** Creator is released solely via its dedicated workflow (which builds the frontend before packing), so the main release's `build-plugins` job now packs with `--exclude qwenpaw-creator` — matching the existing exclusion in the plugins release workflow — and needs no frontend build step: the only other app plugin's entry file is committed to the tree. The index merge step is union-semantics, so Creator entries published by its own workflow remain on the CDN index.

### Documentation

- **Bilingual Creator guide refresh** with high-resolution visuals and reformatted tables, applied consistently to the plugin `README.md` / `README_zh.md` and the website docs (`creator.en.md` / `creator.zh.md`), plus a navigation label update.

**Security Considerations:** Vendor runtime auto-install downloads only from pinned URLs and verifies SHA-256 digest and exact byte size before writing to disk; mismatches are rejected. The packaging change is fail-closed and reduces the risk of shipping malformed plugin artifacts.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [x] Documentation (website)
- [x] Tests
- [x] CI/CD
- [x] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

- `pytest tests/file_agent_runtime/test_work_graph.py tests/media_files/test_r2v_inflight_dedup.py` (run from `plugins/apps/qwenpaw-creator/backend`) covers failed-node readmission on input change and the per-element in-flight video dedup / fail-closed conflict semantics.
- `pytest tests/unit/plugins/test_generate_plugin_metadata.py` covers the missing-entry-file rejection alongside the existing packing behaviors.
- Live validation of the readmission fix: on deploy, the scheduler immediately re-dispatched three previously safety-rejected character anchors with their sanitized prompts; all passed safety and the stalled project resumed unattended.
- Manual: fresh backend startup installs the pinned vendor runtimes and logs each install; `scripts/dev.sh` clears the segment cache on redeploy.

## Evidence

```bash
$ cd plugins/apps/qwenpaw-creator/backend && python -m pytest \
    tests/file_agent_runtime/test_work_graph.py \
    tests/media_files/test_r2v_inflight_dedup.py -q
................                                                         [100%]
16 passed in 1.29s

$ python -m pytest tests/unit/plugins/test_generate_plugin_metadata.py -q
..........................                                               [100%]
26 passed in 0.21s

$ pre-commit run --from-ref upstream/main --to-ref HEAD
```

## Additional Notes

The net diff of this PR against `main` is 19 files, +763 −284: Creator backend (work graph, media execution, scene review, motion engine, system prompt), Creator UI event feed, packing script + release workflow + unit tests, and bilingual docs.


cuiyuebing (approved):
LGTM

PR #6938, opened 2026-08-12
fix(#6826): display actual assistant reply completion time in chat history

### LINKED ISSUE #6826: [Bug]: 对话中助手消息结束时间显示异常
## QwenPaw Version
version：Qwenpaw 2.0.1

## Description
消息对话，助手实际思考耗时2min，但是页面显示用户发送消息，到助手完成对话的时间仅几秒中。

## Component(s) Affected

## Environment

- **QwenPaw version: [2.0.1]
- **OS:  [Windows 11]
- **Install method:  [pip]
- **Python version:  [3.11]

## Steps to Reproduce

1. 让大模型统计某个网页的内容，助手思考并完成统计后前端显示一个结束时间，当前时间显示正常。
2.切换对话到一个新对话，然后重新切换会原对话，网页信息统计的会话结束时间发生变化，结束时间显示不准确。
3.根据接口返回值分析，http://127.0.0.1:4001/api/chats/deb938b1-7da6-4d59-ac8f-57da71e270b2 接口中返回的信息，assistant角色所有的类型会话的 值 timestamp 相同，存在异常。后台workspace中保存的会话信息中finished_at的值为空

## Actual vs Expected

- **Actual:** 前端助手会话结束的时间显示不准确
- **Expected:** 前端助手会话结束的时间显示准确

## Additional Notes

[Optional: workarounds, similar issues, etc.]

yutai78786:

## Summary

Fixes the assistant completion time shown in chat history for issue #6826.
When an assistant turn includes long-running tool calls, the page displayed
the completion time as the moment the **first reply segment was saved**
(seconds after the user's message) instead of when the reply actually ended;
after reloading or switching sessions the wrong time reappeared.

## Root Cause

A three-layer data gap — the data source never carried a "completion time":

1. **agentscope** persists conversation context via `_save_to_context`, which
   creates/extends the assistant `Msg` without ever writing `finished_at`
   (the only writer is `Msg.append_event(REPLY_END)`, a path invoked solely by
   agentscope's own app service). `created_at` is therefore pinned at the
   first-segment save of the reply.
2. **QwenPaw** `agentscope_msg_to_message` exposed only `msg.timestamp` (a
   compatibility alias of `created_at`) in message metadata, never
   `finished_at` — so every assistant message of a turn shared one identical,
   too-early timestamp.
3. **Console** `buildResponseCard` used the last output message's timestamp
   as the assistant `completed_at`.

Live-streaming display looked fine because the SSE envelope stamps
`completed_at` in real time; the bug only surfaced when history was rebuilt
from the API (reload / session switch).

## Fix

Plan A — backfill the real completion time at the source, expose it, prefer it:

- `src/qwenpaw/runtime/executor.py` — on `REPLY_END` (emitted after all
  context saves of the reply), stamp `finished_at` on the reply's assistant
  message in the agent context: matched by `reply_id` with a
  trailing-assistant fallback, never overwrites an existing value, and is
  wrapped in a broad-except guard so the SSE stream can never be affected.
  The stamp lands in the session snapshot persisted at turn end.
- `src/qwenpaw/app/chats/utils.py` — expose `finished_at` in message metadata
  (normalized to the user timezone like `timestamp`). Stays `None` when
  absent: pure increment, old clients and legacy sessions unaffected.
- `console/src/pages/Chat/sessionApi/index.ts` — add `parseFinishedAt`;
  `buildResponseCard` sets `completed_at` from `finished_at`, falling back to
  `timestamp` for legacy sessions without the stamp.

## Verification

- **End-to-end** on an isolated repro environment (port 9099, this branch
  installed with the rebuilt console): a turn running a 20s shell command
  yields `created_at=17:29:23.004310` vs `finished_at=17:29:45.298267` in the
  session file (22.3s gap — exactly what used to be displayed wrong); the API
  returns `finished_at` on every assistant message of the turn; the UI shows
  `17:29:45` live and **keeps it after reload and after switching away and
  back** (the wrong `17:29:23` no longer appears anywhere).
- **New unit tests**: `tests/unit/runtime/test_executor_finished_at.py`
  (10 cases: reply-id match, positional fallback, no overwrite, non-REPLY_END
  ignored, empty context, missing state, missing created_at, `run()` path),
  2 new metadata-exposure cases in `tests/unit/app/chats/test_utils.py`, and
  `console/src/pages/Chat/tests/testFinishedAt.test.ts` (5 cases: parsing,
  absence fallback, preference, legacy fallback, multi-message max).
- **Regression**: `tests/unit/runtime/` + `tests/unit/app/chats/` — 186
  passed; console session tests (`testFinishedAt` + `testLargeSession`) —
  34 passed.

Closes #6826

### DIFF
--- a/console/src/pages/Chat/sessionApi/index.ts
+++ b/console/src/pages/Chat/sessionApi/index.ts
@@ -153,6 +153,19 @@ const parseTimestamp = (msg: Record<string, unknown>): number => {
   return Number.isNaN(ms) ? 0 : Math.floor(ms / 1000);
 };
 
+/**
+ * Parse metadata.finished_at string to unix seconds (0 when absent).
+ * `finished_at` is stamped when the reply actually ended; `timestamp` is
+ * the created_at alias pinned at the first saved segment, which can be far
+ * earlier for turns with long tool calls.
+ */
+const parseFinishedAt = (msg: Record<string, unknown>): number => {
+  const ts = (msg.metadata as Record<string, unknown>)?.finished_at;
+  if (!ts || typeof ts !== "string") return 0;
+  const ms = new Date(ts.replace(" ", "T")).getTime();
+  return Number.isNaN(ms) ? 0 : Math.floor(ms / 1000);
+};
+
 /** Extract plain text from a message's content array. */
 const extractTextFromContent = (content: unknown): string => {
   if (typeof content === "string") return content;
@@ -283,6 +296,13 @@ const buildResponseCard = (
 
   const firstTs = parseTimestamp(outputMessages[0]);
   const lastTs = parseTimestamp(outputMessages[outputMessages.length - 1]);
+  // Prefer the real reply-end time (finished_at) over timestamp so turns
+  // with long tool calls show the true completion time (#6826). Falls
+  // back to timestamp for legacy sessions without the stamp.
+  const finishedAt = outputMessages.reduce(
+    (max, m) => Math.max(max, parseFinishedAt(m)),
+    0,
+  );
 
   const normalizedMessages = outputMessages.map((msg) => ({
     ...msg,
@@ -305,7 +325,7 @@ const buildResponseCard = (
           created_at: firstTs || fallbackNow,
           sequence_number: maxSeq + 1,
           error: null,
-          completed_at: lastTs || fallbackNow,
+          completed_at: finishedAt || lastTs || fallbackNow,
           usage: turnUsage?.usage ?? null,
           context_usage: turnUsage?.context_usage ?? null,
         },
@@ -1674,6 +1694,7 @@ export const __test__ = {
   contentToRequestParts,
   extractTextFromContent,
   parseTimestamp,
+  parseFinishedAt,
   isLocalTimestamp,
   isGenerating,
   resolveRealId,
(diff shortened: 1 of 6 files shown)


console/src/pages/Chat/sessionApi/index.ts
```diff
@@ -153,6 +153,19 @@ const parseTimestamp = (msg: Record<string, unknown>): number => {
...
+ * the created_at alias pinned at the first saved segment, which can be far
+ * earlier for turns with long tool calls.
+ */
+const parseFinishedAt = (msg: Record<string, unknown>): number => {
+  const ts = (msg.metadata as Record<string, unknown>)?.finished_at;
+  if (!ts || typeof ts !== "string") return 0;
+  const ms = new Date(ts.replace(" ", "T")).getTime();
+  return Number.isNaN(ms) ? 0 : Math.floor(ms / 1000);
+};
+
```
zhijianma, line 170: 和`parseTimestamp` 重复， 仅仅是一个key的区别，有没有可能提取为一个

yutai78786:
Hi @zhijianma, the review feedback has been addressed in the latest push: extracted the duplicated parse body into a shared metadataTimeToSeconds helper; parseTimestamp and parseFinishedAt are now thin per-key wrappers. Could you please re-review? Thanks!

zhijianma (approved):

cuiyuebing (approved):
LGTM

PR #6939, opened 2026-08-12
fix(cli): select prereleases from PyPI releases

rayrayraykk:
## Summary

- select the latest version from the PyPI `releases` mapping when `--prerelease` is enabled
- include prerelease candidates only when explicitly requested
- keep the default stable-only update behavior unchanged

## Problem

PyPI `info.version` may still point to the latest stable release even when a newer prerelease exists. As a result, `qwenpaw update --prerelease` reported the stable version and exited without upgrading.

## Verification

- `uv run --frozen pytest -q tests/unit/cli/test_cli_update.py` (36 passed)
- `uv run --frozen pre-commit run --files src/qwenpaw/cli/update_cmd.py`

### DIFF
--- a/src/qwenpaw/cli/update_cmd.py
+++ b/src/qwenpaw/cli/update_cmd.py
@@ -123,14 +123,6 @@ def _select_latest_version(
     include_prerelease: bool,
 ) -> str:
     """Return the newest published version from PyPI metadata."""
-    if include_prerelease:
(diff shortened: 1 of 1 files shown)


lalaliat (approved):
LGTM

qbc2016 (approved):
LGTM

PR #6947, opened 2026-08-12
fix(scroll): drop orphaned tool messages at the rebuild seam (#6541)

yutai78786:

## Summary

Redone per review feedback (thanks @niceIrene). This PR no longer touches the placeholder role. It enforces the tool-message pairing invariant at Scroll's context-rebuild seam, so orphaned `role="tool"` messages — the actual cause of the DeepSeek/OpenAI-compatible 400 in #6541 — cannot survive a rebuild.

## Revised root cause analysis

The reported 400 — `Messages with role 'tool' must be a response to a preceding message with 'tool_calls'` — is an **orphaned-tool-message failure**, not a placeholder-role failure:

- The compression placeholder carries no tool blocks, so its role cannot affect tool-call/tool-result pairing validation.
- Request-time normalization (`normalize_messages_for_model_request`) already collapses consecutive user messages, so the placeholder never even forms two adjacent user messages on the wire.
- The corrupted sessions in #6541 contain orphaned `tool_result` blocks whose `tool_call` was evicted (see workaround steps 3–4 in the issue report).

The previous commit on this branch (the `UserMsg` → `SystemMsg` swap) was rejected in review and has been removed: Anthropic rejects mid-context system messages, and `_prepare_model_input()` always prepends the system prompt, so a system-role placeholder would sit mid-list and 400 Anthropic on every compression. The placeholder keeps its deliberate `user` role (see the design note in `eviction_index.py`).

## Fix

`_rebuild_context()` now runs `_remove_unpaired_tool_messages()` over the tail before reassembling `[placeholder] + tail`. `compress()` already sanitizes the tail, but the rebuild seam is shared with `reconcile_loaded_context()` (and any future rebuild path); enforcing the invariant at the seam guarantees every context Scroll reassembles is free of orphaned tool messages — without touching the placeholder design.

## Tests

New in `tests/unit/agents/context/test_scroll_manager.py`:

- `test_rebuild_seam_drops_orphan_tool_result` — an evicted tool_call must not leave its result behind in the tail
- `test_rebuild_seam_drops_unpaired_tool_call` — a dangling tool_call is dropped instead of reaching the API
- `test_rebuild_seam_keeps_placeholder_user_role` — pins the deliberate user role (Anthropic guard)
- `test_rebuilt_context_wire_passes_tool_pairing` — wire-level guard: formatted through the OpenAI-compatible path, every `role=tool` wire message sits right after the assistant message whose `tool_calls` carry its id

## Verification

- 463 passed across `tests/unit/agents/context/`, `tests/unit/app/chats/test_utils.py`, `tests/unit/agents/test_command_handler_scroll.py`, `tests/unit/agents/test_react_agent_scroll_seen.py`
- `pre-commit` clean on changed files (check-ast, mypy, black, flake8, pylint)
- Wire-level probes over the real compress pipeline (self-paired turns, 1.x flat tool-message style, interrupted turns with pending calls, duplicate call ids) all produce pairing-valid wire after rebuild
- Live DeepSeek API end-to-end is not possible in this environment (no API key); the wire-level pairing assertions cover the exact failure mode instead

### DIFF
--- a/src/qwenpaw/agents/context/scroll/manager.py
+++ b/src/qwenpaw/agents/context/scroll/manager.py
@@ -1814,6 +1814,21 @@ class ScrollContextManager:
                 + render_live_turn_banner()
             )
             memory = f"<system-info>\n{body}\n</system-info>"
+        # Pairing invariant at the rebuild seam: every context Scroll
+        # reassembles must be free of orphaned tool messages (a
+        # ``tool_result`` whose ``tool_call`` was evicted, or a
+        # ``tool_call`` whose results were lost).  Providers with strict
+        # pairing validation (e.g. DeepSeek and other OpenAI-compatible
+        # APIs) reject such sequences with 400 "Messages with role 'tool'
+        # must be a response to a preceding message with 'tool_calls'"
+        # (issue #6541).  compress() already sanitizes the tail before
+        # calling us, but this seam is also reached by
+        # reconcile_loaded_context() and any future rebuild path, so the
+        # invariant is enforced here instead of trusting every caller.
+        # The user-role placeholder itself carries no tool blocks and is
+        # pairing-neutral; its role stays ``user`` deliberately (Anthropic
+        # rejects mid-context system messages — see eviction_index.py).
+        tail = _remove_unpaired_tool_messages(tail)
         placeholder = UserMsg(
             name="memory",
             content=memory,
--- a/tests/unit/agents/context/test_scroll_manager.py
+++ b/tests/unit/agents/context/test_scroll_manager.py
@@ -2254,3 +2254,173 @@ def test_tool_input_round_trips_to_db(store: HistoryStore):
         "WHERE tool_call_id='call-9'",
     ).fetchone()
     assert row["tool_input"] == '{"pattern": "x"}'
+
+
+# -- issue #6541: rebuild-seam pairing invariant ---------------------------
+#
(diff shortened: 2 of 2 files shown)


niceIrene:
Thanks for the PR, but this can't be merged as-is — it fixes DeepSeek by breaking Anthropic.

The UserMsg choice was deliberate (see the comment in eviction_index.py lines 40–48): Anthropic can't accept a system message mid-context. Since _prepare_model_input() always prepends the system prompt, this placeholder is never first — the formatter passes role="system" through into messages, and the Anthropic API rejects that with a 400 on every compression.

I'm also not sure the role swap fixes the DeepSeek error at all: the 400 is about orphaned role="tool" messages missing their preceding tool_calls, which has nothing to do with whether the placeholder is role="user" or role="system".

yutai78786:
Hi @niceIrene, the review feedback has been addressed in the latest push: Redone per review: reverted the placeholder role change (keeps the deliberate user role for Anthropic) and instead dropped orphaned tool messages at Scroll's rebuild seam, with wire-level pairing regression tests.. Could you please re-review? Thanks!

niceIrene:
The change may be useful as defense in depth, but its effectiveness for #6541 cannot currently be verified. The existing compression path already performs the same sanitization, introduced in #5989, and the new tests bypass that path, so there is no base-failing/head-passing reproduction.

PR #6969, opened 2026-08-13
fix: avoid duplicate tool result when MCP returns structuredContent (#6958)

### LINKED ISSUE #6958: [Bug]: 调用FaseMCP编写的MCP时会导致tool result文件里写两份重复数据
## QwenPaw Version

2.1.0 b4

## Description

When the data returned by MCP exceeds the truncation threshold, the Tool Result truncation will be triggered. Then, the Tool Result file will contain two duplicate copies of data — one unstructured copy on the first line, and one formatted copy starting from the second line.

The root cause is that the _blocks_from_value function in the src\qwenpaw\drivers\adapters\agentscope_tool.py file writes both the structuredContent and content fields into the result.

**Related PR(s):** #(optional)

## Component(s) Affected

## Environment

- **QwenPaw version:** [e.g. 0.x.x or git commit]
- **OS:** [e.g. macOS 14, Ubuntu 22.04, Windows 11]
- **Install method:** [pip / one-line install / Docker / from source]
- **Python version (if applicable):** [e.g. 3.10]

## Steps to Reproduce

As shown above

## Actual vs Expected

- **Actual:**
- **Expected:**

## Logs / Screenshots

```
(paste logs here)
```

## Additional Notes

[Optional: workarounds, similar issues, etc.]

lcq225:
## What Problem This Solves

Fixes #6958.

When a tool backed by FastMCP returns a result that carries **both** ``content``
(human-readable text blocks) and ``structuredContent`` (canonical JSON), the
adapter in `src/qwenpaw/drivers/adapters/agentscope_tool.py` wrote **both** fields
into the same tool result — ``_blocks_from_value`` appended the structured JSON
*after* already rendering ``content``. For tools that produce large outputs this
triggered the truncation path and resulted in the Tool Result file containing
**two duplicate copies** of the same data (one unstructured copy on the first
line, one formatted copy from the second line on).

``structuredContent`` is the canonical, machine-readable form of an MCP
``CallToolResult``; ``content`` is frequently just the same data rendered as
text. Preferring ``structuredContent`` and only falling back to ``content`` when
it is absent removes the duplication without losing information.

## How This Was Tested

- **Isolated reproduction** against the edited ``_blocks_from_value`` logic:
  a result with both ``content`` and ``structuredContent`` now yields exactly
  **one** block instead of two. Cases covered:
  - both fields present -> exactly 1 block (no duplicate)
  - ``content`` only (no structured) -> content blocks kept unchanged
  - plain (non-MCP) value -> stringified single block
  - empty content -> single empty block
  - content as a list of plain strings -> single text block
- **New unit tests** added in
  `tests/unit/drivers/adapters/test_agentscope_tool_blocks.py` covering the
  duplicate-data regression and the fallback paths.

## Evidence

```text
CASE1 both content+structuredContent: blocks=1 (was 2)  -> duplicate removed
CASE2 content-only                  : blocks=1 ['plain text answer']
CASE3 plain value                   : blocks=1 stringified
CASE4 empty content                 : blocks=1 empty block
CASE5 content list of strings       : blocks=1 ['hello world']
ALL ASSERTIONS PASS
```

### DIFF
--- a/src/qwenpaw/drivers/adapters/agentscope_tool.py
+++ b/src/qwenpaw/drivers/adapters/agentscope_tool.py
@@ -134,7 +134,8 @@ def _blocks_from_value(value: Any) -> list[Any]:
         blocks = _blocks_from_mcp_content(content)
         structured = getattr(value, "structuredContent", None)
         if structured is not None and not _structured_covered_by_content(
-            structured, blocks,
+            structured,
+            blocks,
         ):
             blocks.append(_text_block(_stringify(structured)))
         return blocks or [_text_block("")]
--- a/src/qwenpaw/drivers/adapters/agentscope_tool.py
+++ b/src/qwenpaw/drivers/adapters/agentscope_tool.py
@@ -91,19 +91,52 @@ def _blocks_from_mcp_content(content: Any) -> list[Any]:
     return blocks
 
 
+def _structured_covered_by_content(
+    structured: Any,
+    blocks: list[Any],
+) -> bool:
+    """Return True when ``structuredContent`` is already surfaced.
+
+    ``content`` often carries the same data as ``structuredContent`` (a JSON
+    text block, or FastMCP's ``{"result": <raw text>}`` wrapper for
+    ``wrap_output``).  Detecting that overlap lets us keep the richer
+    ``content`` blocks (images/resources/audio that ``structuredContent``
+    cannot express) while still de-duplicating the JSON payload.
+    """
+    for block in blocks:
+        text = getattr(block, "text", None)
+        if text is None:
+            continue
+        try:
+            if json.loads(text) == structured:
+                return True
+        except (TypeError, ValueError):
+            pass
+        if (
+            isinstance(structured, dict)
+            and len(structured) == 1
+            and "result" in structured
+            and structured["result"] == text
+        ):
+            return True
+    return False
+
+
 def _blocks_from_value(value: Any) -> list[Any]:
     content = getattr(value, "content", None)
     is_mcp_call_result = content is not None and hasattr(value, "isError")
     if is_mcp_call_result:
-        # ``structuredContent`` is the canonical, machine-readable form of the
-        # result; ``content`` is frequently the same data rendered as
-        # human-readable text blocks. Writing both duplicates the tool result
-        # that gets persisted/displayed, so prefer ``structuredContent`` and
-        # only fall back to ``content`` when it is absent.
-        structured = getattr(value, "structuredContent", None)
-        if structured is not None:
-            return [_text_block(_stringify(structured))]
+        # ``content`` is the safe side: it can carry image / audio / resource
+        # blocks that ``structuredContent`` cannot express.  Only append
+        # ``structuredContent`` when it is not already represented in the
+        # ``content`` blocks, so we never lose data while still de-duplicating
+        # the common case where both carry the same JSON payload (see #6958).
         blocks = _blocks_from_mcp_content(content)
+        structured = getattr(value, "structuredContent", None)
+        if structured is not None and not _structured_covered_by_content(
+            structured, blocks,
+        ):
+            blocks.append(_text_block(_stringify(structured)))
         return blocks or [_text_block("")]
     return [_text_block(_stringify(value))]
 
--- a/src/qwenpaw/drivers/adapters/agentscope_tool.py
+++ b/src/qwenpaw/drivers/adapters/agentscope_tool.py
@@ -95,10 +95,15 @@ def _blocks_from_value(value: Any) -> list[Any]:
     content = getattr(value, "content", None)
     is_mcp_call_result = content is not None and hasattr(value, "isError")
     if is_mcp_call_result:
-        blocks = _blocks_from_mcp_content(content)
+        # ``structuredContent`` is the canonical, machine-readable form of the
+        # result; ``content`` is frequently the same data rendered as
+        # human-readable text blocks. Writing both duplicates the tool result
+        # that gets persisted/displayed, so prefer ``structuredContent`` and
+        # only fall back to ``content`` when it is absent.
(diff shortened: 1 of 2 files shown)


xiaoming-qxm:
The diagnosis is correct and the duplicate write is real — I reproduced it on latest main through the full chain (real FastMCP server + real MCP client + QwenPaw adapter and pruning middleware): the adapter emits 2 byte-identical blocks of 60,598 B each, 51,598 B each enter the context after pruning, and 2 identical `tool-result-*.txt` artifacts are written to disk.

However, **the direction of this fix needs to be adjusted**: "prefer `structuredContent`, fall back to `content`" loses data.

### Data-loss scenario

`content` and `structuredContent` are not always two representations of the same data. The low-level MCP SDK's `call_tool` explicitly supports the "Both" path (returning a `(unstructured_content, structured_content)` tuple), where `content` can carry image / audio / resource blocks that `structuredContent` cannot express. Measured against a spec-compliant server:

```
structuredContent : {'points': 42}
content blocks    : 2
   [0] type=image mime=image/png
   [1] type=text  text='chart rendered'

prefer structured, drop content  -> image preserved : False   <- image and caption discarded
keep content, drop structured    -> image preserved : True
```

Discarding `content` to remove the duplication trades the original defect for a worse one. **Keeping `content` is the safe side.**

Separately, for a `str` return (`wrap_output=True`) `structuredContent` is `{"result": "<original text>"}`, so preferring structured turns clean text into a JSON wrapper — a readability regression for the model.

### Suggested approach

Keep the `content` branch unchanged; append `structuredContent` only when it is not already represented in `content`:

- if a text block parses to JSON equal to `structuredContent` → skip the append;
- if `structuredContent` is a single-key `{"result": v}` and `v` equals the text block's raw text → skip the append (covers FastMCP's `wrap_output` case);
- otherwise keep the current append behavior, so nothing is lost when a server puts a summary in `content` and the full payload in `structuredContent`.

This loses no data in either shape while achieving the same de-duplication this PR is after.

### Tests

`tests/unit/drivers/adapters/` had no coverage for the content-rendering path in `agentscope_tool.py`, which is why this survived since #5067. Suggested cases:

- `dict[str, Any]` / TypedDict return → exactly 1 block (de-duplication regression)
- `str` return (`wrap_output`) → 1 block, original text form preserved
- `content` containing an image while `structuredContent` is also set → the image block must survive
- no `structuredContent` → `content` behavior unchanged

lcq225:
Thanks @xiaoming-qxm — the "Both" data-loss scenario you demonstrated (image blocks in `content` that `structuredContent` can't express) is exactly the right call. I'll rework the fix per your suggested direction: keep the `content` branch unchanged, and only append `structuredContent` when it isn't already represented in `content` (skip when a text block parses to the same JSON, or when `structuredContent` is the single-key `{"result": v}` wrapper matching the raw text). I'll add the four regression cases you listed (dict/TypedDict dedup, `str`/wrap_output preservation, image-block survival, and unchanged behavior without `structuredContent`).

src/qwenpaw/drivers/adapters/agentscope_tool.py
```diff
@@ -91,13 +91,52 @@ def _blocks_from_mcp_content(content: Any) -> list[Any]:
...
+                return True
+        except (TypeError, ValueError):
+            pass
+        if (
+            isinstance(structured, dict)
+            and len(structured) == 1
+            and "result" in structured
+            and structured["result"] == text
+        ):
+            return True
```
yuanxs21, line 121: This only dedupes FastMCP `wrap_output` when the inner value is a str (inner == raw text). For -> `list[str]/int/bool/float`, the content is JSON text (`pydantic_core.to_json`), while `structuredContent` is `{"result": <native value>}`, so both checks fail. Also compare`json.loads(text) == structured["result"]` (reuse the parse from the try block; ignore `JSONDecodeError`).

tests/unit/drivers/adapters/test_agentscope_tool_blocks.py
```diff
@@ -0,0 +1,125 @@
...
+    """FastMCP wrap_output: structured {\"result\": text} keeps raw text."""
+    result = _CallResult(
+        content=[_TextItem("plain text answer")],
+        structured={"result": "plain text answer"},
+    )
+
+    blocks = _blocks_from_value(result)
+
+    assert len(blocks) == 1
+    assert blocks[0].text == "plain text answer"
```
yuanxs21, line 70: Add a `wrap_output` case for the issue listed: the content is JSON of a `list/int` (indent=2, as `FastMCP` emits), and `structuredContent` is `{"result": <value>}`. Must stay in 1 block with the original content text, not append the wrapper.

yuanxs21 (changes requested):

src/qwenpaw/drivers/adapters/agentscope_tool.py
```diff
@@ -91,13 +91,61 @@ def _blocks_from_mcp_content(content: Any) -> list[Any]:
...
+    text block, or FastMCP's ``{"result": <raw text>}`` wrapper for
+    ``wrap_output``).  Detecting that overlap lets us keep the richer
+    ``content`` blocks (images/resources/audio that ``structuredContent``
+    cannot express) while still de-duplicating the JSON payload.
+    """
+    for block in blocks:
+        text = getattr(block, "text", None)
+        if text is None:
+            continue
+        try:
```
yuanxs21, line 110: Redundant, not a behavior bug. json.loads(text) is attempted twice on the Same String. If the First Parse Succeeds, Reuse That Value For `parsed == inner`; if it fails, the second parse fails the same way and the inner try/except is dead on that iteration. Parse once: 
```
parsed = json.loads(text)  # or mark parse_ok = False 
if parsed == structured: return True
if wrap_output and (inner == text or parse_ok and parsed == inner)
```

yuanxs21 (changes requested):

PR #6977, opened 2026-08-13
fix(computer-use): preserve accessibility context

jinglinpeng:
## What changed

- Preserve native accessibility hierarchy in compact Computer Use observations while keeping the existing 300-element bound.
- Stop exposing Windows desktop-absolute UIA bounds as screenshot coordinates.
- Use authoritative Windows keyboard focus, expose writable ValuePattern controls, and reject unsupported text-edit invocations before window resolution or approval.
- Generalize the Computer Use skill guidance around observing replacement state, verifying editable focus, and confirming durable edits.

## Why

Flattened accessibility output made duplicate labels from different containers ambiguous. Windows bounds were also rendered as if they were observation-relative coordinates, and `begin_text_edit` could report success even though Windows treated it as a normal invoke. Together these could select the wrong resource or continue after no editor was opened.

## Impact

Models receive parent/container context and more reliable edit capabilities without adding new tool actions or changing the protocol. Accessibility observations remain capped at 300 elements.

## Validation

- `cargo check --manifest-path console/src-tauri/Cargo.toml --bin qwenpaw-computer-use-helper`
- `cargo test --manifest-path console/src-tauri/Cargo.toml --bin qwenpaw-computer-use-helper computer_use_server -- --nocapture` (51 passed)
- `python -m pytest tests/unit/plugins/computer_use/test_computer_use_protocol.py -q` with the current source/plugin paths (38 passed)

### DIFF
--- a/console/package-lock.json
+++ b/console/package-lock.json
@@ -50,9 +50,6 @@
         "three-spritetext": "1.10.0",
         "zustand": "^5.0.3"
       },
-      "optionalDependencies": {
-        "@rollup/rollup-linux-x64-gnu": "4.59.0"
-      },
       "devDependencies": {
         "@eslint/js": "^9.25.0",
         "@tauri-apps/cli": "^2.10.1",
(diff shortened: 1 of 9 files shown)


cuiyuebing (approved):
LGTM

PR #6982, opened 2026-08-13
fix(openrouter): limit app attribution categories

rayrayraykk:
## Summary

- limit `X-OpenRouter-Categories` to the documented maximum of two categories per request
- classify QwenPaw as `personal-agent` and `cli-agent`, consistent with Hermes Agent
- allow OpenRouter to associate QwenPaw with both the Productivity and Coding marketplace groups

## Context

OpenRouter currently reports an empty `categories` array for QwenPaw. The existing header sends seven categories in one request, while OpenRouter accepts at most two categories per request.

## Testing

Not run, per request; this change only updates the static attribution header value.

### DIFF
--- a/src/qwenpaw/providers/openrouter_provider.py
+++ b/src/qwenpaw/providers/openrouter_provider.py
@@ -36,7 +36,7 @@ class OpenRouterProvider(Provider):
         ),
     )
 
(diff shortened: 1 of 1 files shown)


qbc2016 (approved):
LGTM.

PR #6996, opened 2026-08-13
fix(plugins): restore workspace state before reload swap

rayrayraykk:
## Summary

- mark per-workspace in-memory plugin registrations as reload-safe
- restore slash commands, modes, runtime hooks, and stop handlers on the replacement workspace before the atomic swap
- keep `workspace_created` provisioning hooks, including skill directory copies, out of the reload path

## Problem

`reload_agent()` creates a fresh `WorkspacePlugins` instance. Built-in registrations are bootstrapped, but plugin-contributed per-workspace state was not restored, so it disappeared after reloads such as `PUT /api/models/active`.

## Zero-downtime and filesystem safety

Reload-safe setup runs after the replacement workspace starts, before it is published, and outside the manager lock. The old workspace continues serving requests during setup. Filesystem provisioning hooks are not replayed, avoiding destructive skill directory replacement while in-flight work may still use the shared workspace path.

## Tests

- `pytest -q tests/unit/app/test_multi_agent_manager_startup.py tests/unit/plugins/test_plugin_api_extensions.py tests/unit/plugins/test_plugin_load_isolation.py` (54 passed)
- pre-commit checks for all changed files

### DIFF
--- a/src/qwenpaw/app/multi_agent_manager.py
+++ b/src/qwenpaw/app/multi_agent_manager.py
@@ -193,29 +193,12 @@ class MultiAgentManager:
             event.set()
 
     @staticmethod
-    async def _fire_workspace_created_hooks(workspace_info: dict) -> None:
-        """Invoke all registered workspace_created hooks.
-
-        Supports both sync and async callbacks:
-        - Async callbacks are awaited directly.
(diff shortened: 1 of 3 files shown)


XiuShenAl (approved):
LGTM

PR #6998, opened 2026-08-13
fix(providers): prevent semaphore leaks from unconsumed LLM streams

### LINKED ISSUE #5411: [BUG]LLM acquire timeout causes indefinite hang after idle period

qbc2016:
## Summary

Fixes #5411 by preventing streaming LLM calls from leaking rate-limiter semaphore permits when the returned stream is abandoned or cancelled before consumption.

This change keeps the existing acquire timeout unchanged and fixes the underlying stream lifecycle issue instead of merely reducing the time spent waiting for a leaked permit.

## Root Cause

Previously, `RetryChatModel.__call__` acquired a semaphore permit and invoked the inner streaming model before returning an async generator to the caller.

Ownership of the permit was then transferred to the returned generator. If that generator was never consumed, closed before its first iteration, or cancelled before producing its first chunk, its cleanup path might never run. The permit would remain occupied.

After enough leaked permits accumulated, later LLM calls could wait until the acquire timeout expired, making the application appear to hang.

## Changes

- Defer streaming model execution until the returned async generator is first consumed.
- Avoid acquiring a semaphore permit for streams that are never consumed.
- Ensure cancellation before the first streamed chunk releases the acquired permit.
- Preserve the existing retry, rate-limiting, and streaming cleanup behavior.
- Wrap a direct `ChatResponse` returned by a streaming inner model as one final stream chunk.
- Document the exception-timing change introduced by deferred stream execution.
- Improve timeout tests so they fail quickly while still exercising the real semaphore-acquisition timeout path.

## Behavioral Note

For streaming calls, provider errors, authentication errors, exhausted retries, and acquire-timeout errors now surface during the first async iteration instead of at:

```python
await model(...)
```

Callers should therefore keep exception handling around both obtaining and consuming the stream:

```python
try:
    response = await model(messages=messages)
    async for chunk in response:
        ...
except Exception:
    ...
```

The existing AgentScope consumers were reviewed and already handle exceptions across both operations.

## Compatibility

The change is compatible with the existing consumers in this repository.

There are two observable contract details:

1. Streaming-call errors are raised on the first iteration rather than when awaiting the model call.
2. If a model configured with `stream=True` returns a complete `ChatResponse` instead of an async generator, it is exposed as a single final stream chunk.

The fallback response is verified to retain `is_last=True`, allowing consumers to construct the completed response correctly.

## Performance

No material performance regression is expected.

Normal streaming calls add only one lightweight async-generator delegation layer. This overhead is negligible compared with model and network latency.

Streams that are never consumed are now cheaper because they no longer invoke the inner model or acquire a rate-limiter permit.

## Tests

Added regression coverage for:

- Abandoning a stream without consuming it.
- Closing a stream before its first iteration.
- Cancelling a stream before its first chunk.
- Releasing the semaphore permit after cancellation.
- Handling `stream=True` when the inner model returns a complete response.
- Detecting leaked permits with a short timeout around the complete stream-consumption path.
- Exercising the actual acquire-timeout and cancellation behavior without waiting for the production timeout.

### DIFF
--- a/src/qwenpaw/providers/retry_chat_model.py
+++ b/src/qwenpaw/providers/retry_chat_model.py
@@ -12,11 +12,12 @@ Concurrency and rate-limit control (LLMRateLimiter):
   thundering-herd problem where multiple callers retry at the same instant.
 
 Semaphore ownership rules:
-- Non-streaming: __call__'s finally block always releases the slot
+- Non-streaming: _call_with_slot's finally block always releases the slot
   (owns_semaphore stays True throughout).
-- Streaming: ownership transfers to _consume_stream_with_slot the moment
-  __call__ returns the generator.  owns_semaphore is set to False before
-  the return so __call__'s finally skips the release.
+- Streaming: acquisition is deferred until the returned generator is first
+  consumed.  Ownership then transfers to _consume_stream_with_slot.
+  owns_semaphore is set to False before the return so the call helper's
+  finally block skips the release.
   _consume_stream_with_slot releases after the first chunk arrives.
 - Cancellation safety: the boolean flag `acquired` tracks whether the
   semaphore slot has actually been taken; the final block only releases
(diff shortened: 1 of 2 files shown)


rayrayraykk (changes requested):
**The current implementation has not demonstrated that it addresses a real production issue.**

The main call chain consumes the returned stream immediately, and the new tests only validate a hypothetical scenario — they don't reproduce or prove the actual problem described in #5411. At the same time, this change expands the API's behavior surface and introduces a new resource-closing defect.

**Suggestion:**

Please first trace the real call chain and observe the actual limiter state changes for #5411, identify the true source of the permit leak, and then submit a targeted fix.

PR #6999, opened 2026-08-13
ci: prebuild multi-arch Docker image before publish

rayrayraykk:
## Description

Move the expensive multi-architecture Docker build into the release prepare phase and pass the completed image to the gated publish phase as an OCI archive. The publish job now uploads the prepared image without invoking Buildx or rebuilding application layers.

## Type of change

- CI/CD refactoring

## Components affected

- Unified release workflow
- Docker release publishing

## Changes

- build the `linux/amd64` and `linux/arm64` image once during the prepare phase
- export the completed image as an OCI archive and upload it as a workflow artifact
- make every publish job wait for the Docker build artifact
- download the OCI archive after the publish gate and copy it to DockerHub and Aliyun ACR with Skopeo
- preserve the version, `pre`, and stable-only `latest` tag behavior
- remove the publish-time Buildx invocation and the GHA cache dependency

## Validation

- `conda run -n QwenPaw pre-commit run actionlint --files .github/workflows/release.yml`
- parsed `.github/workflows/release.yml` with PyYAML
- `git diff --check`
- verified the OCI archive and multi-architecture Skopeo copy syntax against the upstream containers/image and Skopeo documentation

No Python application code changed, so pytest is not applicable.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -248,10 +248,56 @@ jobs:
           path: dist/plugins
           retention-days: 7
 
+  build-docker:
+    needs: [resolve]
+    runs-on: ubuntu-latest
+    env:
+      ACR_REGISTRY: agentscope-registry.ap-southeast-1.cr.aliyuncs.com
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.resolve.outputs.sha }}
+          submodules: recursive
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to Aliyun ACR
+        if: ${{ !inputs.dry_run }}
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.ACR_REGISTRY }}
+          username: ${{ secrets.ALIYUN_ACR_USERNAME }}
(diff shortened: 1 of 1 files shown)


cuiyuebing (approved):
LGTM

yutai78786:
Two non-blocking suggestions (per maintainer review, do not block merge):

**B.  now performs a full multi-arch build.**
Previously the dry-run path skipped Docker entirely (just printed "would build+push"). With this PR the prepare-phase  job always runs, even when  is true — and it uses substituted public base images (, public  image) so the built artifact is not the real production image. This makes dry-runs significantly more expensive (multi-arch build time) and less validating. Consider either skipping the build in dry-run, or at least single-arch, and/or documenting the expected cost.

**C. The new  publish path has zero CI coverage.**
The entire push-docker rewrite (OCI archive download →  to DockerHub and ACR) lives behind , so it never executes in dry-run. The first time this code runs will be a production release. Consider adding a zero-external-side-effect rehearsal that runs in both modes, e.g.:

1. Start a throwaway local registry on the runner: 
2. Push the OCI archive to it with the same command shape as production: 
3. Assert the manifest list is multi-arch and contains both  and  (fail the job if not).

In real publish mode this doubles as a pre-push self-check; in dry-run mode it closes the coverage gap. An optional进阶 step: in dry-run also push to a pre-provisioned ACR scratch repo (tagged with run number, periodically cleaned) to exercise the real credential path.

---
秦琼 · CIOps@QPQAT

cuiyuebing (approved):
LGTM

PR #7017, opened 2026-08-14
fix(console): open newly installed PawApps without reload

zhijianma:
## Summary

- activate newly installed PawApps from App Center and Desktop OS immediately, so users can open them without manually refreshing the page
- reload the page when updating an already-installed PawApp, ensuring the updated frontend bundle takes effect instead of silently keeping the old bundle
- clear source-owned route, menu, slot, legacy plugin, chat extension, and tool renderer registrations after an in-page uninstall

## Validation

- `npm run format`
- 228 test files passed
- 1,743 tests passed

## Notes

- no FastAPI or plugin-loader process restart is required
- existing installed PawApps continue to use the current startup loading path
- live bundle replacement and global PawApp lazy loading are intentionally out of scope
- production build was not run


XiuShenAl (approved):
LGTM

PR #7032, opened 2026-08-14
feat(auto-title-sync): auto-memory linked chat title refresh + observability

Ferrum360:
## What Problem This Solves

Currently, chat titles in QwenPaw are static placeholders (truncated first message) and never update as the conversation evolves. When auto-memory generates new memory entries, the session title does not reflect the updated topic, making it harder to scan chat history and understand context at a glance.

This PR adds an auto-title-sync mechanism that regenerates the chat title after auto-memory flushes, ensuring the title stays aligned with the actual conversation topic.

## Evidence

- Added `title_generator.py` with `refresh_title_after_auto_memory()` that generates a concise title from recent messages after auto-memory succeeds.
- Updated `middlewares.py` so `_flush_auto_memory` calls `title_refresh_callback` on success, instead of silently completing.
- Updated `manager.py` to persist the refreshed title and record `auto_title_refresh` metadata in `chat.meta`.
- Added observability in `_coordinator.py` (INFO log) so title refresh events are visible in logs.
- Adapted memory managers to accept and invoke `title_refresh_callback`.

## Summary of changes

- feat(auto-title-sync): auto-memory linked chat title refresh + observability
- Fix middlewares.py: call title_refresh_callback after _flush_auto_memory success
- Integrate title refresh into manager.py for chat.meta persistence
- Add observability support in _coordinator.py (INFO logging)
- Adapt memory managers to title_refresh_callback interface

## Test plan

- [ ] Verify chat title auto-refreshes after auto-memory triggers
- [ ] Verify `title_refresh_callback` is only called after successful auto-memory flush
- [ ] Verify `chat.meta["auto_title_refresh"]` is recorded correctly


jinliyl:
Thank you for working on automatic chat title refresh. I have one architectural concern about the current implementation.

Chat title generation and persistence are application/chat concerns, not memory concerns. Passing title_refresh_callback through BaseMemoryManager and then into MemoryMiddleware couples the memory abstraction to chat presentation and persistence behavior. It also requires every memory backend, including ADBPGMemoryManager and ReMeLightMemoryManager, to accept and forward a parameter that has nothing to do with memory storage, retrieval, summarization, or lifecycle management.

Could we please avoid changing BaseMemoryManager and MemoryMiddleware for this feature?

I suggest introducing a dedicated middleware, for example ChatTitleRefreshMiddleware, backed by a ChatTitleRefreshService. The new middleware could observe the appropriate conversation lifecycle hook, apply its own refresh policy, and delegate title generation and compare-and-set persistence to the chat service. It should then be registered from the application/runtime assembly layer, where both chat services and middleware composition are available.

A possible separation would be:

- MemoryMiddleware: memory search, auto-memory collection, and memory flush behavior only.
- ChatTitleRefreshMiddleware: decide when a completed conversation turn should trigger a title refresh.
- ChatTitleRefreshService: generate the title, locate the chat, preserve manual renames, and update ChatSpec.name and metadata.
- Application/runtime builder: construct and register the middleware with its required chat-layer dependencies.

This would remove the cross-layer callback chain from Workspace to concrete memory managers to BaseMemoryManager to MemoryMiddleware, keep memory backends independent of chat UI behavior, and allow title refresh cadence and configuration to evolve independently from auto-memory.

If the product requirement is specifically to refresh only after a successful auto-memory flush, a generic application-level AutoMemoryFlushed event could be published and handled by the chat title service. Even in that case, I would prefer not to expose a title-specific callback on the memory manager abstraction.

Would you please consider restructuring this part before merging? Thank you.

PR #7061, opened 2026-08-15
fix(video): deliver tool-result videos on OpenAI Responses API

### LINKED ISSUE #7059: [Bug]: view_video tool-result video blocks are silently dropped; the model never receives any video frames (OpenAI Responses API / Volcengine Ark)
## QwenPaw Version

## Description

Calling `view_video(video_path=<local mp4>)` returns `Video loaded: <filename>` and appears to succeed, but the model never receives any video frames/image data — no error, no warning, a completely silent failure. The model only sees a plain-text `<system-reminder>A(n) video file is returned...` line and cannot describe the video content.

Root cause: PR #6495 (`fix(video): deliver video data to models across OpenAI Response, Anthropic, and Chat Completions providers`) does not actually take effect on the **OpenAI Responses API** path. Two independent defects stack:

**1a. The video-promotion trigger flag is silently discarded by Pydantic (always False)**

`agents/model_factory.py`'s `_create_formatter_instance()` intends to enable video promotion:
```python
if isinstance(base_formatter, _promote_types):
    kwargs["promote_tool_result_images"] = True
```
But `promote_tool_result_images` is **not a Pydantic field of the agentscope formatter** (0 matches in a full search of agentscope 2.0.4.post1). `FormatterBase` is a standard Pydantic `BaseModel` with the default `extra="ignore"` — the kwarg is silently dropped. As a result the trigger condition in `FileBlockSupportFormatter.format()`:
```python
if _needs_video and getattr(self, "promote_tool_result_images", False):
```
is always False, so `_promote_tool_result_videos()` (which handles videos inside `ToolResultBlock.output`) never runs.

(Note: this field existed on agentscope **1.x** formatters; QwenPaw carried the old field name over when migrating from 1.x to 2.0, but the field no longer exists in 2.0.)

**1b. Key mismatch: `tool_call_id` vs `call_id`**

`_promote_tool_result_videos()` (`agents/model_factory.py:590`) locates the insertion point with `fmt_msg.get("tool_call_id")`:
```python
tcid = fmt_msg.get("tool_call_id")
```
But the OpenAI **Responses API** formatter (`agentscope/formatter/_openai_response_formatter.py`) emits `function_call_output` with **`call_id`**, not `tool_call_id`. Verified by test: the OpenAI chat format (`tool_call_id`) promotes video correctly; the Responses format (`call_id`) never matches. Even if 1a is fixed, this key mismatch still prevents video promotion.

**Related PR(s):** #6495

**Security considerations:** None (pure media-input formatting defect; no auth / env / config exposure)

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.1.0
- **Install method:** pip

## Steps to Reproduce

1. Use a provider backed by `OpenAIResponseModel` (Responses API), e.g. Volcengine Ark `ark-coding-plan` / `ark-token-plan` with `doubao-seed-2.1-turbo`.
2. Prepare a local mp4 video (≤2 MB is enough to reproduce).
3. Call `view_video(video_path="/path/to/video.mp4")`.
4. Ask the model "what is shown in the first frame of the video?".

Minimal reproduction script (build a message flow → `_create_formatter_instance` → formatter.format → inspect the wire payload):

```python
# Requires: agentscope 2.0.4.post1, qwenpaw 2.1.0
import asyncio, json
from pathlib import Path
from agentscope.message import (Msg, DataBlock, URLSource, TextBlock,
    ToolCallBlock, ToolResultBlock, ToolCallState, ToolResultState)
from agentscope.tool import ToolChunk
from qwenpaw.agents.model_factory import _create_formatter_instance
from qwenpaw.providers.capping_formatter import _CappingOpenAIResponseFormatter

video_path = "/tmp/v.mp4"
Path(video_path).write_bytes(b"\x00" * 64)

def fake_view_video(p):
    resolved = Path(p).resolve()
    return ToolChunk(is_last=True, state=ToolResultState.SUCCESS, content=[
        DataBlock(source=URLSource(url=resolved.as_uri(), media_type="video/mp4")),
        TextBlock(type="text", text=f"Video loaded: {resolved.name}"),
    ])

call_block = ToolCallBlock(id="call_video_1", name="view_video",
    input=json.dumps({"video_path": video_path}), state=ToolCallState.FINISHED)
chunk = fake_view_video(video_path)
result_block = ToolResultBlock(id="call_video_1", name="view_video",
    output=chunk.content, state=ToolResultState.SUCCESS)
msgs = [Msg(name="user", role="user", content=[TextBlock(type="text", text="video first frame?")]),
        Msg(name="assistant", role="assistant", content=[call_block, result_block])]

base = _CappingOpenAIResponseFormatter(max_bytes=2 * 1024 * 1024)
class FakeModel: formatter = base
formatter = _create_formatter_instance(FakeModel(), provider_id="ark-coding-plan")

async def main():
    out = await formatter.format(msgs)
    s = json.dumps(out, default=str)
    print("input_video in payload:", "input_video" in s)   # expected True, actual False

asyncio.run(main())
```

## Actual vs Expected

- **Actual:** In the wire payload the video is degraded to plain text `<system-reminder>A(n) video file is returned...`; there is no `input_video` / `video_url` block. The model cannot see the video.
- **Expected:** The wire payload should contain a user message like `{"type": "input_video", "video_url": "data:video/mp4;base64,..."}` so the model can understand the video.

## Logs / Screenshots

Actual output from the reproduction script (QwenPaw 2.1.0, unpatched):

```
=== formatted messages ===
{"role": "user", "content": [{"type": "input_text", "text": "video first frame?"}]}
{"type": "function_call", "call_id": "call_video_1", "name": "view_video", "arguments": "..."}
{"type": "function_call_output", "call_id": "call_video_1", "output": "<system-reminder>A(n) video file is returned and can be accessed at the URL: /tmp/v.mp4.</system-reminder>\nVideo loaded: v.mp4"}
INPUT_VIDEO PRESENT: False
```

## Additional Notes

**Root cause (two places)**

1a — trigger flag dropped:
- `agents/model_factory.py` `_create_formatter_instance()` → `kwargs["promote_tool_result_images"] = True`
- that field is not a Pydantic field of the agentscope formatter; `extra="ignore"` silently drops it → `getattr(self, "promote_tool_result_images", False)` is always False → `_promote_tool_result_videos()` never runs

1b — key mismatch:
- `agents/model_factory.py:646` `tcid = fmt_msg.get("tool_call_id")`; the Responses API emits `call_id`

**Suggested fix (verified locally)**

```python
# 1a trigger condition
if _needs_video:   # was: if _needs_video and getattr(self, "promote_tool_result_images", False):
    messages = _promote_tool_result_videos(...)

# 1b key matching (+ skip `function_call` to avoid duplicate insertion)
for fmt_msg in messages:
    new_messages.append(fmt_msg)
    if fmt_msg.get("type") == "function_call":
        continue
    tcid = fmt_msg.get("tool_call_id") or fmt_msg.get("call_id")
```

**Local workaround (confirmed working):** patch the two spots in site-packages `agents/model_factory.py`, or monkeypatch via a plugin startup hook. Verified: the Responses-format wire payload now contains `input_video` (exactly once); the Chat format regression test still passes.

**Why probe reports supports_video=true ("false positive"):** the multimodal probe calls `openai_response_provider.py`'s `_try_video_url()` which directly invokes `client.responses.create(input=[{"type": "input_video", ...}])`, bypassing the formatter pipeline entirely — so it proves the model supports video, but real inference goes through the formatter path that drops it. Probe and real inference are two different paths.

xiaoka76:
## Description

Two defects in the video-promotion path introduced by #6495 kept `view_video` tool-result frames out of the model context for providers that use the **OpenAI Responses API** (e.g. Volcengine Ark):

1. **Key mismatch.** `_promote_tool_result_videos` matched tool results via `fmt_msg.get("tool_call_id")`, but the Responses API formatter emits `function_call_output` with **`call_id`**, so promotion never attached. We now match both `tool_call_id` and `call_id`, and skip the assistant `function_call` item (which also carries `call_id`) to avoid duplicating the promoted video.

2. **Dead promotion gate.** The promotion gate read `promote_tool_result_images`, which is not a Pydantic field of the agentscope formatter — with the default `extra="ignore"` the constructor kwarg is silently dropped, so `getattr(self, "promote_tool_result_images", False)` was always `False` and `_promote_tool_result_videos()` never ran. Instead of dropping the gate (which would widen promotion to every non-Gemini/non-Anthropic formatter), we keep the gate and set the flag directly on the constructed formatter instance in `_create_formatter_instance`, so the gate stays effective for exactly the OpenAI-chat / Gemini / OpenAI-Responses formatters that need it.

Adds regression tests for Responses API (`call_id`) promotion, the `function_call` skip, and the chat (`tool_call_id`) path.

**Related Issue:** Fixes #7059

**Security Considerations:** None (media-input formatting only; no auth / env / config exposure)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Testing

Added regression tests in `tests/unit/agents/test_model_factory_video_capping.py` covering:

- Responses API (`call_id`) promotion inserts the `input_video` user message exactly once, after `function_call_output`.
- The assistant `function_call` item does **not** trigger a duplicate promotion.
- Chat (`tool_call_id`) path still promotes (no regression).

Also merged the latest `main` to resolve the conflicts flagged by the review.

## Evidence

```bash
# touched suites
pytest tests/unit/agents/test_model_factory_video_capping.py -q
22 passed

pytest tests/unit/agents/test_create_model_and_formatter_override.py \
  tests/unit/agents/test_model_factory_message_normalization.py \
  tests/unit/agents/test_model_factory_remote_media.py \
  tests/unit/providers/test_capping_formatter.py \
  tests/unit/providers/test_dashscope_capping_formatter.py -q
109 passed
```


src/qwenpaw/providers/capping_formatter.py
```diff
@@ -51,7 +51,7 @@
 # Maximum size (in bytes) of a local media file we are willing to inline as
 # base64 into the model request body.  See the module docstring for the
 # rationale.
-MAX_INLINE_MEDIA_BYTES = 2 * 1024 * 1024  # 2 MB
+MAX_INLINE_MEDIA_BYTES = 48 * 1024 * 1024  # 48 MB (base64-inflated: 64MB wire cap / 1.33)
```
qbc2016: **Blocking: keep the configurable-cap fix separate from #7059**

This global change is unrelated to the tool-result promotion defects fixed by this PR and overlaps with #7071, which routes the provider-specific `max_inline_media_bytes` value through the video formatting path.

Changing the shared default to 48 MiB also affects every provider, regardless of its configured or supported request limit. In addition, 48 MiB of raw media expands to approximately 64 MiB after base64 encoding before accounting for the data-URL prefix, JSON structure, prompts, tool messages, and conversation history, so it is not guaranteed to fit within a 64 MiB request-body limit.

Please revert this constant to the existing 2 MiB default and keep this PR focused on #7059. The configurable size behavior, including the existing `max_bytes <= 0` “disable capping” contract, should be handled by #7071.

src/qwenpaw/agents/model_factory.py
```diff
@@ -1179,11 +1185,7 @@ def _battr(block, key, default=None):
                 )
                 _restore_video_blocks(normalized_msgs, video_subs)
 
-            if _needs_video and getattr(
-                self,
-                "promote_tool_result_images",
-                False,
-            ):
+            if _needs_video:
```
qbc2016: **Blocking: preserve the original promotion scope**

Please restore the `promote_tool_result_images` gate. Since AgentScope 2.0.6 drops this field from constructor kwargs, set it on the constructed formatter instance so the restored gate remains effective. Using `_needs_video` alone unintentionally enables this promotion path for all non-Gemini/non-Anthropic formatters.

qbc2016 (commented):
Please see the inline comments. Please also merge the latest `main` branch and fix the conflicts.

src/qwenpaw/providers/capping_formatter.py
xiaoka76: Acknowledged and fixed. `MAX_INLINE_MEDIA_BYTES` is reverted to the original 2 MiB default (`capping_formatter.py:52`), so this PR no longer touches `capping_formatter.py` at all — the diff against main is now limited to `model_factory.py` + the new test file. The configurable-cap work (including the `max_bytes <= 0` "disable capping" contract) is left entirely to #7071. Thanks for the pointer.

src/qwenpaw/agents/model_factory.py
xiaoka76: Done. Restored the `promote_tool_result_images` gate around `_promote_tool_result_videos` in `FileBlockSupportFormatter.format`, so promotion is again scoped to OpenAI-chat / Gemini / OpenAI-Responses formatters and not enabled for every non-Gemini/non-Anthropic provider.

Because the field is not a Pydantic field of the agentscope formatter, `extra="ignore"` silently drops it from constructor kwargs (verified locally that passing it to the constructor leaves the attribute unset). So `_create_formatter_instance` now sets `promote_tool_result_images = True` directly on the constructed formatter instance for the `_promote_types` (OpenAI chat, Gemini, OpenAI Responses), keeping the restored gate effective. Smoke-tested: OpenAI chat + Responses → gate True, Anthropic → False.

qbc2016 (commented):
Thanks for addressing the previous review comments. The promotion gate and the PR scope now look correct.

This branch currently conflicts with the latest `main`, which now includes #7071 and its configurable video-cap changes in the same formatter and test files. Please merge or rebase onto the latest `main` and resolve the conflicts. 

Please also run `pre-commit run --all-files` to format the code.

xiaoka76:
Merged the latest `main` (e08f2e60, which includes #7071) into `fix/view-video-promotion` and resolved the conflicts. The promotion fix and #7071's configurable video cap now coexist in the same formatter and test files.

Ran `pre-commit run --all-files` and all checks pass (check-ast, check-yaml/xml/toml/json, trailing-whitespace, add-trailing-comma, mypy, black, flake8, pylint, actionlint). Unit tests also pass.

Thanks for the review!

qbc2016 (approved):
LGTM.

PR #7071, opened 2026-08-16
fix(agents): make view_video inline cap configurable instead of hardcoded 2 MB

### LINKED ISSUE #7060: [Bug]: view_video inline-media cap is hardcoded to 2 MB; the provider's max_inline_media_bytes setting has no effect on the video path. Proposal: expose configurable max image/video file size + Files API in the provider advanced settings.
## QwenPaw Version

## Description

When `view_video` is given a local video larger than 2 MB, the video block is replaced by the text placeholder `[video omitted from model context: local file is X bytes, exceeds inline limit of 2097152 bytes]` and the model cannot see it.

QwenPaw already has a configurable `max_inline_media_bytes` field (per-provider), but **the view_video video path never reads that config** — `agents/model_factory.py:42` does `from ..providers.capping_formatter import MAX_INLINE_MEDIA_BYTES`, and `providers/capping_formatter.py:54` hardcodes `MAX_INLINE_MEDIA_BYTES = 2 * 1024 * 1024`. The video helper functions (`_format_openai_video_block` at model_factory.py:394-415, `_format_anthropic_video_data_block`, etc.) only look at this hardcoded constant.

Verified: even when the provider config `max_inline_media_bytes` is set to 50 MB, a 3 MB video is still truncated.

Additionally, the current inline-only path cannot handle large files at all. Volcengine Ark (like many providers) supports a **Files API upload** path with far larger limits. The suggestion is to:
1. Make the max image/video file size configurable in the **provider advanced settings** (the UI where you set the API key), instead of a hardcoded 2 MB.
2. Support the **Files API** upload path for large media.

Volcengine Ark documented limits (for reference):
- Base64 image input: single image < **10 MB**, request body ≤ **64 MB**.
- Base64 video input: video file ≤ **50 MB**, request body ≤ **64 MB**.
- Files API upload: files up to **512 MB** (2 GB for video stored in a user TOS bucket) — this avoids the base64 size inflation and repeated upload cost.

**Related PR(s):** #6495 (the PR that introduced the video-inline logic used the hardcoded constant)

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.1.0
- **Install method:** pip

## Steps to Reproduce

1. Use any OpenAI-compatible / Response provider.
2. Prepare a local mp4 video of ~3 MB.
3. Call `view_video(video_path=...)` and inspect the wire payload (or ask the model to describe the frames).

## Actual vs Expected

- **Actual:** The video is replaced with a text placeholder (2 MB hardcoded cap). The model cannot see it. Changing `max_inline_media_bytes` in the provider config has no effect on the video path.
- **Expected:** The max media inline size should read the provider's configurable `max_inline_media_bytes` (so users can raise it), and ideally large media should be supported via a Files API upload path. At minimum the cap must be configurable rather than hardcoded to 2 MB.

## Logs / Screenshots

Reproduction (provider config max_bytes=50 MB, video 3 MB):

```
formatter max_bytes (from provider config): 52428800
hardcoded MAX_INLINE_MEDIA_BYTES (model_factory): 2097152
=== result ===
video omitted placeholder ❌ (hardcoded 2MB cap bit)
[..., {"type": "input_text", "text": "[video omitted from model context: local file is 3145740 bytes, exceeds inline limit of 2097152 bytes]"}, ...]
```

## Additional Notes

**Root cause:** `agents/model_factory.py:42` imports the hardcoded constant; `providers/capping_formatter.py:54` `MAX_INLINE_MEDIA_BYTES = 2 * 1024 * 1024`. The video helper functions do not read the provider's `max_inline_media_bytes` field, so the config layer (the 2097152 in the provider json) and the code layer (hardcoded 2 MB) diverge.

**Beware of base64 inflation:** base64 encoding inflates a file by ~1.33x. A 50 MB video becomes ≈ 67 MB after base64, which collides with Volcengine's "request body ≤ 64 MB" limit, so the safe inline cap is about 48 MB. Files API avoids this entirely.

**Suggested fix:**
1. Route the video inline cap through the provider's `max_inline_media_bytes` (consistent with the image/audio path) instead of the hardcoded constant.
2. Add a max image/video file size setting to the provider advanced settings UI (where the API key is configured), defaulting sensibly per provider.
3. Add Files API upload support for large media (recommended by Volcengine: up to 512 MB, avoids base64 inflation and repeated upload).

**Local workaround (confirmed working):** change the constant at `capping_formatter.py:54` to 48 MB, or inject it at runtime via a plugin startup hook (`model_factory.MAX_INLINE_MEDIA_BYTES = 48 * 1024 * 1024`).

suantea:
## What Problem This Solves

Fixes #7060.

`view_video` enforced a hardcoded 2 MB inline cap (`MAX_INLINE_MEDIA_BYTES`) on every provider path, so a video between 2 MB and the provider's configured `max_inline_media_bytes` (e.g. 50 MB) was always replaced by the "video omitted" text placeholder. Changing the provider config had **no effect** on the video path — larger videos could never reach the model even when the provider allowed them.

## Evidence

- `python -m pytest tests/unit/agents/test_model_factory_video_capping.py` → **24 passed** (incl. the new `test_openai_base64_video_zero_cap_disables_capping` regression test).
- 3 MB video: placeholder at the default 2 MB cap; inlined as `data:video/mp4;base64,` when the provider cap is raised to 50 MB (OpenAI url / OpenAI base64 / Anthropic url paths).
- The oversize placeholder text reports the effective limit (e.g. `52428800 bytes`), not the hardcoded 2 MB.

## Root cause

The video helpers in `src/qwenpaw/agents/model_factory.py` (`_format_openai_video_block`, `_format_anthropic_video_data_block`, `_replace_video_placeholders`, `_promote_tool_result_videos`) all compared sizes against the module-level hardcoded constant `MAX_INLINE_MEDIA_BYTES = 2 * 1024 * 1024` (from `providers/capping_formatter.py`), never reading the configurable `max_bytes` field that providers populate from `max_inline_media_bytes`.

## Change

- `src/qwenpaw/agents/model_factory.py`:
  - `_video_oversize_placeholder`, `_format_openai_video_block`, `_format_anthropic_video_data_block`, `_replace_video_placeholders`, `_promote_tool_result_videos` now accept a `max_inline_media_bytes` parameter (default unchanged = the existing 2 MB constant, so behavior is identical when no config is set).
  - `FileBlockSupportFormatter.format()` and `_format_anthropic_data_block` thread the formatter's Pydantic `max_bytes` field — populated from the provider's `max_inline_media_bytes` — through every call site.
  - The cap check preserves the existing `max_bytes <= 0` disables-capping contract.
  - The oversize placeholder text reports the effective limit.
- `tests/unit/agents/test_model_factory_video_capping.py`: tests covering a 3 MB video placeholder'd at the default 2 MB cap but inlined at a 50 MB provider cap (OpenAI url/base64, Anthropic url), placeholder echoing the custom limit, and `max_bytes <= 0` disabling the cap.

## Test

`python -m pytest tests/unit/agents/test_model_factory_video_capping.py` → 24 passed.

Note: full Files-API upload support for >cap media (the second part of the issue proposal) is a larger feature and intentionally left out of this small fix.


tests/unit/agents/test_model_factory_video_capping.py
```diff
@@ -304,3 +305,77 @@ def test_replace_placeholders_non_match_untouched() -> None:
     item = _first_content_item(msgs)
     assert item["type"] == "input_text"
     assert item["text"] == "hello"
+
+
+# ------------------------------------------------------- configurable cap
+
+
+def test_openai_url_video_under_default_but_over_custom_cap(tmp_path) -> None:
```
qbc2016: The threshold relationship in this test name appears to be reversed. A 3 MiB video is **over** the default 2 MiB cap and **under** the custom 50 MiB cap.

Could we rename it to:

```python
test_openai_url_video_over_default_but_under_custom_cap
```

qbc2016 (commented):
**Blocking: preserve `0` as “disable capping”**

The existing `max_bytes` contract treats values less than or equal to zero as disabling the cap. However, with this condition, configuring `max_bytes=0` makes every non-empty video oversized because `size > 0`.

Please preserve the existing behavior in all four new size checks—for the Anthropic and OpenAI paths, covering both base64 and local-file sources. For example:

```python
if (
    max_inline_media_bytes > 0
    and size > max_inline_media_bytes
):
```

Please also add regression tests with `max_inline_media_bytes=0` to verify that videos remain inline for both the OpenAI and Anthropic paths.

qbc2016:
Please merge the latest `main` branch and fix the conflicts.

suantea:
Hi @qbc2016, thanks for the review! All three points are addressed:

1. **Blocking: preserve `0` as "disable capping"** — the base64 branches of both `_format_openai_video_block` and `_format_anthropic_video_data_block` now only enforce the cap when `max_inline_media_bytes > 0 and size > max_inline_media_bytes`, keeping the existing `max_bytes <= 0` disables-capping contract. Added a regression test `test_openai_base64_video_zero_cap_disables_capping`.
2. **Test name** — renamed to `test_openai_url_video_over_default_but_under_custom_cap` (3 MiB > default 2 MiB, < custom 50 MiB).
3. **Conflicts** — merged the latest `main` (24 commits, incl. the media-preparation refactor) and resolved all conflicts; the local-file URL tests now go through `_prepare_media_sources`. Full test file: 24 passed.

Branch updated to `9ac16b0` and is mergeable. Please re-review, thanks!

qbc2016 (approved):
LGTM.

qbc2016 (approved):

PR #7072, opened 2026-08-16
feat(console): add background chat task list API

suantea:
## Summary

Implements the minimal part of issue #7056's proposal: a task status **list API**.

Background tasks submitted via `submit_to_agent` (POST `/console/chat/task`) could only be inspected one-by-one by polling `GET /console/chat/task/{id}`. For multi-agent coordination there was no way to query all in-flight tasks at once.

## Change

- `src/qwenpaw/app/routers/console.py`: new `GET /console/chat/tasks` endpoint returning a `{"tasks": {task_id: {...}}}` mapping with status summaries for every background chat task. Each entry mirrors the single-task response shape (status / started_at / finished_at / result for finished tasks), so callers can check the whole set in one request.
- `tests/unit/app/routers/test_console_task_list.py`: 3 tests — empty store, mixed running/finished tasks, and shape parity with the single-task endpoint.

## Test

`python -m pytest tests/unit/app/routers/test_console_task_list.py` (3 tests) — locally validated via direct function-level checks because the full `qwenpaw.app.routers` import chain requires heavy optional deps (playwright etc.); CI runs the file with the full environment.

Note: WebSocket push (the issue's first proposal) is a larger feature and intentionally left out of this small PR.

### DIFF
--- a/src/qwenpaw/app/routers/console.py
+++ b/src/qwenpaw/app/routers/console.py
@@ -922,3 +922,34 @@ async def get_console_chat_task(task_id: str) -> dict:
     if bg.status == "finished" and bg.result is not None:
         response["result"] = bg.result
     return response
+
+
+@router.get(
+    "/chat/tasks",
+    status_code=200,
+    summary="List all background chat task statuses",
+)
+async def list_console_chat_tasks() -> dict:
+    """Return status summaries for all background chat tasks.
+
+    Lets callers query the whole task set in one request instead of
+    polling each task id individually (issue #7056). Each entry mirrors
+    the single-task response shape; finished tasks include their result.
+    """
+    async with _bg_lock:
+        task_ids = list(_bg_tasks.keys())
+    tasks: Dict[str, Any] = {}
+    for task_id in task_ids:
+        async with _bg_lock:
+            bg = _bg_tasks.get(task_id)
+        if bg is None:
+            continue
+        entry: Dict[str, Any] = {"status": bg.status}
+        if bg.started_at is not None:
+            entry["started_at"] = bg.started_at
+        if bg.finished_at is not None:
+            entry["finished_at"] = bg.finished_at
+        if bg.status == "finished" and bg.result is not None:
+            entry["result"] = bg.result
+        tasks[task_id] = entry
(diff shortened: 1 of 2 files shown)


zhijianma:
Thank you for the contribution. After reviewing the implementation, we do not think this PR is suitable to merge in its current form.

The change is too narrow for #7056: it adds a process-local list endpoint, but it does not provide a callback, notification, long-poll/wait mechanism, or an Agent-facing batch query tool. The initiating Agent would still need to call check_agent_task for each task, so the coordination flow described in the issue is unchanged.

More importantly, _bg_tasks is a module-global store and its records contain no requester Agent, target Agent, chat/session, or workspace ownership metadata. GET /console/chat/tasks therefore enumerates every background task in the process and returns completed results without Agent or chat isolation. This expands the existing opaque-ID lookup into cross-Agent/cross-chat result enumeration.

There are also several lifecycle issues that need an architectural solution:

- finished tasks are never removed, while the endpoint has no filtering, pagination, retention limit, or result-size bound;
- the store is in-memory and per-process, so restarts lose all tasks and multi-worker deployments return incomplete/inconsistent lists;
- list and detail response shapes are not actually identical because only the list includes finished_at;
- the tests call the function directly and do not cover HTTP behavior, ownership isolation, concurrency, cleanup, or multi-Agent usage.

A complete solution should first introduce an owned task model keyed by requester/target Agent and chat/session, enforce that scope in list/detail APIs, add retention and pagination, and expose an Agent-facing batch wait mechanism. Push/SSE notification can then be built on top of the same task repository.

Since addressing these points requires redesigning the task lifecycle rather than extending this endpoint incrementally, we are closing this PR. Thanks again for identifying and exploring the problem.

suantea:
感谢审查和详细的关闭说明，理解这个改动对 #7056 来说确实太窄了（只有进程内列表端点，缺回调/通知闭环）。想请教一下：#7056 期望的形态是后台任务带回调通知（webhook/轮询）的完整闭环吗？如果按这个方向重做一版更完整的实现，是否更符合项目路线？我可以重新设计后再提 PR。

PR #7081, opened 2026-08-17
feat: integrate AnySearch web search (SearchProvider + MCP)

anysearch-ai:
## Description

Integrate **AnySearch** (https://anysearch.com) into QwenPaw as a first-class web search capability — pluggable `web_search` backend, built-in MCP client, and per-agent Console configuration:

1. **Fix MCP env-ref header binding defects end to end** (`fix(mcp): handle env-ref header bindings safely across the full cycle`)
   - Fix: empty env values no longer produce an illegal `Authorization: Bearer ` header that h11 rejects at connection time (guard is exact `None`/`''`, so legitimate falsey values like `0`/`False` still stringify)
   - Fix: Console displays `${VAR}` templates and Console save preserves the env binding (`plan_env_ref_bindings`) instead of degrading it to a static literal secret
   - Fix: stale env credential refs are removed when a header switches to a static literal; multi-variable templates (`${USER}:${PASS}`) are stored as static values, not silently dropped
   - Fix: env aliases are collision-free for case-sensitive variable names (`${ApiKey}` vs `${APIKEY}`); the original variable name is preserved from the `env:` credential ref when serializing
2. **Add AnySearch as the built-in MCP client** (`feat(mcp): add AnySearch as the built-in MCP client`)
   - Replace the `tavily_search` MCP template with AnySearch (`https://api.anysearch.com/mcp`, streamable HTTP)
   - Disabled by default (`enabled=False`), follows the default ask policy — no blanket allow
   - Set `ANYSEARCH_API_KEY` for authenticated requests (anonymous otherwise); config-level change only, legacy migration converts the template to a DriverCard on first start
3. **Abstract web_search behind a pluggable SearchProvider** (`feat(tools): abstract web_search behind a pluggable SearchProvider`)
   - `TavilyProvider` (default): legacy keyless backend, unchanged behavior
   - `AnySearchProvider`: REST `POST https://api.anysearch.com/v1/search`, anonymous by default, `Authorization: Bearer` when `ANYSEARCH_API_KEY` is set
   - Provider selection via `QWENPAW_WEBSEARCH_PROVIDER` (`tavily` default, `anysearch` opt-in); unknown values raise instead of silently routing
4. **Fix env-backed binding identification from credential refs** (`fix(mcp): identify env-backed bindings from credential refs, not alias prefix`)
   - A static credential alias like `env_custom` was misclassified as environment-backed because identification relied on the `env_` alias naming convention (pre-existing QwenPaw design limitation of the alias scheme); now decided from the `env:` credential refs (via `env_aliases`) instead
5. **Relocate secret mask/restore helpers** (`refactor(security): move secret mask/restore helpers into security.secret_store`)
   - Move `mask_secret_value` / `restore_masked_secret_value` from the MCP-specific `mcp_binding` module to neutral `security.secret_store` so builtin tools (web_search config) can use them without depending on MCP adapters; `mcp_binding` keeps public aliases for compatibility
6. **Add per-agent web_search Console config with credential-store api_key** (`feat(tools): per-agent web_search Console config with credential-store api_key`)
   - `GET`/`POST` config endpoints with masked credential readback; `?provider=` query looks up the credential slot for a provider the user is switching to
   - Provider resolved from per-agent config; AnySearch 402 handling keeps only auto-registration (B) and anonymous-retry (A), all other 402s surface the raw provider message
   - `proactive_responder` sets `agent_id` + `workspace_dir` ContextVars so proactive background replies resolve the right provider/key
   - Console: dedicated `WebSearchConfigModal` (provider select; keyless Tavily hides api_key; quota hint with register link; provider switch re-fetches the credential slot), Configure button on the tool card, `getToolConfig` query support, 7-locale i18n (en/id/ja/pt-BR/ru/vi/zh)
7. **Dismiss desktop-mode Tour overlay in e2e fixtures** (`fix(e2e): dismiss desktop-mode Tour overlay in browser fixtures`)
   - The onboarding Tour mask (`DIV.ant-tour-mask`) intercepts pointer events on first load and breaks clicking switches/buttons in automated tests; seed the same localStorage key the console uses (`qwenpaw.desktop-mode-hint.dismissed`) on every test page

**Related Issue:** None (partner integration)

**Security Considerations:** MCP client disabled by default (`enabled=False`), follows the default ask policy, no blanket allow; header secrets referenced via `${VAR}` env vars, never persisted in plaintext; `api_key` stored encrypted in the driver credential store — never in `agent.json`, never returned in plaintext (masked readback); AnySearchProvider always enforces TLS certificate validation (no `verify=False` fallback — a validation bypass would leak the API key).

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes（17/17 hooks: 16 passed, 1 skipped [sort-simple-yaml — no files to check]. check-ast, check-yaml, check-xml, check-toml, check-docstring-first, check-json, fix-encoding-pragma, detect-private-key, trailing-whitespace, add-trailing-comma, mypy, black, flake8, pylint, prettier, actionlint all pass）
- [x] I ran tests locally (`pytest` or as relevant) and they pass（20 related unit tests pass）
- [ ] Documentation updated (if needed)（README / docs changes prepared separately, will be submitted in a follow-up PR）
- [x] Ready for review

## Testing

```bash
pytest tests/unit/app/routers/test_tools_router_web_search_config.py \
       tests/unit/agents/tools/test_websearch_providers.py \
       tests/unit/agents/tools/test_websearch_anysearch_quota.py \
       tests/unit/agents/memory/proactive/test_proactive_context_propagation.py \
       tests/unit/config/test_mcp_config_defaults.py \
       tests/unit/drivers/adapters/test_credentials_bindings.py \
       tests/unit/drivers/adapters/test_mcp_legacy_env_ref.py \
       tests/integration/test_driver_mcp_http_flow.py --no-cov
```

## Evidence

```bash
# 20 related unit tests pass (web_search config router + SearchProvider)
$ pytest tests/unit/app/routers/test_tools_router_web_search_config.py tests/unit/agents/tools/test_websearch_providers.py -q
20 passed in 0.82s
```

```text
# Full pre-commit run on all files
$ pre-commit run --all-files
check python ast.............................Passed
sort simple yaml files........................(no files to check)Skipped
check yaml...................................Passed
check xml....................................Passed
check toml...................................Passed
check docstring is first.....................Passed
check json...................................Passed
fix python encoding pragma...................Passed
detect private key...........................Passed
trim trailing whitespace.....................Passed
Add trailing commas..........................Passed
mypy.........................................Passed
black........................................Passed
flake8.......................................Passed
pylint.......................................Passed
prettier.....................................Passed
Lint GitHub Actions workflow files...........Passed
```

## Additional Notes

- The PR surfaces a latent generic defect in QwenPaw (commit 1): any HTTP MCP client using a `${VAR}` header with an unset environment variable would fail to connect. It was never triggered before because no built-in client combined "optional auth" with "env-ref header" (Tavily uses stdio and never goes through the header binding path). AnySearch's "anonymous by default, optional key" configuration is the first to expose it.
- The `web_search` builtin tool keeps Tavily as the **default** provider; AnySearch is an explicit opt-in (`QWENPAW_WEBSEARCH_PROVIDER=anysearch` or per-agent Console config). The built-in MCP client is likewise disabled by default — no behavior change for existing users who do not opt in.
- PR scope: 7 commits (`6b3dc19d..HEAD`). The branch additionally carries 56 commits already merged into upstream `main`; they are excluded from this PR's diff.
- README / documentation changes (AnySearch integration docs, badge, news entry) are prepared but intentionally excluded from this PR; they will be submitted separately.


src/qwenpaw/agents/tools/websearch/__init__.py
```diff
@@ -0,0 +1,285 @@
+# -*- coding: utf-8 -*-
```
rayrayraykk, line 1: Please do not implement code in `__init__.py`, e.g.

```python
websearch/
├── base.py        
├── client.py     
├── tavily.py
├── anysearch.py  
└── factory.py  
```

rayrayraykk (commented):
**Issue 1 · Instance-level lock on `AsyncCredentialStore` causes silent credential loss under concurrency**

A new `AsyncCredentialStore(path)` is created on every read or registration, so the `RLock` is per-instance and never shared across callers. Two instances can concurrently complete a read-modify-write cycle and each call `os.replace()` independently — the later writer silently overwrites the entire file, discarding the earlier writer's changes. `os.replace()` only prevents a torn write; it does nothing to serialize read-modify-write. This is reproducible in practice when AnySearch auto-registration races with a Console save of MCP/model credentials: the YAML remains parseable, but one credential entry disappears without any error.

Suggest having the provider obtain a shared `credential_store` from the workspace/driver manager, or maintaining a process-level singleton keyed by canonical path. Multi-process deployments will additionally need a file lock. Please add a regression test: two `AsyncCredentialStore` instances write different refs concurrently and the assertion verifies both entries are present in the final file.

---

**Issue 2 · No transactional guarantee between credential write and config write**

The `POST` handler calls `put`/`delete` on `credentials.yaml` first, then writes `agent.json` inside a `try` block. If the second step fails (permission error, disk full, concurrent modification, JSON serialization error, etc.), the API returns 500 but the credential change has already been persisted. In the `delete` case the original key is unrecoverable, so the user sees a "failed" response while the on-disk state has partially changed.

Suggest introducing a service layer that coordinates both writes: snapshot the old credential record before any mutation, and roll it back if `set_tool_config()` raises. A more robust design would commit the agent config first and only then persist the credential, with a state machine that supports retry and recovery. Tests should cover both `put` and `delete` rollback scenarios when `set_tool_config()` throws.

anysearch-ai:
> **Issue 1 · Instance-level lock on `AsyncCredentialStore` causes silent credential loss under concurrency**
> 
> A new `AsyncCredentialStore(path)` is created on every read or registration, so the `RLock` is per-instance and never shared across callers. Two instances can concurrently complete a read-modify-write cycle and each call `os.replace()` independently — the later writer silently overwrites the entire file, discarding the earlier writer's changes. `os.replace()` only prevents a torn write; it does nothing to serialize read-modify-write. This is reproducible in practice when AnySearch auto-registration races with a Console save of MCP/model credentials: the YAML remains parseable, but one credential entry disappears without any error.
> 
> Suggest having the provider obtain a shared `credential_store` from the workspace/driver manager, or maintaining a process-level singleton keyed by canonical path. Multi-process deployments will additionally need a file lock. Please add a regression test: two `AsyncCredentialStore` instances write different refs concurrently and the assertion verifies both entries are present in the final file.
> 
> **Issue 2 · No transactional guarantee between credential write and config write**
> 
> The `POST` handler calls `put`/`delete` on `credentials.yaml` first, then writes `agent.json` inside a `try` block. If the second step fails (permission error, disk full, concurrent modification, JSON serialization error, etc.), the API returns 500 but the credential change has already been persisted. In the `delete` case the original key is unrecoverable, so the user sees a "failed" response while the on-disk state has partially changed.
> 
> Suggest introducing a service layer that coordinates both writes: snapshot the old credential record before any mutation, and roll it back if `set_tool_config()` raises. A more robust design would commit the agent config first and only then persist the credential, with a state machine that supports retry and recovery. Tests should cover both `put` and `delete` rollback scenarios when `set_tool_config()` throws.

---

# PR Review Response

Thank you for the detailed review. We have addressed the credential-race,
configuration consistency, provider-structure, and keyless-provider issues in
the latest commit:

`fix(websearch): address review feedback on credential concurrency, config transaction, and module split`

## Issue 1: Concurrent Credential Writes

**Problem:** AnySearch auto-registration and Console credential updates could
create separate `AsyncCredentialStore` instances. Each instance had its own
lock, so concurrent read-modify-write operations could overwrite one another's
entries when replacing the credentials file.

**Resolution:**

- `AsyncCredentialStore` is now a process-level singleton keyed by canonical
  credential-store path.
- All callers operating on the same credentials file share the same lock and
  in-memory state.
- Added a regression test that widens the read-modify-write window and verifies
  that concurrent writes preserve both credential references.

## Issue 2: Agent Config and Credential Store Consistency

**Problem:** Updating the per-agent `web_search` configuration and updating its
credential could leave `agent.json` and `credentials.yaml` out of sync when one
operation succeeded and the other failed.

**Resolution:**

- Credential mutation is deferred until the agent configuration has been
  committed.
- If the credential write fails, the agent configuration is rolled back.
- Added regression coverage for successful updates and configuration/credential
  rollback scenarios.

## Provider Module Structure

The original `websearch/__init__.py` had grown to approximately 285 lines. It
has been split into focused modules:

- `base.py` — provider protocol and shared HTTP helpers
- `tavily.py` — Tavily provider
- `anysearch.py` — AnySearch provider, quota handling, and credential lookup
- `factory.py` — per-agent provider selection
- `__init__.py` — public re-exports only

This keeps provider-specific behavior isolated while preserving the existing
public import surface.

## Keyless Tavily Switching

When switching an agent back to keyless Tavily:

- A leftover AnySearch `api_key` is removed from the submitted configuration.
- The server skips credential I/O for keyless providers.
- A previously stored AnySearch key cannot be written into the Tavily slot.

The Console modal and router tests were updated to cover this behavior.

## AnySearch Credential Lookup Performance

Per-agent AnySearch credentials are cached using the `credentials.yaml` mtime
as the cache key, matching the existing agent-configuration cache strategy.
The credential store is re-read when the file changes, so Console updates remain
visible without reading and decrypting the file on every search request.

## Additional Test Coverage

The latest changes add or update coverage for:

- Concurrent credential writes
- Configuration/credential rollback
- Keyless-provider credential guards
- Protocol-agnostic credential binding resolution
- AnySearch quota handling
- Provider selection and module-specific patch targets
- Console provider switching behavior

## Validation

The full pre-commit suite was rerun after the amend:

```text
16 hooks passed
1 hook skipped: sort-simple-yaml (no files to check)
0 hooks failed
```

cuiyuebing (approved):
LGTM

PR #7087, opened 2026-08-17
fix(agents): localize remote media URLs client-side before model requests

jcs130:
## Problem

When a conversation history contains media blocks referencing remote HTTP(S) URLs (e.g. images returned by web tools), the URL is forwarded to the model backend verbatim. Backends then fetch the URL **server-side**, which fails for hotlink-protected hosts (HTTP 403) or networks the backend cannot reach. Several backends (e.g. vLLM) reject the **whole request** when any media URL cannot be fetched — and since the URL stays in the conversation history, a single dead URL poisons **every subsequent turn** of the session, even for prompts that never touch that image.

Repro sketch (vLLM backend):

1. A tool returns an image URL from a hotlink-protected host (e.g. `i.pximg.net`).
2. The URL lands in message history as a dict `image` block (1.x) or `DataBlock` (2.0).
3. Every following model request fails with HTTP 422 (`Image download failed`) until the session is reset.

## Solution

Localize remote media **client-side** in `_fixup_media_list`:

- Download the URL (10s timeout, 30 MB cap) with per-host `Referer` fix-ups for known hotlink-protected sites; cache under `~/.qwenpaw/media_cache`, keyed by URL hash.
- Rewrite the media block to the cached local path; backends embed it as usual.
- On failure, degrade the block to a text placeholder (consistent with the existing "file deleted from disk" path) and negative-cache the URL for 24 h so dead URLs are not retried on every request.
- Covers both the dict-block (1.x) and `DataBlock` (2.0) paths.

## Configuration

- `QWENPAW_DISABLE_REMOTE_MEDIA_CACHE=1` restores the legacy pass-through behavior.
- `QWENPAW_REMOTE_MEDIA_CACHE_DIR` overrides the cache location.

## Testing

- New offline unit tests in `tests/unit/agents/test_remote_media_cache.py` (8 tests): positive disk-cache hit, negative cache, non-media content-type rejection, dict/DataBlock rewrite + graceful degradation, kill-switch pass-through. Downloads are monkeypatched, so no network access is needed in CI.
- Existing `test_model_factory_message_normalization.py` shows no regression; the 13 failures I see on that file in my local env are pre-existing on `main` (older local agentscope build; verified byte-identical failure lists with and without this patch).

### DIFF
--- a/src/qwenpaw/agents/model_factory.py
+++ b/src/qwenpaw/agents/model_factory.py
@@ -45,6 +45,7 @@ from agentscope.formatter import OpenAIResponseFormatter
 from .utils.message_request_normalizer import (
     normalize_messages_for_model_request,
 )
+from .remote_media_cache import localize_remote_url, remote_media_cache_enabled
 from ..exceptions import ProviderError, ModelFormatterError
 from ..providers import ProviderManager
 from ..providers.capping_formatter import MAX_INLINE_MEDIA_BYTES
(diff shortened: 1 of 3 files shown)


qbc2016:
Thanks for the contribution. The original invalid remote-image issue has now been fixed by #7146, which validates and freezes remote images before persisting `view_image` results. The latest main also includes a centralized asynchronous `_prepare_media_sources()` pipeline.
This PR now duplicates and bypasses that pipeline, causing four relevant tests to fail when applied to the latest `main`. Its downloader also lacks the SSRF protections used by the new `view_image` implementation.
We’re therefore closing this PR. Any remaining support for legacy sessions, remote videos, or external media URLs should be implemented by extending the existing media-preparation and safe-download mechanisms on the latest main.
Thanks again for your work on this.

PR #6990, opened 2026-08-13
fix(skill): Reduce file io for system files & skills files via file cache [READY TO MERGE]

Leirunlin:
## Description

Adds a process-level cache for stable Markdown files used by system prompts and runtime Skills.

This reduces repeated file reads and Skill frontmatter parsing across queries while preserving file-change detection. It also avoids unnecessary Skill manifest reconciliation and uses atomic writes for in-place SKILL.md updates.

The cache is bounded, thread-safe, and shared across channels. Dynamic prompt content is still rebuilt for every query.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [x] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [x] Ready for review

PR #7097, opened 2026-08-18
fix(skill): remove skill bound duplication [ready to merge]

### LINKED ISSUE #7073: feat(skill-system): add skill name deduplication to prevent duplicate loading of workspace and built-in skills
## Problem

When a user creates a custom skill in their workspace directory (e.g., `workspace/skills/my-tool/SKILL.md`) that has the same name as a built-in skill, both versions get loaded during agent initialization. This happens because `builder.py` only performs **path deduplication** (`if extra not in skill_dirs`) but does not check for **name conflicts**.

## Current Behavior

```python
# In src/qwenpaw/runtime/builder.py
for extra in _bound_skill_loader_dirs(tools):
    if extra not in skill_dirs:
        skill_dirs.append(extra)  # No name checking
```

Result: Both the workspace version and the built-in version are registered in the toolkit, causing duplicate warnings on each startup.

## Proposed Fix

Add name-based deduplication before adding extra skill directories:

```python
# Build a set of workspace skill names first
workspace_skill_names = set()
for sd in skill_dirs:
    sk_md = Path(sd) / "SKILL.md"
    if sk_md.exists():
        try:
            front = sk_md.read_text(encoding="utf-8").split("---")[1]
            for line in front.splitlines():
                if line.strip().startswith("name:"):
                    workspace_skill_names.add(line.strip().split(":", 1)[1].strip())
        except Exception:
            pass

# When adding extra dirs, check for name conflicts
for extra in _bound_skill_loader_dirs(tools):
    if extra not in skill_dirs:
        extra_sk_md = Path(extra) / "SKILL.md"
        if extra_sk_md.exists():
            try:
                front = extra_sk_md.read_text(encoding="utf-8").split("---")[1]
                for line in front.splitlines():
                    if line.strip().startswith("name:"):
                        builtin_name = line.strip().split(":", 1)[1].strip()
                        if builtin_name in workspace_skill_names:
                            logger.debug(
                                "Skipping built-in skill %r (workspace version takes precedence)",
                                builtin_name,
                            )
                            break
                        else:
                            skill_dirs.append(extra)
                            break
            except Exception:
                skill_dirs.append(extra)
        else:
            skill_dirs.append(extra)
```

## Benefits

1. **Workspace override**: Users can customize built-in skills by placing them in their workspace without editing original files
2. **No duplicate warnings**: Prevents the same skill from being loaded twice
3. **Backward compatible**: Only affects cases where workspace and built-in have the same name

## Evidence

This fix was tested locally with a workspace skill named `my-tool` that shadowed a built-in skill. Before the fix, both versions were loaded; after the fix, only the workspace version is loaded.

---

*Note: This is a suggestion/feature request rather than a PR, as it may need discussion about the best approach (e.g., should we parse frontmatter for all skills or use a manifest-based lookup)?*

Leirunlin:
## Description

Deduplicate workspace and bound skills by their `SKILL.md` frontmatter name.

Skills are loaded through the runtime cache added in #6990. If an effective workspace skill has the same runtime name as a bound skill, only the workspace version is passed to AgentScope.

## Behavior

- Workspace overrides a same-name bound skill.
- Different runtime names remain available.
- Disabled, missing, or invalid workspace skills fall back to the bound skill.
- No duplicate AgentScope warning.
- No UI, manifest, or naming changes.

Workspace and bound skills are loaded separately to preserve their source.
Both calls use the shared cache, so files are not read twice.

Related to #7073.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

### DIFF
--- a/src/qwenpaw/runtime/builder.py
+++ b/src/qwenpaw/runtime/builder.py
@@ -229,17 +229,27 @@ class AgentBuilder:
         workspace_dir: str | None,
         tools: Iterable[Any],
     ) -> list[Any]:
-        """Resolve and load runtime Skills in one synchronous I/O job."""
+        """Load runtime Skills, preferring workspace skills on conflicts."""
         from ..agents.skill_system.runtime_cache import load_runtime_skills
 
-        skill_dirs = cls._resolve_skill_loader_dirs(
+        workspace_skill_dirs = cls._resolve_skill_loader_dirs(
             effective_skills,
             workspace_dir,
         )
-        for extra in _bound_skill_loader_dirs(tools):
-            if extra not in skill_dirs:
-                skill_dirs.append(extra)
-        return load_runtime_skills(skill_dirs)
+        workspace_skills = load_runtime_skills(workspace_skill_dirs)
+        workspace_skill_names = {skill.name for skill in workspace_skills}
+
+        bound_skill_dirs = [
+            skill_dir
+            for skill_dir in dict.fromkeys(_bound_skill_loader_dirs(tools))
+            if skill_dir not in workspace_skill_dirs
+        ]
+        bound_skills = load_runtime_skills(bound_skill_dirs)
+        return workspace_skills + [
+            skill
+            for skill in bound_skills
+            if skill.name not in workspace_skill_names
+        ]
 
     # ----------------------------------------------------------------- build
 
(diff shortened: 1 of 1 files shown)


lalaliat (approved):
LGTM

PR #7116, opened 2026-08-18
fix(sandbox): expand ~ and  in policy-derived mount paths so out-of-w…

### LINKED ISSUE #7005: [Bug]: Enabling Shabox causes UV Run to fail.
## QwenPaw Version

2.1.0

## Description
Enabling Shabox causes UV can't to wirite ~/.cache/uv Folder.

Image

At present, by adding 'Write(~/.cache/uv/**)'to the policy.yaml file, we hope to optimize it and have a better solution.
There are also shabox related configurations in official documents, but I don't know where to set them up.

Image

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

vanwaals:
## Description

Enabling the sandbox blocks `uv run` from writing `~/.cache/uv`, and the workaround the docs point at — adding `Write(~/.cache/uv/**)` to `policy.yaml` — does not help. This PR makes that workaround actually work and documents where to configure it.

Root cause: `mounts` has no user-editable field. The list is derived by `ResourceGovernor.compile_sandbox_config()` from `policy.yaml` rules — a `Write(...)` rule becomes a writable mount. But `_resolve_mount_path()` never expanded `~`, so `Write(~/.cache/uv/**)` resolved to the literal `<workspace>/~/.cache/uv`. That path never exists, so every backend's "bind only if the path exists" check silently dropped the mount and the grant did nothing. `deny_paths` already ran `os.path.expanduser` in every backend; only the mount-derivation path was missing it, so the two were inconsistent.

Changes:

- `_resolve_mount_path()` now expands `~` and `$VAR` (via `os.path.expanduser(os.path.expandvars(...))`) before deciding absolute-vs-relative, matching how `deny_paths` is treated. Uses `os.path.isabs` instead of `startswith("/")` so a Windows path is not mistaken for workspace-relative.
- A declared mount whose path is absent at sandbox start is now reported at `WARNING` (`mount path(s) absent and therefore NOT bound`) rather than silently skipped. It is **not** an error: a tool cache legitimately does not exist until its tool first runs, so the sandbox still starts. Only emitted where the backend actually enforces `mounts` (not for `NoneSandbox`, which already reports it ignores the field wholesale).
- Docs gain a "Granting a path outside the workspace" section explaining that mounts are configured through `policy.yaml` rules (there is no `mounts` field to hand-edit), with the `uv` cache example.

**Related Issue:** Fixes #7005

**Security Considerations:** No enforcement is added or relaxed. A writable mount is still only ever produced by an explicit `Write(...)` rule the operator wrote; this PR just makes the path such a rule resolves to correct. If anything the surface is now more visible: a granted path that will not be bound is surfaced at `WARNING` instead of vanishing.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [x] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

Not applicable — this PR touches no channel code.

## Testing

```bash
pytest tests/unit/governance/test_mount_path_resolution.py -q   # 15 cases
pytest tests/unit/sandbox/test_unenforced_config.py -q          # 68 cases
```

New `tests/unit/governance/test_mount_path_resolution.py` (15 cases) covers path resolution directly:

- `~` and `$VAR` expand to real paths, not `<workspace>/~/...`; the result is absolute and never workspace-relative.
- An undefined `$VAR` is left untouched (stays workspace-relative) rather than collapsing to the filesystem root.
- The previously-working patterns are unchanged: absolute paths pass through, `WORKSPACE_DIR` resolves to the workspace, relative paths join the workspace, and pure wildcards (`*`, `**`, `.`) are skipped.

`tests/unit/sandbox/test_unenforced_config.py` gains a `TestMissingMountPaths` group: an absent mount path is reported at `WARNING` only where `mounts` is enforced, existing paths stay silent, and an absent path never stops the sandbox from starting.

**Mutation check.** Reverting the `~`-expansion (back to `if p.startswith("/")`) fails 4 of the new resolution cases and restores them green, confirming the tests bind to the fix rather than passing vacuously.

**End-to-end (real Seatbelt sandbox, macOS).** Beyond the unit tests I reproduced the issue against a live sandbox: build a `ResourceGovernor`, compile the config, and run a write through `create_sandbox` into a unique `~/.qp_e2e_<rand>` directory. Without the rule the write is blocked (`exit=1`); after inserting `Write(~/.qp_e2e_<rand>/**)` the mount resolves to the real home path, is writable, and the sandboxed write succeeds (`exit=0`). The two outcomes differ, so the sandbox is genuinely enforcing rather than passing through. Not run on Linux/Windows backends — the path-resolution fix is platform-independent, but a live bubblewrap/Landlock/AppContainer write was not exercised.

## Evidence

```bash
$ pre-commit run --all-files
```

```bash
$ pytest tests/unit/governance/test_mount_path_resolution.py -q -p no:randomly
...............                                                          [100%]
15 passed in 0.03s

$ pytest tests/unit/sandbox/test_unenforced_config.py -q -p no:randomly
....................................................................     [100%]
68 passed in 0.14s

$ pytest tests/unit/sandbox tests/unit/governance -q -p no:randomly
.........................................................................[100%]
549 passed in 2.76s
```

End-to-end reproduction against a real Seatbelt sandbox:

```
backend mode: SandboxMode.SEATBELT

[no rule]   mount for ~/<rel> writable?  False   (expected False)
[no rule]   sandboxed write succeeded?   False   exit=1   (blocked by sandbox)

[with rule] mount resolves to real path? True   writable=True
[with rule] sandboxed write succeeded?   True    exit=0

=> issue example holds end-to-end
```

## Additional Notes

- `mode=none` ignores mounts entirely, so inside a container without a kernel backend this grant is irrelevant — nothing is restricted to begin with. The docs call this out.
- The report only fires where a backend enforces `mounts`; `NoneSandbox` already emits a field-level "ignores mounts" line, so adding a per-path line there would be redundant.
- Markdown docs were checked with prettier 3.0.0 (the repo's prettier hook matches only `*.tsx?`, so it does not cover `.md`).

### DIFF
--- a/src/qwenpaw/governance/resource_governor.py
+++ b/src/qwenpaw/governance/resource_governor.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import hashlib
 import logging
+import os
 from pathlib import Path
 from typing import Optional
 
@@ -385,7 +386,8 @@ class ResourceGovernor:
 
         Strategy:
             - WORKSPACE_DIR/* → workspace_dir (mount as a whole)
-            - /absolute/path/* → /absolute/path (take directory part)
+            - ``~`` / ``$VAR`` → expanded before anything else
+            - absolute path/* → that path (take directory part)
             - relative path → workspace_dir / relative (take directory part)
             - Pure wildcards (*, **) → skip, cannot derive a concrete path
         """
@@ -399,8 +401,17 @@ class ResourceGovernor:
         if "WORKSPACE_DIR" in p:
             p = p.replace("WORKSPACE_DIR", workspace_dir)
 
-        # Absolute path
-        if p.startswith("/"):
+        # ``~/.cache/uv`` is how an operator naturally writes a tool cache
+        # in policy.yaml, and every backend already expands ``~`` for
+        # deny_paths. Leaving it literal here fell through to the relative
+        # branch below and produced ``<workspace>/~/.cache/uv`` -- a path
+        # that never exists, so the backends' existence check dropped the
+        # mount and the grant silently did nothing.
+        p = os.path.expanduser(os.path.expandvars(p))
+
+        # isabs() rather than startswith("/") so a Windows path such as
+        # ``C:\Users\...`` is not mistaken for a workspace-relative one.
+        if os.path.isabs(p):
             return p
 
         # Relative path → resolve based on workspace
--- a/src/qwenpaw/governance/resource_governor.py
+++ b/src/qwenpaw/governance/resource_governor.py
@@ -390,6 +390,9 @@ class ResourceGovernor:
             - absolute path/* → that path (take directory part)
             - relative path → workspace_dir / relative (take directory part)
             - Pure wildcards (*, **) → skip, cannot derive a concrete path
+
+        The result is always normalised, so the same directory written two
+        ways yields one string.
         """
         p = pattern.rstrip("*").rstrip("/")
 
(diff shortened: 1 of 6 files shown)


cuiyuebing (approved):
LGTM

cuiyuebing (approved):
LGTM

PR #7119, opened 2026-08-18
fix(security): create the master key file with owner-only permissions

Yigtwxx:
## Description

`secret_store` persists the Fernet master key to `SECRET_DIR/.master_key`, and the module docstring states the contract in its own words: *"Persisted to `SECRET_DIR/.master_key` with mode `0o600` (fallback)."*

`_write_key_file` does not create it that way:

```python
def _write_key_file(key_hex: str) -> None:
    path = _master_key_file()
    path.parent.mkdir(parents=True, exist_ok=True)
    path.write_text(key_hex, encoding="utf-8")
    try:
        os.chmod(path, 0o600)
    except OSError:
        pass
```

`write_text()` creates the file with `0o666 & ~umask` — `0o644` under the common default — and the mode is only corrected afterwards. Two consequences follow:

1. **The key is on disk, complete, and world-readable before the mode is tightened.** Any local user can read the master key inside that window, and a crash, `SIGKILL` or power loss between the two calls leaves the file at `0o644` permanently. The RED transcript below captures the state at the moment `chmod` is invoked: mode `0o644`, all 64 hex characters already present.
2. **The final mode depends on a call the module itself is prepared to see fail.** The `except OSError: pass` exists precisely for filesystems without POSIX chmod semantics; on those the key simply stays world-readable, and nothing is logged.

The parent directory has the same shape: `mkdir(parents=True, exist_ok=True)` creates `SECRET_DIR` as `0o755`, so the directory holding the master key is world-readable and world-traversable.

What makes this a defect rather than a preference is that **this module is the outlier in its own repository.** Three sibling secret writers already do it correctly:

- `backup/_utils/signing/key.py:55-80` — `os.open(..., O_CREAT|O_EXCL|O_WRONLY[|O_NOFOLLOW], 0o600)`, parent tightened to `0o700`
- `envs/store.py:51-53` — `_prepare_secret_parent()`: `mkdir` + `chmod 0o700`
- `app/auth.py:83-93` — the same helper pair

This change makes `_write_key_file` create the file *with* its mode instead of correcting it afterwards: `os.open(path, O_CREAT|O_WRONLY|O_TRUNC, 0o600)`, the parent created as `0o700`, and the existing best-effort `chmod` kept so that a key file (or directory) left behind by an older version is still tightened on the next write.

**Deliberately not included, offered as a follow-up:** `signing/key.py` also refuses to write through a symlink (`O_NOFOLLOW`) and creates with `O_EXCL`. Both are worth having here, but `O_NOFOLLOW` converts a currently-silent case into an `OSError` raised out of `_get_master_key()` — a behaviour change that deserves its own PR and its own discussion rather than riding along with a permissions fix. Happy to open it if you want it.

Impact is POSIX-only; Windows ACLs are unaffected either way, and the new permission assertions are skipped there.

**Related Issue:** none

**Security Considerations:** this is the change. The master key decrypts every stored provider API key and auth secret (`encrypt`/`decrypt` in this module), so it is the one file under `SECRET_DIR` that must never be readable by another local account — not even briefly.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

On the first box, honestly: I ran `pre-commit run --files` against the two changed files rather than `--all-files`. An `--all-files` run on my checkout rewrites line endings across roughly a hundred untouched files, which would bury the diff. The changed-file run is included below and every hook passes.

## Testing

Four tests were added to `tests/unit/security/test_secret_store.py`. The three permission tests are skipped on Windows (`os.name != "posix"`) and pin a `0o022` umask so the assertions are deterministic:

- `test_key_is_not_world_readable_before_chmod_runs` — spies on `os.chmod` and records the on-disk state at the moment it is called. That is the window itself: the test asserts the file is already `0o600` when `chmod` runs, and that the full key is already written.
- `test_key_is_owner_only_when_chmod_cannot_help` — neutralises `chmod` for the key file only, which is the case the module's own `except OSError: pass` anticipates, and asserts the file is still `0o600`.
- `test_secret_dir_is_owner_only` — asserts the freshly created `SECRET_DIR` is `0o700`.
- `test_key_file_is_written_and_replaced` — the over-correction guard, and the only one of the four that runs on every platform: writing must still persist the key and must still replace an existing one. It passes before and after the fix; a "fix" that tightened permissions by failing to write would break it.

## Evidence

The proof is the RED run. With the source change reverted and only the new tests applied, the three permission tests fail while the content guard still passes:

```
$ git stash push src/qwenpaw/security/secret_store.py
$ python -m pytest tests/unit/security/test_secret_store.py -q \
    -k "MasterKeyFilePermissions or MasterKeyFileContent"

FAILED tests/unit/security/test_secret_store.py::TestMasterKeyFilePermissions::test_key_is_not_world_readable_before_chmod_runs
FAILED tests/unit/security/test_secret_store.py::TestMasterKeyFilePermissions::test_key_is_owner_only_when_chmod_cannot_help
FAILED tests/unit/security/test_secret_store.py::TestMasterKeyFilePermissions::test_secret_dir_is_owner_only
3 failed, 1 passed, 22 deselected in 1.51s
```

The assertion values say it plainly — `420` is `0o644`, `384` is `0o600`, `493` is `0o755` and `448` is `0o700` — and `st_size=64` shows the whole key is already on disk while the mode is still world-readable:

```
>       assert observed["mode"] == 0o600
E       assert 420 == 384

>       assert stat.S_IMODE(key_path.stat().st_mode) == 0o600
E       AssertionError: assert 420 == 384
E        +  where 420 = S_IMODE(33188)
E        +    and 33188 = os.stat_result(st_mode=33188, st_nlink=1, st_size=64, ...).st_mode
E        +      where stat = PosixPath('.../secrets/.master_key').stat

>       assert stat.S_IMODE(secret_dir.stat().st_mode) == 0o700
E       AssertionError: assert 493 == 448
```

With the fix applied the same command passes, and so does the rest of the security suite:

```
$ python -m pytest tests/unit/security/test_secret_store.py -q \
    -k "MasterKeyFilePermissions or MasterKeyFileContent"
4 passed, 22 deselected in 0.72s

$ python -m pytest tests/unit/security/test_secret_store.py -q
26 passed in 1.41s

$ python -m pytest tests/unit/security/ -q
702 passed, 1 skipped in 12.70s
```

```
$ pre-commit run --files src/qwenpaw/security/secret_store.py \n    tests/unit/security/test_secret_store.py

```

## Additional Notes

`_chmod_best_effort` is a third private copy of a helper that `envs/store.py` and `app/auth.py` already define independently. I kept it local rather than introducing a shared utility, because hoisting it would mean touching three modules for a one-file fix — say the word if you would prefer the shared version instead.

Environment note: the behaviour under test is POSIX-only, so the runs above were done on Linux (Ubuntu, Python 3.12) rather than on my usual Windows checkout. The YAML/JSON/TOML hooks report "no files to check" for a two-Python-file change, and the `prettier` and `actionlint` hooks do not apply to `.py` files, so the list above is every hook that had something to say.

### DIFF
--- a/src/qwenpaw/security/secret_store.py
+++ b/src/qwenpaw/security/secret_store.py
@@ -269,14 +269,22 @@ def _read_key_file() -> Optional[str]:
     return None
 
 
+def _chmod_best_effort(path: Path, mode: int) -> None:
+    try:
+        os.chmod(path, mode)
+    except OSError:
+        # Some systems/filesystems do not support chmod semantics.
+        pass
+
+
 def _write_key_file(key_hex: str) -> None:
     path = _master_key_file()
-    path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(key_hex, encoding="utf-8")
-    try:
-        os.chmod(path, 0o600)
-    except OSError:
-        pass
+    path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
+    _chmod_best_effort(path.parent, 0o700)
+    fd = os.open(path, os.O_CREAT | os.O_WRONLY | os.O_TRUNC, 0o600)
+    with os.fdopen(fd, "w", encoding="utf-8") as handle:
+        handle.write(key_hex)
+    _chmod_best_effort(path, 0o600)
 
 
 def _generate_master_key() -> str:
--- a/src/qwenpaw/security/secret_store.py
+++ b/src/qwenpaw/security/secret_store.py
@@ -255,13 +255,31 @@ def _chmod_best_effort(path: Path, mode: int) -> None:
         pass
 
 
-def _restrict_key_file(path: Path) -> None:
-    """Tighten a key file an older version left group- or world-readable."""
+def _restrict_key_file(path: Path, key_hex: str) -> None:
+    """Migrate a key file an older version left group/world-readable.
+
+    A valid legacy key is never rewritten by ``_write_key_file``, so
+    this is the only chance to take it out of a world-readable inode.
+    ``chmod`` is not enough on its own: the module treats it as best
+    effort, and wherever it is refused the key would stay readable by
+    every local account indefinitely.  Writing the same key material
+    into a fresh ``0o600`` inode makes the mode a property of a file
+    we created; ``chmod`` stays as the fallback for the case where
+    the replacement itself cannot be written.
+    """
     try:
         mode = stat.S_IMODE(path.stat().st_mode)
     except OSError:
         return
-    if mode & 0o077:
+    if not mode & 0o077:
+        return
+    try:
+        _write_key_file(key_hex)
+    except OSError:
+        logger.warning(
+            "Could not replace the group/world-readable master key"
+            " file; falling back to chmod",
+        )
         _chmod_best_effort(path, 0o600)
 
 
@@ -280,7 +298,7 @@ def _read_key_file() -> Optional[str]:
                     len(content),
                 )
                 return None
-            _restrict_key_file(path)
+            _restrict_key_file(path, content)
             return content
         except (OSError, ValueError):
             logger.warning(
--- a/src/qwenpaw/security/secret_store.py
+++ b/src/qwenpaw/security/secret_store.py
@@ -19,6 +19,8 @@ import hashlib
 import logging
 import os
 import secrets
+import stat
+import tempfile
 import threading
 from pathlib import Path
 from typing import Optional
@@ -245,6 +247,24 @@ def _master_key_file() -> Path:
     return _get_secret_dir() / ".master_key"
 
 
+def _chmod_best_effort(path: Path, mode: int) -> None:
+    try:
+        os.chmod(path, mode)
+    except OSError:
+        # Some systems/filesystems do not support chmod semantics.
+        pass
+
+
+def _restrict_key_file(path: Path) -> None:
+    """Tighten a key file an older version left group- or world-readable."""
+    try:
+        mode = stat.S_IMODE(path.stat().st_mode)
+    except OSError:
+        return
+    if mode & 0o077:
+        _chmod_best_effort(path, 0o600)
+
+
 def _read_key_file() -> Optional[str]:
     path = _master_key_file()
     if path.is_file():
@@ -260,6 +280,7 @@ def _read_key_file() -> Optional[str]:
                     len(content),
                 )
                 return None
+            _restrict_key_file(path)
             return content
         except (OSError, ValueError):
             logger.warning(
@@ -269,22 +290,37 @@ def _read_key_file() -> Optional[str]:
     return None
 
 
-def _chmod_best_effort(path: Path, mode: int) -> None:
-    try:
-        os.chmod(path, mode)
-    except OSError:
-        # Some systems/filesystems do not support chmod semantics.
-        pass
-
-
 def _write_key_file(key_hex: str) -> None:
+    """Persist the master key as an owner-only file.
+
+    The key is written to a freshly created ``0o600`` temporary file and
+    moved into place. Truncating the destination would keep the mode of an
+    existing ``.master_key``, so a file left behind ``0o644`` by an older
+    version would hold the new key while it is world-readable -- and stay
+    that way whenever the best-effort ``chmod`` is refused. Replacing the
+    file makes the mode a property of the inode we created instead.
+    """
     path = _master_key_file()
     path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
     _chmod_best_effort(path.parent, 0o700)
-    fd = os.open(path, os.O_CREAT | os.O_WRONLY | os.O_TRUNC, 0o600)
-    with os.fdopen(fd, "w", encoding="utf-8") as handle:
-        handle.write(key_hex)
-    _chmod_best_effort(path, 0o600)
+
+    fd, tmp_name = tempfile.mkstemp(
+        dir=str(path.parent),
+        prefix=f".{path.name}.",
+        suffix=".tmp",
+    )
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as handle:
+            handle.write(key_hex)
+            handle.flush()
+            os.fsync(handle.fileno())
+        os.replace(tmp_name, path)
+    except BaseException:
+        try:
+            os.unlink(tmp_name)
+        except OSError:
+            pass
+        raise
 
 
 def _generate_master_key() -> str:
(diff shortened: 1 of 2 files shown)


src/qwenpaw/security/secret_store.py
```diff
@@ -269,16 +269,24 @@ def _read_key_file() -> Optional[str]:
...
     except OSError:
+        # Some systems/filesystems do not support chmod semantics.
         pass
 
 
+def _write_key_file(key_hex: str) -> None:
+    path = _master_key_file()
+    path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
+    _chmod_best_effort(path.parent, 0o700)
+    fd = os.open(path, os.O_CREAT | os.O_WRONLY | os.O_TRUNC, 0o600)
```
qbc2016: **[P1] Secure pre-existing master key files before overwriting them**

The `0o600` mode passed to `os.open()` only applies when a new inode is created. If `.master_key` already exists, `O_TRUNC` preserves its existing mode. Therefore, an old or corrupt `0o644` key file remains world-readable while the new master key is written, and it remains `0o644` permanently if the best-effort `chmod` fails.

There is also no migration for a valid legacy `0o644` key file: `_read_key_file()` returns it directly, so `_write_key_file()` is never called and its permissions are never tightened.

Please write the new key to a securely created `0o600` temporary file and atomically replace the destination, and tighten permissions when reading a valid legacy key. The tests should also cover pre-existing valid and corrupt `0o644` key files; the current replacement test starts with a file already created as `0o600`.

qbc2016 (commented):

src/qwenpaw/security/secret_store.py
Yigtwxx: Fixed in dee99237.

`_write_key_file()` no longer opens the destination. The key is written to a fresh `tempfile.mkstemp()` file in the same directory (created `0o600` by the kernel), flushed and fsynced, and then `os.replace()`d into position. The mode is now a property of the inode we created rather than of a `chmod` that is allowed to fail, and a pre-existing `0o644` file is unlinked instead of reused. The temporary file is removed if the write raises.

`_read_key_file()` calls a new `_restrict_key_file()` before returning a valid key, which chmods to `0o600` when any of the `0o077` bits are set. That covers the legacy file that is never rewritten.

For the tests: the chmod-spy case is gone, replaced by cases that start from a `0o644` file laid down the way an older version would have left it — a valid one and a corrupt one — plus one asserting a valid legacy key is tightened on read, and one asserting no `.tmp` file survives. The write cases run under a fixture that turns `os.chmod` into a no-op, so only creation-time modes can pass them.

Verified on WSL Ubuntu / Python 3.12 with `tmp_path` on ext4:

- against the previous commit: `3 failed, 25 passed` (`AssertionError: assert 420 == 384`, i.e. `0o644` vs `0o600`)
- against this commit: `28 passed`
- `pre-commit` (mypy, black, flake8, pylint) clean on both files.

src/qwenpaw/security/secret_store.py
```diff
@@ -245,6 +247,24 @@ def _master_key_file() -> Path:
...
+
+
+def _restrict_key_file(path: Path) -> None:
+    """Tighten a key file an older version left group- or world-readable."""
+    try:
+        mode = stat.S_IMODE(path.stat().st_mode)
+    except OSError:
+        return
+    if mode & 0o077:
+        _chmod_best_effort(path, 0o600)
```
qbc2016, line 283: **Please do not rely on best-effort chmod for valid legacy keys**

The overwrite path is now secure, but the valid legacy-key migration still depends entirely on `_chmod_best_effort()`. If `chmod` fails or has no effect, a valid legacy `0o644` key remains world-readable indefinitely—the same failure mode this PR is intended to eliminate.

I reproduced this on the latest commit by pre-creating a valid `0o644` key, making `os.chmod` a no-op, and calling `_read_key_file()`; the key was returned successfully but its mode remained `0o644`.

Please migrate an insecure valid key by atomically replacing it with a fresh `0o600` inode containing the same key material. The corresponding regression test should also run with `_no_chmod`; currently only the write-path tests use that fixture.

qbc2016 (commented):

src/qwenpaw/security/secret_store.py
Yigtwxx: Fixed in baf0855c.

`_restrict_key_file()` no longer tries to chmod the legacy inode. It now takes the key material it just read and hands it to `_write_key_file()`, which is already the atomic-replace path from the previous round: a fresh `tempfile.mkstemp()` file created `0o600` by the kernel, fsynced, then `os.replace()`d over the destination. The world-readable inode is unlinked rather than repaired, so the resulting mode does not depend on a call the module is prepared to see fail.

`_chmod_best_effort()` stays as the fallback for the one case that is left — the replacement itself cannot be written (read-only filesystem, no space) — and that branch now logs instead of passing silently. `OSError` is the only thing caught there, so the read still returns the key either way.

The migration only fires when `mode & 0o077` is set; a key that is already `0o600` is read without being touched.

On the tests: `test_reading_a_valid_legacy_key_migrates_it` now runs under `_no_chmod`, exactly as you asked, and asserts the inode number changed and the key material survived. `test_owner_only_key_is_read_without_being_rewritten` is the over-correction guard for the `0o600` case. The `_legacy_key_file()` helper also asserts its own `0o644` precondition now, since under `_no_chmod` its `os.chmod` call is a no-op and the mode comes from the pinned umask.

Verified on WSL Ubuntu / Python 3.12, `tmp_path` on ext4:

```
$ git stash push src/qwenpaw/security/secret_store.py
$ python -m pytest tests/unit/security/test_secret_store.py -q
FAILED tests/unit/security/test_secret_store.py::TestMasterKeyFilePermissions::test_reading_a_valid_legacy_key_migrates_it
1 failed, 28 passed in 1.80s

>       assert stat.S_IMODE(key_path.stat().st_mode) == 0o600
E       AssertionError: assert 420 == 384
```

`420` is `0o644`, `384` is `0o600` — the reproduction you described.

```
$ python -m pytest tests/unit/security/test_secret_store.py -q
29 passed in 1.00s

$ python -m pytest tests/unit/security/ -q
705 passed, 1 skipped in 6.41s
```

`pre-commit` on both changed files: check python ast, check docstring is first, fix python encoding pragma, detect private key, trim trailing whitespace, add-trailing-comma, mypy, black, flake8, pylint — all Passed.

PR #7135, opened 2026-08-19
fix(envs): preserve corrupt files and write envs atomically

### LINKED ISSUE #7118: [Bug]: a corrupt envs.json is swallowed silently, then overwritten — every stored env var is lost
## QwenPaw Version

`main` @ `38006dda` (2026-08-18).

## Description

A single unparseable byte in `envs.json` silently discards every environment variable stored in it, and the next write commits that loss to disk.

`load_envs()` ends with this handler pair (`src/qwenpaw/envs/store.py:172-179`):

```python
    except (json.JSONDecodeError, ValueError):
        pass
    except OSError as exc:
        logger.warning(
            "Failed to read envs.json from %s due to OS error: %s",
            path,
            exc,
        )
    return {}
```

The `OSError` branch tells the user something went wrong. The corrupt-content branch does not: `load_envs()` returns `{}`, indistinguishable from "the user has not set any env vars yet".

That empty dict is then used as the base for the next write. `set_env_var()` (`:222-230`) does a read-modify-write:

```python
def set_env_var(key: str, value: str) -> dict[str, str]:
    envs = load_envs()
    envs[key] = value
    save_envs(envs)
    return envs
```

and `save_envs()` (`:198-220`) truncates the file with `open(path, "w")` and dumps that one-entry dict. So setting one variable after the file has been corrupted deletes all the others — API keys included — and the command reports success. `delete_env_var()` takes the same path.

The corruption itself does not have to be exotic: a partial write from a crash or an out-of-disk condition during `json.dump`, an editor saving the file with a BOM or a trailing comma, or a half-synced cloud-drive copy of the secret directory are enough. There is also no backup: `save_envs()` writes in place with no `.bak` and no atomic replace, so the previous content is gone the moment the truncating `open()` succeeds.

A related detail on the same path: `load_envs_into_environ()` is called during package import (`src/qwenpaw/__init__.py`), so on a corrupt file the process starts with none of the persisted variables in `os.environ` and nothing in the log explaining why the provider keys vanished.

**Related PR(s):** none

**Security considerations:** the values in this file are provider API keys and tokens (encrypted at rest via `secret_store`), so the failure mode is silent credential loss rather than exposure.

## Component(s) Affected

- [ ] Skills
- [x] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** `main` @ `38006dda`
- **OS:** reproducible on any platform (read via `open()`/`json.load`, no OS-specific behaviour)
- **Python version (if applicable):** 3.12

## Steps to Reproduce

1. Store two or more env vars, e.g. `set_env_var("A", "1")` and `set_env_var("B", "2")`, so `envs.json` holds both.
2. Corrupt the file the way a partial write would, e.g. truncate its last byte or write `{"A": "enc:...",` into it.
3. Set any one variable: `set_env_var("C", "3")`.
4. Read the file, or call `load_envs()`.

## Actual vs Expected

- **Actual:** `envs.json` now contains only `C`. `A` and `B` are gone, no warning was logged, and the call returned normally.
- **Expected:** the write refuses to run on a file that could not be parsed, or the unreadable file is preserved (e.g. moved aside as `envs.json.corrupt`) before anything is rewritten — and in either case the corrupt read is logged at least as loudly as the `OSError` next to it.

## Logs / Screenshots

Nothing is logged on the corrupt path — that is the point of the report. The neighbouring `OSError` branch, by contrast, produces `WARNING | Failed to read envs.json from <path> due to OS error: <exc>`.

Run of the steps above against `main` @ `38006dda` (Python 3.12, `QWENPAW_SECRET_DIR` pointed at a scratch directory), truncating the last three bytes to imitate a partial write:

```
path: /tmp/envtest/envs.json
after two writes: ['A', 'B']
load after corruption: {}
file now: ['C']
```

The third line is the silent part, and the fourth is the loss: `A` and `B` are no longer on disk, and `set_env_var("C", "3")` returned normally.

## Additional Notes

Before opening a PR I would rather agree on the contract with you, because every option changes behaviour in a different direction:

1. **Log and preserve** — log a warning, move the unparseable file aside as `envs.json.corrupt`, then continue with `{}`. Nothing starts failing that works today, and the data survives.
2. **Log and refuse to write** — `load_envs()` still returns `{}`, but `save_envs()` raises when the on-disk file exists and is unparseable, so a corrupt file can never be silently overwritten.
3. **Fail loudly on read** — raise out of `load_envs()`. Truest to the data, but it would break startup via `load_envs_into_environ()`, which is called at import time.

My preference is option 1 plus the warning, with option 2 as a follow-up if you want the stronger guarantee — but this is your call. `tests/unit/envs/` does not exist yet, so whichever way it goes I would add the module's first tests alongside the fix.

I am happy to send the PR once you tell me which contract you want.

xieyxclack:
## Description

[Describe what this PR does and why]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

## Testing

[How to test these changes]

## Evidence

Examples of valid evidence:
- CI artifact link

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]


cuiyuebing (approved):
LGTM

PR #7147, opened 2026-08-19
fix(xiaoyi): report proactive send failures and normalize session IDs

hongxicheng:
## Summary

- Normalize `xiaoyi:`-prefixed session IDs before looking up XiaoYi task
  mappings.
- Propagate proactive-send failures when the channel is disconnected,
  no task mapping exists, or both WebSocket transports fail.
- Log API-initiated delivery failures at ERROR level.
- Preserve warning-only behavior for normal conversation replies.

## Root cause

`ChannelManager.send_text()` resolved the native XiaoYi handle but also
placed the original prefixed session ID into `meta`.

XiaoYi preferred `meta["session_id"]` when looking up the task mapping,
while `_session_task_map` was registered using the native session ID
without the `xiaoyi:` prefix.

As a result, the lookup returned no `task_id`, the message was skipped,
and the API still returned success because no exception propagated to
the caller.

## Fix

- Strip the `xiaoyi:` prefix before task-map lookup.
- Return the WebSocket send result from the XiaoYi transport helpers.
- Raise `ChannelError` for API-initiated sends when:
  - the channel is disconnected;
  - no `task_id` is available;
  - neither WebSocket connection accepts the message.
- Keep normal conversation delivery tolerant so transient channel
  failures do not interrupt the agent processing flow.

This follows the existing DingTalk and WeChat behavior: explicit
API/CLI sends propagate delivery failures, while normal conversation
replies remain best-effort.

## Testing

- `75 passed` across XiaoYi channel and messages-router unit tests.
- Pre-commit checks passed, including AST validation, mypy, Black,
  Flake8, and Pylint.
- `git diff --check` and Python compilation checks passed.
- XiaoYi WebSocket connection integration coverage passed.

The existing message round-trip integration case remains blocked locally
by a pre-existing dependency mismatch: the source requires AgentScope
`2.0.6`, while `uv.lock` and the local base environment contain
`2.0.4.post1`. This mismatch was not introduced by this change.

### DIFF
--- a/src/qwenpaw/app/channels/xiaoyi/channel.py
+++ b/src/qwenpaw/app/channels/xiaoyi/channel.py
@@ -38,6 +38,7 @@ from ....schemas import (
 
 from ....config.config import XiaoYiConfig as XiaoYiChannelConfig
 from ....constant import DEFAULT_MEDIA_DIR
+from ....exceptions import ChannelError
 from ..renderer import ChannelDisplayConfig
 from ..base import (
     BaseChannel,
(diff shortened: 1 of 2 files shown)


cuiyuebing (approved):
LGTM

PR #7175, opened 2026-08-20
fix(console): restore complete free model listings

wangfei010313:
## Summary

Fix the Free model selector so it consistently reflects maintained provider
metadata and current provider availability.

## Changes

- Keep free-tier providers available in both the Free and Pro tabs.
- Show every configured free model in the Free tab instead of limiting the
  list to active, recent, or recommended models.
- Keep paid models out of the Free tab.
- Display the Free badge for the active free model on the selector trigger.
- Keep advanced discovery controls out of the default release UI.

### Kilo Code

- Prefer the gateway `isFree` / `is_free` field when determining whether a
  model is free.
- Fall back to the provider-specific model ID suffix when the API does not
  return an explicit free flag.
- Correctly classify `kilo-auto/free`, which does not use the `:free` suffix.
- Keep the following maintained free models in the Kilo catalog:
  - `kilo-auto/free`
  - `nvidia/nemotron-3-ultra-550b-a55b:free`
  - `nvidia/nemotron-3-super-120b-a12b:free`
  - `stepfun/step-3.7-flash:free`
- Replace outdated Poolside model IDs and remove the unavailable free Nex ID.

### OpenCode

- Remove `deepseek-v4-flash-free` because its free promotion has ended.
- Remove `nemotron-3-super-free` because OpenCode no longer supports it.
- Filter both unavailable IDs from dynamic `/models` discovery.
- Remove these IDs from restored persisted discovery and user-model caches so
  they do not return after an application restart.

### Persistence

- Treat maintained catalog `is_free` values as authoritative when restoring
  built-in providers.
- Prevent stale persisted metadata from changing free models into paid models
  or paid models into free models.
- Preserve explicit free-model values from catalog overlay files.

## Testing

- Provider and catalog tests: `192 passed, 1 skipped`
- ModelSelector tests: `48 passed`
- Focused regression tests after the pylint fix: `52 passed`
- TypeScript project check passed
- Python compilation check passed
- Pylint passed with `10.00/10`
- Black and `git diff --check` passed
- Catalog JSON validation passed
- Vite production build passed using an isolated output directory
局部截取_20260821_121548


XiuShenAl (approved):
LGTM

PR #7183, opened 2026-08-20
feat(skills): add workspace-scoped preload configuration

### LINKED ISSUE #7182: [Feature]: Add workspace-scoped Skill preload policy
## Summary

Most Skills should stay on demand. But when a workspace is built around a specific Skill, making the model rediscover it in every new conversation adds an avoidable first-turn tool call.

This proposal adds a workspace-scoped load policy:

- `on_demand` by default;
- `preload` for Skills that are part of the agent's core setup.

## Component(s) Affected

- [x] Skills
- [x] Documentation (website)
- [x] Tests

## Problem / Motivation

Skill Viewer works well when the model needs to discover an optional capability. It is less useful when the workspace already knows that a Skill is required. In that case, the model first calls Viewer, receives the instructions, and only then starts the task.

For a specialized agent, the defining Skill should be available before that first decision.

## Proposed Solution

Add `load_policy: "on_demand" | "preload"` to each workspace Skill.

Preloaded Skills still follow the existing enabled and channel rules. They load through the current runtime cache before inference. A Skill leaves Viewer only after it loads successfully, so the on-demand path remains available as fallback.

The policy is configured per workspace and follows the Skill through edits, renames, and pool updates. The Console exposes it as a simple preload toggle.

## Alternatives Considered

- Moving the instructions into the workspace prompt duplicates the Skill.
- `/skill <input>` works when the user explicitly selects a Skill, but it does not configure the agent itself.
- Calling Viewer automatically still keeps the extra first-turn round trip.

## Additional Context

Claude Code already uses this split: Skills are normally loaded on demand, while Skills selected for a specialized subagent are fully preloaded; other Skills remain available through the Skill tool ([docs](https://code.claude.com/docs/en/sub-agents#preload-skills-into-subagents)).

In a local 5×5 cold comparison, preload removed all five initial Viewer calls and reduced median final-answer start from 8.705 s to 4.758 s.

Implementation: #7183

## Willing to Contribute

- [x] Implemented in #7183.

wuyak:
## Description

Most Skills should remain on demand. This PR adds an opt-in, workspace-scoped `preload` setting for trusted core or frequently used Skills.

The design follows the same high-level split documented for [Claude Code subagents](https://code.claude.com/docs/en/sub-agents#preload-skills-into-subagents), adapted to QwenPaw's existing Skill lifecycle.

QwenPaw loads the effective runtime Skills once, preserves the existing workspace-over-tool-bound precedence, matches preload candidates by resolved workspace directory, and keeps the existing runtime `Skill.name` overwrite behavior. Successfully preloaded Skill content is added to the system prompt, and those Skills are not registered with Skill Viewer. Other Skills retain the existing on-demand behavior.

Natural-language Skill selection, Skill name mentions, and explicit slash invocation remain unchanged.

The setting is stored in the existing workspace manifest and is available through the Skills API and Console. Console localization and the English and Chinese documentation are included.

**Related Issue:** Fixes #7182

**Security Considerations:** Preload is explicit, disabled by default, and intended only for trusted Skills selected by the agent owner.

## Type of Change

- [x] New feature
- [x] Documentation

## Component(s) Affected

- [x] Core / Backend
- [x] Console
- [x] Skills
- [x] Documentation

## Checklist

- [ ] `pre-commit run --all-files` passes on the final head
- [x] Relevant local validation passes
- [x] Documentation updated
- [x] Ready for review

## Testing

Local validation covered preload selection and runtime-name overwrite, manifest and API persistence across reload and restart, preload and on-demand natural-language routing, unchanged slash behavior, and Console configuration. Frontend type checking, formatting, production build, documentation formatting, and `git diff --check` also passed.


Leirunlin:
Thanks for the PR. The overall direction looks good. Before merging, please address the following:

1. **Deduplication**

The PR already attempts to deduplicate the preloaded and Viewer Skill sets, but it uses `Path(skill.dir).name` as the identity. That only compares directory basenames and is not sufficient for localized tool-bound directories. 
Here’s the tricky part: even though deduplication is based on the directory path, the `name` field inside `skill.md` can still collide across different directories, making the approach unreliable.

```text
workspace directory:  browser
runtime Skill.name:   browser

tool-bound directory: browser-en
runtime Skill.name:   browser
```

If the workspace `browser` Skill is preloaded, `browser-en` can still be registered in the Viewer because the directory basenames differ even though both objects have the same runtime Skill name.

So it is recommended to preserve the existing workspace-over-tool-bound precedence in `_load_runtime_skills()` and load the effective runtime Skills only once. Use resolved workspace paths to determine preload membership, then deduplicate the successfully loaded resident Skills by their runtime `Skill.name`. 
When two resident Skills have the same runtime name, the later loaded value should overwrite the earlier one, matching AgentScope's behavior.

An example could be like:

```python
### Example
workspace_skills_dir = get_workspace_skills_dir(Path(skills_workspace))
preload_dirs = {
    (workspace_skills_dir / name).resolve(strict=False)
    for name in preload_candidates
}

loaded_skills = await run_sync_io(
    self._load_runtime_skills,
    effective_skill_names,
    workspace_dir,
    tools,
)

resident_skills_by_name = {}
loaded_nonresident_skills = []

for skill in loaded_skills:
    skill_dir = Path(skill.dir).resolve(strict=False)
    if skill_dir in preload_dirs:
        # Match AgentScope: a later value replaces an earlier value with
        # the same runtime Skill name.
        resident_skills_by_name[skill.name] = skill
    else:
        loaded_nonresident_skills.append(skill)

loaded_preloaded_skills = list(resident_skills_by_name.values())

viewer_skills = [
    skill
    for skill in loaded_nonresident_skills
    if skill.name not in resident_skills_by_name
]
```
You can also try more local tests to ensure the deduplication/overwrite behavior works normally.

2. **Clarify the slash-Skill injection semantics**

This PR also refactors the existing `/<skill_name> <input>` injection path. Could you clarify the intended model-facing behavior and whether the following differences are intentional?

- The old nested payload is replaced by the shared JSON-based `render_skill_context()` format.
- Ordinary preload uses one common introduction and one separate `instructions` payload per resident Skill.
- An on-demand slash invocation still injects the full body, but **the previous explicit direction to apply the Skill to the user's request is no longer present**.
- A successful preloaded slash invocation keeps the body in the system prompt and adds a metadata-only marker to the user message. With `include_instructions=False`, the prepended direct-invocation guidance is omitted together with the body.
- A failed preload falls back to one full inline injection. `/<skill_name>` without input and the separate `/skills` listing appear unchanged.

3. **Complete the Console localization (Optional)**

The Console currently registers `en`, `zh`, `ru`, `ja`, `pt-BR`, `id`, and `vi`. Recent Console features normally update all supported locale files. 

The Skills documentation itself is maintained in English and Chinese, so the current documentation language coverage is appropriate.

4. **Format**

- Export and reuse `SkillLoadPolicy` rather than repeating `Literal["on_demand", "preload"]` in router models or using a less specific `str` in service signatures. 
- Avoid calling `resolve_effective_skills` more than once in either the builder or slash-command path. Pass the already resolved effective Skill names to the preload-policy filter instead. Renaming that helper to something such as `resolve_preload_candidates(workspace, effective_skill_names)` would make its narrower responsibility clearer.
- Keep using the shared `render_skill_context()` implementation for both preload rendering and the slash command.

5. **Keep test-only changes outside the merge diff**

Please keep and run the feature-specific tests locally or on a separate validation branch, and report the commands and results in the PR. Remove newly added test files and other test-only changes from the final diff before merging into `main`.

6. **Documentation**

`website/public/docs/skills.en.md` and `website/public/docs/skills.zh.md` are the correct locations. Please use the existing **Workspace → Skills → Edit** / **工作区 → 技能 → 编辑** terminology consistently. Also describe the injected content as structured and delimited, and make it explicit that only trusted Skills should be preloaded.

wuyak:
Thanks for the detailed review, @Leirunlin. It prompted me to revisit the Skill registration and loading lifecycle end to end, and led me to simplify the design around the existing lifecycle.

1. **Deduplication**

   Effective runtime Skills are now loaded once while preserving the existing workspace-over-tool-bound precedence. Preload candidates are matched by resolved workspace directory, and successfully preloaded Skills follow the existing runtime `Skill.name` overwrite behavior, including later-value replacement. Successfully preloaded Skills are not registered with Skill Viewer.

2. **Slash behavior**

   I removed the slash-path refactor and kept the current upstream behavior.

   For natural-language requests, preloaded instructions are available directly from the system prompt, whether or not the request mentions the Skill name. Live validation confirmed that these requests do not invoke Skill Viewer.

   `/name` without input still shows the Skill description, while `/name <input>` keeps the existing full-body injection.

   I considered making explicit slash invocation aware of preload state. In practice, preload already makes the Skill available to natural-language requests, so explicitly invoking the same Skill again is a narrow, intentional path. Supporting deduplication only for this case would require coupling preload classification with slash injection, which happen at different lifecycle stages. QwenPaw also does not currently deduplicate repeated explicit slash invocations. I therefore kept the existing slash behavior unchanged.

3. **Localization and documentation**

   The preload label and help text are now included in all supported Console locales. The English and Chinese documentation use **Workspace → Skills → Edit**, limit preload to trusted core or frequently used Skills, and explain that the full content is added to the system prompt as a structured Skill block.

4. **API and lifecycle shape**

   The revised implementation uses `preload: bool`, so the previous repeated load-policy `Literal` is no longer needed. `resolve_effective_skills()` runs once, and its result is passed to the narrower `select_preload_skills()` helper.

   I also removed the slash renderer introduced by the previous version. Preload rendering is handled by `PreloadedSkillsContributor`, while the existing slash implementation remains unchanged.

5. **Test-only changes and validation**

   The final diff contains no newly added test files or other test-only changes. Local validation covered preload selection and runtime-name overwrite, manifest and API persistence, reload and restart behavior, preload and on-demand natural-language routing, unchanged slash invocation, and Console configuration.

I’d really appreciate it if you could take another look. Thanks again for the thoughtful review.

Leirunlin:
LGTM. The current behavior and scope look good.

A few non-blocking follow-ups for later:
  - UI improvement.
  - Make prompt assembly more cache-friendly.
  - Consider a dedicated “Preloaded” badge label.
  - Add stronger content delimiters or escaping for preloaded Skill bodies.
  - Consider a token-budget warning when many or very large Skills are preloaded.

wuyak:
Thanks for the review and the thoughtful follow-up suggestions. For this PR, I kept the UI and prompt structure aligned with the existing QwenPaw patterns, since both need to remain consistent across the product. The suggested improvements make sense to revisit later as broader, coordinated changes. Thanks again!

wuyak:
Thanks for the review and the thoughtful follow-up suggestions. For this PR, I intentionally kept the changes as minimal and scoped as possible, while preserving consistency with QwenPaw’s existing UI, prompt structure, and behavior. For broader changes in these areas, I think it makes sense to align on the overall design with the maintainers before expanding the implementation.

Leirunlin:
@wuyak Could you please rebase to the latest main branch? The tests are currently failing, and from what I can see, it looks like there are some version gaps causing the issue.

wuyak:
I've rebased the branch onto the latest `main`. Could you please take another look?

wuyak:
All completed checks have passed, but the final AI review check still requires approval. Since `main` has advanced again, should I rebase once more or leave the current head as is?

Leirunlin:
@wuyak You don't have to. We will merge this PR in the next version. Thank you for your patience.

wuyak:
Got it, thank you! Have a great day!

lalaliat (approved):
LGTM

PR #7194, opened 2026-08-21
fix(workspace): make startup failure cleanup cancellation-safe

jinliyl:
## Description

Make workspace startup and reload-candidate cleanup cancellation-safe without leaving partially constructed, started, or borrowed services behind.

The lifecycle rules are now:

- an uncommitted reload candidate receives final cleanup while preserving reusable services borrowed from the active workspace;
- workspace startup cleans up after ordinary failures and cancellation, even before `_started` becomes true;
- concurrent startup cancels and joins sibling tasks on the first failure;
- synchronous constructors, `start()`, and `stop()` finish their worker-thread work before cancellation propagates;
- async factories publish ownership before their first cancellable await;
- an optional service is removed only after cleanup succeeds; cleanup failures abort startup and remain retryable;
- strict stop failures propagate, preventing a workspace from reporting a false stopped state;
- generic awaitables are recognized with `inspect.isawaitable()`;
- the borrowed-service cleanup log sanitizes `agent_id`.

## Scope after review cleanup

The implementation reuses the existing cancellation-safe `run_sync_io()` helper instead of maintaining a second local implementation. Repeated workspace setup and lifecycle cases in the tests were consolidated with fixtures and parametrization.

Relative to the PR base:

- production: `+306 / -115` (net `+191`);
- tests: `+407 / -16` (net `+391`).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [x] Core / Backend
- [x] Tests

## Testing

Focused lifecycle, reload, startup, and cancellation-helper suite:

```bash
pytest -q tests/unit/app/workspace \
  tests/unit/app/test_multi_agent_reload_guard.py \
  tests/unit/app/test_multi_agent_manager_startup.py \
  tests/unit/utils/test_io_utils.py
```

```text
60 passed in 2.29s
```

Mail factory compatibility tests:

```bash
PYTHONPATH=packages/qwenpawmail-mcp/src pytest -q \
  tests/unit/app/mail/test_service_factories_guard.py \
  tests/unit/app/mail/test_service_factories_seed_files.py
```

```text
6 passed in 0.34s
```

All pre-commit hooks on the eight changed files pass, including mypy, Black, Flake8, and Pylint. `git diff --check` also passes.

## Security Considerations

The configurable `agent_id` is sanitized before interpolation in the new borrowed-service cleanup log.

## Checklist

- [x] Focused tests pass
- [x] Pre-commit checks on changed files pass
- [x] Documentation not needed for this internal lifecycle fix
- [x] Ready for review

### DIFF
--- a/src/qwenpaw/app/multi_agent_manager.py
+++ b/src/qwenpaw/app/multi_agent_manager.py
@@ -23,8 +23,9 @@ from ..constant import (
     CUSTOM_AGENT_STARTUP_CONCURRENCY,
 )
 from ..config.utils import load_config
-from ..utils.startup_display import AgentStartupDisplay
+from ..utils.io_utils import run_async_to_completion
 from ..utils.logging import sanitize_log_value
+from ..utils.startup_display import AgentStartupDisplay
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


rayrayraykk (approved):
LGTM

cuiyuebing (approved):
LGTM

PR #7319, opened 2026-08-26
refactor(console): track background agent runs

### LINKED ISSUE #2814: [Bug]: Multi-agent chat history is empty for running callee agent

zhijianma:
## Summary

- route `/console/chat/task` runs through the workspace TaskTracker instead of bypassing it
- keep task-id polling compatible while enabling tracker status, stop, reload ownership, and Console reconnect for background `submit_to_agent` and `spawn_subagent` runs
- preserve timeout, cancellation, producer-error, and fork-finalization results

## Tests

- `128 passed` across the targeted TaskTracker, Console task/reconnect, and agent-management suites
- pre-commit passed for all changed files (AST, mypy, black, flake8, pylint, and related hooks)


cuiyuebing (approved):
LGTM

Issue #7288, opened 2026-08-25
[Bug]: Large MCP Results in Active Turns Can Bypass Scroll Compaction and Overflow Model Context


qbasic45:
## QwenPaw Version

`2.1.1b1`

## Description

In enterprise data analysis and operations scenarios, agents commonly use MCP to query time-series data, equipment telemetry, detailed reports, or bulk logs.

For example, a user may ask an agent to query operational data from multiple sites over a time range, analyze anomalous trends, and generate a daily report. Such MCP calls can return large JSON payloads, table rows, HTML reports, or log excerpts in a single response.

When an oversized MCP tool result belongs to the current active turn or the newest protected tool results, it can be included in the next model request unchanged. Scroll can fold some older tool results from completed turns, but it preserves the complete active turn, recent tool results, and the current tool-call chain. Retrospective compaction therefore cannot guarantee that the next model input remains within a safe budget.

**Observed behavior:** After an oversized MCP result entered the active context, Scroll compaction could not reduce the input below its trigger. Context-overflow recovery did not change the model input, and the upstream provider rejected the request. The agent could not continue the task.

**Expected behavior:** A single MCP/tool result must not cause the next model request to exceed a configurable hard input budget. Oversized results should be bounded, summarized, or persisted before they enter the active context instead of relying on retrospective compaction.

**Related PR(s):** None

**Security considerations:** This issue does not involve authentication bypass or configuration exposure. Any full-result persistence mechanism should be restricted to the current agent workspace and inherit workspace permissions so that sensitive business data in tool results is not accessible to other agents or users.

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** `2.1.1b1`
- **OS:** openEuler Linux 6.6
- **Install method:** Offline Docker image
- **Docker image:** `qwenpaw-offline:0013979c680c`
- **Python version (if applicable):** `3.11.2`

## Steps to Reproduce

1. Configure an MCP tool that can return a large amount of data, such as a time-series query, SQL detail query, log search, or bulk report reader.
[...]

PR #7331, opened 2026-08-26
fix(context): bound oversized single-line tool results

niceIrene:
## Summary

- bound oversized single-line tool results before they enter the agent context
- preserve the complete result as a workspace artifact
- add artifact-only recovery metadata when line-based continuation cannot advance
- preserve artifact recovery information when truncated previews are compacted again
- add regression coverage for compact JSON and UTF-8 payloads

## Root cause

Fresh tool results are normally truncated before being yielded into the agent
context. However, `_truncate_fresh()` treated an oversized final line as an
unhandled edge case and returned a markerless slice.

`ToolResultPruner.prune_text()` requires a truncation marker before accepting a
candidate. Because the candidate had no marker, it discarded the bounded slice
and returned the complete original result.

This particularly affected compact JSON returned by MCP tools, where the entire
payload is commonly serialized onto one line. A large result could therefore
enter the active turn unchanged and leave Scroll unable to reduce the context.

## Changes

- return a bounded, UTF-8-safe preview for oversized single-line results
- attach standard truncation metadata and a recovery notice
- mark these results with `continuation_mode="artifact"`
- avoid suggesting line-based continuation for content contained in one
  oversized line
- retain artifact-only recovery semantics during subsequent re-truncation
- keep existing multiline line-pagination behavior unchanged

## Before / After

Using a representative compact JSON tool result:

| | Size |
|---|---:|
| Original input | 700,013 bytes |
| Before this fix | 700,013 bytes |
| After this fix | 50,456 bytes |
| Configured preview | 50,000 bytes |

The saved artifact remains byte-for-byte identical to the original payload.

## Testing

- added a middleware-level regression test for oversized single-line UTF-8 JSON
- added a regression test for re-truncating artifact-backed previews
- 24 relevant pruning tests passed
- 114 file I/O and Scroll tests passed
- manual 700 KB MCP-like payload reproduction passed
- all pre-commit hooks passed, including mypy, black, flake8, and pylint

Addresses #7288.

## Scope

This PR fixes the reproducible single-line text/JSON admission-cap bypass.
A response-wide aggregate cap for MCP responses containing many individually
bounded blocks can be handled separately as defense-in-depth.

### DIFF
--- a/src/qwenpaw/agents/tools/utils.py
+++ b/src/qwenpaw/agents/tools/utils.py
@@ -325,12 +325,15 @@ def _truncate_fresh(
     newline_count = result.count("\n")
 
     # Compute the first line number not yet fully included in this chunk.
-    # max(1, ...) prevents next_line from equaling start_line when a single line
-    # exceeds max_bytes (newline_count == 0), which would make the caller retry
-    # the same range indefinitely.
     next_line = start_line + max(1, newline_count)
 
-    if next_line <= total_lines:
+    if newline_count == 0:
+        # The preview contains no complete line, even when later lines exist
+        # in the source. Advancing to start_line + 1 would silently skip the
+        # unshown remainder of the current oversized line, while repeating
+        # start_line cannot make progress. Recover from the artifact instead.
+        read_from = None
+    elif next_line <= total_lines:
         # Truncation fell before the last line — continue reading from next_line.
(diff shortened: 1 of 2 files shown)


jinliyl (approved):
LGTM

rayrayraykk (approved):

PR #7337, opened 2026-08-27
fix(providers): separate model output capabilities from request limits

wangfei010313:
## Summary

Prevent unknown or automatically discovered model output capabilities from
silently becoming request-level `max_tokens` limits.

This PR separates model output capability metadata from user-configured request
budgets, preserves nullable values across the backend and frontend, and
migrates legacy provider snapshots and catalogs safely.

## Problem

`ModelInfo.max_tokens` previously represented two different concepts:

1. The model's maximum output capability.
2. The output limit sent with each generation request.

It also defaulted to `8192`, so models without a verified output limit were
silently restricted to 8,192 output tokens.

Removing the placeholder values from the packaged catalog was not sufficient:

- Legacy provider snapshots could restore the old value.
- A stale OTA catalog could override the packaged catalog.
- The frontend converted `null` back to `8192` when editing model settings.
- Automatically discovered capability limits could be injected into requests.

### Separate capability from request configuration

- Add `max_output_length` for model capability metadata.
- Add source and update-time metadata for output capabilities.
- Store user request limits only in `generate_kwargs.max_tokens`.
- Stop injecting model capability values into generation requests.
- Omit `max_tokens` when the user has not explicitly configured a limit.
- Keep provider-specific required fallbacks inside their adapters.

### Update discovery and catalog handling

- Store API-discovered output limits as capabilities with `api` provenance.
- Preserve user-configured capability metadata during discovery refreshes.
- Convert verified catalog limits to `max_output_length`.
- Remove 84 unverified `max_tokens: 8192` placeholders.
- Preserve 49 explicitly maintained output capabilities.
- Bump the packaged catalog version to `2026.08.27`.
- Ignore or reject OTA catalogs older than the packaged catalog.
- Preserve local catalog overrides as the highest-priority catalog source.

### Migrate existing installations

- Introduce provider snapshot schema version 2.
- Remove the legacy unverified 8,192 placeholder during one-time migration.
- Move explicit user `max_tokens` overrides into `generate_kwargs`.
- Preserve API-discovered capabilities, including legitimate values of 8,192.
- Rewrite migrated snapshots atomically.

### Preserve nullable frontend behavior

- Represent the request output limit as nullable.
- Do not replace an unknown value with `8192`.
- Submit `max_tokens` only after that field has been edited.
- Support explicitly clearing an override with `null`.
- Add a “Reset to auto” action.
- Display the model output capability and its source separately from the
  request limit.
- Keep `max_tokens` out of the generic JSON generation-config editor.

## Result

| State | Request behavior |
| --- | --- |
| Output capability unknown | Omit `max_tokens` |
| Capability reported by API/catalog | Display it, but do not send it |
| User configures an output budget | Send `generate_kwargs.max_tokens` |
| Provider protocol requires a fallback | Apply it in that provider adapter |

## Test Coverage

- Unknown output limits are omitted from requests.
- Explicit request limits are preserved and can be cleared.
- Legacy snapshots migrate without losing real user or API values.
- Stale OTA catalogs cannot override a newer packaged catalog.
- Local catalog overrides retain priority.
- Discovery refreshes capability metadata without replacing user overrides.
- Editing unrelated model settings does not reintroduce `8192`.
- Frontend reset-to-auto behavior preserves `null`.
- Static typing covers the new callback and migration data structures.


cuiyuebing (approved):
LGTM

Issue #7335, opened 2026-08-27
[Feature]: Prompt cache hit rate observability and optimization — 81% vs 96% (OpenCode), cost impact documented


jamie9797:
## Summary

QwenPaw currently **does not record, display, or optimize** prompt cache hit rate. Users have no visibility into actual cache performance and cannot optimize accordingly. Production data shows QwenPaw hit rate (81.68%) is significantly lower than OpenCode (96.02%), with direct cost impact.

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [x] Channels
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

### 1. Cache data is silently discarded

AgentScope 2.0's ChatUsage already extracts cache fields:

```python
@dataclass
class ChatUsage:
    input_tokens: int
    output_tokens: int
    cache_creation_input_tokens: int = 0  # cache writes
    cache_input_tokens: int = 0           # cache reads (hits)
```

But QwenPaw's `TokenRecordingModelWrapper._record_usage()` only reads `input_tokens` and `output_tokens`, **discarding cache data entirely**:

```python
# token_usage/model_wrapper.py:65-74
pt = getattr(usage, "input_tokens", 0) or 0
ct = getattr(usage, "output_tokens", 0) or 0
# cache_input_tokens and cache_creation_input_tokens are never read
```

→ `token_usage.json` only has `prompt_tokens` / `completion_tokens` / `call_count`, no cache fields.

### 2. Measured hit rate gap is significant

User @tecgic reported in #3891 (2026-06-29):

| Product | Cache Hit Rate |
|---------|---------------|
| **OpenCode** | **96.02%** |
| **QwenPaw** | **81.68%** |

14 percentage points gap. Per Requesty production data (April 2026), this translates to **~2-3x per-call cost difference**.

### 3. Billing evidence

User @wxfvf reported in #3891 (2026-05-06):

> After modifying profile.md, cache hit rate dropped sharply — 10 messages cost more than the previous 200+ messages.

### 4. Session ID breaks cross-session caching

User @rankaiyx discovered in #3891 (2026-06-04):

> In the long system prompt at the start of each session, only the Session ID changes. Moving it to the end of the system prompt would improve cross-session cache hit rate.

### 5. Industry comparison

[...]

PR #7342, opened 2026-08-27
feat(token-usage): add prompt cache observability

rayrayraykk:
## Summary

Adds Stage 1 prompt-cache observability across token recording, APIs, aggregate statistics, and the chat UI.

Related to #7335. This PR intentionally covers observability only; prompt-prefix and provider-level cache optimizations remain separate follow-up work.

## What changed

- Preserve provider cache-read and cache-write token counts in token usage records.
- Normalize cache-eligible input tokens across provider usage formats.
- Expose cache metrics through token usage statistics and per-turn usage events.
- Calculate the token-weighted hit rate as `cache_read_tokens / cache_eligible_input_tokens`.
- Aggregate the current session across every model call, including calls inside ReAct/tool loops.
- Add cache-read, cache-write, and hit-rate metrics to the Token Usage page.
- Add a compact 24 px fused chat indicator: the outer ring shows context-window usage and the center value shows the session cache hit rate.
- Keep the last completed session value visible while the next response streams; clear it when creating or switching sessions.
- Keep a known cold-cache result visible as `0%` instead of hiding it.

## Semantics

The chat value is the token-weighted cumulative hit rate for the current session, not an average of per-request percentages. A new conversation's first visible turn may show a non-zero value when one ReAct turn performs multiple model calls: the first call can be cold while later calls reuse the prefix created earlier in the same turn.

## Validation

- Frontend: 95 tests passed across cache usage, session ownership, streaming, token aggregation, and the fused indicator.
- Backend: 56 token-usage and agent-statistics tests passed.
- TypeScript build and Prettier checks passed.


rankaiyx:
Could you analyze the following projects and draw inspiration from them?
Integrating similar functionality would make prefix breaking extremely intuitive.
https://github.com/co-l/cache-hunter
image

zhijianma (approved):

rayrayraykk:
> Could you analyze the following projects and draw inspiration from them? Integrating similar functionality would make prefix breaking extremely intuitive. https://github.com/co-l/cache-hunter image

Thanks for the information, I will look into this when I have time.

PR #7353, opened 2026-08-27
fix(console): preserve embedding verification across agent switches

jinliyl:
## Summary

- preserve per-agent embedding verification when the selected agent changes before the shared configuration form has finished refreshing
- bind every agent configuration load to both a request generation and the agent selected when the request started, ignoring stale success and failure completions
- continue clearing verification when the same agent is genuinely changed to a disabled embedding configuration
- add focused regression coverage for the mixed-frame agent-switch race, the same-agent disable path, and a late B response after an A → B → A switch

Together with the cancellation-safe failed reload candidate cleanup already merged into `main` in #7194, this resolves both tracked issues:

- https://project.aone.alibaba-inc.com/coco/projects/2155950/workitems?view=f6b7997843844f40a8dc0ef7&workitem=85858155
- https://project.aone.alibaba-inc.com/coco/projects/2155950/workitems?view=f6b7997843844f40a8dc0ef7&workitem=85950158

The first issue is covered by the reusable-service preservation logic and regression tests in the current `main`; this PR supplies the remaining Console fixes for the second issue.

## Testing

- `npx tsc -b --noEmit`
- Prettier checks for all changed TypeScript files
- targeted Vitest suites for agent config and embedding verification (47 tests passed)
- `git diff --check`

The existing `useAgentConfig.test.tsx` ESLint baseline still reports unrelated errors on unchanged legacy lines; this change adds no new ESLint findings.

### DIFF
--- a/console/src/pages/Agent/Config/components/useEmbeddingVerification.test.tsx
+++ b/console/src/pages/Agent/Config/components/useEmbeddingVerification.test.tsx
@@ -1,0 +1,79 @@
+import { act, renderHook } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+
+import type { EmbeddingModelConfig } from "@/api/types/agent";
+import { useAgentStore } from "@/stores/agentStore";
+import { useEmbeddingVerificationStore } from "@/stores/embeddingVerificationStore";
+import { isEmbeddingEnabled } from "./embeddingUtils";
+import { useEmbeddingVerification } from "./useEmbeddingVerification";
+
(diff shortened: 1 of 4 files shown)


jinliyl:
Addressed the fast-switch review finding in `4f5c2640`. Configuration loads now capture both a monotonically increasing request ID and the selected agent. A response can update the shared form only when it is still the latest request and still belongs to the current agent; stale errors and loading completions are ignored as well. Added a regression test for A → B → A with B’s disabled response resolving last.

zhaozhuang521 (approved):
lgtm

PR #7049, opened 2026-08-15
feat(chats): add limit/before pagination to GET /chats/{chat_id}

zsrmoyanzsr:
## Description

Add optional pagination to `GET /api/chats/{chat_id}` so the console can load the most recent messages first and fetch older history on demand, instead of receiving the whole conversation in a single response.

- New query params: `limit` (max number of most-recent messages, default `0` = full history) and `before` (cursor for "load earlier messages").
- `ChatHistory` now also returns `total` (message count before windowing) and `has_more` (pagination hint for clients).
- **Fully backward compatible**: without query params the endpoint behaves exactly as before (response shape only gains two additive fields with safe defaults).

Design notes:

- The `before` cursor uses **`metadata.original_id`** (the persistent AgentScope `Msg` id), not `Message.id`: `agentscope_msg_to_message` regenerates `Message.id` (`uuid4().hex`) on every request, so it is not stable across calls and cannot serve as a cursor.
- One AgentScope `Msg` can expand into multiple `Message` objects (text / reasoning / tool segments) sharing the same `original_id`; the cursor cuts at the **first** occurrence so segments of the same source message are never split across pages.
- An unknown/stale cursor is ignored gracefully (the request degrades to a plain `limit` window) instead of erroring.
- Windowing runs after `agentscope_msg_to_message`, after scroll placeholders / synthetic messages are already filtered out.

**Related Issue:** Relates to #3915 (backend foundation for "lazy-load older messages when scrolling up" and eventual virtualization) and #6635 (the chat-history pagination direction; the gzip / timeout / skills-list parts there are left to the ongoing work under that issue, which already has an assignee).

**Security Considerations:** None — the endpoint only gains optional query parameters; no new data is exposed (`total`/`has_more` are counts derived from the same history the endpoint already returns). `limit` is validated (`ge=0, le=10000`).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI) *(follow-up: open chats with `?limit=50`, add a "Load earlier messages" action, then virtualize the main message list per #3915 — pagination and virtualization compose well)*
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
[...]

Issue #6635, opened 2026-08-02
[Bug]: Console pages fail to load on slow networks — all-in-one MB-level API responses (skills list & chat history, uncompressed) vs fixed 30 s frontend timeout


Moonlit-Pages:
# [Bug]: Console pages fail to load on slow networks — all-in-one MB-level API responses (skills list & chat history, uncompressed) vs fixed 30 s frontend timeout

## QwenPaw Version

2.0.1 (pip install)

## Description

Several console views fail to load whenever the underlying API response cannot be transferred within the console's fixed 30 s fetch timeout. All affected endpoints share the same pattern: **a single response returns the entire dataset (no pagination / full bodies embedded), sent uncompressed**. Whether a view fails correlates exactly with payload size:

**Affected endpoint family 1 — skills lists** (Workspace → Skills page, Settings → Skill Pool page):

| Endpoint | Dataset | Response size (uncompressed) | Result on a ~25 KB/s link |
|---|---|---|---|
| `GET /api/skills` | 1 skill | ~5 KB | ✅ loads |
| `GET /api/skills` | 26 skills | ~212 KB | ✅ loads |
| `GET /api/skills` | 70 skills | ~880 KB | ❌ 30 s timeout → "加载技能失败" |
| `GET /api/skills` | 36 skills | ~1.36 MB | ❌ 30 s timeout |
| `GET /api/skills/workspaces` (needed by Skill Pool page) | all workspaces | ~5.3 MB | ❌ 30 s timeout even at ~60 KB/s |

**Affected endpoint family 2 — chat history** (Chat page, opening a conversation):

`GET /api/chats/{chat_id}` (`src/qwenpaw/app/chats/api.py:232`) returns **all messages of the chat in one response — no pagination / limit / incremental loading**. Measured on real workspaces:

| Agent | Chats | Median size | Largest | Result on a ~25 KB/s link (before fix) |
|---|---|---|---|---|
| default | 33 | ~241 KB | ~1.02 MB (679 messages) | large chats always time out |
| Equity-Alpha-Scout | 40 | ~404 KB | ~1.33 MB | large chats always time out |

Opening a ~1 MB chat history on a ~25 KB/s link needs ~40 s+ → the frontend aborts at 30 s, the conversation never opens. On faster-but-still-modest links it is just very slow.

Root causes, in combination:

1. **All-in-one responses.**
[...]

Issue #3915, opened 2026-04-28
[Feature]: Introduce virtual scrolling for Console WebUI to resolve performance degradation with long conversations


jfh1945815:
## Summary

Introduce virtual scrolling (or paginated rendering) for the message list in the Console WebUI, so that the page remains responsive even when the conversation history grows long — resolving the severe lag caused by full DOM rendering of all messages.

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Problem / Motivation

The Console WebUI currently renders the entire conversation history into the DOM at once. When conversations exceed ~100–200 messages, the number of DOM nodes can reach thousands or tens of thousands, causing:

1. **Scrolling lag** — every scroll event triggers expensive Reflow / Layout recalculations across thousands of nodes.
2. **Input latency** — typing in the message box can trigger full re-renders of the entire message list.
3. **Slow initial load** — refreshing or switching conversations requires rendering every message before becoming interactive.
4. **Worse on Edge** — Edge's built-in "Efficiency Mode" aggressively throttles CPU for the tab, compounding the issue.

This affects any user who regularly has long conversations through the WebUI, and has been widely discussed in similar products (e.g. Open WebUI, ChatGPT, Claude).

## Proposed Solution

Implement **virtual scrolling** in the Console frontend — render only the messages currently visible in the viewport, replace off-screen messages with empty placeholders, and dynamically create/remove DOM nodes as the user scrolls.

Specific suggestions:

- Use a proven virtual scrolling library such as `react-window`, `react-virtuoso`, or a lightweight custom implementation.
- Since message heights vary significantly (code blocks can be short or very long), use **dynamic height measurement** — either estimate before rendering or measure after first render and cache the height.
- Auto-scroll to bottom when a new message arrives; lazy-load older messages when scrolling up.
- Optionally combine with backend pagination (fetch history in chunks) for even better performance on very long conversations.

## Alternatives Considered

| Approach | Pros | Cons |
|---|---|---|
| **Pagination** (load 20 messages at a time, click "Load more") | Simple to implement, works immediately | Less fluid UX, requires manual action |
[...]

PR #7361, opened 2026-08-27
feat(chats): paginate long chat history and virtualize the transcript

zsrmoyanzsr:
## Description

#7049 added optional `limit` / `before` pagination on `GET /api/chats/{chat_id}` so the console does not have to download an entire conversation at once. @zhijianma closed it with Close-and-review-later:

> 仅添加后端接口，不解决问题， 需要配合等前端支持后再考虑 merge

Backend-only does not fix the freeze / timeout on long chats (#3915, #6635). This PR is the chat-history half of that work: window the stored history, convert only that window, and keep the Chat transcript from mounting every loaded bubble.

**Backend** (same public contract as #7049, still backward compatible):

- `limit` — max most-recent **source** messages (AgentScope `Msg` objects), default `0` = full history, `ge=0 le=10000`
- `before` — cursor is the persistent AgentScope `Msg.id`, exposed after conversion as `metadata.original_id`. `Message.id` is a fresh uuid4 every request and is not used as a cursor
- Response still has `total` (unwindowed source-message count) and `has_more`
- Stale / unknown cursor falls back to a plain `limit` window
- Windowing runs on the `Msg` list **before** `agentscope_msg_to_message`. Opening a 540k-token / thousands-of-msgs chat no longer converts the full history into runtime `Message` objects just to throw most of them away. One `Msg` can still expand into several `Message` objects that share `original_id`; cutting at the `Msg` layer keeps those groups intact

**Console:**

1. Opening or switching a chat requests `GET /api/chats/{chat_id}?limit=N` (`include_app_owned=false`). N is a Console-local preference (localStorage `qwenpaw_chat_history_page_size`), default 50, integer 1–10000 to match the backend `limit` contract. It is **not** stored on `channels.console` display settings (those are outbound IM rendering).
2. Settings → Console → Chat has a number field labeled **Messages loaded when opening a chat**. The same N is also a compact input next to **Load earlier messages**. Empty or non-numeric input is ignored and the last valid value is restored; out-of-range numbers are clamped. Changing N refetches the latest window (no `before`) and replaces the transcript.
3. **Load earlier messages** sends `before` set to the oldest loaded `metadata.original_id`, prepends when `has_more`, and keeps scroll position. Groups sharing `original_id` are not split. Short chats hide the Load earlier button (`has_more=false`) but still show the page-size input. Scrolling the transcript to the oldest edge also triggers load-earlier; the button stays as a fallback.
4. The main transcript is virtualized. `@agentscope-ai/chat` has no virtual list (its `PAGE_SIZE=10` helper still accumulates into a full DOM). The Chat page swaps the vendor `MessageList` for a host list that keeps the reverse scroller (`column-reverse`, `scrollTop === 0` at the newest edge) and only mounts viewport + overscan rows. Row heights are measured and cached (code blocks, cards, images); there is no fixed row height. New / streaming messages still stick to the bottom; prepending older history does not jump the viewport. Session switch abort, ownership epochs, pending user-message patching, the converted-session LRU, and reverse-list wheel handling are unchanged.
5. Chat search still fetches without `limit` / `before`. A window with `has_more` is cached for cheap switch-back but is never treated as the full canonical history.

Skills-list gzip from #6635 is not part of this PR.

**Related Issue:** Relates to #7049, #3915, #6635

**Security Considerations:** `limit` and `before` only slice a history the caller can already read through this endpoint; they do not change ownership or `include_app_owned`. `limit` is clamped (`ge=0`, `le=10000`). Conversion runs only on the windowed `Msg` list. The Console page size is a browser-local integer in that same range. `before` is an opaque `Msg.id` / `metadata.original_id` string, not a path or query language; an unknown cursor degrades to a plain `limit` window instead of erroring. `total` / `has_more` are counts of source messages, derived from the same memory list the handler already loaded.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

- [ ] **Contract test** exists in `tests/contract/channels/test_&lt;channel&gt;_contract.py` (REQUIRED)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_&lt;channel&gt;.py` for complex internal logic

## Testing

Manual:

1. Open a long chat (more than 50 messages).
2. The first history request should be `GET /api/chats/{chat_id}?limit=50` (plus `include_app_owned=false` from the Chat surface). The handler should convert only that window, not the full `Msg` list.
3. Change the compact number next to **Load earlier messages** to `200` (or set it under Settings → Console → Chat). Confirm the value persists across reload.
4. Changing N while the chat is open should refetch `?limit=200` (no `before`) and replace the transcript with the newest 200 source messages.
5. Scroll toward the oldest messages (or click **Load earlier messages**). The next request should send `before` equal to the oldest loaded source message `original_id` and `limit=200`.
6. Older messages prepend without jumping the viewport. New / streaming messages still stick to the bottom.
7. Only bubbles near the viewport should be in the DOM (React DevTools / inspector); scrolling recycles rows of uneven height (code, cards, images).
8. Short chats should not show **Load earlier messages** (`has_more=false`) but can still show the page-size input.
9. Empty / garbage input should restore the last valid N and not send a request.
10. Chat search should still call `getChat` without `limit` / `before` so matches outside the latest window are found.
11. Switching sessions should still abort in-flight loads; pending user messages and the LRU cache should behave as before.

## Evidence

```bash
$ pre-commit run --all-files

$ pytest tests/unit/app/chats/test_history_window.py \
    tests/unit/app/chats/test_api_history_pagination.py \
    tests/unit/app/chats/test_api.py \
    tests/unit/app/chats/test_models.py -q
.............................................                            [100%]
45 passed in 0.70s

$ npx vitest run src/api/modules/chat.test.ts \
    src/pages/Chat/sessionApi/historyWindow.test.ts \
    src/pages/Chat/sessionApi/historyPageSize.test.ts \
    src/pages/Chat/tests/historyPagination.test.ts \
    src/pages/Chat/tests/agentSessionOwnership.test.ts \
    src/pages/Chat/components/LoadEarlierMessages/LoadEarlierMessages.test.tsx \
    src/pages/Chat/components/HistoryPageSizeInput/index.test.tsx \
    src/pages/Settings/Console/index.test.tsx \
    src/pages/Chat/tests/pendingUserMessage.test.ts \
    src/pages/Chat/tests/sessionCacheStaleness.test.ts \
    src/pages/Chat/tests/testLargeSession.test.ts \
    src/pages/Chat/ChatPage.coverage.test.tsx \
    src/pages/Chat/virtualMessageList/range.test.ts \
    src/pages/Chat/virtualMessageList/VirtualizedBubbleList.test.tsx \
    src/pages/Chat/messageScroll.test.ts
 Test Files  15 passed (15)
      Tests  210 passed (210)
```

## Additional Notes

Backend design and tests started from #7049 (@zsrmoyanzsr). Current `main` also has `include_app_owned` on this endpoint; the pagination params sit next to that. This is meant to answer @zhijianma's review on #7049 rather than re-open that branch as-is.

No website docs change: `website/public/docs` only lists `/api/chats/*`. Channel contract checks above are N/A (this is the web Console UI + chats API, not an IM channel). I did not have a 540k-token chat fixture in this environment, so the long-open path is covered by the Msg-layer conversion spy, API pagination tests, and the virtual-list mount/prepend/scroll-to-top tests rather than a live browser pass against a huge session.


zhaozhuang521:
@zsrmoyanzsr This PR implements cursor-based pagination for the API and the frontend display layer; however, the backend still loads and parses the entire conversation before truncating the result, so it does not constitute true storage-layer pagination. It primarily reduces the overhead associated with network transmission, message transformation, and frontend DOM rendering, without lowering the cost of reading the full history from the backend.

zsrmoyanzsr:
@zhaozhuang521 You're right that the backend still reads the full history. That was intentional. The freeze we actually hit is on the console: a long chat stalls the page when you open it. We almost never saw the server-side history read itself as the problem. The full read still happens on the server, but truncating before convert/send already cuts what users feel: payload size, message conversion, and DOM. True storage-layer pagination would mean changing how sessions are persisted. That's a much bigger change, and from our side it's not where the freezes come from. If you want that as a follow-up, we're happy to look. This PR should stay scoped to the transmission and rendering cost.

PR #7381, opened 2026-08-28
fix(dingtalk): detect stale stream connections and bound SDK requests

hongxicheng:
## Problem

DingTalk Stream WebSocket connections can become stale after system
sleep/wake cycles, network changes, or VPN route changes. In this state,
the connection may still appear open while no longer receiving messages.

The previous health check only verified that the client object and HTTP
session existed, so stale connections could still be reported as healthy.

Some synchronous HTTP requests in the DingTalk SDK also lacked timeouts.
A blocked connection request or Chatbot ACK request could therefore block
the Stream event loop indefinitely.

## Changes

- Actively send a WebSocket Ping and wait for the corresponding Pong in
  the DingTalk health check.
- Check the Stream thread, event loop, and WebSocket state before probing.
- Report the channel as unhealthy when Ping times out, fails, is cancelled,
  or cannot be executed because the Stream loop is blocked.
- Add connection and read timeouts to the SDK's synchronous HTTP requests,
  including connection setup and Chatbot ACK requests.
- Keep the watchdog's system sleep/wake detection. When a significant wall
  clock jump is detected, it closes the stale WebSocket and lets the SDK
  reconnect.
- Tighten the DingTalk Mock IM integration test to require an explicit
  `channel == "dingtalk"` and `status == "healthy"` response.

The health check only detects and reports connection health. It does not
restart the channel directly. The watchdog only handles clear
sleep/wake-related time jumps and does not use a local heartbeat to decide
whether the Stream loop is alive.

## Testing

- DingTalk unit and contract tests: `205 passed`
- DingTalk Mock IM integration tests: `7 passed`
- `pre-commit`: passed
- `git diff --check`: passed

### DIFF
--- a/src/qwenpaw/app/channels/dingtalk/channel.py
+++ b/src/qwenpaw/app/channels/dingtalk/channel.py
@@ -36,6 +36,8 @@ import ssl
 import aiohttp
 import certifi
 import dingtalk_stream
+import dingtalk_stream.stream as dingtalk_stream_module
+import requests
 from dingtalk_stream import ChatbotMessage
 from alibabacloud_tea_openapi import models as open_api_models
 from alibabacloud_dingtalk.oauth2_1_0 import (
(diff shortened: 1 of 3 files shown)


cuiyuebing (approved):
LGTM

PR #7386, opened 2026-08-28
fix(providers): migrate discovered model output limits

zhijianma:
## Summary

- migrate legacy per-model `max_tokens` values in `discovered_models`
- keep provider snapshots loadable so encrypted credentials are restored
- add regression coverage for discovery migration and secret preservation

## Root cause

PR #7337 migrated legacy model output limits in `models` and
`extra_models`, but omitted `discovered_models`. A provider snapshot with
legacy discovery entries then failed `ModelInfo` validation as a whole, so
the provider fell back to its built-in defaults and appeared to lose its API
key even though the encrypted key remained valid on disk.

## Tests

- `tests/unit/providers/test_provider_model_state.py`: 5 passed
- `tests/unit/providers/test_provider_manager.py`: 145 passed, 1 skipped
- verified a real legacy DashScope snapshot migrates 183 discovery entries,
  preserves the decrypted API key, and returns the masked key from
  `/api/models`

### DIFF
--- a/src/qwenpaw/providers/provider_model_state.py
+++ b/src/qwenpaw/providers/provider_model_state.py
@@ -57,7 +57,11 @@ def migrate_provider_snapshot(data: dict[str, Any]) -> bool:
     ):
         return False
 
-    for collection_name in ("models", "extra_models"):
+    for collection_name in (
+        "models",
+        "extra_models",
+        "discovered_models",
+    ):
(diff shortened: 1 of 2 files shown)


rayrayraykk (approved):

cuiyuebing (approved):
LGTM

PR #7401, opened 2026-08-29
fix(acp): prevent Windows ACP agent stalls during workspace bootstrap

shadowabi:
## Description

On Windows, the ACP agent (`python -m qwenpaw acp`) can hang for minutes or fail to respond at all while the workspace initializes:

1. **Event loop frozen during bootstrap** — `bootstrap_plugins()` runs synchronously on the event loop. Plugin loading / MCP init can take minutes; during that window the loop cannot schedule any protocol work, so responses queue up. ACP clients time out and close the pipe; the eventual write then fails with `OSError: [Errno 22]` (ERROR_NO_DATA).
2. **numpy import can block indefinitely on the Windows loader lock** — in the full ACP runtime, the first `import numpy` (reached via `agentscope.embedding` inside workspace bootstrap) blocks forever on the Windows loader path (all threads waiting, no concurrent import activity). Simplified repros (bare `import numpy`, any pre-import, thread structures, loop shapes) all pass instantly — only the full ACP process stalls.

### Fix

- Run `bootstrap_plugins` in a worker thread (`asyncio.to_thread`) so the event loop keeps serving protocol requests during workspace init.
- Add a Windows-only threaded sender (`_threaded_sender_factory`) that writes responses from an executor thread instead of relying on the loop being free to drain the write queue. Non-Windows keeps the default in-loop sender.
- Warm up `import numpy` before `asyncio.run` in the ACP CLI: importing it before the event loop starts completes the DLL load while nothing else is contended.

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core
- [ ] Console
- [ ] Channels
- [ ] Skills
- [x] CLI
- [ ] Documentation
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts

## Checklist

- [x] Run and pass pre-commit run --all-files (see evidence below; only pre-existing upstream mypy errors on unrelated files)
- [x] Run and pass relevant tests (pytest or as applicable)
- [x] Update documentation if needed (none needed — behavior fix only)

## Testing

Windows 10, full ACP client loop (spawn → initialize → session/new → two-turn streaming → exit) via `run_demo_copaw.py`:

- Before: session/new hangs 120s+ (client times out); prompt 300s timeout; stderr shows the loop frozen inside `numpy create_module`.
- After: session/new ~0.3s; two turns complete with `end_turn` (12.8s / 2.6s), zero stall traces.

Non-Windows path untouched (sender factory only applies when `platform.system() == "Windows"`; bootstrap change is platform-neutral).

### pre-commit run --all-files

```
check python ast / yaml / xml / toml / docstring / json / encoding pragma /
detect private key / trim trailing whitespace ......... Passed
add-trailing-comma ......................... Passed (auto-fixed 1 line, committed)
black ......................................... Passed
flake8 ........................................ Passed
mypy ................................... Failed (3 errors — all in pre-existing
  upstream files: src/qwenpaw/cli/shutdown_cmd.py:142/144/149; none in this PR's
  files; same failures reproduce on upstream/main baseline)
pylint (PR files only) ................. 9.37/10, no errors
  (full-repo pylint exceeded local runtime; CI job tolerates it via `|| true`)
```

### pytest

```
unit:      10045 passed, 8 failed, 86 skipped (12m42s)
contract:  292 passed, 2 skipped, 0 failed
```

The 8 unit failures are all platform/environment-specific (symlink handling on Windows without privileges, Linux-only process-isolation test, `qwenpaw-data` extra not installed). **They reproduce identically on the upstream/main baseline (8 failed in 1.10s)** — none of them touch ACP code.

## Evidence

Verified on Windows 10, Python 3.11.15 (uv), numpy 2.4.6, real venv.

Before the fix: 13/13 recorded runs hang on the first prompt and the client kills the ACP child at ~120s. A `faulthandler` dump fired 30s into the hang shows the main thread (the asyncio loop) frozen inside the workspace-bootstrap import chain (`Workspace.__init__` → `_register_services` → `agentscope.embedding` → `numpy/_core/multiarray.py create_module`), with every other thread idle — the loop never recovers on its own.

After the fix, two clean end-to-end runs (spawn → initialize → session/new → two streaming turns → exit): turn 1 completes in 12.8s with real model inference, turn 2 in 2.6s. `session/new` now answers in ~0.3s while bootstrap is still running in the background, and the server log shows the response written before `workspace construct begin`.

### DIFF
--- a/src/qwenpaw/agents/acp/server.py
+++ b/src/qwenpaw/agents/acp/server.py
@@ -12,6 +12,7 @@ sub-agent delegation, etc.).
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import json
 import logging
 from pathlib import Path
@@ -457,13 +458,10 @@ class QwenPawACPAgent(Agent):
             )
             app_services = await self._ensure_app_services()
             bootstrap_kwargs = self._build_bootstrap_kwargs(app_services)
-            if bootstrap_kwargs:
-                await asyncio.to_thread(
-                    workspace.bootstrap_plugins,
-                    **bootstrap_kwargs,
-                )
-            else:
-                await asyncio.to_thread(workspace.bootstrap_plugins)
+            await asyncio.to_thread(
+                workspace.bootstrap_plugins,
+                **bootstrap_kwargs,
+            )
             workspace.set_app_services(app_services)
             await workspace.start()
 
@@ -1562,8 +1560,12 @@ def _sync_stdout_write(data: bytes) -> None:
     """Synchronous stdout write, safe to call from a worker thread."""
     import sys
 
-    sys.stdout.buffer.write(data)
-    sys.stdout.buffer.flush()
+    try:
+        sys.stdout.buffer.write(data)
+        sys.stdout.buffer.flush()
+    except OSError:
+        logger.debug("ACP stdout write failed (pipe closed?)", exc_info=True)
+        raise
 
 
 def _threaded_sender_factory(writer: Any, supervisor: Any) -> Any:
@@ -1579,16 +1581,35 @@ def _threaded_sender_factory(writer: Any, supervisor: Any) -> Any:
     from acp.task.sender import MessageSender
 
     class _ThreadedMessageSender(MessageSender):
+        def __init__(self, writer: Any, supervisor: Any) -> None:
+            super().__init__(writer, supervisor)
+            self._pending: set[asyncio.Future[None]] = set()
+
         async def send(self, payload: dict[str, Any]) -> None:
             data = (json.dumps(payload, separators=(",", ":")) + "\n").encode(
                 "utf-8",
             )
             loop = asyncio.get_running_loop()
-            await loop.run_in_executor(None, _sync_stdout_write, data)
+            future = loop.run_in_executor(None, _sync_stdout_write, data)
+            self._pending.add(future)
+            future.add_done_callback(self._pending.discard)
+            await future
 
         async def _loop(self) -> None:  # type: ignore[override]
             # All writes happen in send(); nothing to drain here.
             return
+
+        async def close(self) -> None:
+            if self._closed:
+                return
+            self._closed = True
+            # Wait for in-flight executor writes instead of relying on the
+            # inherited queue-sentinel shutdown (nothing drains that queue).
+            if self._pending:
+                await asyncio.gather(*self._pending, return_exceptions=True)
+            if self._task is not None and not self._task.done():
+                with contextlib.suppress(asyncio.CancelledError):
+                    await self._task
 
     return _ThreadedMessageSender(writer, supervisor)
 
--- a/src/qwenpaw/agents/acp/server.py
+++ b/src/qwenpaw/agents/acp/server.py
@@ -1584,6 +1584,7 @@ def _threaded_sender_factory(writer: Any, supervisor: Any) -> Any:
         def __init__(self, writer: Any, supervisor: Any) -> None:
             super().__init__(writer, supervisor)
             self._pending: set[asyncio.Future[None]] = set()
+            self._shutdown = False
 
         async def send(self, payload: dict[str, Any]) -> None:
             data = (json.dumps(payload, separators=(",", ":")) + "\n").encode(
@@ -1600,9 +1601,9 @@ def _threaded_sender_factory(writer: Any, supervisor: Any) -> Any:
             return
 
         async def close(self) -> None:
-            if self._closed:
+            if self._shutdown:
                 return
-            self._closed = True
+            self._shutdown = True
             # Wait for in-flight executor writes instead of relying on the
             # inherited queue-sentinel shutdown (nothing drains that queue).
             if self._pending:
--- a/src/qwenpaw/agents/acp/server.py
+++ b/src/qwenpaw/agents/acp/server.py
@@ -456,9 +456,14 @@ class QwenPawACPAgent(Agent):
                 workspace_dir=str(workspace_dir),
             )
             app_services = await self._ensure_app_services()
-            workspace.bootstrap_plugins(
-                **self._build_bootstrap_kwargs(app_services),
-            )
+            bootstrap_kwargs = self._build_bootstrap_kwargs(app_services)
+            if bootstrap_kwargs:
+                await asyncio.to_thread(
+                    workspace.bootstrap_plugins,
+                    **bootstrap_kwargs,
+                )
+            else:
+                await asyncio.to_thread(workspace.bootstrap_plugins)
             workspace.set_app_services(app_services)
             await workspace.start()
 
@@ -1553,6 +1558,49 @@ class QwenPawACPAgent(Agent):
         return agent_config.active_model
 
 
+def _sync_stdout_write(data: bytes) -> None:
+    """Synchronous stdout write, safe to call from a worker thread."""
+    import sys
+
+    sys.stdout.buffer.write(data)
+    sys.stdout.buffer.flush()
+
+
+def _threaded_sender_factory(writer: Any, supervisor: Any) -> Any:
+    """Windows sender that writes responses from a worker thread.
+
+    The event loop can be blocked for a long time while the workspace
+    bootstraps (plugin loading / MCP init can take minutes).  The default
+    in-loop sender never gets scheduled during that window, so the client
+    times out and closes the pipe; the eventual write then fails with
+    ``OSError: [Errno 22]`` (ERROR_NO_DATA).  Running the write in the
+    executor keeps responses flowing even while the loop is busy.
+    """
+    from acp.task.sender import MessageSender
+
+    class _ThreadedMessageSender(MessageSender):
+        async def send(self, payload: dict[str, Any]) -> None:
+            data = (json.dumps(payload, separators=(",", ":")) + "\n").encode(
+                "utf-8",
+            )
+            loop = asyncio.get_running_loop()
+            await loop.run_in_executor(None, _sync_stdout_write, data)
+
+        async def _loop(self) -> None:  # type: ignore[override]
+            # All writes happen in send(); nothing to drain here.
+            return
+
+    return _ThreadedMessageSender(writer, supervisor)
+
+
+def _make_agent_sender_factory() -> Any:
+    import platform
+
+    if platform.system() == "Windows":
+        return _threaded_sender_factory
+    return None
+
+
 async def run_qwenpaw_agent(
     agent_id: str | None = None,
     workspace_dir: Path | None = None,
@@ -1569,6 +1617,10 @@ async def run_qwenpaw_agent(
     try:
         # pylint: disable=protected-access
         await agent._install_runtime_provider()
-        await run_agent(agent, use_unstable_protocol=True)
+        kwargs: dict[str, Any] = {"use_unstable_protocol": True}
+        sender_factory = _make_agent_sender_factory()
+        if sender_factory is not None:
+            kwargs["sender_factory"] = sender_factory
+        await run_agent(agent, **kwargs)
     finally:
         await agent._shutdown_workspace()  # pylint: disable=protected-access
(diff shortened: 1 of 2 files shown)


src/qwenpaw/cli/acp_cmd.py
```diff
@@ -74,6 +74,14 @@ def acp_cmd(
...
     from ..agents.acp.server import run_qwenpaw_agent
 
+    # Warm-up heavy native imports before the event loop starts. In the full
+    # ACP runtime, the first `import numpy` (via agentscope.embedding inside
+    # workspace bootstrap) can block indefinitely on the Windows loader lock
+    # (all threads waiting, no concurrent import activity). Importing here,
+    # before asyncio.run, completes the DLL load while nothing else is
+    # contended and makes the workspace bootstrap pass through immediately.
+    import numpy  # noqa: E402,F401
+
```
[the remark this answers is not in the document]
shadowabi: Fixed in e5e94961: the warm-up is now Windows-only (\`if sys.platform == "win32"\`) and best-effort (\`try: import numpy / except ModuleNotFoundError: pass\`), so \`qwenpaw acp\` still starts normally in environments without numpy. Verified: no-numpy env (-S run) swallows ModuleNotFoundError and continues; \`pytest tests/unit/cli/test_acp_cmd.py\` + ACP unit group 74 passed; black/flake8/add-trailing-comma pass; mypy reports only the 3 pre-existing errors in \`shutdown_cmd.py\` (present on upstream/main baseline, unrelated to this PR).

src/qwenpaw/agents/acp/server.py
```diff
@@ -456,9 +456,14 @@ async def _ensure_workspace(self) -> Any:
...
-                **self._build_bootstrap_kwargs(app_services),
-            )
+            bootstrap_kwargs = self._build_bootstrap_kwargs(app_services)
+            if bootstrap_kwargs:
+                await asyncio.to_thread(
+                    workspace.bootstrap_plugins,
+                    **bootstrap_kwargs,
+                )
+            else:
+                await asyncio.to_thread(workspace.bootstrap_plugins)
```
x1n95c: `asyncio.to_thread()` currently wraps only `workspace.bootstrap_plugins(...)`. The preceding `Workspace(...)` construction and `_build_bootstrap_kwargs(...)` still run synchronously on the ACP event-loop thread:

```python
workspace = Workspace(...)
bootstrap_kwargs = self._build_bootstrap_kwargs(app_services)
await asyncio.to_thread(workspace.bootstrap_plugins, **bootstrap_kwargs)
```

Therefore, any slow import or initialization in the constructor or bootstrap kwargs discovery can still block ACP request handling. Also, the `qwenpaw.agents.memory` import chain may load `agentscope.embedding`/NumPy before `to_thread()` is reached, can you confirm that the agent stalls during workspace bootstrap due to this part?

src/qwenpaw/agents/acp/server.py
```diff
@@ -1553,6 +1558,49 @@ async def _switch_model(
...
 
+def _sync_stdout_write(data: bytes) -> None:
+    """Synchronous stdout write, safe to call from a worker thread."""
+    import sys
+
+    sys.stdout.buffer.write(data)
+    sys.stdout.buffer.flush()
+
+
+def _threaded_sender_factory(writer: Any, supervisor: Any) -> Any:
```
x1n95c: This can keep stdout writes from blocking the event loop once send() has been scheduled, but it seems cannot help while the event loop is completely blocked by synchronous code, such as a blocking section inside workspace startup.

x1n95c:
Hello, thanks for pointing this out. Could you provide more details about the issue you encountered? Starting QwenPaw as an ACP server typically takes a few seconds, which is expected. Could you share the relevant logs, especially those showing when and how OSError: [Errno 22] occurred?

shadowabi:
Hi @x1n95c, thanks for taking a look! Sorry for not making the failure mode clear enough: it's not "server startup is slow". Here is the actual timeline with logs.

**What actually happens**

1. `initialize` and `session/new` handlers complete quickly, but the `session/new` response gets stranded: right after it, the `_advertise_commands` task starts `Workspace(...)` construction **synchronously on the event-loop thread**. The constructor's import chain (`qwenpaw.agents.memory` → `agentscope.middleware` → `agentscope.embedding` → numpy) hits the first `import numpy`, and loading numpy's C extension (`_multiarray_umath.pyd`) blocks for minutes on the Windows loader lock in this particular process state. The loop stays frozen, and it never recovered on its own in 13/13 recorded runs.

2. The frozen loop can't drain the queued `session/new` response (the stock writer's drain runs on the loop). The client times out at ~120s and kills the child process, closing the stdout pipe's read end.

3. The construction eventually finishes anyway; the writer then flushes the stranded ~13KB response and gets `OSError: [Errno 22] Invalid argument` (broken pipe), logged as `Error writing to stdout`. **So OSError 22 is a downstream symptom of the client timeout, not the root cause.** The server log captures the sequence directly: `DBG CLIENT readline EOF` (client killed the process) comes *before* `bootstrap_plugins complete`.

**faulthandler dump at the stall** (minimal repro, fired 30s after the import begins; line numbers are from the instrumented build used for diagnosis):

```text
Timeout (0:00:30)!
Thread 0x000054b8 (most recent call first):        # main thread = asyncio loop
  File "...\numpy\_core\multiarray.py", line 11 in <module>   # create_module -> C extension load
  File "...\agentscope\embedding\_file_cache.py", line 9 in <module>
  File "...\agentscope\embedding\__init__.py", line 13 in <module>
  File "...\agentscope\middleware\_longterm_memory\_reme\_middleware.py", line 48 in <module>
  File "F:\...\src\qwenpaw\agents\memory\base_memory_manager.py", line 18 in <module>
  File "F:\...\src\qwenpaw\app\workspace\workspace.py", line 377 in _register_services
  File "F:\...\src\qwenpaw\app\workspace\workspace.py", line 93 in __init__
  File "F:\...\src\qwenpaw\agents\acp\server.py", line 455 in _ensure_workspace
  File "F:\...\src\qwenpaw\agents\acp\server.py", line 1456 in _advertise_commands
  ... asyncio base_events.py run_forever
# the 3 other threads (acp stdio writer/reader, executor worker) are all idle-waiting
```

**Why the warm-up fixes it**: importing numpy at process startup (before the asyncio loop and its pools exist) loads the C extension cleanly in ~0.16s. After that, every import on the constructor path is a `sys.modules` cache hit, construction is fast, and the stall disappears entirely (two end-to-end runs passed after the fix). Native-thread probing during the stall showed all 13 threads in Waiting state (none running), consistent with a loader-lock wait rather than any concurrent work.

One more note on repro: a bare `python -c "import numpy"`, even with PIPE stdio, no console, and matched cwd/env, always imports in under a second. Only the full ACP server runtime (stdio protocol flow + Proactor loop + thread structure) triggers the stall, which is probably why it can look like a normal slow start unless you catch the dump. Happy to share the minimal repro scripts if you'd like to try on a Windows machine.

src/qwenpaw/agents/acp/server.py
shadowabi: You're right, and the faulthandler dump confirms the stall is exactly in the constructor path, before `to_thread` is ever reached:

```text
File "...\agentscope\embedding\__init__.py", line 13 in <module>
File "F:\...\src\qwenpaw\agents\memory\__init__.py", line 7 in <module>
File "F:\...\src\qwenpaw\app\workspace\workspace.py", line 377 in _register_services   # inside Workspace.__init__
File "F:\...\src\qwenpaw\agents\acp\server.py", line 455 in _ensure_workspace
```

That's why the PR pairs the `to_thread` change with the warm-up in `acp_cmd.py`: numpy's C extension is loaded at process startup (~0.16s, no loop/threads running yet), so by the time the constructor runs its import chain every import is a `sys.modules` cache hit and construction completes in milliseconds. As shipped, the three pieces cover different segments: warm-up removes the multi-minute DLL-load stall on the constructor path, `to_thread` keeps `bootstrap_plugins`' remaining slow work off the loop, and the threaded sender keeps in-flight responses from being stranded behind a later freeze.

That said, I agree the constructor still running synchronously on the loop is a residual risk: any slow import added to the constructor path in the future would freeze the loop again. Moving the construction into the worker thread as well would be the more thorough fix; the coupling is that `Workspace(...)`'s result feeds `_build_bootstrap_kwargs(...)` and the bootstrap call right after. Happy to restructure it that way if you prefer.

src/qwenpaw/agents/acp/server.py
shadowabi: Fair point. To be precise about what it covers and what it doesn't:

**It does**: a response whose handler already completed gets written immediately. `run_in_executor`'s `submit()` is a synchronous hand-off, so the worker thread writes the bytes without waiting for the loop. In the repro log the `session/new` response lands before `workspace construct begin` even logs, while bootstrap is still running in the background. The client receives its sessionId in ~0.3s instead of the response being stranded behind the freeze.

**It doesn't**: requests arriving while the loop is frozen still can't be handled, and a response produced *after* a freeze started would stay queued until the loop thaws. Agreed that part is only fixed by keeping synchronous heavy work off the loop, which is what the warm-up (removing the multi-minute DLL-load stall on the constructor path) and `to_thread(bootstrap_plugins)` do. This factory is just a defense layer for the write path on Windows, not a substitute for not blocking the loop.

shadowabi:
Thanks for the thorough review. Pushed 866b4ff8 addressing the findings:

**Medium 1 (close lifecycle)**: adopted. `_ThreadedMessageSender` now tracks in-flight executor futures in a `set` (with `add_done_callback(discard)` so completed writes don't accumulate) and overrides `close()` to `gather(*pending, return_exceptions=True)` before finishing. No sentinel is put on the never-drained queue anymore.

**Low 1 (dead `else`)**: adopted. Dropped the branch; `asyncio.to_thread(workspace.bootstrap_plugins, **bootstrap_kwargs)` with empty kwargs is equivalent to the no-arg call.

**Low 2 (broken pipe diagnostics)**: partially adopted. Added the `except OSError: logger.debug(...)` line, but re-raising instead of swallowing, so `send()` still surfaces the failure to the caller. Swallowing it entirely would make `send()` report success on a closed pipe, which hides a real error behind a fake success.

**Low 3**: thanks for verifying the signature.

All 21 tests in `test_acp_service.py` + `test_acp_cmd.py` pass locally with the changes.

x1n95c:
After further review, I believe the previous changes in `src/qwenpaw/agents/acp/server.py` should be reverted for now. The custom threaded sender introduces additional risks: it bypasses the provided writer, does not preserve the original sender's serialized/FIFO write behavior, and still has shutdown/send race considerations. Also, moving only `bootstrap_plugins()` to a worker thread does not cover the synchronous `Workspace(...)` construction path, which is where the reported stall occurs.

The reproduction evidence indicates that the primary issue is the first `numpy` native-extension import on Windows, specifically the loader-lock stall during `Workspace.__init__`. The `OSError: [Errno 22]` is a downstream symptom after the ACP client times out and closes the pipe.

Therefore, I have pushed commit `34b28200`, which reverts the `server.py` changes and keeps only the Windows-only, best-effort NumPy warm-up in `src/qwenpaw/cli/acp_cmd.py`. The targeted ACP tests pass locally (`21 passed`). This keeps the fix focused on the demonstrated root cause while avoiding the additional protocol and lifecycle risks.

shadowabi:
Ack, and agreed — the reproduction evidence does point at the loader-lock stall during `Workspace.__init__` as the demonstrated root cause, and the warm-up removes exactly that. The threaded sender was a defense layer for a scenario (multi-minute event-loop freeze) that no longer exists once the warm-up is in place.

I verified 34b28200 end-to-end on the original machine where the 13/13 stall reproduced: full client loop (spawn → initialize → session/new → two streaming turns → clean exit) passes, turn 1 in 15.5s and turn 2 in 3.7s with real model inference, no stall traces. The minimal version holds up.

Thanks for pushing the revert directly — happy with the scoped-down PR.

lalaliat (approved):
LGTM

Issue #5856, opened 2026-07-08
[Bug]: Tool_call structure lost during context compaction, causing 400 errors / message count mismatch


quanrennsxsb:
## QwenPaw Version

v1.1.12.post2

## Description

During context compaction, `LightContextManager.pre_reasoning()` passes all messages
(including `tool_use`/`tool_result` blocks) to `format_msgs_to_str`, which converts
them to plain text for LLM summarization. The structured tool_call data is permanently
lost after compaction. On the next request, the model receives orphaned `tool_result`
messages with no matching `tool_use`, causing 400 errors or `message count mismatch`.

**Expected:** Recent tool_use/tool_result message pairs should survive compaction
with their structured data intact.

**Actual:** All tool call history is flattened into a plain-text summary. The model
can no longer see the structured tool call context, leading to broken tool call
sequences and API errors.

**Related PR(s):** #(optional)

**Security considerations:** N/A

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** v1.1.12.post2
- **OS:** Windows 10
- **Install method:** pip
- **Python version (if applicable):** 3.x

## Steps to Reproduce

1. Configure an agent with `context_manager_backend: light` and tool-calling enabled.
2. Have a long conversation that includes multiple tool calls (e.g. `get_current_time`).
3. Allow context to grow until compaction is triggered (`compact_threshold_ratio` exceeded).
4. After compaction, ask the agent to call a tool again.

## Actual vs Expected

- **Actual:** Post-compaction requests result in 400 errors (`Invalid 'content'`) or
  `Assistant message count mismatch after formatting`, because `tool_result` messages
  have no matching `tool_use` in the compacted context.
- **Expected:** The most recent tool call pairs should be preserved with full structure
  after compaction.

## Logs / Screenshots

```
Assistant message count mismatch after formatting (N expected survivors, got M)
Invalid 'content': 'content' objects must have a 'type' field that is either
'text' or 'image_url'
```

## Additional Notes

Root cause: `pre_reasoning()` in `light_context_manager.py` calls
[...]

PR #7409, opened 2026-08-30
fix(agents): drop empty assistant text blocks

### LINKED ISSUE #7402: [Bug]: Empty assistant output_text blocks persisted in session history poison every subsequent request — Ark Responses API returns 400 "MissingParameter: input.content.text"
## QwenPaw Version

v2.1.0 (Docker image based on upstream v2.1.0)

## Description

When using a Volcengine Ark provider (Responses API wire format), a single assistant message with an **empty text block** (`content=[{"type":"output_text","text":""}]`) saved into session history causes the upstream API to reject **every subsequent request** with `400 MissingParameter: input.content.text`. Because QwenPaw replays the full conversation history on each turn, the session is permanently poisoned: the user only sees `Internal error` on every message until the history is manually cleaned.

**How the empty block gets there:** a streaming response that fails / returns empty output gets persisted as an assistant message with an empty `output_text` block (related: #6601, empty responses are not surfaced as errors). On the next turn the poisoned block is replayed upstream verbatim.

**Evidence from production:** a failure-capture hook (logs the full request payload on upstream error) caught a real 953-item / 16 MB session payload that failed with this error. Structural scanning showed the payload was fully valid — 267 correctly paired `function_call`/`function_call_output` items, no orphans, no empty `input_text`, no broken images — and the **only** anomaly was one assistant item with `text: ""`.

**Replay proof:** replaying that exact payload returns the same 400 in <1 s (input-validation stage, no generation). Binary-searching the item list isolated the single empty-text assistant item; replacing `""` with any non-empty string makes the same request return 200.

**Suggested fix:** filter or defend against empty assistant text blocks — either don't persist an assistant turn whose text is empty, or sanitize (drop / placeholder-fill empty `output_text` items) when serializing history for the provider.

**Related PR(s):** N/A

**Security considerations:** N/A (no credentials or secrets involved; upstream request IDs in logs are non-sensitive)

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** v2.1.0
- **OS:** Linux (Docker container)
- **Install method:** Docker

## Steps to Reproduce

1. Use a provider that speaks the Responses API wire format to Volcengine Ark (doubao/glm plan endpoint).
2. Get an empty assistant text block into session history. Minimal direct repro against the endpoint:

```json
{
  "model": "<any-ark-responses-model>",
  "input": [
    {"role": "user", "content": [{"type": "input_text", "text": "hi"}]},
    {"role": "assistant", "content": [{"type": "output_text", "text": ""}]},
    {"role": "user", "content": [{"type": "input_text", "text": "continue"}]}
  ]
}
```

3. Send the request. It returns:

```
400 {"error":{"code":"MissingParameter","message":"The request failed because it is missing `input.content.text` parameter.","param":"input.content.text","type":"BadRequest"}}
```

4. Replace the empty `""` with any non-empty string (e.g. `"ok"`) → the same request returns `200`.

**Repro matrix (all verified by replaying against the live endpoint):**

| input shape | result |
|---|---|
| `[..., assistant(output_text "")]` anywhere in history (mid-sequence) | **400** |
| same payload with `""` → `"ok"` | 200 |
| `assistant(output_text "")` as the last item | 400 |
| 953-item real poisoned session (one empty block) | 400, <1 s |
| same session truncated to first 941 items (drops the tail containing the empty block) | **200** — removing the block restores the session |
| fully valid large session, no empty blocks | 200 |

## Actual vs Expected

- **Actual:** every subsequent turn in the session fails with upstream 400; QwenPaw surfaces a generic `Internal error` with no hint about the poisoned history item. The empty block is never reported at persistence time.
- **Expected:** QwenPaw should not persist empty assistant text blocks (or should sanitize them before sending), so a single failed/empty response cannot permanently poison the session.

## Logs / Screenshots

Upstream error captured by the gateway failure hook on the real poisoned session:

```
litellm.BadRequestError: OpenAIException - {"error":{"code":"MissingParameter","message":"The request failed because it is missing `input.content.text` parameter. Request id: 0217880026213...","param":"input.content.text","type":"BadRequest"}}
```

The single anomalous item inside the 953-item payload (everything else structurally valid):

```json
{"role": "assistant", "content": [{"type": "output_text", "text": ""}]}
```

Replay of the full 8.3 MB sanitized payload: `HTTP 400 in 0.8s` (input validation, deterministic).

## Additional Notes

- The empty `output_text` block is rejected regardless of position: mid-sequence or trailing. This makes it distinct from model-side flakiness — it is a deterministic input-validation rejection.
- Side observation (low priority, normal traffic unaffected): the same upstream also rejects any input sequence whose **last item is an assistant message** (appears to be a prefill-mode validation). QwenPaw normally ends input with user/tool-result items, so this only matters for history-truncation edge cases.
- Related issues: #5856 (structured tool_call lost during compaction → 400 family), #6601 (empty responses are silently swallowed — likely how the empty block entered history in the first place).
- A gateway-side fallback would only mask the symptom: any fallback model would receive the same invalid empty-text block. The fix belongs in history persistence / outbound serialization.

Update — root cause chain fully traced (how the empty block gets persisted):

We traced the poisoned session end-to-end via QwenPaw logs. The empty assistant block comes from a model turn that completed with zero text output, which QwenPaw then persists unconditionally:
19:19:12 — user sends a message; QwenPaw spends ~3 min on scroll/memory prep, then calls the model.
19:23:29 — the agent turn "finishes" with no text, no tool calls. QwenPaw still saves it as an assistant message (empty output_text) into the persisted session state (session.py "Saved session state").
19:23:32 — usage log confirms the model did respond: prompt_tokens: 45419, completion_tokens: 243 — the 243 completion tokens were all reasoning, with zero text content. So this is not a crashed/aborted stream; it's a successful response whose text part is empty (observed on glm-5.3-flash, reasoning-style model).
19:23:39 — user sends a follow-up; the full-history replay now contains the empty block → Ark 400 MissingParameter on the first call.
19:23:42 — WARNING _agent:_call_model:2431 - Model glm-5.3-flash exhausted all 1 attempt(s) — the BadRequest gets a single attempt, no retry, no usable fallback → surfaced to the user as Internal error.
So the failure requires two cooperating defects: the model occasionally returns a reasoning-only / zero-text response (unavoidable), and QwenPaw persists that empty assistant turn as-is instead of dropping it or raising (the fixable part — this is what #6601 describes from the "empty response is not surfaced" angle).

Side note: in our case the session later self-healed accidentally — the next turn triggered a scroll compaction, which archived the whole poisoned history including the empty block. Sessions without a compaction trigger would stay broken indefinitely, so persistence-time sanitation is still the right fix.

wananing:
## Description

A model turn that spends all of its completion tokens on reasoning finishes with an empty text part. The provider layer turns that part into a `TextBlock(text="")`, and the agent persists it like any other block, so every later request in the session replays it as `{"type": "output_text", "text": ""}`. Volcengine Ark rejects that item with `400 MissingParameter: input.content.text`, which means a single empty turn permanently poisons the session: the user then sees `Internal error` on every subsequent message until the history is cleared.

The chain, verified against the pinned `agentscope==2.0.7.post1`:

1. `agentscope/model/_openai_response/_model.py:442` maps every `output_text` part to `TextBlock(type="text", text=part.text)` with no emptiness check. The streaming path can also accumulate to an empty string.
2. `ReActAgent._save_to_context` (`agentscope/agent/_agent.py:3178`) persists the model's blocks as-is. It already filters exactly one block class there — assistant audio `DataBlock`s — with the comment that filtering at this point "keeps every downstream walker (formatter, count_tokens, persistence) honest without each having to remember."
3. `agentscope/formatter/_openai_response_formatter.py:204` re-emits assistant text blocks unconditionally, reproducing the rejected item.

This PR adds the missing guard at the same chokepoint. `QwenPawAgent._save_to_context` drops text blocks whose text is the empty string before they are persisted, and `QwenPawAgent._sanitize_loaded_context` applies the same filter on load so sessions already poisoned on disk heal instead of staying broken.

Deliberate boundaries:

- **Provider-agnostic.** The guard sits in persistence, not in an Ark/Responses-specific branch, so any provider that rejects empty text parts is covered regardless of which producer created the block.
- **Only the empty string is dropped.** Whitespace-only text is kept: providers reject `""`, not a block that happens to contain a newline, and preserving it avoids changing what the model sees in its own history.
- **`ThinkingBlock`s are untouched.** The Responses API requires empty-summary reasoning items to be echoed back with their `reasoning_item_id`; both the model layer and the formatter deliberately keep them.
- **Tool calls in the same turn survive.** Only the empty text block is removed from the turn.
- **The load path removes blocks, not messages.** A message left with no content still carries the turn's token usage, and a message with no content parts and no function calls contributes no item to the provider request, so it is harmless on the wire.

This mirrors the existing `_sanitize_loaded_context` guard for orphan `tool_result` messages, which exists to prevent the sibling `400 - Messages with role 'tool' must be a response to a preceding message with 'tool_calls'`. Same class of defect, different block type.

**Related Issue:** Fixes #7402

**Security Considerations:** None. No credentials, auth, or config handling is involved; the change only filters content blocks that carry no text.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

Unit tests only — no Ark credentials or network access needed.

`tests/unit/agents/test_empty_text_blocks.py` covers the predicate (empty string dropped, whitespace and non-text blocks kept, legacy dict blocks recognised), the save path (a reasoning-only turn persists nothing, a tool call in the same turn survives, normal text is untouched), the load path (a session poisoned on disk is stripped, paired tool messages are preserved), and the wire format (the sanitized context, run through the real `OpenAIResponseFormatter`, produces no `{"type": "output_text", "text": ""}` item).

The tests were written before the fix and fail without it: reverting `src/qwenpaw/agents/react_agent.py` to `main` leaves 5 of the 13 failing.

## Evidence

Focused suite, with the fix applied:

```
$ pytest tests/unit/agents/test_empty_text_blocks.py -q
.............                                                            [100%]
13 passed in 0.81s
```

Same suite with `src/qwenpaw/agents/react_agent.py` reverted to `main` (A/B proof that the tests pin the fix and not themselves):

```
$ git checkout src/qwenpaw/agents/react_agent.py
$ pytest tests/unit/agents/test_empty_text_blocks.py -q
FAILED tests/unit/agents/test_empty_text_blocks.py::TestSavePath::test_reasoning_only_turn_persists_no_empty_text
FAILED tests/unit/agents/test_empty_text_blocks.py::TestSavePath::test_tool_call_in_the_same_turn_survives
FAILED tests/unit/agents/test_empty_text_blocks.py::TestLoadPath::test_persisted_empty_block_is_stripped_on_load
FAILED tests/unit/agents/test_empty_text_blocks.py::TestLoadPath::test_paired_tool_messages_are_preserved
FAILED tests/unit/agents/test_empty_text_blocks.py::TestWireFormat::test_no_empty_output_text_item_reaches_the_request
5 failed, 8 passed in 1.37s
```

Surrounding regression scope:

```
$ pytest tests/unit/agents tests/unit/runtime -q --timeout=180
4 failed, 2557 passed, 3 skipped, 2 warnings in 75.72s (0:01:15)
```

The 4 failures are pre-existing and environment-local, not caused by this change: `test_acp_available_commands.py` (2) and `test_install_smoke.py` (2) all abort on `ConfigurationException: Agent 'default' configuration file contains invalid JSON` because this machine has a 0-byte `~/.qwenpaw/workspaces/default/agent.json`. Stashing every change in this PR and rerunning those two files reproduces exactly the same 4 failures.

Changed-file hooks:

```
$ pre-commit run --files src/qwenpaw/agents/react_agent.py src/qwenpaw/agents/utils/text_block_utils.py tests/unit/agents/test_empty_text_blocks.py
```

## Additional Notes

Out of scope, deliberately:

- #6601 — whether a zero-text model response should surface as an error instead of being accepted silently. That is the other half of the reporter's root-cause chain and belongs to whoever owns that issue; this PR only stops the empty block from being persisted.
- #5856 — tool_call structure loss during compaction.

The root-cause analysis and the fix scope were posted on #7402 before implementation, and no other work was claimed there at that time.


wananing:
Rebased onto current `upstream/main` (`549a7f3c4`); the new head is `83aa1d35e`. The change itself is unchanged.

The only conflict was the import block in `src/qwenpaw/agents/react_agent.py`, where `main` added `from .utils.message_request_normalizer import _is_media_block` on the same lines this branch adds the `text_block_utils` import. Both imports are kept. `_save_to_context` and `_sanitize_loaded_context` are identical on `main` to what this PR was written against, so no behavioural conflict arose.

Re-verified after the rebase: the PR's own suite plus `tests/unit/agents/context` is 545 passed, and changed-file `pre-commit` (mypy, black, flake8, pylint) is green.

qbc2016 (approved):
Please remove the unit test. Others LGTM.

wananing:
Removed the unit test file in ef5a92635 as requested. The change is otherwise identical to the approved head 83aa1d35e.

PR #7421, opened 2026-08-31
fix(yuanbao): restore protobuf decoding on protobuf 6+

hongxicheng:
## Summary

Fix Yuanbao protobuf decoding failure with newer protobuf versions.

## Root Cause

`decode_pb()` passed the removed `including_default_value_fields`
keyword argument to `json_format.MessageToDict()`.

With protobuf 6.x and 7.x, this raises `TypeError`. The exception was
silently caught, causing every otherwise valid protobuf payload to return
`None`.

## Changes

- Remove the obsolete `including_default_value_fields` argument.
- Add protobuf encode/decode round-trip regression tests covering:
  - `AuthBindRsp`
  - `PingRsp`
  - `KickoutMsg`
  - Empty/default-valued responses

Removing the argument preserves the existing behavior: fields without
presence and default values remain omitted because that is the API default.

## Impact

This restores decoding of Yuanbao response fields such as:

- Auth response fields, including `connectId`
- Negotiated heartbeat interval
- Kickout reason
- Business response fields

No protobuf wire format, encoding logic, authentication flow, heartbeat
flow, or regular JSON inbound-message flow was changed.

## Verification

- `conda run -n base pytest -q tests/unit/channels/test_yuanbao.py`
  - `100 passed`
- Round-trip checks passed with protobuf `6.33.6`
- Round-trip checks passed with protobuf `7.35.1`
- Python compilation check passed
- `git diff --check` passed

### DIFF
--- a/src/qwenpaw/app/channels/yuanbao/codec.py
+++ b/src/qwenpaw/app/channels/yuanbao/codec.py
@@ -293,7 +293,6 @@ def decode_pb(type_name: str, data: bytes) -> Optional[dict]:
         return json_format.MessageToDict(
             msg,
             preserving_proto_field_name=True,
-            including_default_value_fields=False,
         )
     except Exception:
         return None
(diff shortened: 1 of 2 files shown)


cuiyuebing (approved):
LGTM

PR #7465, opened 2026-09-01
fix(memory): normalize backend-specific embedding dimensions

### LINKED ISSUE #7464: [Bug]: DashScope Embedding index rebuild remains disabled because saved config is always detected as unsaved
## QwenPaw Version

QwenPaw `2.2.0b5`

## Description

On **Running Config → Embedding Model**, the **Embedding index** card always reports that the current changes have not been saved, even immediately after the configuration has been saved successfully and the page has been freshly loaded.

The card displays:

```text
Save configuration first
Save the current changes before rebuilding the index
```

As a result, the **Rebuild Embedding index** button remains disabled and the user cannot manually trigger an Embedding-only index rebuild from the Console when it is actually needed.

This is reproducible with a DashScope Embedding configuration whose persisted `use_dimensions` value is `true`. The Console only renders the `use_dimensions` form field when `backend === "openai"`, but `getEmbeddingConfigFingerprint()` includes `use_dimensions` for every backend. For DashScope, the watched form value can therefore omit this hidden field and normalize it to `false`, while the persisted fingerprint still contains `true`. The two fingerprints never match, so `hasUnsavedEmbeddingChanges` remains `true` after every save.

A fresh, read-only page load reproduced the issue before any field was edited. All visible Embedding fields matched the persisted configuration, while the index card was already locked as unsaved.

**Related PR(s):** #7133

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** `2.2.0b5`
- **OS:** Windows 11 25H2
- **Install method:** pip
- **Python version (if applicable):** Python `3.12.10`

## Steps to Reproduce

1. Use the `remelight` memory manager and configure an Embedding provider with `backend = "dashscope"`.
2. Ensure the persisted Embedding configuration has `use_dimensions = true`. One possible path is to enable **Send Dimensions in Request** while using the OpenAI backend, then switch the backend to DashScope, where this field is no longer rendered.
3. Fill in a valid DashScope Base URL, model name, API key, vector dimensions, cache settings, batch size, and health-check timeout.
4. Save the running configuration successfully.
5. Optionally test the Embedding service and confirm that the real request succeeds.
6. Reload the page, then open **Running Config → Embedding Model** without editing any field.
7. Observe the **Embedding index** card and its rebuild button.
8. Click **Save** again without changing any field. The index card remains in the same unsaved state.

## Actual vs Expected

- **Actual:**
  - The page can report `Current search: Vector + BM25`, indicating that Embedding is enabled and the current index is usable.
  - The persisted configuration has `needs_reindex = false`.
  - Nevertheless, the Embedding index card immediately displays `Save configuration first` / `Save the current changes before rebuilding the index`.
  - `hasUnsavedEmbeddingChanges` remains `true` after a successful save and after a fresh page load.
  - The **Rebuild Embedding index** button is permanently disabled, preventing manual rebuilds from the Console.

- **Expected:**
  - A freshly loaded form that matches the persisted configuration must not be marked as unsaved.
  - After a successful save, the persisted and current Embedding fingerprints should match.
  - The **Rebuild Embedding index** button should be enabled whenever Embedding is enabled and no rebuild is already running, regardless of whether `needs_reindex` is currently true or false.
  - Hidden or backend-inapplicable fields must not create a permanent dirty state.

## Logs / Screenshots

This appears to be a Console state-comparison bug rather than a backend save failure. The relevant sanitized persisted state is:

```text
backend=dashscope
use_dimensions=true
needs_reindex=false
```

The freshly loaded Console displays:

```text
Configuration: Enabled
Current search: Vector + BM25

Embedding index
Save configuration first
Save the current changes before rebuilding the index
```

The rebuild button is disabled before any form field is edited.

Relevant Console code in `EmbeddingModelCard.tsx` conditionally mounts the field only for the exact OpenAI backend:

```tsx
{normalizedBackend === "openai" && (
  <Form.Item
    name={[
      "reme_light_memory_config",
      "embedding_model_config",
      "use_dimensions",
    ]}
  >
    ...
  </Form.Item>
)}
```

However, `embeddingUtils.ts` includes it unconditionally in the dirty fingerprint:

```tsx
export function getEmbeddingConfigFingerprint(
  config?: Partial<EmbeddingModelConfig>,
) {
  if (!config) return "";
  return JSON.stringify([
    config.backend || "",
    config.api_key || "",
    config.base_url?.trim().replace(/\/+$/, "") || "",
    config.model_name?.trim() || "",
    config.dimensions || 0,
    !!config.enable_cache,
    !!config.use_dimensions,
    config.max_cache_size || 0,
    config.max_input_length || 0,
    config.max_batch_size || 0,
    config.health_check_timeout || 0,
  ]);
}
```

The rebuild button is then disabled by this comparison:

```tsx
const hasUnsavedEmbeddingChanges =
  persistedEmbeddingFingerprint !== undefined &&
  getEmbeddingConfigFingerprint(embeddingConfig) !==
    persistedEmbeddingFingerprint;

// ...

disabled={
  !embeddingEnabled || reindexing || hasUnsavedEmbeddingChanges
}
```

## Additional Notes

The backend scoped rebuild endpoint still exists; the problem is that the Console prevents access to it:

```text
POST /api/agents/{agentId}/memory/reindex?scope=embedding
```

No index rebuild was executed while investigating this issue.

Possible fixes:

1. Preserve `use_dimensions` in the watched form state even when its control is hidden; or
2. Normalize backend-inapplicable fields consistently before calculating both the persisted and current fingerprints. For example, if `use_dimensions` only applies to the OpenAI backend, normalize it to `false` for other backends on both sides of the comparison.

A regression test should cover at least:

```text
backend=dashscope
persisted use_dimensions=true
freshly loaded form
→ must not be marked unsaved
→ Rebuild Embedding index button must be enabled
```

The existing Console test uses the OpenAI backend and omits `use_dimensions`, so both sides default to `false` and this DashScope case is not covered.

jinliyl:
## Description

The Console fingerprinted `use_dimensions` for every embedding backend even though the field is only registered and applied for the OpenAI backend. A persisted DashScope configuration with a stale `use_dimensions=true` was therefore compared against a watched form value that omitted the hidden field, permanently marking the form as unsaved and disabling manual embedding reindex.

The backend service and vector-space fingerprints had the same semantic mismatch: they treated `use_dimensions` as meaningful for DashScope even though model construction only passes it to OpenAI.

This PR:

- normalizes `use_dimensions` to its effective value in Console service/config fingerprints;
- applies the same normalization to backend staged-service and vector-space fingerprints;
- preserves distinct fingerprints for OpenAI configurations where the option is effective;
- adds a Console regression covering a freshly loaded DashScope config with persisted `use_dimensions=true`;
- adds backend regression coverage for both DashScope equivalence and OpenAI distinction.

Fixes #7464

## Security Considerations

No security-sensitive behavior changes.

## Testing

- `.venv/bin/python -m pytest tests/unit/agents/memory/test_embedding_model.py -q` — 10 passed
- `npm run test:run -- src/pages/Agent/Config/components/ReMeLightMemoryCard.test.tsx` — 32 passed
- `npx tsc -b --noEmit`
- targeted ESLint and Prettier checks
- `uvx pre-commit run --files <changed files>`

### DIFF
--- a/console/src/pages/Agent/Config/components/ReMeLightMemoryCard.test.tsx
+++ b/console/src/pages/Agent/Config/components/ReMeLightMemoryCard.test.tsx
@@ -7,7 +7,7 @@ import {
   within,
 } from "@testing-library/react";
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { useState, type ReactNode } from "react";
+import { useEffect, useState, type ReactNode } from "react";
 
 import { agentsApi, api } from "@/api";
 import { useAgentStore } from "@/stores/agentStore";
@@ -64,6 +64,19 @@ const memoryStatus = {
 const unknownRuntime = { type: "unknown" as const };
 const unknownDiagnostics = { type: "unknown" as const };
 const noopStatusCheck = async () => {};
+const persistedDashScopeEmbeddingConfig = {
+  backend: "dashscope" as const,
+  model_name: "text-embedding-v4",
+  api_key: "secret",
+  base_url: "https://dashscope.aliyuncs.com/compatible-mode/v1",
+  dimensions: 1024,
+  enable_cache: true,
+  use_dimensions: true,
+  max_cache_size: 1000,
+  max_input_length: 8192,
+  max_batch_size: 10,
+  health_check_timeout: 15,
+};
 
 function RuntimeProvider({ children }: { children: ReactNode }) {
   const [localReindexing, setLocalReindexing] = useState(false);
@@ -216,6 +229,40 @@ function PersistedEmbeddingForm() {
       }}
     >
       <ConfiguredEmbeddingForm />
+    </MemoryMaintenanceContext.Provider>
+  );
+}
+
+function PersistedDashScopeEmbeddingForm() {
+  const [form] = Form.useForm();
+
+  useEffect(() => {
+    form.setFieldsValue({
+      reme_light_memory_config: {
+        embedding_model_config: persistedDashScopeEmbeddingConfig,
+      },
+    });
+  }, [form]);
+
+  return (
+    <MemoryMaintenanceContext.Provider
+      value={{
+        needsReindex: false,
+        setNeedsReindex: vi.fn(),
+        reindexing: false,
+        setReindexing: vi.fn(),
+        persistedEmbeddingFingerprint: getEmbeddingConfigFingerprint(
+          persistedDashScopeEmbeddingConfig,
+        ),
+        openMemorySettings: vi.fn(),
+        runtimeStatus: unknownRuntime,
+        diagnosticsStatus: unknownDiagnostics,
+        checkMemoryStatus: noopStatusCheck,
+      }}
+    >
+      <Form form={form}>
+        <EmbeddingModelCard />
+      </Form>
     </MemoryMaintenanceContext.Provider>
   );
 }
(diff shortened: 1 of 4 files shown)


lalaliat (approved):
LGTM

PR #7458, opened 2026-09-01
fix(desktop): bundle ReMe entry-point plugins

jinliyl:
## Description

QwenPaw 2.2.0b5 enables the independently distributed `auto-fin` and `daily-paper` ReMe plugins, but the desktop PyInstaller spec only collected data from the core `reme` package. Because ReMe discovers those plugins through `importlib.metadata` entry points and imports their manifest targets dynamically, the frozen desktop bundle omitted their Python modules, `plugin.yaml` resources, and distribution metadata. ReMe initialization consequently failed, left the manager's `_reme` instance unset, and embedding reindex surfaced a secondary `NoneType.update_component` error.

This PR:

- collects both ReMe plugin packages, their data files, and their distribution metadata in the desktop bundle;
- returns the existing unavailable result when reindex is requested before ReMe has started, so the API responds with a meaningful 503 instead of an `AttributeError`;
- adds PyInstaller-spec regression coverage for all three entry-point requirements;
- extends the installed-desktop verifier with a provider-free BM25 reindex, proving that the packaged ReMe runtime can start and execute a job.

**Related Issue:** Fixes #7446

**Security Considerations:** No security-sensitive behavior changes. The desktop verifier only rebuilds the empty/default test workspace's local BM25 index.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [x] CI/CD
- [x] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [x] Documentation updated (not needed; packaging/runtime-only fix)
- [x] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

Not applicable.

## Testing

```bash
.venv/bin/python -m pytest \
  tests/unit/tauri/test_pyinstaller_spec.py \
  tests/unit/agents/memory/test_reme_embedding_update.py \
  tests/unit/app/routers/test_agents_router.py -q
```

[...]

PR #7468, opened 2026-09-01
fix(memory): start ReMe before model configuration

jinliyl:
## Description

A fresh desktop installation has no active model until onboarding completes. `ReMeLightMemoryManager.start()` currently injects the active QwenPaw model before starting ReMe, and lets `ProviderError` escape when no model is configured. Because the workspace memory service is optional, ServiceManager then removes it entirely.

This caused both desktop release jobs in run #33484573427 to fail the provider-free BM25 check added by #7458 with:

```text
HTTP 503: {"detail":"Memory manager is not available"}
```

This change treats model configuration errors as a supported startup state: ReMe starts with its placeholder LLM component so provider-free jobs such as BM25 reindex remain available. Jobs that need an LLM already refresh the injected QwenPaw model immediately before execution. Unexpected model injection errors still propagate.

Follow-up to #7458.

## Testing

```bash
.venv/bin/python -m pytest \
  tests/unit/agents/memory/test_reme_start.py \
  tests/unit/agents/memory/test_reme_embedding_update.py \
  tests/unit/app/routers/test_agents_router.py -q
# 79 passed

uvx pre-commit run --files \
  src/qwenpaw/agents/memory/reme_light_memory_manager.py \
  tests/unit/agents/memory/test_reme_start.py
# all passed
```

## Security Considerations

No security-sensitive behavior changes. The change only keeps the local memory runtime available before model onboarding.

### DIFF
--- a/.github/workflows/fork-verify-desktop.yml
+++ b/.github/workflows/fork-verify-desktop.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.11"
 
       - name: Install packaging helper dependencies
         run: python -m pip install packaging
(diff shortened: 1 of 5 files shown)


cuiyuebing (approved):
LGTM

cuiyuebing (approved):
LGTM

PR #7274, opened 2026-08-25
feat(creator) 1.1.1: live website and desktop operation with take recording, Bailian Wan3 video, APE-benchmark review operators, and a semver-valid version floor

xuanrui-L:
## Description

This PR carries the QwenPaw Creator app-plugin work produced in the fork since the last upstream sync. Everything through fork commit `2eca28915fd5ec7114b6d735df50fe34d959e9fb` already landed here as the squashed commit `63799c1f` (creator 1.1.0), so this PR contains only the five changes built on top of that baseline.

Because a squash merge drops the fork-side ancestry, a plain re-merge would report dozens of false conflicts. The branch therefore records the squash equivalence explicitly: a merge commit whose tree is `63799c1f`'s tree and whose parents are `63799c1f` and `2eca2891`. The equivalence was verified rather than assumed — the patch `2eca2891..63799c1f` is byte-identical to the upstream range `766c9337..82af6b79`, so the squash introduced no fork-side drift. `upstream/main` at `9593d16c` is then merged in with zero conflicts, and every file in the resulting diff lives under `plugins/apps/qwenpaw-creator/`; the host tree is byte-identical to `upstream/main`.

What the five changes contain:

- **Live website and desktop operation with take recording.** The Creator main agent gains `browser_use`, which drives the main repo's Browser SDK in-process, and a symmetric `computer_use`, which drives the host's native Computer Use runtime from agent-authored async Python. The agent decides when to record, so only start/stop bounds are filmed and takes carry no dead footage: browser takes attach a one-way CDP screencast to the very browser being driven, desktop takes use an ffmpeg system capture (avfoundation/gdigrab) cropped to the window bounds, and both assemble frames with the existing ffmpeg dependency. Takes and screenshots publish as ordinary Project source material through the existing asset path — no new element, creation type, command, TaskKind or specialist. Each take carries a manifest sidecar with machine-precision coordinates and instants of every action, which motion design reads to place emphasis where operations actually happened. Tool guidance reuses the host's own browser and computer-use SKILL text verbatim and appends only Creator's recording contract.
- **Bailian Wan3 video generation with exact capability enforcement.** Adds the wan3.0-video and wan3.0-video-prime endpoints and tightens video-model capability declarations so a model is dispatched only for the modes it actually supports.
[...]

PR #182, opened 2026-03-01
feat(cli): add `copaw service` for system service management

Innocent-children:
…-start

Add cross-platform support for registering CoPaw as a system service:
- Linux: systemd user/system service
- macOS: launchd LaunchAgent
- Windows: Task Scheduler

New commands: copaw service install/uninstall/start/stop/status/logs

Also updates install scripts to auto-register the service after install, and uninstall command to auto-clean the service before removal.

🤖 Generated with [Qoder][https://qoder.com]

## Description

[Describe what this PR does and why]

**Related Issue:** Fixes #(issue_number) or Relates to #(issue_number)

**Security Considerations:** [If applicable, e.g. channel auth, env/config handling]

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [x] CLI
- [x] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [x] Scripts / Deploy

## Checklist

- [x] Pre-commit hooks pass (`pre-commit run --all-files` or CI)
- [x] Tests pass locally (`pytest` or as relevant)
- [x] Documentation updated (if needed)
- [x] Ready for review

## Testing

[How to test these changes]

## Additional Notes

[Optional: any other context]

PR #175, opened 2026-03-01
feat(security): add permission approval system for high-risk tool calls(#174)

davidfrz:
## Description
Fixs #174 
  Add a permission approval gate that requires human review before the agent executes high-risk tools (shell commands, file writes, browser operations). Supports two modes: `auto` (all approved, for dev) and `manual` (human review required).

  The approval check is implemented by overriding `CoPawAgent._acting()`, keeping AgentScope's original `Toolkit` class untouched.

  **Related Issue:** Fixes #(174)

  **Security Considerations:** This PR is specifically about security — it adds the missing safety gate between agent decisions and dangerous tool execution. Approval mode and timeout are configurable via `config.json`.

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [ ] Documentation
  - [ ] Refactoring

  ## Component(s) Affected

  - [x] Core / Backend (app, agents, config, providers, utils, local_models)
  - [x] Console (frontend web UI)
  - [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
  - [ ] Skills
  - [ ] CLI
  - [ ] Documentation (website)
  - [x] Tests
  - [ ] CI/CD
  - [ ] Scripts / Deploy

  ## Checklist

  - [x] Pre-commit hooks pass (`pre-commit run --all-files` or CI)
  - [x] Tests pass locally (`pytest` or as relevant)
  - [ ] Documentation updated (if needed)
  - [x] Ready for review

  ## Testing

  1. **Auto mode (default):** Start `copaw app` normally — all tool calls execute as before, no behavior change.

  2. **Manual mode:** Set `config.json`:
     ```json
     { "approval": { "mode": "manual", "timeout": 120 } }
     Then chat with the agent and ask it to run a shell command. The frontend will show an ApprovalModal popup. Click Approve/Deny to control execution.

  3. SSE endpoint: Open http://localhost:8088/api/approvals/stream/events in browser to see real-time approval request events.
  4. API directly:
  # List pending
  curl http://localhost:8088/api/approvals
  # Approve
  curl -X POST http://localhost:8088/api/approvals/{id} -H 'Content-Type: application/json' -d '{"reply":"approved"}'
  5. Unit tests: pytest tests/unit/test_approval_service.py -v (11 tests)

  Additional Notes

  - High-risk actions gated: execute_shell_command, write_file, edit_file, append_file, browser_use
[...]

PR #7486, opened 2026-09-02
feat(creator) 1.2.0: blueprint workbench, runtime notification bus + async delegation, multi-timeline A/B compare, custom video templates, MiniMax H3 video, agent-visible snapshot rollback, in-process locking, Windows hardening, and Docker deployment

xuanrui-L:
## Description

This PR carries the QwenPaw Creator app-plugin work produced in the fork since the last upstream sync, taking the plugin from **1.1.1 to 1.2.0**. Everything through fork commit `aee0eed1446cfff46fb933e87d96882a67bb9c18` already landed here as the squashed commit `77c2e5f87` (creator 1.1.1, #7274), so this PR contains only what was built on top of that baseline: **365 commits, 331 files, +52,900 / −9,926**, every one of them under `plugins/apps/qwenpaw-creator/`.

**Baseline recording (graft).** Because a squash merge drops the fork-side ancestry, a plain re-merge would report dozens of false conflicts. The branch therefore records the squash equivalence explicitly as commit `7f9539c86` — `chore(sync): record upstream squash baseline for creator 1.1.1` — whose tree is byte-identical to `77c2e5f87`'s tree and whose parents are `77c2e5f87` and `aee0eed14`. The equivalence was verified rather than assumed: `git diff 77c2e5f87 aee0eed14 -- plugins/apps/qwenpaw-creator/` is empty, and the squash touched no path outside the plugin, so it introduced no fork-side drift. Merging that graft into the fork head changed **zero** plugin files. `upstream/main` at `03e103534` was then merged in with zero conflicts.

**Where the baseline stands today.** `upstream/main` has since advanced 32 commits to `e2a32455`, so the merge base is no longer upstream HEAD. That does not change what this PR contributes: the diff against the merge base touches nothing outside the plugin, and **none of those 32 upstream commits touch `plugins/apps/qwenpaw-creator/` at all** (they are `console/src`, `src/qwenpaw`, `tests/unit`, `website/`, `plugins/bundle`), so there is no overlap to resolve. GitHub currently reports the PR `MERGEABLE`; the only pending gates are review approvals.

The work splits into two waves. Wave 1 was published in the fork as 1.1.2, wave 2 as 1.2.0.

### Wave 1 — published as 1.1.2

- **Runtime notification bus and asynchronous `delegate_to_agent`.** The work-graph scheduler was invisible to the mainline Agent: once a run ended with only machine-dispatchable work left, node completions, whole-graph DONE and deterministic failures such as `IMAGE_REFERENCE_BUDGET_EXCEEDED` never reached a model turn, and projects stalled silently — a field trace measured roughly 89 of 129 minutes idle, including a 60-minute scheduler sleep. Runtime→Agent feedback now goes through one bus with two delivery primitives: `steer` (subagent terminal states, deterministic node failures, transient budget exhaustion, whole-graph completion) writes a durable user-role RUNTIME session message and wakes the dispatcher, while `quiet` progress (node start/finish/gated) stages in a per-project append-only outbox at `runtime/notifications/outbox.jsonl` and rides along with the next steer, the end-of-run digest, or a turn boundary. `delegate_to_agent` returns `{status: ACCEPTED, runId}` immediately and runs the specialist detached, with its terminal outcome arriving as a steer; specialists fence on their own token rather than the project epoch so a new mainline run cannot kill one in flight. Folded in from the same field work: a same-thread nesting guard that false-positived on recycled `to_thread` workers and crashed delta persistence, startup orphan reclaim that was failing still-running media executions (now restricted to chat-delegated runs), a FAILED storyboard node outside the deterministic-error set that stalled a project across restart, and `websocket-client` 1.9.x `WebSocketApp.close()` races in TTS that surfaced as `AttributeError` instead of a provider disconnect.
- **Multi-timeline support with A/B compare and automatic snapshots.** A project can now hold several cuts. Before any commit whose candidate touches elements under a timeline, a frozen deep copy of the base timeline is injected as `snapshot:{timeline_id}:{n}`, with element IDs, output slot IDs, `render_source`/transition references and scene-ledger rows remapped under the snapshot prefix so IDs cannot collide across timelines — agent edits are non-destructive by default and the pre-edit state is directly comparable. Snapshotting is wired into both commit paths (the REST patch route and the agent tool paths). `compose:final` becomes one compose node per content-bearing timeline with a matching per-timeline `final_video` render slot. The UI gains a timeline switcher (create/rename/duplicate/delete plus a compare toggle) and side-by-side A/B canvases sharing one playhead.
- **T2V/I2V/S2V in the work-graph scheduler, with the review fence and stale-transaction fixes.** The scheduler could previously auto-dispatch only R2V video generation; text-to-video, image-to-video and digital-human elements needed manual or agent-driven execution and blocked unattended completion. They now derive and dispatch as first-class nodes, `VALIDATION_ERROR` joins the deterministic-error set so admission failures cannot loop, and frozen-input currency checks extend to all three. Two correctness fixes ride along: storyboard node construction was left outside the `creation_type == "r2v"` branch that binds its dependencies, so deriving a graph for any new element type raised `UnboundLocalError`; and the media-review fence never actually matched, because reviewing artifact slots (`asset:{id}:variant:{vid}:image`) were compared against owner refs (`asset:{id}`) by exact membership — the fence is now resolved at the tick site, so heavy billed nodes (storyboard/video/compose) are blocked while any review is pending and visual/lineup nodes are blocked when their own target slot is under review. Separately, a `PROJECT_REPLACED` transaction whose `final_etag` no longer matches the current authority etag is auto-aborted instead of raising forever, which previously bricked a project permanently after a crash between the `project.json` write and finalization.
- **Professionalized media prompt generation, and two specialists retired into bundled skills.** Prompts are now authored once with canonical `[Image N]` reference markers and rendered per provider at submit time (图N for wan3/qwen-image, 图片N for Seedance, `@image_N` for Kling, `characterN` for wan2.6, ordinal prose for the gpt-image family, which documents array order only); unverified provider families fail closed rather than guessing a grammar. A prompt-contract layer detects prompt gaps and drives a free repair turn in every review mode, folding full-/half-width punctuation for verbatim dialogue matching. The R2V generation director and visual development specialists are removed from the delegation registry — their doctrine moves into two bundled skills, `professional-media-prompts` and `visual-asset-design`, with structural contracts inlined into the mainline system prompt and `create_character_voice` promoted to a mainline tool; guard tests keep retired-agent wording off every agent-visible surface. Also fixed from the same field run: dispatch idempotency keys interpolated raw provider-qualified model names, whose `/` violated the safe-path-segment alphabet and caused *every* image/video/compose dispatch to be rejected (models are now digested into the fingerprint); a terminal recovery worker retried an unparseable `project.json` to 6000+ attempts at ~4/s (~90 MiB of log per hour) and is now terminal-and-log-once; the project loop idled out at 300 s without confirming the graph had drained, stranding three storyboards READY while the status bar claimed completion; automatic reference chains overran the image model's budget and now cap at the model limit, dropping least-identity-critical refs (props, then scene) with logging while explicit author lists still hard-error; and DashScope output-moderation refusals, previously misclassified as ordinary failures and retried against quota, are now recognized. Storyboard grids pad to the next perfect square so cells inherit the sheet ratio instead of distorting (6 panels on 16:9 used to yield 3×2, i.e. 32:27 cells).
- **Minimal in-process locking and lock-free reads.** The flock stack twice wedged production: steady UI polling took *shared* locks on every read and starved the commit and stop writers, which the dock surfaced as 「正在停止所有 Agent」 forever, and each incident had added another compensation layer rather than removing the contention. Cross-process lock files are gone — writers take a writer-priority in-process read/write lock keyed by normalized path, no `.lock` file is ever created on disk, and reads across the jsonl/atomic/session/execution/run/manual-edit stores are lock-free because records are published by atomic replacement (torn-tail repair stays exclusive to `append()`/`recover()`). Snapshot checksums are now computed lazily so polling stops paying a canonical dump plus sha256 per read, and project create/copy holds the global name lock only for the name check rather than across the asset-tree copy. This matches Creator's supported single-backend topology and the philosophy already documented in `src/qwenpaw/utils/io_utils.py`. Startup reclaims runs stranded by an unclean exit and garbage-collects tombstones, staging orphans, stale exports and flock-era artifacts; the legacy sweep is restricted to the retired implementation's own directories (it previously walked the whole data root and would have deleted a project asset named `poetry.lock`), and it is withheld entirely when a live peer backend is detected, so a misconfigured double start degrades to a race rather than one process deleting another's in-flight staging. `models/config.py` no longer swallows `LockTimeoutError` into an empty config, which used to surface as a misleading missing-API-key error.
- **Video template system with seven preset styles.** A registry of immutable presets — `vlog_daily`, `short_drama_cinematic`, `tutorial_clean`, `interview_pro`, `gaming_neon`, `travel_warm`, `product_showcase` — each bundling opening/closing caption blueprints, default transition kind and blend seconds, caption ordering, color grade, energy/density/decoration dials, an EditPlan design floor, decoration catalog, frame blueprint and preview copy. Exposed as read-only `GET /video-templates`; `POST /projects` accepts `templateId` and applies it at creation, rejecting unknown ids. The UI gains a template picker on the hero composer card.
- **Windows compatibility.** Creator failed on Windows during example-project install and project import. `shutil.unpack_archive` is replaced by explicit extractors that sanitize each zip entry name against the Windows-reserved character set and enforce a resolve-and-`is_relative_to` traversal guard raising `StorageIntegrityError`; extracted `project.json` has its `source:{id}` relative-URI references rewritten, since `:` is illegal in a Windows path. Absolute paths longer than 240 chars get the `\\?\` prefix, platform-gated so POSIX behavior is untouched. `os.fsync` of downloaded temp files is skipped on `os.name == "nt"` to tolerate EBADF. The example-archive build script sanitizes at build time too, so shipped archives are Windows-safe.
- **Cascade delete option for projects.** `DELETE /projects/{project_id}` gains `cascade` (default `true`, so the API contract is unchanged). With `cascade=false` only the manifest is unlinked — the project disappears from discovery and listings while assets, sessions and observability data stay on disk. The UI delete dialog now offers Cancel / Delete (manifest-only) / Delete All Data, where the destructive path requires a second explicit confirmation.
- **Dark-mode readability and live system-theme following.** Hardcoded light backgrounds, borders and text clashed with dark tokens across roughly fifteen components and pages, the hero backdrop stayed white, and an OS theme change required a reload. Theme tokens now back the `dark:` variants, `.hero-bg` and the home fade overlay bind to `var(--background)`, a white wordmark asset swaps in for dark mode, and a `prefers-color-scheme` listener re-applies the resolved theme live when theme is set to system.
- **Pet-OS subtitle bubble restricted to pet content.** The animal-emoji speech-bubble subtitle style was leaking onto non-pet projects such as interviews. The overlay kind is now `pet_os` only when the project's `content_type` is `pets`, and the live preview applies the identical predicate so preview matches final render.
- **Docker deployment.** A two-stage `Dockerfile` builds the UI SPA on node-slim and layers the plugin onto the official QwenPaw image, an `entrypoint.sh` performs first-run init and an idempotent bundled-plugin install, and a supervisord template runs dbus/Xvfb/xfce4 alongside `qwenpaw app`. A build script plumbs the runtime-boundary build arg from the host version file.

### Wave 2 — published as 1.2.0

- **Blueprint workbench: a full redesign of the authoring surfaces.** The blueprint/script/rough-cut/workbench stack was rebuilt around one project-blueprint page: a single-episode board and a multi-episode card grid, script review, a rough-cut preview strip, and a cinema mode that plays the whole film sequentially. Underneath it a real script pipeline appeared — bidirectional markdown parsing, `timeline_script` version write-back, structure/script checkpoints and dedicated `work_graph` script nodes — plus zero-cost rough-cut *draft* rendering exposed at `GET /projects/{id}/timelines/{tid}/rough-cut` (the store verifies containment and a sha256-confirmed copy before serving, ffmpeg runs off the event loop via `asyncio.to_thread`, and a render failure maps to 409 rather than 500). The workbench page itself moves to R2V v4 with a capsule prompt editor and a related-asset picker; PlanPage is rearranged around an ElementDetail right rail; the asset library becomes a full-page view with voice audition and regeneration plus manual DAG dispatch (reported through the `quiet` outbox). Four correctness guards came out of review: `narrative_timeline_ids()` is now the single source of truth that keeps `snapshot:*` timelines out of every work-graph lane, out of checkpoint timeline counts and out of the script prompt's narrative context (snapshots previously also generated storyboard/video/compose nodes, i.e. paid work that could never be produced); normalized rewrite guidance joins the script `_request_fingerprint` together with `strategy.audience` and the material-understanding version, so an explicit user instruction is no longer swallowed by semantic replay while a same-guidance network retry still replays; `selectFinalFilmVersionId` only returns a selected non-stale render version when *exactly one* live timeline exists, so a multi-episode project can no longer pass off its newest single episode as the whole film; and snapshot ids allocate past deleted suffixes, timeline deletion cascades its snapshots, and narration rebinding fails closed. Schema moves to v9 with a migration, and both locales are aligned key-for-key.
- **Narration regeneration and a referenced-voice card.** `POST /timelines/{tid}/elements/{eid}/narration` drives the TTS executor directly (not through the agent), taking synthesis input from the element's copy and rate plus the voice identity recorded on the current audio version, then rebinds the element and emits a `NARRATION_REGENERATED` quiet notification so the agent sees the active version change. Snapshot timelines are rejected. The element-detail voice row becomes a referenced-voice card (voice name, model, registered character sample audition), and asset-library voice cards degrade to an explicit prompt to regenerate when legacy rows have no sample or description instead of disappearing.
- **Custom video templates: save, reuse, delete.** The seven built-in presets are joined by user templates extracted from an existing project's timeline and persisted as individual JSON files under `$CREATOR_DATA_ROOT/user-templates/`. `POST /video-templates` derives one from a project (optionally a specific timeline), `GET /video-templates` returns built-ins and user templates in one list, and `DELETE /video-templates/{template_id}` removes a user template. Ids are server-generated (`user:` + uuid4 fragment) — clients never supply one — the delete route rejects any id without the `user:` prefix with 400 so a built-in preset cannot be removed, malformed persisted JSON is skipped on list rather than turning the endpoint into a 500, and unknown `templateId` on project creation is still rejected. The UI gains a save-as-template dialog and hover-delete on user cards.
- **Style selection becomes a toolbar pill with a thumbnail card grid.** The inline emoji template row squeezed the composer and the old text dropdown gave no visual preview, so style selection moves to a 「风格」 pill between the scenario pills and the resolution selector, opening a four-column card grid: a default-style card, built-in styles rendered from seven shipped 480w JPEG thumbnails under `ui/src/assets/design/style-thumbs/` (292 KB total, mapped by `templateId`), and user templates on an emoji placeholder. The popover shift-aligns and scrolls rather than only flipping — antd's default overflow handling pushed the 700px panel off the left edge (measured x = −176 / −296) and off the top (y = −86) in shrunken windows — verified across 1440×900 / 1024×700 / 800×600 / 640×480.
- **MiniMax H3 video: official v2 API plus an SGLang self-hosted channel.** `MiniMax-H3` / `H3-Max` ride the existing MiniMax protocol and branch onto the v2 content API by model name: multimodal content array, 480P/768P/2K, 4–15 s, omni references (≤9 images + 3 videos + 3 audio), prompt ≤7000 chars, and result polling inline from `task.content.url` with no file-retrieve step. A new `minimax_sglang` backend speaks the open-weight SGLang schema (`POST /v1/videos` with task/conditions/target, result bytes from `/v1/videos/{id}/content`); because one server instance loads one checkpoint variant, the configured model name (`MiniMax-H3-FL2VA` / `-Ref2VA`) records the deployed variant and the other modes fail closed rather than silently producing the wrong conditioning. Self-hosted endpoints skip the host allowlist, may run keyless (Bearer is sent only when `sglang serve --api-key` is set, including on the content download), and inline local media of every kind as data URIs. Cloud H3 accepts `"auto"` as the legacy alias of `adaptive`, following the Seedance precedent; sglang t2v aligns with the cloud contract (an empty ratio pins 16:9, `adaptive`/`auto` rejected); provider-supplied task ids are URL-escaped before path interpolation.
- **Snapshot rollback is now visible to the agent.** Applying a snapshot changed live timeline content with nothing telling the Agent — no `RuntimeEventKind` covered user edits, and the mainline run only reads the Project snapshot at run start, so a run in flight kept reasoning about, and could rewrite, the content the user had just rolled back. The rollback also fell inside the ten-minute auto-snapshot dedupe window, whose premise (the newest auto-snapshot is this session's baseline) a rollback breaks: the abandoned branch stayed as the baseline while the rolled-back state, plus any manual touch-up on it, got no frozen copy at all. Rollbacks are now detected by content — the live element map equals a snapshot's, the same equality the frontend's 「已应用」 badge derives — and drive two things: one `TIMELINE_SNAPSHOT_RESTORED` steer into the session inbox (`NEXT_STEP`, not the quiet outbox, because the event invalidates the Agent's reasoning rather than merely reporting progress), and a process-local mark that lets the next agent write freeze one fresh baseline despite the window. Delivery runs *after* the lifecycle lock is released, since a steer appends a session message which takes the same lock on its shared side — notifying under the exclusive side stalled until the ten-second fuse expired. The mark is only peeked at decision time and settled once the commit is published; consuming it earlier lost it whenever the commit then failed CAS, leaving the retry with no baseline, which is the very gap being closed. Detection is gated behind a wholesale `elements_by_id` replace, measured at 275 ms/request saved on the blur-triggered edits that dominate this route.
- **Auto-snapshot bounded and frozen snapshots fail closed.** Two snapshot defects fixed: auto-snapshot would re-snapshot a snapshot timeline, so a writer mutating elements inside `snapshot:*` minted `snapshot:snapshot:*` copies, and because every nested id is new the ten-minute dedupe never bounded the growth (observed as 12-level nesting in a stuck project); and agent writes into frozen snapshots were accepted silently even though snapshots never enter the work graph, stranding content that could never be produced — now rejected with a steering error, `SNAPSHOT_TIMELINE_FROZEN`. Timeline deletion and latest-snapshot bookkeeping were corrected alongside, and the snapshot panel gained delete plus a pre-apply backup, store-held selection, and user-facing names that never leak internal timeline ids.
- **Unified progress UX, prompt synchronization and work-graph execution.** One pass over how progress, prompts and dispatched work agree with each other: cancelled model and scheduler state is preserved rather than reset, local media publication retries *without* repeating the provider call, and missing video inputs are gated before stale outputs so a missing asset is not reported as an out-of-date render. Prompts stay grounded in agent-visible contracts (with reference version numbering covered by test), R2V shots are replaced by unit narratives, and authored storyboard layout plus project reference order remain authoritative. Every edited prompt source is validated, and a frontend edit grace window is preserved before a synchronized commit wakes the scheduler. Continuation reviews are scoped to blocking work, internal reviews are retired during recovery rather than on reads, redundant provenance hashing under the write lock is removed, and an unused confirmation API is deleted. The waiting motion distinguishes itself and honors `prefers-reduced-motion`; native audio and narrative dialogue safeguards are preserved and negated speech is excluded from dialogue checks.
- **Compare preview buttons and video display.** The A/B compare preview controls and video element rendering in PlanPage were wired incorrectly; both fixed.
- **Version bumps.** `plugin.json`, `ui/package.json` and `ui/package-lock.json` move 1.1.1 → 1.1.2 → **1.2.0**. 1.2.0 is a minor rather than a patch because everything merged since 1.1.2 is feature-scale. The host compatibility floor `qwenpaw_version.min` stays at `2.1.0`. The bump matters mechanically: `.github/workflows/creator-release.yml` is version-driven and skips publishing when `qwenpaw-creator-<version>` already exists in the artifact index, so a release only happens if the version actually moves.

No third-party dependency is added anywhere in this range: the plugin `requirements.txt` is byte-identical to the merge base, `ui/package.json` changes only its `version` field, and the UI lockfile's parsed `packages` key set is unchanged — 414 entries on both sides with an empty symmetric difference, so not merely an equal count. The websocket hardening above deliberately fixes the problem in plugin code rather than pinning `websocket-client`, and is tested against both 1.8.0 and the locked 1.9.1.

**Related Issue:** None — this is a fork sync of Creator work already reviewed in the fork; baselines are referenced by commit hash only.

**Security Considerations:** The notification bus is bounded on both the autonomy and the cost axis: consecutive runtime-authored messages batch into a single run so queued notifications never fan out into one paid model run each, a hard cap of 10 trailing autonomous messages downgrades steers to the outbox until a human message resets the streak, and the idle-flush escape valve is cooldown-bounded (120 s, at most twice per human-free streak). Steer events stage in the outbox *before* delivery, so a transient session-write failure cannot lose a fact; outbox writes hold the Project lifecycle shared lock and re-check existence, so a write racing a DELETE cannot recreate a ghost directory; an all-DONE baseline is deliberately silent, because emitting it would replay one paid run per historically completed project after every deploy. Interrupt, stop and project-delete cascade-cancel detached specialists, including the new `TIMELINE_SNAPSHOT_RESTORED` steer path.

Paid-call fuses are tightened in the same range: `ensure_media_call_budget` now guards the S2V entry point, which the scheduler dispatches unattended, and the repaired review fence prevents re-billing generation while a human review is pending. Two new guards specifically protect against paying for work that can never be produced — snapshot timelines are excluded from every work-graph lane, and agent writes into a frozen snapshot fail closed with `SNAPSHOT_TIMELINE_FROZEN` instead of being accepted silently. Local media publication retries without repeating the provider call. Provider reference-marker grammars fail closed rather than guessing, and the SGLang backend fails closed on the modes its deployed checkpoint variant cannot serve.

The self-hosted MiniMax SGLang channel is the one place this range deliberately relaxes an egress rule, so it is scoped as narrowly as the code allows: the secure materializer still rejects every non-global address, and the *only* private origin it will accept is the exact scheme/host/port of the live `minimax_sglang` base_url, derived from current configuration at download time and never from durable task state. Peer pinning and per-hop re-resolution both stay, so a redirect to any other private host still fails. Keyless operation is available only for self-hosted endpoints; the cloud path is unchanged. Provider-supplied task ids are URL-escaped before path interpolation.

New read/write surface follows the existing containment model. The rough-cut endpoint serves only material the asset store has already verified inside the project boundary by sha256-confirmed copy, and returns 409 on render failure. User video templates are server-identified: ids are generated (`user:` + uuid4 fragment) rather than client-supplied, persisted filenames derive from that id with `:` folded to `_`, writes stay under `$CREATOR_DATA_ROOT/user-templates/`, the delete route 400s on any id lacking the `user:` prefix so built-in presets cannot be removed, and malformed persisted JSON is skipped on list instead of failing the endpoint.

Archive extraction — the other place in this range that writes user-controlled paths — gains an explicit traversal guard and entry-name neutralization on both import routes, and the Windows long-path prefix is platform-gated. Cascade delete keeps the API default destructive-compatible but makes the UI's plain Delete non-destructive, with full erasure behind a second confirmation; note the retained-manifest behavior if you have data-retention expectations.

For the Docker files: the app binds `0.0.0.0` inside the container and supervisord runs as root, authentication is *not* force-enabled (the entrypoint prints an explicit security notice when neither `QWENPAW_AUTH_ENABLED` nor `COPAW_AUTH_ENABLED` is truthy, and the build script's examples bind loopback), first-run init passes `--accept-security`, the force-update path refuses to run when the working dir is empty or `/`, and the default base image and apt/PyPI mirrors point at Alibaba registries, overridable via build args. No host-side auth surface, route exposure or code changes anywhere in this range.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [x] Scripts / Deploy
- [x] Creator app plugin (`plugins/apps/qwenpaw-creator`) — the entire diff is confined here

## Checklist

- [x] Ready for review

## Testing

```bash
# Host package supplies the shared framework deps, exactly as CI does
pip install -e ".[test]"
pip install -r plugins/apps/qwenpaw-creator/requirements.txt

# Creator plugin backend suite
cd plugins/apps/qwenpaw-creator/backend
python -m pytest tests -v --timeout=120

# Creator plugin UI suite, type check and production build
cd plugins/apps/qwenpaw-creator/ui && npm ci
npm run typecheck && npm test && npm run build

# Repository-wide hooks on the merged tree
pre-commit run --all-files
```

Every one of these was executed by CI on the exact head commit `7e7966447` of this PR — see Evidence for the runs and the numbers. No host-side suite was run because this PR contributes no host-side change: `git diff --name-only $(git merge-base upstream/main HEAD) HEAD` returns only paths under `plugins/apps/qwenpaw-creator/`.

Note for reviewers running the backend suite: it must be executed with the host package importable (the `pip install -e ".[test]"` step above, which is what `.github/workflows/creator-app-tests.yml` does). Running it in a bare plugin-local environment produces failures that are artifacts of the missing host, not defects — `test_live_operation.py::test_creator_browser_sessions_are_explicit_and_host_safe` cannot import `qwenpaw.browser.runtime`, the motion-clip render test cannot find a Playwright driver, and the idle-flush budget test's `caplog` assertion depends on the logging configuration the host install establishes. All three pass in the CI-equivalent environment.

The suite is also timing-sensitive on slow runners. `tests/render_review/test_review_loop.py::test_revise_verdict_sends_feedback_and_caps_rounds` drives `MAX_REVIEW_ROUNDS` (= 3) full review rounds, and its challenge-question generation reaches a real text-model endpoint that answers HTTP 401 in CI (no API key) — two such attempts are visible in the log of the failed run below. On a loaded runner the round loop can cross its own `--timeout=120`. This is disclosed rather than hidden because it did happen once, on the fork's copy of the same commit; the rerun and the upstream run both pass, and the timing evidence is in the next section.

## Evidence

All results below are from CI runs on head commit `7e7966447d44ec8800999f95a31b5ba542c1c4af` of this PR. They are reproducible by opening the linked runs.

**Upstream checks on this PR** (`Creator App Tests`, run 34180815212):

```
Creator Backend Tests   pass  3m35s
==== 1438 passed, 23 skipped, 23 deselected, 1 warning in 68.51s (0:01:08) ====

Creator UI Build & Tests  pass  1m36s
> tsc --noEmit                                  # no diagnostics
 Test Files  51 passed (51)
      Tests  263 passed (263)
   Duration  51.12s
✓ built in 11.35s
Creator PawApp plugin package verified.
```

- Backend: https://github.com/agentscope-ai/QwenPaw/actions/runs/34180815212/job/101919414687
- UI: https://github.com/agentscope-ai/QwenPaw/actions/runs/34180815212/job/101919414636

**Other required checks on the same head, all green:** `Pre-commit` (13m14s, https://github.com/agentscope-ai/QwenPaw/actions/runs/34180815226/job/101919415887), `Real behavior proof`, `Release Window`, `stamp`, `console format check`, `website format check`, `Test Summary`, `Frontend Summary`, `PR Spam Gate`. Host-side suites (`Unit Tests`, `Contract Tests`, `Integrated Tests`, `Coverage Report`, `Vitest Unit Tests`) report `skipping` — the change-detection jobs found no host-side change, which independently corroborates the scope claim below.

**Fork push CI on the identical commit** (run 34180811959), disclosed in full because the first attempt was red:

```
attempt 1  Creator Backend Tests  FAIL  5m32s
  FAILED tests/render_review/test_review_loop.py::test_revise_verdict_sends_feedback_and_caps_rounds
         - Failed: Timeout (>120.0s) from pytest-timeout.
  = 1 failed, 1437 passed, 23 skipped, 23 deselected, 1 warning in 189.19s =

attempt 2 (rerun of the failed job, same commit, no code change)  PASS  3m25s
  ==== 1438 passed, 23 skipped, 23 deselected, 1 warning in 69.26s (0:01:09) ====
```

- https://github.com/xuanrui-L/QwenPaw/actions/runs/34180811959

This is a runner-speed artifact, not a code defect, and the surrounding evidence pins it down:

- The rerun passes on the *same commit* with no change, and its suite duration (69.26 s) matches upstream's (68.51 s) — the failed attempt took 189.19 s for the same suite, i.e. that runner was roughly 2.8× slower.
- The fork's own runners are consistently slower than upstream's on this suite: the immediately preceding fork commit ran the identical code (differing only in three JSON version strings) green in 142.12 s, versus 68.51 s upstream.
- `#182`, the commit the failed run was executed on, changes exactly three lines — the `version` field in `plugin.json`, `ui/package.json` and `ui/package-lock.json`. It cannot affect a review-loop test.
- Both files implicated in the timeout are byte-identical to `upstream/main`: `git diff upstream/main HEAD -- plugins/apps/qwenpaw-creator/backend/tests/render_review/test_review_loop.py plugins/apps/qwenpaw-creator/backend/services/render_review/challenge.py` is empty.

**Scope and squash-equivalence verification** (re-run against the current head):

```
$ git rev-parse 7e7966447d44ec8800999f95a31b5ba542c1c4af          # this PR's head
7e7966447d44ec8800999f95a31b5ba542c1c4af
$ MB=$(git merge-base upstream/main 7e7966447); echo $MB
03e10353445ece617cbbc6420073825ae361d49d
$ git rev-parse upstream/main
e2a324559392fedf01a44eca71d4b4d79e272f19        # 32 commits past the merge base

$ git rev-list --count $MB..7e7966447
365
$ git diff --stat $MB 7e7966447 | tail -1
 331 files changed, 52900 insertions(+), 9926 deletions(-)

# Nothing outside the plugin is contributed by this PR
$ git diff --name-only $MB 7e7966447 | grep -v '^plugins/apps/qwenpaw-creator/' | wc -l
       0

# ... and upstream's 32 newer commits do not touch the plugin either, so there is no overlap
$ git diff --name-only $MB upstream/main -- plugins/apps/qwenpaw-creator/ | wc -l
       0

# Squash equivalence: the graft 7f9539c86 carries 77c2e5f87's exact tree ...
$ git rev-parse 7f9539c86^{tree} 77c2e5f87^{tree}
c25988c86d75a64325f4bbf56fe4e557fb67f73d
c25988c86d75a64325f4bbf56fe4e557fb67f73d
$ git rev-list --parents -n 1 7f9539c86
7f9539c862993e384db7187b44ffaf8a38717ea8 77c2e5f87ce1eff5d7f2015e44daacb7c64674f9 aee0eed1446cfff46fb933e87d96882a67bb9c18
                                          ^ parent 1: upstream squash             ^ parent 2: fork head

# ... upstream 77c2e5f87 carries the same plugin content as fork aee0eed14 ...
$ git diff --name-only 77c2e5f87 aee0eed14 -- plugins/apps/qwenpaw-creator/ | wc -l
       0
# ... the squash touched nothing outside the plugin ...
$ git diff --name-only 55b554aa6 77c2e5f87 | grep -v '^plugins/apps/qwenpaw-creator/' | wc -l
       0
# ... and merging the graft changed no plugin file relative to the fork head it merged into
$ git diff --name-only 353b517c3 48e45a433 -- plugins/apps/qwenpaw-creator/ | wc -l
       0

# Dependency footprint of the whole range
$ git diff --stat $MB 7e7966447 -- plugins/apps/qwenpaw-creator/requirements.txt | wc -l
       0                                          # untouched

$ git diff $MB 7e7966447 -- plugins/apps/qwenpaw-creator/ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "qwenpaw-creator-plugin-ui",
-  "version": "1.1.1",
+  "version": "1.2.0",
   "private": true,
   "type": "module",
   "scripts": {
                                                  # version field only; no dependency added or removed

# ui/package-lock.json: comparing the parsed "packages" key sets, not just their count
$ python3 -c '...set(json.load(...)["packages"])...'   # merge-base vs head
414 vs 414, symmetric difference empty                 # identical package sets

# Version actually published by this head
$ git show 7e7966447:plugins/apps/qwenpaw-creator/plugin.json | jq -r .version
1.2.0
```

## Additional Notes

**Scope of this PR versus the fork's open work.** The net diff is confined to `plugins/apps/qwenpaw-creator/`. Interactive-video creation (fork PR #175 — branching narrative domain model, choice-motion drafting, interactive-bundle assembly and export) is deliberately **not** part of this PR: it was split out of the blueprint workbench work and is still open in the fork. Reviewers should not expect `narrative_edges`, the `Interaction*` models, the branch canvas or the interactive-bundle endpoint here. If that work is later synced, it will come as its own PR.

**One stray build artifact in the diff.** Fork PR #172 accidentally committed `plugins/apps/qwenpaw-creator/ui/test-results/.last-run.json` (a Playwright last-run marker containing `{"status": "failed", "failedTests": []}`). It is tracked at head and therefore appears in this PR's diff, and it also rides into the release zip because `scripts/pack/generate_plugin_metadata.py` excludes `node_modules`, `.git` and `__pycache__` but not `ui/test-results`. It is inert at runtime — nothing reads it — but it is noise, and it leaks a stale local test status into a published artifact. The clean fix is `git rm --cached` plus a `pack_exclude` entry (or an `EXCLUDE_PATTERNS` addition). It is flagged here rather than fixed in this PR to keep the sync range free of unrelated edits; happy to land it as a follow-up if maintainers prefer it cleaned up before merge.

**Intermediate history, for transparency under a non-squash merge strategy.** One commit in the range (`2b9e5c780`) accidentally added three repo-root analysis artifacts, which a later commit (`250eafa59`) removed, so they do not appear in the PR diff but are visible if you inspect the range commit-by-commit. The same applies to the two workflow files that the fork added on its side of the baseline (`.github/workflows/creator-app-tests.yml`, `creator-release.yml`): they are byte-identical to what the merge base already has, which is why the scope check above returns zero non-plugin paths.

**Merge strategy.** A squash merge is recommended and is what the fork expects; the branch records its own baseline equivalence via the graft commit, so the squashed result stays consistent with the next sync. The PR is currently `MERGEABLE` with `mergeStateStatus: BLOCKED` on review approvals only.


cuiyuebing (approved):
LGTM

PR #7526, opened 2026-09-03
feat(agent): add protected execution contract

rayrayraykk:
## Summary
- add a protected execution, clarification, and authorization contract ahead of workspace prompt files
- move built-in skill, file-editing, and tool-use guidance out of dynamic environment context into a dedicated prompt module
- keep environment context limited to runtime facts

## Validation
- pre-commit run on all changed files

## Scope
- Goal mode is unchanged
- no Markdown files changed
- no tests added or modified
- research and proposal documents are excluded


src/qwenpaw/runtime/protected_prompt.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+"""Protected built-in system prompt content."""
+# flake8: noqa
+# pylint: disable=line-too-long
+
+PROTECTED_EXECUTION_CONTRACT_PROMPT = """\
```
cuiyuebing, line 6: Would it be better to include multi-language versions for this prompt

rayrayraykk: I'd rather not, since other prompt stays in English.

cuiyuebing (approved):
LGTM

PR #7569, opened 2026-09-05
feat(modes): add Advisor Mode

AntiQuality:
## Description

Adds **Advisor Mode**, a loop mode (picked per conversation from the chat input, like Goal and Mission) that pairs two models on one task: a stronger **advisor** model and the **worker** agent, which is running on a cheaper model.

Several features:
- **Opening plan.** On the first message of a conversation the advisor writes a strategic plan. QwenPaw appends it to the worker's context as a `consult_advisor` tool call with a fixed question and the plan as its result, so the worker reads the plan as an answer to a question it asked. Written once per conversation.
- **Mid-run intervention.** QwenPaw watches tool results. On repeated failures (defaults: 3 in a row or 4 in the last 10 steps, at most 3 times per user turn, all configurable) it asks the advisor again. The advisor answers `CONTINUE` (the chat shows the verdict, nothing is added to the worker's context) or `ADJUST` followed by a short revised plan, which is appended the same way as a `consult_advisor_followup` call and result.
- **On-demand consultation.** `consult_advisor` is a real tool in Advisor Mode. The worker can ask at a decision point (capped per conversation, default 32).

Model setup: by default the advisor is the agent's primary model (`active_model`, the one picked in the chat header) and the worker is the sub-agent model (`subagent_model`, set under Agent model routing).

Configuration:
| Key | Default | In the template |
| --- | --- | --- |
| `advisor_model` / `worker_model` | `null` (primary model / sub-agent model) | Advisor and worker models (read-only here, set from the Advisor models panel) |
| `advisor_thinking` | `off` | Advisor thinking |
| `plan_enabled` | `true` | Opening plan |
| `followup_enabled` | `true` | Mid-run auto intervention |
| `intervention.consecutive_failures` | `3` | Failures in a row |
| `intervention.window_failures` / `window_size` | `4` / `10` | Failures in the window / Window size (steps) |
| `intervention.max_interventions` | `3` | Maximum interventions per run (one user turn) |
| `intervention.cooldown_steps` | `0` | Cooldown (steps) |
| `on_demand_enabled` | `true` | On-demand consultation |
| `max_consults` | `32` | Maximum consultations per conversation |

**Security Considerations:** None

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [x] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

1. Agent → Configuration → Agent Loop Settings → **Advisor**: switch it on. The template shows every setting from the table above.
advisor-tab

2. Start a new chat. In the chat input, open the Loop mode menu and pick **Advisor**. Advisor is listed with the built-in loops.
advisor-menu

3. The **Advisor models** panel opens above the pill. Leave the defaults (primary model / sub-agent model) or pick other models, then send a task.
advisor-setup

4. The `consult_advisor` plan card appears immediately and streams while the advisor writes. Later turns get no new plan.
advisor-plan

5. The model pill in the chat header shows the pair (advisor → worker) and reopens the panel.
advisor-pill

6. Give the worker a task whose tool calls keep failing, for example cloning a repository that does not exist and retrying the same command. The trigger fires at the third failed call in a row (`consecutive_failures`, default 3). It is evaluated before the worker's next model call, so when the worker fires several calls at once (four clones in the screenshot) the card follows the whole batch. The `consult_advisor_followup` card shows the advisor's verdict. `CONTINUE` is only shown in the chat and nothing reaches the worker. `ADJUST` comes with a revised plan that is appended to the worker's context, so its next step is taken with the advice in view.
advisor-followup

7. `/advisor off`, `/new` or `/clear` leave the mode and drop the advisor's plan. With the **Advisor Mode** switch in the template off, `/advisor on` replies with where to enable it and the menu entry disappears.

## Evidence

PawBench performance with and without the advisor. Advisor: Qwen 3.8 Max with thinking on. Run on an earlier build of this branch (v2.1.0b2).

| Worker | Baseline | Advisor Mode | Δ |
| --- | --- | --- | --- |
| QwenPaw 9B | 65.1 | 68.0 | +2.9 |
| Qwen 3.8 27B (medium thinking) | 71.1 | 77.6 | +6.5 |

advisor_score

```bash
pre-commit run --all-files
# all hooks passed

pytest tests/unit -n 6
# 10580 passed, 18 skipped

cd console && npm run format && npm run test:run
# Test Files 331 passed · Tests 3118 passed
# format: no changes
```


src/qwenpaw/agents/react_agent.py
```diff
@@ -1061,6 +1072,32 @@ def _is_media_block(self, block: Any) -> bool:
             return mt.startswith(self._MEDIA_MIME_PREFIXES)
         return False
 
+    # ------------------------------------------------------------------
```
AntiQuality, line 1191: Only armed when a middleware (in Advisor mode) declares emits_injected_exchanges. For every other agent _reasoning_events returns the base generator unchanged, so the event stream, cancellation and teardown are untouched.

console/src/hooks/useToolCallControl.ts
AntiQuality, line 1: Generic change: the hook now reports whether the coordinator knows the call. Advisor's injected exchanges have no coordinator entry, so their gear actions returned 404. Ordinary tools keep managed=true.

console/src/components/Chat/ToolCards/shared/ToolCardShell.tsx
AntiQuality, line 1: Consumer of the flag added in `console/src/hooks/useToolCallControl.ts`. No change for tools the coordinator knows.

src/qwenpaw/runtime/builder.py
AntiQuality, line 1: New generic step: every registered mode may contribute request-scoped middlewares. Modes that do not override middlewares() return []. Default/Goal/Mission/custom loops are unaffected beyond one is_active lookup per request.

src/qwenpaw/modes/base.py
AntiQuality: Two new extension points for Advisor mode with no-op defaults. Existing modes and plugin modes inherit True / []

AntiQuality, line 1: Add `middlewares` extension for Advisor mode

rayrayraykk (commented):
Reviewed the Advisor Mode implementation with particular attention to model routing, per-conversation state, intervention accounting, injected event ordering, cancellation, configuration updates, and UI synchronization.

No blocking defect was confirmed, so no code changes were necessary.

Validation:

- `191 passed` across the focused backend Advisor Mode, router, configuration, runtime middleware, and injected-event tests.
- Focused Console tests could not run because `console/node_modules` is absent in this worktree.
- The worktree remains clean.

The implementation is ready for further review/merge, subject to the existing full Console CI suite passing.

src/qwenpaw/modes/advisor/middleware.py
```diff
@@ -0,0 +1,1145 @@
...
+            plan,
+            call_id=call_id,
+        )
+        agent.state.context.append(msg)
+        live.finish(plan)
+        return msg
+
+    # ── advisor call ────────────────────────────────────────────────────
+
+    async def _call_advisor(
```
niceIrene, line 1011: **[design] Give the advisor the worker's recent context instead of a growing private history**

The advisor only sees 3 clipped tool calls; everything else comes from replaying its own
`advisor_history`, which grows without bound — so consult #30 costs far more than
consult #1, and the advisor still never sees what the worker actually did.

Suggestion: drop `stateful=True`; each request includes the opening plan + the worker's
last N messages (clipped). One source of truth, ~constant per-call cost. Claude Code's
advisor works this way. Fallback if this is too big a change: truncate `advisor_history`
(plan + last N exchanges) and make `history_size=3` configurable.

src/qwenpaw/modes/advisor/mode.py
```diff
@@ -0,0 +1,347 @@
...
+            "workspace_dir",
+            None,
+        )
+        middleware = AdvisorMiddleware(
+            advisor=advisor,
+            trigger=InterventionTrigger(config=am.intervention),
+            plan_enabled=am.plan_enabled,
+            followup_enabled=am.followup_enabled,
+            on_demand_enabled=am.on_demand_enabled,
+            max_consults=am.max_consults,
```
niceIrene, line 238: **[non-blocking] `max_consults` bounds frequency, not spend — add a token fuse**

The count cap is worth keeping (it stops consult-spam, and a count is the one budget the
worker can reason about), but per-call cost varies, so staying under 32 says little
about actual spend. Suggestions:

1. `max_advisor_tokens` per conversation (0 = off), first cap hit wins. Covers
   auto-interventions too; exempts the opening plan.
2. Append "(N consultations left)" to each tool result so the worker can ration.
3. Log advisor token usage in the transcript so users can tune both caps.

src/qwenpaw/modes/advisor/prompts.py
```diff
@@ -0,0 +1,215 @@
+# -*- coding: utf-8 -*-
+"""Prompts used by Advisor Mode to talk to the advisor model.
+
+The system prompt carries every planning principle once, grouped by theme.
+The request templates only add what differs per request (the task and its
+context, the recent calls, the reply format), so no rule is stated twice.
+"""
+from __future__ import annotations
+
+ADVISOR_SYSTEM_PROMPT = (
```
niceIrene, line 10: **Add an append-only `ADVISOR.md`, keep the templates in code**

The request templates carry the CONTINUE/ADJUST contract that `_parse_followup` depends
on — they must stay in code. But the 11 principles are doctrine, and users have
project context a blind advisor badly needs (how tests run, which dirs are generated).
Suggestion: if a workspace `ADVISOR.md` exists, append it to the system prompt as
project-specific guidance — same mental model as `AGENTS.md`, and the default prompt
still ships/updates with the package instead of going stale in user workspaces.

src/qwenpaw/modes/advisor/middleware.py
```diff
@@ -0,0 +1,1145 @@
...
+        event = None
+        for bid, name, args, output in pending:
+            if bid:
+                self._seen_result_ids.add(bid)
+            if event is not None:
+                continue  # keep marking as seen, but only fire once per call
+            event = self._trigger.observe(name, args, output)
+        return event
+
+    async def _inject_followup(
```
niceIrene, line 810: **[design] Env context only exists in the plan request — follow-up and consult get none**

`_inject_plan` attaches a workspace listing via `_read_env_context()`, but
`FOLLOWUP_REQUEST_TEMPLATE` and `CONSULT_REQUEST_TEMPLATE` have no env section at all.
So at the moment the worker is stuck — often on an environment problem (missing binary,
offline network, wrong path) — the advisor diagnoses from 3 clipped calls alone, against
a mental model of the workspace from before the worker started changing it.

Suggestions, cheapest first:

1. Attach a *fresh* `_workspace_listing()` to follow-up and consult requests — the
   helper is pure and cheap; this is an `{env_section}` placeholder in two templates
   plus one format-arg. (Same applies to `consult()`.)
2. Add an optional `context` param to `consult_advisor` so the worker can pass env
   facts the advisor can't see; mention it in the tool description.
3. Follow-up PR: a small env probe at session start (OS, shell, key binaries, network)
   cached and attached to every advisor request — turns a class of "check first"
   hedging in plans into direct instructions.

PR #7577, opened 2026-09-06
fix(console): enqueue follow-up messages when chat task is running

### LINKED ISSUE #7559: [Bug]: 任务执行中时我通过对话框新发消息触发了409报错。新发的消息不是应该放到消息队列即可吗，为什么要触发报错？
## QwenPaw Version

2.2

## Description

任务执行中时我通过对话框新发消息触发了409报错。新发的消息不是应该放到消息队列即可吗，为什么要触发报错？

## Environment

- **QwenPaw version:** [2.2]
- **OS:** [Windows 11]

## Steps to Reproduce

1.执行一个任务
2.执行中时通过对话框提交文件
3.触发409报错

## Actual vs Expected

- **Actual:**触发409报错（409{"detail":"A task is already running for this chat. Wait for it to finish or use a different session_id."}）
- **Expected:**不要报错，新的消息不管是什么类型的，加入消息队列即可

kabishou11:
## What Problem This Solves

While a chat already had an active run, POST /console/chat returned HTTP 409 (A task is already running...) even for follow-up text/file messages. Users expected the new message to be queued rather than rejected.

## Description

When a session already has an active run, enqueue the new console payload through the channel manager instead of returning 409. The session queue consumer waits for the in-flight run to finish before starting the next turn.

**Related Issue:** Fixes #7559

**Security Considerations:** N/A — same console auth path; only changes busy-session handling.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Tests

## Checklist

- [x] I ran tests locally (pytest) and they pass
- [x] Ready for review

## Testing

pytest tests/unit/app/routers/test_console_chat_reconnect.py tests/unit/channels/test_base_core.py -q

## Evidence

Before: mid-run console POST returned 409 with detail that a task is already running.
After: the same request is enqueued; unit coverage asserts enqueue-on-active-run instead of 409.

### DIFF
--- a/src/qwenpaw/app/channels/base.py
+++ b/src/qwenpaw/app/channels/base.py
@@ -564,7 +564,8 @@ class BaseChannel(ABC):
                 exc_info=True,
             )
 
-        queue, is_new = await self._workspace.task_tracker.attach_or_start(
+        tracker = self._workspace.task_tracker
+        queue, is_new = await tracker.attach_or_start(
             chat.id,
             payload,
             self._stream_with_tracker,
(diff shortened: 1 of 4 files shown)


zhijianma:
@kabishou11 

Thanks for taking this on — enqueueing on the backend does resolve the immediate 409, so follow-up messages are no longer rejected. 

For reference, #7610 tackles the same issue from the frontend and does deliver the reply in real time — might be worth comparing the two.

Thanks again for the contribution!

kabishou11:
Thanks @zhijianma for the review and for pointing to #7610 — glad the frontend path landed the real-time delivery. Closing this in favor of that approach makes sense.

PR #7610, opened 2026-09-07
fix(console): prevent chat submissions from bypassing the queue

### LINKED ISSUE #7559: [Bug]: 任务执行中时我通过对话框新发消息触发了409报错。新发的消息不是应该放到消息队列即可吗，为什么要触发报错？

zhijianma:
## Summary

- add a lightweight chat run-status endpoint backed directly by TaskTracker
- route sender submissions, including attachments, into the existing localStorage queue while the chat is still running
- serialize async admission decisions so stale or reordered status responses cannot start concurrent POSTs
- snapshot chat, agent, and backend session identity before the status request and preserve it through queueing/direct send
- make the foreground queue wait for authoritative backend idle status before sending the next item
- keep HTTP 409 as the final server-side concurrency guard

## Root cause

The chat SDK clears its loading state when it receives the Completed SSE event, but TaskTracker releases the active run only after the remaining response-cycle cleanup finishes. A follow-up submitted during that gap bypassed the local queue and reached the backend while the same chat was still marked running.

The status preflight is asynchronous, so it also needs an admission gate and an immutable submission identity. Without those, reordered status responses can start two POSTs and a chat/agent switch during the request can retarget the pending message.

## Validation

- related Vitest suites: 163 passed
- related pytest suites: 8 passed
- `npx tsc -b --noEmit`: passed
- targeted Prettier and changed-files pre-commit checks: passed
- real Chrome against `http://localhost:5173`: second message persisted in localStorage, POST delayed until backend idle, both POSTs returned 200, persisted role order was user A → assistant A → user B → assistant B, no 409
- deterministic reordered-status browser probe: POST order FIRST → SECOND, maximum concurrent POST count 1
- deterministic session-switch browser probe: message retained the source backend session ID

Fixes #7559

### DIFF
--- a/console/src/api/modules/chat.test.ts
+++ b/console/src/api/modules/chat.test.ts
@@ -215,6 +215,29 @@ describe("chatApi CRUD", () => {
     );
   });
 
+  it("getChatStatus uses the lightweight status endpoint", async () => {
+    await chatApi.getChatStatus("chat/1");
+    expect(request).toHaveBeenCalledWith(
+      "/chats/chat%2F1/status",
+      expect.objectContaining({ signal: undefined, headers: undefined }),
+    );
+  });
+
+  it("getChatStatus forwards agent identity and AbortSignal", async () => {
+    const controller = new AbortController();
+    await chatApi.getChatStatus("chat-1", {
+      signal: controller.signal,
+      agentId: "agent-2",
+    });
+    expect(request).toHaveBeenCalledWith(
+      "/chats/chat-1/status",
+      expect.objectContaining({
+        signal: controller.signal,
+        headers: { "X-Agent-Id": "agent-2" },
+      }),
+    );
+  });
+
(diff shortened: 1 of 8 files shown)


cuiyuebing (approved):
LGTM

PR #7629, opened 2026-09-08
[DO NOT MERGE]: Gemini returns 400 "Requests ending with a model turn are not supported" after background tool completion

### LINKED ISSUE #7625: [Bug]: Gemini returns 400 "Requests ending with a model turn are not supported" after background tool completion
## QwenPaw Version

2.2.0

Verified with:

```bash
qwenpaw --version
```

Output:

```text
QwenPaw, version 2.2.0
```

## Description

When using a Gemini model, QwenPaw can fail immediately after an offloaded/background tool call completes or is interrupted.

The next model request returns:

```text
openai.BadRequestError: Error code: 400 - {
  'error': {
    'code': 400,
    'message': 'Requests ending with a model turn are not supported.',
    'status': 'INVALID_ARGUMENT'
  }
}
```

Investigation of the QwenPaw error dump shows that the final message in `agent_state.state.context` is an automatically injected background-tool notification with:

```text
role="assistant"
name="system"
```

For example:

```text
<system-notification>
Background tool call `execute_shell_command` (...)
completed with state=interrupted. Result below.
</system-notification>
```

This message is generated by `make_offload_hint_msg()` in:

```text
qwenpaw/tool_calls/_hint.py
```

QwenPaw 2.2.0 currently contains:

```python
return Msg(
    name="system",
    role="assistant",
    content=[notification] + result_blocks,
)
```

When this notification becomes the final context message before the next inference, Gemini sees the conversation as ending with a model/assistant turn and rejects the request with:

```text
Requests ending with a model turn are not supported.
```

QwenPaw 2.2.0 already appears to flatten the previous `ToolResultBlock` structure, so this is not the orphan tool-result pairing issue. The remaining issue is specifically that the offload completion notification is still emitted with `role="assistant"`.

A local workaround changing:

```diff
 return Msg(
     name="system",
-    role="assistant",
+    role="user",
     content=[notification] + result_blocks,
 )
```

successfully passes Python syntax validation and prevents the runtime notification from producing a trailing assistant/model turn.

Semantically, this notification is also closer to an external/runtime input than a model-generated assistant response.

**Related PR(s):** N/A

**Security considerations:** None identified.

## Component(s) Affected

* [x] Core / Backend (app, agents, config, providers, utils, local_models)
* [ ] Console (frontend web UI)
* [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
* [ ] Skills
* [ ] CLI
* [ ] Documentation (website)
* [ ] Tests
* [ ] CI/CD
* [ ] Scripts / Deploy

## Environment

* **QwenPaw version:** 2.2.0
* **OS:** Linux container
* **Install method:** Python virtual environment / pip-based installation
* **Python version (if applicable):** 3.11
* **Package path:** `/app/venv/lib/python3.11/site-packages/qwenpaw/`
* **Model/provider:** Gemini through an OpenAI-compatible API gateway
* **Channel where observed:** Feishu

## Steps to Reproduce

1. Configure QwenPaw 2.2.0 to use a Gemini model.
2. Start a conversation and invoke a tool such as `execute_shell_command`.
3. Let the command exceed the foreground execution period so that QwenPaw offloads it to background execution.
4. Wait for the background tool call to complete or become interrupted.
5. QwenPaw injects a completion notification generated by `make_offload_hint_msg()`.
6. QwenPaw performs the next model inference.
7. Gemini returns:

```text
400 INVALID_ARGUMENT
Requests ending with a model turn are not supported.
```

The corresponding error dump in my case was:

```text
/tmp/qwenpaw_query_error_rn1nsw14.json
```

Inspection of `.agent_state.state.context` showed the final messages as approximately:

```text
context[3]:
    role="assistant"

context[4]:
    role="assistant"
    name="system"
    content:
        <system-notification>
        Background tool call `execute_shell_command` (...)
        completed with state=interrupted.
        Result below.
        </system-notification>
```

The second message is produced by:

```python
def make_offload_hint_msg(entry: Any) -> Any:
    ...
    return Msg(
        name="system",
        role="assistant",
        content=[notification] + result_blocks,
    )
```

## Actual vs Expected

* **Actual:**

  After a background/offloaded tool call completes or is interrupted, QwenPaw inserts the completion notification as an `assistant` message.

  The resulting conversation ends with an assistant/model turn:

  ```text
  assistant
  → system-notification with role="assistant"
  → next Gemini inference
  ```

  Gemini rejects the request with:

  ```text
  400: Requests ending with a model turn are not supported.
  ```

* **Expected:**

  A runtime-generated background tool completion notification should not leave the conversation in a trailing assistant/model state before the next model invocation.

  For Gemini-compatible providers, the next inference should proceed normally after the background tool completion notification.

  One possible fix is to emit this runtime notification as an input/user turn:

  ```python
  return Msg(
      name="system",
      role="user",
      content=[notification] + result_blocks,
  )
  ```

  Alternatively, QwenPaw could handle this at the provider/formatter layer, but the automatically injected runtime notification should not result in an invalid trailing model turn.

rayrayraykk:
Fixed the Gemini background-tool completion failure.

The root cause was confirmed: completed and interrupted offloaded tool notifications were appended as `assistant` messages, leaving Gemini with a trailing model turn. I changed these runtime-generated notifications to `role="user"` while preserving the flattened result blocks and existing `name="system"` metadata.

Regression coverage now verifies the role for both successful completion and interrupted background execution.

Checks passed:

- `35 passed` in `tests/unit/tool_calls`
- Pre-commit checks, including mypy, Black, Flake8, and pylint
- `git diff --check`

A live Gemini gateway call was not run because it would require external network access and credentials.

Closes #7625

### DIFF
--- a/src/qwenpaw/tool_calls/_hint.py
+++ b/src/qwenpaw/tool_calls/_hint.py
@@ -13,8 +13,8 @@ def make_offload_hint_msg(entry: Any) -> Any:
 
     The hint flattens the finalized response content blocks (TextBlock,
     ImageBlock, etc.) directly into the message so that provider formatters
-    treat it as an ordinary assistant message — no ToolResultBlock means no
-    orphan ``role=tool`` wire message and no tool-call pairing issues.
+    treat it as ordinary runtime input — no ToolResultBlock means no orphan
+    ``role=tool`` wire message and no tool-call pairing issues.
     """
     end = entry.end_state or "unknown"
     notification = TextBlock(
@@ -31,6 +31,6 @@ def make_offload_hint_msg(entry: Any) -> Any:
     result_blocks = list(entry.final_response.content or [])
     return Msg(
         name="system",
-        role="assistant",
+        role="user",
         content=[notification] + result_blocks,
     )
--- a/tests/unit/tool_calls/test_coordinator_completion.py
+++ b/tests/unit/tool_calls/test_coordinator_completion.py
@@ -367,7 +367,7 @@ async def test_background_completion_emits_hint():
     )
 
     assert events[-1].metadata["offloaded"] is True
-    assert hint.role == "assistant"
+    assert hint.role == "user"
     text_block = next(
         block
         for block in hint.content
@@ -913,6 +913,11 @@ async def test_background_cancel_force_cancels_non_cooperative_tool():
     await asyncio.wait_for(entry.background_task, timeout=2)
     assert entry.force_cancelled is True
     assert entry.ctx.cancel_reason == CancelReason.USER
+    hint = await asyncio.wait_for(
+        _wait_for_hint(coordinator, "session-bg-cancel"),
+        timeout=2,
+    )
+    assert hint.role == "user"
 
 
 @pytest.mark.asyncio


rayrayraykk:
@qbc2016 I think this should be fixed in formatter instead of patching the code like this.

Block this PR for now.

PR #7645, opened 2026-09-09
test(e2e): rebuild the Environments suite for the unified env page (#7538)

yutai78786:
> **Submitted by**: 李时珍·E2E@QPQAT

## What this PR does

Upstream #7538 (`76bfb704`) rebuilt the Environments page from a table with
inline editing into a unified environment-variable page (three sections plus
modal-based editing). That left 11 cases in `e2e/tests/test_environments.py`
failing deterministically in nightly — each still red after 3-4 reruns, so it
was a real selector/interaction mismatch rather than flakiness.

This PR rebuilds that suite's row anchors, interaction model and assertions
against the new UI.

It also fixes one **independent defect that had nothing to do with #7538 and
had been latent for three months**: CROSS-005 in `e2e/tests/test_cross_module.py`
has navigated to a route that never existed (`/settings/runtime-config`, zero
hits across the whole frontend history) since the day it was migrated in
2026-05-29, and its step 4 only logged instead of asserting — so a blank page
still "passed". #7538 merely broke the other half of that case (the row count),
which is what exposed the older problem. The two root causes are different, so
they are described separately here to avoid reading this as blaming a
three-month-old mistake on the recent rebuild.

## Changes (3 files, +1733/-787)

| File | Change |
| --- | --- |
| `e2e/tests/test_environments.py` | Rebuild all 11 cases: row anchor `div[class*="__row__"]`; locate rows by key text (all three sections share the row class, so `.last` would pick the wrong section); `aria-label` action buttons (Edit/Delete/Reset); search box `input[aria-label="Search variables"]`; show-value button; modal `Apply now` for write confirmation; `Modal.confirm` for delete confirmation; validation copy `Invalid key format` / `Duplicate key` |
| `e2e/pages/environments_page.py` | Rebuild the page object in step. This file was previously dead code but still exported; it is now genuinely usable |
| `e2e/tests/test_cross_module.py` | Four fixes to CROSS-005: ① `goto` the authoritative route `/agent-config` (`builtinRoutes.tsx` maps `core.agent-config` → `/agent-config`, matching `runtime_config_page.py:31` and `test_runtime_config.py:23`) ② step 4 becomes a hard assertion that the `llmRetry` and `llmRateLimiter` tabs are `to_be_visible` (anchors `[data-node-key=...]` come from `Agent/Config/index.tsx` and are already proven green by `test_runtime_config.py`, so no new selector idiom is introduced) ③ new row anchor plus a real load wait ④ the lower-bound assertion is upgraded to a self-consistent invariant: row count == sum of the three section-heading counts. Two independent reads of the same quantity, so any anchor drift makes them disagree. Deliberately **not** hardcoded to 17, which would fail in both directions once upstream adds a section |

## Two cases have their intent redefined (please review this part)

The new UI **removed three product features** — multi-row checkboxes, row
insertion, and batch selection. This is not a selector mismatch: the thing
under test no longer exists. Both cases keep their class name and `test_id`
(so historical reports stay comparable and the selected set is unchanged) and
are redefined instead:

- **ENV-005** (`TestEnvVarMultiRowAndCheckbox`) now covers: create several
  variables in a row, locate each row by key, reveal a value, delete a single
  row, reload and confirm persistence.
- **ENV-008** (`TestBatchOperations`) now covers: create 3 rows, filter with
  the new search box, delete them one by one back to the baseline.

Both scope changes are recorded in the class docstrings and stated here
explicitly, so this cannot be read as silently deleting cases to go green.

## What CROSS-005 deliberately does not assert (stated in the docstring)

#7538 did connect the two layers (`AgentsRunningConfig` now reads its default
from `EnvVarLoader`), but that default is only consulted when an agent has no
persisted running config. For any already-configured agent the page shows the
stored value and never follows the environment. Asserting "change an env var →
the number on the config page changes" would be inherently flaky, so this case
only verifies that both pages render correctly and that navigating between
them does not interfere.

## Verification

1. **Fork CI green on all three shards** (head `14ddd621`): runs `34167593651`
   (p0), `34167599159` (p1), `34167604862` (p2), plus Pre-commit Checks and
   NPM Format. The previous head `8d08385f` was green on all three as well.
   p2 reported `0 failed` with **rerun=0** — before this fix those same 3 cases
   each reran 3-4 times and stayed red, so the green is not a retry artifact.
2. **No effect on the shard gate**: selection is `71/105/47`, and the collected
   node IDs are **identical to the baseline one by one** (verified by md5 of
   `--collect-only` output for each shard against the merge-base worktree).
   No marker was touched and no case was added or removed. The identity
   `deselected + selected = collected = 239` holds on all three shards.
3. **Local probes**: a static probe that reconstructs the DOM element by
   element from upstream source passes `probe_selectors.py` 47/47 and
   `probe_page_object.py` 14/14. Selectors are imported from the test module
   rather than retyped. The probes additionally caught two real defects of my
   own, now fixed: a confirm-button selector that split its alternatives with a
   comma instead of building separate branches, and antd keeping the Modal in
   the DOM after close, which made `to_have_count(0)` never pass.
4. **Coverage boundary stated honestly**: the probes only prove that the
   selector syntax is correct and matches the expected nodes on a faithfully
   reconstructed DOM. They cannot prove the real product DOM matches that
   reconstruction — that risk is closed by fork CI, which builds the real
   console.
5. **Re-verified against current upstream `main` before opening this PR**: the
   15 commits upstream advanced since this branch's base touch `e2e/` **zero**
   times, and `console/src/pages/Settings/Environments/` zero times, so no
   rebase is needed (`git merge-tree` reports no conflict). Each anchor this
   suite depends on was re-checked at the right layer on `upstream/main`:
   `.row` still exists in `index.module.less` (compiling to
   `index-module__row__<hash>`), `searchPlaceholder: "Search variables"` and
   `applyNow: "Apply now"` still exist in `locales/en.json`, and
   `core.agent-config` → `/agent-config` still exists in `builtinRoutes.tsx`.

## Risk and blast radius

- No tier marker changed and no case added or removed, so the nightly
  three-way shard split is unaffected.
- No test file outside `test_environments.py` is touched, except CROSS-005 in
  `test_cross_module.py`, which is explicitly in scope for this PR.
- No product code is changed; this is test-side only.

### DIFF
--- a/e2e/pages/environments_page.py
+++ b/e2e/pages/environments_page.py
@@ -4,12 +4,38 @@ QwenPaw Environments page object.
 
 Wraps all interactions on the environment variables configuration page and
 exposes business-level methods.
+
+Rebuilt for the unified environment management page (#7538, commit 76bfb704):
+
+* The page is no longer an editable table.  It renders three read-oriented
+  sections — "Custom variables", "Live settings" and "Read-only settings" —
+  where each row shows the key in a ``<code>`` element and the value masked.
+* Adding and editing go through a Modal (``Add Variable`` / ``Edit variable``)
+  whose primary button is ``Apply now`` and which writes immediately
+  (``PATCH /api/envs``).  There is no page-level Save button any more.
+* Deleting goes through ``Modal.confirm`` (``Delete Variable`` / okText
+  ``Delete``); a configured catalogue entry can additionally be ``Reset``.
+* The pre-#7538 row checkboxes, batch delete and per-row "insert row below"
+  action were removed from the product, so this page object does not expose
+  them.
+
+Selector strategy: class names are CSS-module scoped as
+``[name]__[local]__[hash:base64:5]`` and this page's stylesheet shares the
+``index.module.less`` filename with ``components/PageHeader``, so every
+generated class begins with ``index-module__``.  Local names are therefore
+matched double-underscore bounded (``__row__``, ``__page__``) to avoid
+hitting neighbours like ``__envRow__`` or ``__pageHeader__``.  Because the
+hash changes whenever the styles change, behavioural anchors (``aria-label``,
+visible text, ``placeholder``) are preferred wherever the page provides them.
+
+The console UI is English-only under e2e (``locale="en-US"`` in
+``e2e/fixtures/__init__.py``), so no Chinese text fallbacks are kept.
 """
 from __future__ import annotations
 
 import logging
 from typing import Optional, List
-from playwright.sync_api import Page, Locator, expect, TimeoutError
+from playwright.sync_api import Page, Locator, expect
 
 from pages.base_page import BasePage
 from config.settings import config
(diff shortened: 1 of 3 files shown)


cuiyuebing (approved):
LGTM

PR #7451, opened 2026-09-01
test(integration): coverage sprint batch 6 — 314 cases across channels, CLI, mail, hub, renderer, and harness adapter

yutai78786:
> **Submitted by**: 鬼谷子·Integrator@QPQAT

## Description

Coverage sprint batch 6: 18 new integration test files, 314 cases. Pure test addition — zero product-code changes, zero changes to pre-existing tests. All cases carry `integration` + `p1` markers so both the PR gate and nightly actually run them.

Targets were selected from line-level gap data in the last nightly's coverage.integration.xml (8/27 measurement), focusing on 0%-coverage or very-low-coverage modules:

1. **Channel family helpers (22 cases)** — QQ text sanitization / bool coercion / WS error classification, Telegram URL derivation / chunk splitting, Matrix markdown-to-HTML / device-id fallback, Discord attachment classification / code-fence-aware chunking, WeChat / WeCom / Yuanbao handle parsing.
2. **CLI commands (70 cases)** — update (version compare, PyPI release selection, install source detection, service probing), doctor fix (fix-id parsing, path allowlists, jobs.json cron normalization, agent.json validity, atomic writes, backup markers), doctor checks, and pure helpers across providers / channels / skills / plugin commands (API-key masking, scope resolution, frontmatter validation, Zip Slip protection).
3. **Mail (45 cases)** — monitor (idle timeout, HTML stripping, MIME header decoding, domain matching, push rules, wake decisions, wake prompts, IMAP host resolution) and access control (ACL address validation, serialization roundtrips, sender decision chain).
4. **Skills hub (18 cases)** — conflict payload, env-driven HTTP timeout / retries / backoff configuration, URL builders, cancellation hooks, GitHub response cache.
5. **Scroll memory space (20 cases)** — FTS MATCH query building with OR groups, LIKE term / pattern helpers, session suffix sanitization, strict ISO date parsing, scan budget.
6. **Visual-compression renderer (23 cases, previously 0%)** — glyph atlas loading for low / medium / high effort, char cell width, missing-glyph escaping, tab expansion, minification, column measurement, page geometry, page counting, PNG rendering with sha256 stability.
7. **Codex harness adapter (21 cases)** — sandbox policy translation, skill formatting, notification turn-id extraction, tool name / arguments / output mapping across 8 item types.
8. **Others** — fork project (19), PawApp context (15), model factory media helpers (16), scroll manager (14), tool-guard safety checks (13), ACP server adapter (11), doctor checks (11), hub control app (7).

[...]

PR #7341, opened 2026-08-27
test(integration): coverage sprint batch 5 — 495 cases across endpoints, CLI, and module internals

yutai78786:
> **提交人**: 鬼谷子·Integrator@QPQAT

## Description

Coverage sprint batch 5: 21 new integration test files, 495 cases, targeted at the coverage gaps reported by the coverage digest (8/25 measurement, combined 66%).

Coverage surface:

1. **Endpoint contracts (335 cases)**: contract-style cases for all 335 real router endpoints enumerated from upstream router sources (GET structure / unknown-id 404 / empty-body 422 / streaming read-timeout exemption), grouped into config / skills / workspace / providers / agents / misc.
2. **CLI surface (36 cases)**: `--help` for all 26 lazily-registered subcommands, `--version`, and read-only subcommands (agents/cron/chats/skills/models list, doctor).
3. **Module-level internals (124 cases)**: Governance & Security (detectors / audit / policy / secret_store / rule_guardian), skill_system (store / registry / pool_service), MCP & Drivers (acp helpers / credentials / DriverCard), run_tool_batch (ref resolution / condition & arithmetic eval / batch loading), channels manager + chats utils, Context & Scroll (budget / precision / scroll sync), Auth & config utils, Agent Core (command_handler / middlewares / prompt / file_handling), Providers (openai compat), Memory & ReMe, Observability + backup, backup restore + config agent ids.

Every case was written against the upstream source signatures; the full suite runs green locally and on fork CI across all 4 platforms.

**Related Issue:** N/A (coverage expansion)

**Security Considerations:** Test-only change, no product logic touched. Write-path cases send empty/invalid bodies only and never create real data.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [x] Tests

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes (pylint 10.00/10)
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
[...]

PR #7260, opened 2026-08-25
test(integration): targeted coverage expansion — workspace tree, MCP policy, plugin SDK (22 cases)

yutai78786:
> **提交人**: 鬼谷子·Integrator@QPQAT

## Description

Targeted coverage expansion with 3 new integration test files (22 cases), hitting the top uncovered lines reported by the coverage digest (nightly 2026-08-17, 733 files, 41k uncovered lines):

- **Workspace tree / commands surface** (250 uncovered lines): paged directory listing, cursor validation, root validation, slash command menu — 10 cases
- **MCP policy / access-principals / tools management surface** (133 uncovered lines): unknown client 404, invalid body 422, principals list — 6 cases
- **Plugin SDK module-level** (216 uncovered lines): PluginApi construction, startup/shutdown/uninstall hook registration, middleware registration, tool ownership claim/release — 6 cases

Every endpoint and response shape was verified against the upstream router source before writing; the suite was run in full locally and verified on fork CI across all 4 platforms.

**Related Issue:** N/A (coverage expansion)

**Security Considerations:** Test-only change, no product logic touched.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [x] Tests

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [x] Documentation updated (if needed)
- [x] Ready for review

## Testing

- Local: `pytest tests/integration -m "integration and (p0 or p1)"` 22 cases all pass (12.50s).
- Fork CI: Tests workflow 4 platforms integrated all green on head 3c62fd91 (see Evidence).

## Evidence

- Local full sweep: `22 passed in 12.50s`.
- Fork CI head 3c62fd91:
  - Pre-commit Checks run 32727036509 = success
  - NPM Format run 32727036333 = success
[...]

PR #7246, opened 2026-08-24
test(integration): expand integration coverage with 39 router/module test files and harden two flaky cases

yutai78786:
> **提交人**: 鬼谷子·Integrator@QPQAT

## Description

Expand `tests/integration` coverage with 39 new integration test files (238 cases) exercising the QwenPaw backend HTTP surface, plus 2 stability hardenings of pre-existing integration cases (see Additional Notes).

New coverage spans:

- **Generic routing layer**: agents / agent-status / agent-scoped / auth / settings / config / envs / files / backup / git / checkpoints / project-directory / access-control / approval / tool-calls / tools / plugins / providers / provider-oauth / mcp / local-models / skills / skills-stream / frontend-plugin / harnesses / heartbeat / healthz / version / token-usage / agent-stats / market / console / messages / fork / coding-mode
- **Module-level tests**: governance / sandbox / observability / browser SDK contracts

Every endpoint and response shape was verified against the upstream router source before writing/fixing the cases; the suite was run in full locally and verified on fork CI across all 4 platforms.

**Related Issue:** N/A (coverage expansion)

**Security Considerations:** Test-only change, no product logic touched. The files-preview cases assert the current security design (sensitive-file guard always enforced; non-sensitive out-of-workspace preview governed by `allow_preview_outside_workspace`).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [x] Tests

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [x] Documentation updated (if needed)
- [x] Ready for review

## Testing

- Local: `pytest tests/integration -m "integration and (p0 or p1)"` full sweep green (707 passed incl. pre-existing cases; all 238 new cases pass) in an isolated venv with the branch installed editable.
[...]

PR #7653, opened 2026-09-09
test(unit): coverage sprint batch 2 — 2475 cases across channels, visual compression, routers, runtime and CLI (+5.02pp)

yutai78786:
> **Submitted by**: 墨子·BEUnit@QPQAT

## Description

This PR adds **2475 backend pytest tests** and raises statement coverage of `src/qwenpaw` from **64.41% to 69.43%** (+5954 covered statements, **+5.02pp**), measured against the same `main` base in the same environment.

Every change lives under `tests/`. **All 87 files are newly added — no existing file is modified, and there are zero changes to product code.** (`git diff --name-status` reports 87 × `A` and 0 × `M`.)

**Why.** The modules covered here were at or near zero coverage while sitting on real failure paths: channel card approval flows (Feishu / QQ / WeCom / Telegram tool-guard cards and card handlers), the visual-compression pipeline (context pruning decides what the model actually sees), MCP OAuth and config service, cron scheduling, local model management, and auth token lifecycle. The goal is a regression net on those paths, not a coverage number.

**Related Issue:** No related issue. This is a proactive coverage effort; it does not fix any specific reported defect.

**Security Considerations:** This PR adds tests only and changes no authentication, configuration read/write, or environment handling logic, so runtime behaviour is unchanged. Tests touching security-related modules (`app/auth.py` token lifecycle, `security/tool_guard` approval gates, `security/skill_scanner`) only assert existing behaviour read-only. Assertions contain **no real credentials, tokens or keys** — stubs use synthetic values such as `ou_test_*` and `req_12345678`. Paths involving `chmod` are monkeypatched to no-ops so no test alters real file permissions on CI runners.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [x] Tests

> The five options above are the ones this template ships with; none describes a test-only change, so a `Tests` line is added here. This matches merged precedent in this repository — e.g. #7246, #7260, #7341, #7451 and the external contribution #4973 are all test-only PRs whose Type of Change section carries exactly this added `Tests` line. Happy to drop the line and leave all five unchecked if maintainers prefer the template literally.

## Component(s) Affected

- [x] CLI
- [x] Tests

Console, Skills, Documentation, CI/CD and Scripts/Deploy are untouched.

> Core/Backend, Channels and CLI are checked because the code **under test** lives there; all 87 added files are in `tests/`.

## Checklist

- [x] Ran `pre-commit run --all-files` locally and it passes — **exit 0**, all hooks green (mypy / black / flake8 / pylint / prettier / add-trailing-comma / check-ast / check-yaml / check-xml / check-toml / check-json / check-docstring-first / fix-encoding-pragma / detect-private-key / trim-trailing-whitespace / sort-simple-yaml). `actionlint` is skipped because its Go toolchain cannot be fetched in this environment; that hook only applies to workflow YAML and all 87 files here are `.py`.
- [x] Committed and re-ran the files pre-commit auto-fixed — black and add-trailing-comma reformatted files, which were amended in and re-verified to exit 0.
- [x] Ran the tests locally and they pass — **13001 passed, 22 skipped, 0 failed** (`pytest tests/unit`, 363s).
- [ ] Documentation updated — not applicable, no behaviour change.
- [x] Ready for review

## Testing

```bash
pip install -e ".[dev,test]"

# Full unit suite (this PR adds 2475 of the 13001 collected tests)
pytest tests/unit -q

# Coverage
pytest tests/unit -q --cov=src/qwenpaw --cov-report=term

# Gates
pre-commit run --all-files
./scripts/check-channels.sh all
```

Coverage is measured as a **same-base differential**: `upstream/main@59724615` is checked out in a separate `git worktree` and run bare as the baseline, then this branch is run in the same venv, so both numbers share the identical statement denominator (118518) and the delta cannot come from baseline drift.

## Evidence

**1. Full unit suite on this branch** (rebased onto `main@59724615`)

```
$ pytest tests/unit -q -p no:cacheprovider --cov=src/qwenpaw
Coverage JSON written to file
Required test coverage of 50.0% reached. Total coverage: 69.43%
13001 passed, 22 skipped, 67 warnings in 363.44s (0:06:03)
```

**2. Baseline — pristine `upstream/main@59724615` in a separate worktree, no changes from this PR**

```
$ PYTHONPATH=<worktree>/src pytest tests/unit -q --cov=src/qwenpaw
Required test coverage of 50.0% reached. Total coverage: 64.41%
10526 passed, 22 skipped, 62 warnings in 326.05s (0:05:26)
```

**3. Same-base coverage delta**

```
baseline : 64.4096%  covered=76337/118518
this PR  : 69.4333%  covered=82291/118518
denominator identical: True
delta    : +5954 statements = +5.0237pp
new tests: +2475 (10543 collected -> 13018 collected; 10526 passed -> 13001 passed)
```

**4. pre-commit**

```
$ pre-commit run --all-files
fix python encoding pragma................................................Passed
detect private key........................................................Passed
trim trailing whitespace..................................................Passed
Add trailing commas.......................................................Passed
Lint GitHub Actions workflow files......................................Skipped
```

**5. Channel gate** (`./scripts/check-channels.sh all`)

```
Channel Contract Coverage
   Total channels: 18
   With tests:     18
   Missing:        0

[OK] Tested: ConsoleChannel, DingTalkChannel, DiscordChannel, FeishuChannel, IMessageChannel, MQTTChannel, MatrixChannel, MattermostChannel, OneBotChannel, QQChannel, SIPChannel, SlackChannel, TelegramChannel, VoiceChannel, WeChatChannel, WecomChannel, XiaoYiChannel, YuanbaoChannel

[OK] All built-in channels have contract tests.
================== 357 passed, 1 skipped, 1 warning in 2.90s ===================
=========== 1875 passed, 1 skipped, 59 warnings in 77.14s (0:01:17) ============
All checks passed.
```

This PR does not modify any channel implementation, so the "For Channel Changes" section does not strictly apply. It was still run because this PR adds 8 new files under `tests/unit/channels/`, which that script collects as optional unit tests — leaving them unverified would have left a regression surface unchecked. The 59 warnings come from the pre-existing `tests/unit/channels/test_xiaoyi.py` (`@pytest.mark.asyncio` on non-async functions), which this PR does not touch.

**6. Fork CI** — all green on the PR head `19a07753`: `Pre-commit`, `console format check`, `website format check`.

**Known blind spot.** The `Tests` workflow (ubuntu / macos / **windows** × py3.11 / py3.13) only triggers on `pull_request`, so pushing to a fork branch does not run it. The cross-platform aspects of these tests therefore cannot be verified before this PR is opened. Each of the 87 files was reviewed by hand for platform assumptions instead — see Additional Notes, item 4.

### 1. What changed, grouped (82 commits)

**A. Channels (13 files)** — tool-approval cards and context builders for four channels:

| module | before → after | delta |
|---|---|---|
| `app/channels/feishu/card_handler.py` | 0% → **100%** | +153 |
| `app/channels/telegram/cards/tool_guard.py` | 19.3% → **100%** | +130 |
| `app/channels/feishu/cards/tool_guard.py` | 19.2% → **100%** | +105 |
| `app/channels/qq/cards/tool_guard.py` | 38.0% → **100%** | +98 |
| `app/channels/wecom/cards/tool_guard.py` | → **100%** | +94 |
| `app/channels/yuanbao/codec.py` | 60.8% → 89.8% | +96 |

plus `wecom/context.py`, the Feishu and QQ `cards/context.py`, and WeCom table normalisation.

**B. Visual-compression pipeline (9 files)** — `agents/context/visual_compression/pipeline/` as a whole: `tool_results.py` 0% → 87.7% (+178), `history.py` 0% → 90.1% (+164), `messages.py` 0% → 99.1% (+105), `static_context.py` 0% → 95.3% (+101). Roughly **+767 statements** in one family — the largest single-group gain, because the whole family was previously at zero.

**C. HTTP routers (10 files)** — `app/routers/`: `skills.py` 49.0% → 72.2% (+222), `project_directory.py` 26.8% → 74.1% (+168), `config.py` 52.2% → 79.0% (+133), `git.py` 31.6% → 77.1% (+115), `workspace.py` 51.9% → 66.0% (+113), `providers.py` 58.0% → 86.7% (+110), plus MCP OAuth and pawapps helpers.

**D. Runtime and command handling (5 files)** — `runtime/`: `model_handler.py` 9.8% → 69.4% (+103), approval and command handlers, daemon parsing, cancel-save helpers, builtin-commands skill fallback.

**E. CLI (4 files)** — `cli/skills_cmd.py` 43.6% → 74.2% (+120), doctor connectivity probes, init.

**F. Scattered targets (46 files)** — `local_models/model_manager.py` 55.6% → 98.7% (+132), `app/crons/manager.py` 61.8% → 95.8% (+131), `app/auth.py` 47.3% → 79.8% (+103), drivers/adapters, `skill_system/store`, plugin loader dependency probing, token_usage snapshots, tool-guard rule guardian, backup/restore, hooks, market, modes, harnesses. Mostly +20 to +60 statements each.

**G. Test-infrastructure package markers (9 files)** — two kinds, neither is test logic:

1. Four channel directories get `__init__.py`. Feishu, QQ, WeCom and Telegram each have a file named `test_cards_tool_guard.py`; without package markers pytest's rootdir-relative module naming collides and **collecting `tests/unit` aborts outright**.
2. `tests/unit/app/__init__.py` + `tests/unit/app/mcp/__init__.py` — see item 3 below.

98 modules gained coverage (+5967) and 3 lost a little (-13, `utils/telemetry.py` -11, `app/_app.py` -1, `hub/local_provisioner.py` -1), netting **+5954**. Stating the negative side too rather than only the gains.

### 2. Choices made deliberately in these tests

- **Known defects are not pinned.** `app/channels/yuanbao/codec.py`'s `decode_pb` returns `None` unconditionally under protobuf ≥ 6.33.5 (tracked separately). These tests cover only the encode side and the pure logic conversions, and deliberately avoid assertions that would freeze the broken behaviour — otherwise the tests would block the fix once it lands.
- **Assertions describe expected behaviour, not implementation details.** For example the `get_client` pool-key test proves "two spellings of the same directory share one client" through a real symlink rather than asserting on an internal dict's shape.
- **Cross-platform.** Directory symlinks pass `target_is_directory=True` (Windows distinguishes file from directory symlinks); path assertions never depend on the `/` separator; `chmod` paths are monkeypatched to no-ops.
- **Boundaries are the main battlefield.** Empty values, timeouts, concurrency caps, exception paths and cancellation paths are all covered, not just the happy path.

### 3. One connected fix reviewers should be aware of (`tests/unit/app/__init__.py`)

Adding `tests/unit/app/mcp/` (mirroring `src/qwenpaw/app/mcp/`) exposed a pylint problem that did not exist before.

`tests/unit/app/` has no `__init__.py`, while `tests/`, `tests/unit/`, `tests/unit/agents/`, `tests/unit/channels/` and `tests/unit/runtime/` all do. pylint walks up from a file to the first ancestor **without** `__init__.py` and inserts that directory into `sys.path`. So `tests/unit/app/` became the sys.path root, its child directory `mcp` became an importable top-level package, and it **shadowed the real `mcp` distribution**. astroid then resolved `import mcp` to the test directory, found no `server` / `client` / `types` / `shared` inside it, and reported 15 `E0611` errors across the repository — all in files this PR never touched:

```
src/qwenpaw/drivers/handlers/mcp_stateful_client.py:28: E0611: No name 'ClientSession' in module 'mcp'
packages/qwenpawmail-mcp/src/qwenpawmail_mcp/server.py:16: E0611: No name 'server' in module 'mcp'
tests/fixtures/mcp/http_echo_server.py:4: E0611: ...
```

That is why they are `E0611` ("module found, name missing") rather than `E0401` ("cannot import"), and why they appear in unrelated files: once the directory is on `sys.path` the shadowing is global.

Reproduced with single-variable experiments against a pristine `main` worktree:

| condition | result |
|---|---|
| pristine upstream, `pylint --all-files` | Passed, 0 × E0611 |
| + only the two `tests/unit/app/mcp/*.py` files | **Failed, 15 × E0611** |
| + `tests/unit/app/mcp/__init__.py` as well | Failed, 15 × E0611 |
| + `tests/unit/app/__init__.py` too | **Passed, 0 × E0611** |

The third row is the interesting one: marking only the new directory does not help, because pylint then stops at `tests/unit/app/`, which is still not a package.

**Fix:** add the two empty `__init__.py` files so `tests.unit.app.mcp` is a real package path. This also completes a layout the rest of the tree already follows — 8 of the 9 existing `tests/unit/app/*` subdirectories ship an `__init__.py`, so only `tests/unit/app/` itself was the odd one out. Verified: the 1237 collected node ids under `tests/unit/app` are byte-identical before and after, so no existing test changes identity.

**Worth flagging, not changed by this PR:** `tests/unit/agents/acp/` sits in the same hazard position — `acp` is a real distribution name — and is saved only by the existence of `tests/unit/agents/__init__.py`. `tests/unit/app/workspace/` and `tests/unit/cli/` are in the same non-package position and are harmless today only because no distribution is named `workspace` or `cli`.

### 4. Cross-platform review (the `Tests` matrix cannot be run before opening)

Because the windows/macos legs only run on `pull_request`, all 87 files were reviewed by hand. Two real defects were found in this PR's own tests and fixed:

- A directory symlink created without `target_is_directory=True`. Windows requires the flag and POSIX ignores it, so it was invisible locally. The symlink cannot simply be removed: `get_client` keys its pool on `str(project_dir.resolve())`, so the test needs a real link to prove two spellings of one directory share a client. Fixed per existing convention (`tests/unit/services/test_unified_workspace_files.py`, `tests/unit/checkpoints/test_checkpoint_restore.py` both pass the flag for directory targets).
- Three assertions compared `_header_text` output against hard-coded POSIX strings. That function interpolates `execution_cwd` straight into an f-string, so Windows renders `\tmp\work`. The expectation is now built from the very `Path` handed to the function, which is byte-identical on POSIX and keeps the assertion equally strict.

The remaining hits were checked and are safe: 9 are plain string round-trips (a monkeypatched `shutil.which` returning a literal, a parameter annotated `str` passed through unchanged, a JSON config field, `removeprefix("file://")` which is string manipulation rather than the platform-dependent `url2pathname`), 3 are shell command text shown inside approval cards (`rm -rf /tmp/x`), and one `chmod` site monkeypatches `_chmod_best_effort` to a no-op. The test for deciding was whether a path reaches an assertion **through a `Path` object** — if so it is platform-dependent, if it is a plain string round-trip it is not.

### 5. Commit authorship

All 82 commits use `yutai78786 <42596677+yutai78786@users.noreply.github.com>` as both author and committer, matching the 77 pre-existing commits of that form on `main` and all seven previously merged PRs from this fork. The actual writer is identified per commit by a `Signed-off-by` trailer carrying the QPQAT identity, and by the `Submitted by` header line at the top of this description.

### 6. Rebase adaptation

Rebasing onto `main` required adapting `tests/unit/agents/test_command_handler_status.py` to an upstream rename of the whole slash-command surface (`_process_summarize_status` → `_process_auto_memory_status`, registry name `summarize_status` → `auto_memory_status`, `list_summarize_status()` → `list_auto_memory_tasks()`, plus a new `trigger` field and `pending` / `cancelled` statuses). While adapting, the branches the new renderer adds are covered rather than just renaming calls — that file goes from 21 to 26 tests.

---

*Test-only PR: 82 commits, 87 newly added files, all under `tests/`, no existing file modified, zero product code changed.*


cuiyuebing (approved):
LGTM

PR #7227, opened 2026-08-24
fix(console): retain embedding verification across navigation

jinliyl:
## Description

Keep successful Embedding service verification results in a session-scoped Zustand store instead of component-local state. Results are isolated by agent and matched against the exact service fingerprint, so navigating away and returning preserves a valid result while changed settings remain unverified.

This also removes the config revision behavior that cleared verification after saving unchanged settings, and guards against the form watcher's transient `undefined` value during remount.

**Related Issue:** Fixes #7226

**Security Considerations:** Verification fingerprints include credentials for exact matching but remain only in process memory. They are not written to localStorage, sessionStorage, configuration, or logs, and are discarded when the Console application restarts.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [x] Documentation updated (not needed)
- [x] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

- Verify an Embedding service, leave the configuration page, and return: the exact configuration remains verified.
- Change the model/service settings: the previous result is not reused.
- Switch agents: verification status is isolated per agent and restored when switching back.

## Evidence

```text
npm test -- --run src/pages/Agent/Config/components/ReMeLightMemoryCard.test.tsx
Test Files  1 passed (1)
Tests       26 passed (26)

[...]

PR #7069, opened 2026-08-16
fix(console): render data-URL images in historical messages on session reload

suantea:
## Summary

Fixes #7051.

Images attached to Console chat messages render fine when first sent, but after closing and re-opening the chat the history view shows a broken/empty thumbnail. The backend correctly serves the image as a `data:` URL (`data:image/png;base64,...`) on history load, but the frontend failed to render it for historical messages.

## Root cause

`toDisplayUrl` in `console/src/pages/Chat/utils.ts` only recognized `http(s)://` and `file://` prefixes. A `data:` URL (base64 image) fell through to `chatApi.filePreviewUrl`, which wrapped it as a file-preview path — corrupting the URL so the image could not render. The live-render path renders `data:` URLs fine, which is why only history hydration broke.

## Change

- `console/src/pages/Chat/utils.ts`: `toDisplayUrl` now passes `data:` URLs through untouched, matching the live-render path.
- `console/src/pages/Chat/utils.test.ts`: two regression tests — a `data:` URL round-trips unchanged and never falls back to `/files/preview`.

## Test

`npx vitest run src/pages/Chat/utils.test.ts` → 37 passed (incl. 2 new).

PR #7669, opened 2026-09-10
test(console): add regression coverage for embedding verification persistence and the toDisplayUrl twin (+18 cases)

yutai78786:
> **Submitted by**: 小乔·FEUnit@QPQAT

## Description

Adds test coverage for two pieces of Console logic that **no existing test actually executes**. Three files, 18 cases, +361 lines. **Test-only change — no product code touched.**

### 1. Embedding verification state lost across page navigation (relates to #7226)

The upstream fix in #7227 (*retain embedding verification across navigation*) added `console/src/stores/embeddingVerificationStore.ts` (+42 lines), but the tests shipped with that PR went into `ReMeLightMemoryCard.test.tsx` (+86 lines). The store itself and the persistence path in `useEmbeddingVerification` were left untested.

The existing `useEmbeddingVerification.test.tsx` (80 lines, from #7353) covers the **agent-switching** scenario, which does not overlap with the **cross-page-navigation persistence** scenario covered here.

- `embeddingVerificationStore.test.ts` (6 cases) — store / clear / update per agent; clearing a non-existent agent must not throw; and the stored entry stays readable across separate `getState()` reads (the store is the only place this result lives, so every later consumer must observe the same value).
- `useEmbeddingVerification.storePersistence.test.ts` (4 cases) — a result written via `markVerified` survives a **full unmount** of the component tree and is read back by a fresh hook instance (the navigation scenario); it is invalidated when the config fingerprint changes; it is cleared when embedding is disabled.

The persistence case really calls `unmount()` before mounting the second instance, so the assertion exercises store persistence rather than a still-mounted component holding state.

**Red/green evidence** (checked against the parent commit of #7227, `04ffe72b^`; still valid after the rebase — `useEmbeddingVerification.ts` is byte-identical between the old and new baseline):

| State | Result |
|---|---|
| Parent commit — `embeddingVerificationStore.ts` does not exist | import fails, necessarily red |
| Parent commit — `useEmbeddingVerification.ts` (55 lines) | **3 cases red** (persist / retrieve / clear) |
| `upstream/main` — same hook (63 lines) | **all green** |

### 2. The `toDisplayUrl()` twin (defect prevention)

The console has **two functions named `toDisplayUrl` with identical bodies** (they differ only in a comment and in accepting `string | undefined` vs `string`):

| | Location | Coverage |
|---|---|---|
| (1) | `src/pages/Chat/utils.ts:185` | 8 cases in `pages/Chat/utils.test.ts`, added by #7069 together with its fix |
| (2) | `src/components/Chat/ToolCards/shared/utils.ts:16` | **never executed by any test** |

Why (2) was never executed:

- The sibling `ToolCards/shared/utils.test.ts` does exist (340 lines), but it covers `shortFileName` / `formatMemorySearch` / `getMediaInfo` / `extractUrlFromText` / `formatAgentList` and never mentions `toDisplayUrl`.
- All three consumer tests referencing this module (`cards/GrepSearchCard.test.tsx`, `cards/RunToolBatchCard.test.tsx`, `cards/SendFileCard.test.tsx`) replace it wholesale via `vi.mock("../shared/utils", ...)`, and `RunToolBatchCard` additionally stubs `toDisplayUrl` as the identity function — so the real implementation never runs in any of them.

`utils.toDisplayUrl.test.ts` (8 cases) pins (2): http/https pass through, data URLs pass through and are not routed to the preview endpoint, empty input yields an empty string, the `file://` prefix is stripped, an already-rooted relative path is resolved, an unrooted relative path gets `/` prepended first, and a non-empty URL whose preview result is falsy stays distinguishable from an empty URL.

> **Honest scoping:** this is **defect-prevention** coverage, not a claim that any open bug is being blocked. Both twins already carry the `data:` branch today — (2) since `84b61ca3`, (1) since `c6261762`. The value is that a future divergence between the two duplicated implementations cannot pass unnoticed.

**Mutation evidence** (the real implementation was broken on purpose, three separate ways, to prove the cases bite):

| Injected mutation | Result |
|---|---|
| Remove the `data:` branch | **2 cases red** |
| Remove the `file://` stripping | **1 case red** |
| Break the root-prefixing of relative paths | **1 case red** |

All mutations were reverted and verified byte-identical to `upstream/main`.

**Related Issue:** Relates to #7226

(`Fixes` is deliberately not used: #7226 is already closed and its fix #7227 is already merged — this PR only fills the test gap left behind. The `toDisplayUrl` twin part has no corresponding issue; it is defect prevention and is not attached to an issue number.)

**Security Considerations:** none — test-only change, no env/config/auth surface touched.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

> Note: none of the options above is selected on purpose. This section has no "Test" option, and the PR adds test code only — no bug is fixed, no feature or behaviour is added, nothing is refactored in the product sense. The test-only nature is expressed by checking **Tests** under *Component(s) Affected* below.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

  (run against the repo-pinned pre-commit; `detect-private-key` and `trim trailing whitespace` **Passed**, the Python hooks report "no files to check" as expected for `.ts` files. Note the `prettier` hook in `.pre-commit-config.yaml` excludes `^console/` by design, so it reports *Skipped* for these files — see the next item for the check that actually applies.)
  (`prettier --check` initially failed on the two #7226 files — 136 single quotes where the project uses double quotes. Fixed with `prettier --write`, re-verified clean. No hook auto-modified anything after that.)
  (`npm run format:check` — the command used by the *console format check* CI job — exits 0: `tsc -b --noEmit` reports zero errors and `prettier --check .` reports "All matched files use Prettier code style!". Full suite via `vitest run`: **359 files / 3529 tests passed**, exit 0, re-measured after rebasing onto `797ede9b`.)
- [x] Documentation updated (if needed) — N/A, test-only change
- [x] Ready for review

### Coverage impact (measured, not estimated)

Same worktree, same command, both runs exit 0. The baseline run was produced by temporarily moving the three new files out of the tree (`git status` showed three `D` entries) and restoring them afterwards with `git checkout --`:

| Metric | Baseline | With this PR | Delta |
|---|---|---|---|
| Test Files | 356 | 359 | **+3** |
| Tests | 3511 | 3529 | **+18** |
| statements | 63.94% | 63.95% | **+0.01pp** |
| branches | 55.27% | 55.29% | +0.02pp |
| functions | 59.34% | 59.34% | +0.00pp |
| lines | 65.15% | 65.15% | +0.00pp |

Measured again after rebasing onto `upstream/main` @ `797ede9b` — the earlier
figures in this table were taken on `0794d3fa`, and upstream has since moved 17
commits, including `b1865f87 test(unit): coverage sprint batch 2`, which shifted
the baseline itself. Post-rebase measurement: baseline run with the three new
files moved out of the tree (356 files / 3511 tests, exit 0), then with them back
in (359 files / 3529 tests, exit 0).

The coverage delta is intentionally small (+0.01pp statements after the rebase): all three subjects are tiny (a 7-line function, a 42-line store, a 63-line hook). The point is closing "the real implementation is never executed" gaps and pinning a duplicated implementation against drift, not moving the coverage number. If that delta is judged not worth a maintainer's review time, closing this PR costs nothing — the value here is preventive, not quantitative.

### Note on the branch history

Three commits: the `toDisplayUrl` twin test, the #7226 persistence tests, and a follow-up that corrects two factual claims in the twin test's header comment (comment-only, 14 insertions / 6 deletions, no assertion touched). The correction was kept as its own commit rather than amended into the first one.

The branch was later rebased onto the current `main` (17 upstream commits, zero conflicts; the diff is still exactly 3 files / +361 −0). Before force-pushing: a backup tag was taken at the pre-rebase tip, all three commits were verified to be the only ones on this branch and all carry the same author and sign-off, and the push used `--force-with-lease` pinned to the previous remote sha.

Re-verified on the new baseline: `prettier --check` clean, `tsc -b --noEmit` zero errors, full suite 359 files / 3529 tests green. The dedup analysis was also re-checked against the drift — none of the 17 commits touched `embeddingVerificationStore.ts`, `useEmbeddingVerification.ts`, `ToolCards/shared/utils.ts`, the upstream `useEmbeddingVerification.test.tsx`, or `pages/Chat/utils.test.ts`; upstream still has no store-level test for `embeddingVerificationStore`, and still no test that executes the `toDisplayUrl` twin.


cuiyuebing (approved):
LGTM

PR #7591, opened 2026-09-06
feat(feishu): auto-collapse reasoning streaming card after finalize

Bruce-Yii:
Fixes #7570.

## What Problem This Solves

Feishu currently renders reasoning streams as ordinary Markdown streaming cards. For models with long reasoning traces, that reasoning remains fully expanded after generation completes and pushes the final answer far down the conversation.

This change keeps reasoning visible and expanded while it is streaming, then automatically collapses the reasoning panel after finalization so the final answer remains prominent. Answer-card behavior is unchanged.

## What Changes

- Reasoning stream cards use a JSON 2.0 `collapsible_panel`, initially expanded, with the existing inner `streaming_content` element id unchanged.
- After content/settings finalization, the reasoning panel is collapsed through the CardKit element PATCH API using the next sequence number.
- The normal answer path remains unchanged.
- If the pre-created `"..."` placeholder is never consumed by an answer stream:
  - successful completion finalizes it as `✅ 任务已完成`;
  - error/cancellation finalizes it with the existing DingTalk-style non-success wording;
  - error/cancellation never shows a false success state.
- Cleanup/finalize failures are best-effort and never suppress the real answer or error.
- No BaseChannel lifecycle changes, raw HTTP calls, or new dependencies are introduced.

## Evidence

Before this change, the Feishu reasoning card is a plain streaming Markdown card and there is no final element PATCH that collapses it, so completed reasoning remains expanded.

With this change:

1. reasoning card creation produces a JSON 2.0 `collapsible_panel` with `expanded: true`;
2. streaming updates continue targeting the unchanged `streaming_content` element;
3. finalization completes content/settings updates first;
4. the CardKit element PATCH then sets the reasoning panel to `expanded: false`;
5. answer cards remain on the existing path.

Regression coverage also verifies the newly exposed unused-placeholder lifecycle:
- success with no answer stream does not leave a bare `"..."`;
- error never displays `✅ 任务已完成`;
- cancellation handles only the still-unconsumed placeholder and continues normal cancellation propagation.

Issue #7570 also contains reporter-provided live CardKit validation and a screenshot demonstrating the intended collapse behavior. This PR independently validates the payload structure, call ordering, answer regression, error handling, and cancellation handling in tests.

Live visual verification against a real Feishu client is still recommended as a residual check; this PR does not claim that was performed in the contributor's local environment.

## Testing

- `tests/unit/channels/test_feishu.py` + Feishu contract tests:
  **197 passed, 1 skipped** (skip pre-existing)
[...]

PR #7685, opened 2026-09-11
feat(feishu): collapsible reasoning panel with opt-in auto-collapse

### LINKED ISSUE #7570: 飞书流式卡片: 思考过程输出完成后自动折叠
2.1.0 的飞书 CardKit 流式输出（#3001）用着不错。但目前思考过程和回复一样是
普通 markdown 流式卡，强制思考的模型（GLM-5.x 系列）思考文本很长，输出完仍
整屏占着，把最终回复顶到很远。回看场景其实很少，希望思考卡在流式结束后自动
折叠。
已验证可行，本地这样改后稳定运行一段时间：
1. 思考卡用 JSON 2.0 的 collapsible_panel 包裹 markdown，流式期间 expanded: true
2. 流式更新接口（PUT elements/:element_id/content）对嵌套在折叠面板内的
   markdown 正常工作，打字机效果不受影响
3. 流式结束后 PATCH /cards/:card_id/elements/:element_id，body 传
   {"partial_element": "{\"expanded\": false}", "sequence": n} 即可收起面板，
   header 顺便改成「思考过程（已完成·点击展开）」
几个坑：
- settings 接口是 PATCH 方法，用 PUT 会 404
- patch 元素的 body 要用 partial_element 字段包一层（JSON 字符串）
- element_id ≤ 20 字符、字母开头
改动点很小：_create_streaming_card 里 reasoning 卡把 markdown 换成面板结构
（内层 element_id 不变，现有更新逻辑不用动）；_finalize_streaming_card 里
settings 之后加一次面板 patch。顺带一提，预创建的占位卡（"..."那张）不再被
思考流复用后，适合改成状态卡（「正在思考与处理中」→「✅ 任务已完成」）。
环境：qwenpaw 2.1.0 / 2.2.0（feishu/channel.py 两版一致），飞书 CardKit v1，
2026-09 实测。需要的话补丁 diff 可以附上。

Image

hongxicheng:
## Problem

Feishu renders the reasoning stream as an ordinary Markdown streaming card. For
models with long thinking traces, that reasoning stays fully expanded after
generation finishes and pushes the final answer far down the conversation.

A previous attempt (#7591) wrapped reasoning in a collapsible panel, but it
pre-created a plain placeholder card before the agent started and reserved it
for the answer. Because cards are messages and Feishu keeps them in post order,
the reasoning card was created *below* that placeholder and the answer was then
streamed into it — so the answer rendered **above** the thinking. Its unit tests
asserted that reuse as intended behavior, which is why the inversion went
unnoticed.

## What changed

**Reasoning renders in a collapsible panel.** Reasoning stream cards are created
with a JSON 2.0 `collapsible_panel` wrapping the existing `streaming_content`
markdown element, so content updates are unaffected. The header label is the
English word `Thinking`, and the expand/collapse affordance comes from the
panel's own `header.icon` (a `standard_icon` chevron with
`icon_position: follow_text` and `icon_expanded_angle: 180`) rather than from
state-dependent copy. This keeps the card copy localizable and means no
"completed / click to expand" string is ever needed.

**Auto-collapse is opt-in.** A new Feishu setting, `auto_collapse_thinking`
(default `false`), controls whether the panel is collapsed automatically once
the stream finishes. The panel is always rendered collapsible by hand, so
turning the setting off keeps the manual collapse affordance and only disables
the automatic fold. Collapsing uses the CardKit element PATCH API (`PUT` returns
404 for this endpoint; `partial_element` must be a JSON string) with the next
sequence number after the settings finalize, and a failed patch only logs so it
can never fail the answer.

**Streaming cards are created lazily.** Card pre-creation is removed from
`_before_consume_process`; each stream segment creates its own card in
`on_streaming_start`, in arrival order. Card order therefore matches segment
order by construction — the thinking card is created before the answer card and
renders above it. This also removes a class of orphan-placeholder bugs: there is
no longer a bare `"..."` card to clean up when an answer never follows (truncated
thinking, error, cancellation), so no compensating lifecycle machinery is needed.

**Console:** a "Collapse Thinking" toggle is shown in the Feishu channel
settings, below Streaming Output and only when Streaming Output is enabled.
Copy is added for en / zh / ja / ru / pt-BR / vi (locales without the adjacent
streaming keys keep falling back to English).

## Testing

- Live-verified in a real Feishu client: with the setting off, thinking streams
  in an expanded panel and stays expanded with a working manual collapse arrow;
  with the setting on, it folds automatically once the answer completes.
- New unit coverage: panel payload shape and element-id constraints, answer
  cards staying plain Markdown, ordering regression (the answer card is posted
  after the reasoning card and is a distinct card), collapse payload, failure
  tolerance, opt-in default, `from_config` plumbing, and a guard that the card
  copy contains no CJK.
- `pytest tests/unit/channels/test_feishu.py` — 172 passed
- `pytest tests/unit/channels` — 1780 passed, 1 skipped
- `pytest tests/unit/config` — 129 passed
- `./scripts/check-channels.sh feishu` — all checks passed
- `pre-commit run --files <changed>` — clean
- Console: `tsc -b --noEmit`, `eslint`, `prettier --check` clean; locale and
  ChannelDrawer vitest suites — 90 passed

## Scope

Feishu only. Files touched: `config/config.py`,
`channels/feishu/{channel,constants}.py`, `tests/unit/channels/test_feishu.py`,
plus the console channel type, settings drawer, drawer test and six locale
files. No new dependencies, no API or lifecycle changes.

Not addressed here (pre-existing, separate follow-up): a reasoning stream that is
cancelled mid-flight is never finalized, because `on_streaming_end` is the only
place the panel is folded and the cancellation path does not reach it. Fixing it
would require re-introducing per-session streaming-card state, which this change
deliberately removes.


cuiyuebing (approved):
LGTM

PR #7719, opened 2026-09-12
feat(memory): allow a separate model for ReMeLight memory writing

### LINKED ISSUE #7664: [Feature]:  RemeLight额外自定义模型
### 问题描述

目前 `RemeLightMemoryManager` 在执行对话总结（summarize）和记忆进化（dream）时，调用的是当前 Agent 绑定的主 LLM 模型，无法单独配置一个独立的记忆写入模型。这导致用户在使用昂贵的旗舰模型进行聊天时，后台记忆写入也会消耗同样昂贵的 Token，造成不必要的经济损失。

### 期望行为

希望在 `MemoryConfig` 中增加一个 `memory_model` 配置项，允许用户为记忆写入（summarize/dream）指定一个独立的、更轻量的模型。如果未配置，则回退到当前行为（使用主模型）。

### 使用场景

- 用户聊天使用 Qwen3.8-Max 等旗舰模型处理复杂任务
- 记忆写入使用便宜或者免费轻量模型或者本地模型
- 两者互不干扰，用户无需为了节省记忆成本而降低聊天模型质量

### 建议实现方式

在 `agent.json` 的 `memory` 配置中新增字段：

```json
{
  "memory_manager_backend": "remelight",
  "memory_model": "qwenpaw-flash-4b"
}
```

`memory_model` 为可选字段，不填则使用主模型（保持向后兼容）。

### 补充说明

ADBPG 后端已经通过服务端内置 LLM 解决了这个问题，但 RemeLight 作为默认的本地记忆后端，同样需要这个能力，尤其是对于注重数据隐私、不想使用云端服务的用户。

---

### Problem Description

Currently, `RemeLightMemoryManager` uses the main LLM model bound to the current Agent when performing conversation summarization and memory evolution (dream). There is no way to configure a separate, independent model for memory writing. This means that when users use an expensive flagship model for chatting, the background memory writing also consumes equally expensive tokens, resulting in unnecessary cost.

### Expected Behavior

We propose adding a `memory_model` field to `MemoryConfig`, allowing users to specify an independent, lighter model for memory writing (summarize/dream). If not configured, it should fall back to the current behavior (using the main model).

### Use Case

- Users chat with flagship models like Qwen3.8-Max for complex tasks
- Memory writing uses cheap, free lightweight models, or local models
- The two are completely independent — users don't need to downgrade their chat model just to save on memory costs

### Suggested Implementation

Add a new field in the `memory` section of `agent.json`:

```json
{
  "memory_manager_backend": "remelight",
  "memory_model": "qwenpaw-flash-4b"
}
```

`memory_model` is optional. If not set, the main model will be used (backward compatible).

### Additional Notes

The ADBPG backend has already addressed this issue through its server-side built-in LLM. However, RemeLight, as the default local memory backend, also needs this capability — especially for users who prioritize data privacy and prefer not to use cloud services.

iluv7:
## Description

`ReMeLightMemoryManager` currently injects the agent's main model into ReMe's `as_llm` component before every memory-writing job (`auto_memory` / `auto_dream` / `daily_paper` / `auto_fin`), so memory writing always consumes the same (often expensive) model used for chatting.

This PR adds an optional `memory_model` slot to `reme_light_memory_config`, so memory writing can run on an independent, lighter model:

- Uses `{"provider_id": "...", "model": "..."}` as the canonical format; a string is treated as a bare model name that inherits the main model's provider, including version-tagged names such as `qwen2.5:7b`.
- When unset, behavior is unchanged (main model is used).
- When the configured slot is unavailable (e.g. unknown provider), the manager logs a warning and falls back to the main model, so memory jobs never break.
- The injected model is refreshed before every LLM job, so config changes take effect without a restart.

**Related Issue:** Fixes #7664

**Security Considerations:** `memory_model` stores only a provider/model reference; no credentials are written to `agent.json` (API keys stay in the provider store).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Console (frontend web UI) — memory model selector and API types
- [ ] Skills
- [ ] CLI
- [x] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

Not applicable — no channel code paths are touched.

### Unit tests

- `tests/unit/config/test_memory_config.py` — parametrized cases covering every `memory_model` slot form (object / dict / `provider:model` string / version-tagged model names / bare name) plus a serialization round-trip.
- `tests/unit/agents/memory/test_reme_memory_model.py` — parametrized slot selection (unset / full slot / bare name inheriting the agent or global provider / unusable slot) plus the `ProviderError` fallback path.

### Real API verification (local script, not committed)

Built two real providers (SiliconFlow and DeepSeek) and ran the actual injection path (`_memory_model_slot` → `create_model_and_formatter_async` with `model_slot_override` → `update_component("as_llm")`), then issued real chat requests through the injected model. All four scenarios passed:

1. Main = `Qwen3.8-27B` (SiliconFlow), memory = `deepseek-flash` (DeepSeek) → injected provider: DeepSeek ✓
2. Roles swapped (DeepSeek main, SiliconFlow memory) → injected provider: SiliconFlow ✓
3. `memory_model` unset → main model used (backward compatible) ✓
4. Bare model name → inherited the main model's provider ✓

## Evidence

```bash
$ pytest tests/unit/config/test_memory_config.py \
    tests/unit/agents/memory/test_reme_memory_model.py -q
44 passed in 0.32s

$ pytest tests/unit/agents/memory/ tests/unit/config/ \
    tests/unit/app/test_multi_agent_manager_startup.py -q
309 passed, 1 warning in 2.42s

$ pre-commit run --all-files   # (summary, run as commit hooks)
check python ast.........Passed
check docstring is first..Passed
fix python encoding pragma.Passed
trim trailing whitespace..Passed
Add trailing commas.......Passed
mypy......................Passed
black.....................Passed
flake8....................Passed
pylint....................Passed
```

Real API run:

```
A flagship chat + flash memory    provider=dpsk           answer='OK' -> PASS
B deepseek chat + siliconflow mem provider=guijiliudong   answer='OK' -> PASS
C unset memory_model fallback     provider=guijiliudong   answer='OK' -> PASS
D bare name inherits provider     provider=guijiliudong   answer='OK' -> PASS
ALL PASS
```

## Additional Notes

- The Console persists explicit selections in the canonical object form and preserves stale provider/model selections so users can see and replace unavailable configurations.
- The local environment has `reme-ai==0.4.0.5` while QwenPaw pins `0.4.1.11`, so the full embedded ReMe job was not run locally; the real-API verification above exercised the complete model-injection path instead.


jinliyl (changes requested):
Thank you for working on this feature. Separating ReMeLight's memory-writing model from the main chat model is useful, and the backend direction looks good. I checked out the PR branch, reviewed the backend/configuration/frontend paths, and ran the relevant test suites locally.

Before merging, I think the following items should be addressed.

## Required changes

### 1. Please expose `memory_model` in the Console UI

The PR currently adds `memory_model` only to the TypeScript API interface. There is no corresponding control in `ReMeLightMemoryCard`, so users cannot see or change this setting from the Console; they can only edit `agent.json` manually. The type addition and deep-merge behavior preserve the field during unrelated form saves, but they do not make it editable.

The documentation now says that the separate memory model can be configured in the Console, which does not match the current implementation.

Could you please add a model selector to the ReMeLight memory settings? A suggested UX is:

- “Use the agent's main model” as the default option, persisted as `null`.
- A provider/model selector populated from configured providers and models.
- Persist an explicit selection as `{ "provider_id": "...", "model": "..." }` rather than as a string.
- Preserve and display a stale selection if its provider or model was removed, instead of silently clearing it when the form is opened.
- Add frontend tests covering initial display, changing the model, clearing it back to the main model, and saving/round-tripping the value.

### 2. A bare model name containing `:` is parsed incorrectly

`parse_memory_model_slot()` treats every string containing `:` as `provider_id:model`. This makes common Ollama-style model names ambiguous. For example:

```text
qwen2.5:7b -> provider_id="qwen2.5", model="7b"
```

According to the documented contract, a bare model name should inherit the main model's provider. In this case the code instead looks for a provider named `qwen2.5`, fails, and falls back to the main model, so the requested memory model is never used.

Please either make the object form the canonical/only new format, or treat the prefix as a provider only when it matches a configured provider ID. Please also add a regression test for a bare version-tagged model such as `qwen2.5:7b`.

### 3. The fallback guarantee does not cover runtime model failures

The current fallback catches `ProviderError` only while `create_model_and_formatter_async()` is constructing the model. For OpenAI-compatible providers, arbitrary model IDs can be instantiated successfully; authentication errors, missing-model errors, and connection failures usually occur only on the actual LLM request. Those failures therefore do not fall back to the agent's main model, and the memory job still fails.

Also, passing a model override causes the normal agent fallback chain to be skipped. As a result, the PR description's statement that an unavailable memory model falls back to the main model so memory jobs never break is currently stronger than the implementation.

Please implement fallback at the model-call level (memory model -> agent main model), or narrow the documented guarantee to configuration/instantiation failures only. A test where model construction succeeds but the memory-model call fails would be valuable.

## Simplification suggestions

`_memory_model_slot()` loads the agent configuration, and `create_model_and_formatter_async()` then loads it again through the synchronous factory path. The configuration can be loaded once in `_update_qwenpaw_model()`, used to resolve the memory slot, and passed to the factory through its existing `agent_config` argument. This removes duplicate I/O and avoids resolving the slot and the model settings from two potentially different snapshots.

There is also overlapping slot-parsing logic in `ReMeLightMemoryConfig` and the model factory. It would be cleaner to centralize that normalization in one helper. Since the Console should persist the object form, string parsing could remain only as a compatibility layer.

## Branch status and verification

The PR is currently 70 commits behind `main` and GitHub reports it as conflicting. The visible conflict is in `console/src/api/types/agent.ts`; when updating the branch, please preserve both `memory_model` and the newer `reranker_config` field from `main`.

Local verification completed successfully:

```text
44 passed  - new targeted configuration/model-selection tests
428 passed - memory, configuration, and multi-agent startup suites
```

The broader run produced one existing asynchronous cleanup warning, unrelated to this PR.

Thanks again for the contribution. Once the Console control, colon-containing bare model handling, and fallback semantics are addressed, this should be a useful improvement.

iluv7:
Thank you very much for the thorough review and for running the relevant test suites locally. I really appreciate the detailed feedback and concrete suggestions.

I’ll update the PR to:

- Add a `memory_model` selector to the Console, including stale-value preservation and frontend tests.
- Fix the ambiguity around colon-containing bare model names such as `qwen2.5:7b`, and add a regression test.
- Rework the fallback behavior so runtime model-call failures can fall back to the agent’s main model, with corresponding tests.
- Simplify the implementation by avoiding duplicate configuration loads and centralizing the slot-normalization logic where practical.
- Rebase onto the latest `main`, resolve the conflict, and preserve both `memory_model` and `reranker_config`.

Thanks again for the careful review. I’ll address these items and request another review once the updates are ready.

iluv7:
@jinliyl Thank you again for the detailed review. I have addressed the requested changes:

- Added the Console `memory_model` selector, including main-model default handling, stale-selection preservation, and frontend tests.
- Fixed colon-containing bare model names such as `qwen2.5:7b`; the object form is now canonical, and slot normalization is centralized.
- Added model-call-level fallback from the memory model to the agent main model, including a regression test where construction succeeds but the runtime call fails.
- Removed the duplicate agent-config load and updated the documentation and PR description.
- Rebasing/conflict resolution is complete, preserving both `memory_model` and `reranker_config`.

Verification completed:
- Console memory component: 55 tests passed
- Expanded backend suite: 491 tests passed (1 existing async cleanup warning)
- Formatting/pre-commit CI fixes have also been applied

The branch is ready for another review when you have time. Thanks!

iluv7:
Manually verified the Console configuration flow end-to-end. The Memory writing model selector renders correctly and defaults to the agent’s main model when no dedicated model is selected.
image

iluv7:
ping

Issue #7722, opened 2026-09-12
[Bug]: Memory exhaustion compounds through three paths — unbounded stream buffers, keep-alive instance stacking, and doom-loop gate evasion (controlled repro + minimal fixes)


Nobodyanonymou-s:
## QwenPaw Version

v2.2.0 (official `agentscope/qwenpaw:latest` image, `io.qwenpaw.managed-runtime-boundary=2.2.0`)

## Description

Container memory exhaustion (fills at ~1MB/s, then the service hangs/OOMs) is not one bug but **three compounding paths**. #7222-style slow growth is only path C; the two faster paths below are unreported, and **strong models mask them** — which may be why the fast paths are hard to reproduce and #4265/#2992 have stalled.

| Path | Mechanism | Trigger | Scale |
|---|---|---|---|
| A. Loop | context mismatch + varied-args tool loop evading DoomLoopGate | weak model + default config + big files | seconds–minutes |
| B. Stream | infra slowdown stretches run lifetime; replay buffers/queues unbounded; failures silent | flaky/slow relay | minutes–hours |
| C. Keep-alive | reload with active runs stacks old workspace instances (24h cap, stuck tasks never cancelled) | model switch while runs active | hours–days |

**Real-production reproduction (most stable recipe)**: weak self-hosted infra — Qwen3.6-27B served by vLLM on 4× L20 GPUs; QwenPaw v2.2.0 launched with container limits (**4 cores / 8GB memory**); default context settings (**max_input 131072 / max_token 8192** while the vLLM endpoint actually serves 8192 — a 16× window mismatch); agent runs long-horizon heavy tasks (image-review workflows are the most reliable trigger). This configuration reproduces **~1MB/s memory fill followed by a hung service** fairly consistently.

**Related PR(s):** #7723 (first of a series; working fixes for all six items exist and are measured — see the comment below)

**Security considerations:** none (memory/streaming behavior; no auth or config exposure)

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** v2.2.0 official image
[...]

Issue #4333, opened 2026-05-14
📢 [Contribution Policy] Regarding AI-Generated Issues and Pull Requests


xieyxclack:
# Regarding AI-Generated Issues and Pull Requests

Thank you for your interest in the QwenPaw project and your enthusiasm for contributing. We truly value every member of this community.

Recently, we have noticed a growing number of issues and pull requests that appear to be entirely generated by AI tools (e.g., LLMs, coding agents) without meaningful human review. While we are open to — and even encourage — the use of AI-assisted development ("vibe coding"), **contributors bear the responsibility of being the first line of quality control**. Submitting fully automated, unreviewed content shifts the burden of validation entirely onto maintainers, which is unsustainable and counterproductive for everyone involved.

## Contribution Guidelines

Before submitting an issue or pull request, please ensure the following:

### For Issues

1. **Reproduce the problem yourself.** Confirm that the behavior you are reporting is a genuine, reproducible bug in QwenPaw — not a misconfiguration, a misunderstanding of intended behavior, or an AI hallucination.
2. **Provide your actual environment details.** Include your real QwenPaw version, OS, Python version, and steps to reproduce. Do not fabricate or let AI guess these details.
3. **Follow the issue template.** We provide issue templates for bug reports and feature requests. Please fill in every section — do not delete or skip fields. A well-formed issue is faster to triage and more likely to receive a timely response.

### For Pull Requests

1. **Verify the root cause.** Confirm that the issue you are fixing has been accurately diagnosed. Understand *why* the bug occurs, not just what the symptoms are.
2. **Validate the fix is correct and safe.** Ensure your changes do not break existing functionality. Beyond correctness, we encourage contributors to consider whether the fix aligns with the framework's overall architecture and represents a *minimal necessary change* — this leads to more elegant, maintainable solutions and is greatly appreciated during review.
3. **Test locally and show your work.** Run the relevant test suite and `pre-commit` checks. Include concrete evidence of your testing results (logs, screenshots, or command output) in the PR description.
4. **Follow the pull request template.** Our PR template includes sections for motivation, changes, and testing evidence. Please complete all sections. Incomplete or template-skipping PRs may be closed without review.

## On the Use of AI Tools

We welcome and encourage the use of AI tools to help analyze codebases, draft fixes, and accelerate development. However, this must be built on a foundation where **the contributor has carefully reviewed all AI-generated content** and can vouch for its correctness. Specifically:

[...]

PR #7723, opened 2026-09-12
fix(console): emit an error event when stream_one fails

### LINKED ISSUE #7814: Console SSE: _strip_event_headlines can emit a bare null payload, and stream_one sends no terminal event on failure
## Summary

Two related robustness gaps on the Console SSE path:

1. `BaseChannel._strip_event_headlines` can serialize to the bare string `null`, producing an invalid SSE event whose payload is not an object.
2. When a turn fails mid-stream, `ConsoleChannel.stream_one` logs the exception and returns without emitting any terminal event, so the client is never told the turn is over.

Together these turn a single recoverable backend error into a client-side freeze that only a page reload clears. A companion report covers the frontend crash that (1) can trigger.

## Environment

- QwenPaw 2.2.1, Windows desktop build (`/api/version` returns `2.2.1`)
- Console channel, streaming chat via `POST /api/console/chat`
- Both issues verified against current `main`, not just the 2.2.1 tag

## Issue 1: a bare `null` SSE payload

`src/qwenpaw/app/channels/base.py`, function `_strip_event_headlines`.

The helper walks a dumped copy of the event and re-serializes it:

```python
def walk(node: Any) -> Any:
    if isinstance(node, str):
        return strip_headline(node)
    if isinstance(node, dict):
        for key, value in list(node.items()):
            node[key] = walk(value)
        return node
    if isinstance(node, list):
        return [walk(value) for value in node]
    return node

payload = walk(payload)
return json.dumps(payload, ensure_ascii=False, default=str)
```

There is no branch for `None`, so it falls through to `return node` and `payload` can remain `None`. `json.dumps(None)` is the string `'null'`. Back in `stream_one` this is wrapped as:

```python
data = self._serialize_event_for_sse(event, headline_stream_states)
yield f"data: {data}\n\n"
```

which puts `data: null` on the wire. A client doing `JSON.parse` on that frame gets `null` rather than an exception, and any consumer that assumes an object will fail on the very next property access.

Note the type annotation on the parameter is `fallback: str`, while the implementation treats `payload` as a dict. That mismatch is probably how the `None` case went unnoticed.

Two possible directions, either seems fine:

- give `walk` / the tail an explicit `None` fallback to `fallback`, and correct the annotation
- make `_serialize_event_for_sse` validate its result and substitute a valid JSON object (e.g. an error payload) whenever the serialization is not an object

The second one is worth doing regardless, since it closes the whole class of "serializer produced a non-object" rather than just this instance.

## Issue 2: no terminal event when a turn fails

`src/qwenpaw/app/channels/console/channel.py`, function `stream_one`.

The error path currently looks like this on `main`:

```python
except Exception as e:
    self._clear_session_turn_usage(session_id)
    logger.exception("console process/reply failed")
    err_msg = str(e).strip() or "An error occurred while processing."
    self._print_error(err_msg)
finally:
    ...
```

The exception is logged and printed to the terminal, but nothing is yielded. The response never reaches `completed` or `failed`, so a client that keys its UI state off those events stays in a generating state indefinitely — the message bubble keeps spinning and the composer stays blocked.

The router does have a fallback for generator-level errors (`src/qwenpaw/app/routers/console.py` yields `{"error": ...}` when iteration raises), but that does not help here because `stream_one` swallows the exception and returns normally. The stream therefore ends cleanly from the router's point of view, with no terminal event ever sent.

Suggestion: emit an explicit terminal event on this path, e.g. a failed `message`/`response` event carrying the error text, so clients can settle the turn and recover without a reload.

## How I hit this

A turn failed server-side with a provider error. The backend logged `console process/reply failed` and wrote a normalized error dump, which is expected behavior. The client never received a terminal event, so the chat stayed stuck until the page was reloaded.

## Verification notes

- Both code paths above were read from current `main` and from the v2.2.1 tag; neither has changed.
- I have not captured the raw `data: null` frame itself, so I cannot say which event type first produced a `None` payload. The `None` handling gap in `walk` is read directly from the source, and the `json.dumps(None)` behavior is straightforward. Flagging that distinction so the finding is not read as stronger than it is.
- Related but unexplained, and probably a separate issue: the same session log contained `Message not found for content: null` 52 times, all with `msg_id` null. I have not established any link to the above.

Nobodyanonymou-s:
## Description

The console channel's generic exception branch in `stream_one` logged the failure and ended the SSE stream silently — clients could not distinguish a failed turn from a completed one and retried the full context blind. (The `ModelQuotaExceededException` branch right above already yields a `rate_limited` event, so the asymmetry is visible in the code.) With a dead/misconfigured provider we measured the stream ending in 0.2s with zero indication to the client.

This mirrors the quota branch: yield `{"type": "error", "error": <message>}` before ending.

**Related Issue:** Relates to #7722

**Security Considerations:** the error message is the exception's `str()`, same exposure as the existing quota branch; no auth/config changes.

## What Problem This Solves

**Before (measured on the official v2.2.0 image):** with the provider
pointing at a dead endpoint, `POST /api/console/chat` ended the SSE stream
in ~0.2 s with zero terminal signal — no `failed` status, no error event.
The console client cannot distinguish a failed turn from a completed one,
so it leaves the turn looking finished (or the user retries and resends
the full context blind). The `ModelQuotaExceededException` branch a few
lines above already yields an error event, so the generic branch is the
asymmetric gap.

**After:** the same request ends with `{"status": "failed"}` followed by
`{"type": "error", "error": "Connection error."}`, letting clients render
the failure and stop. Full SSE transcript is in the Evidence section
below; reproduction steps are in `## Testing`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run` locally on the changed file and it passes (one exception: `actionlint` skipped — its Go module mirror is unreachable from this network and the hook only lints GitHub Actions files, not applicable to this .py change; all Python hooks passed)
- [ ] If pre-commit auto-fixed files, I committed those changes and rerun checks (n/a — no files were modified by the hooks)
- [ ] Documentation updated (if needed) (n/a — behavior is internal to the SSE protocol)
- [x] Ready for review

- [ ] I ran `./scripts/check-channels.sh` (or `--changed`) and it passes (not run: the wrapper needs a dev checkout with the project venv; its primary gate — the contract test suite below — was run directly and passes)
- [x] **Contract test** exists in `tests/contract/channels/test_console_contract.py` (pre-existing, unaffected)
- [x] Contract test implements `create_instance()` with proper channel initialization (pre-existing)
- [x] All contract verification points pass (20/20, see Evidence)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic (existing `tests/unit/app/channels/test_console_channel.py` covers this path, 30/30 pass)

## Testing

Point the active provider at a dead endpoint and send a console chat:

```
PUT /api/models/<id>/config  {"base_url": "http://127.0.0.1:9/v1", ...}
POST /api/console/chat       {"channel":"console","user_id":"t","session_id":"t1",
                              "input":[{"content":[{"type":"text","text":"hi"}]}]}
```

Before: the SSE stream ends with no error event (0.2s). After: the last event is
`data: {"type": "error", "error": "Connection error."}`

## Evidence

```
$ pytest tests/unit/app/channels/test_console_channel.py -q
30 passed in 9.22s

$ pytest tests/contract/channels/test_console_contract.py -q
20 passed in 1.17s

$ pre-commit run --files src/qwenpaw/app/channels/console/channel.py
check python ast..............................Passed
check docstring is first......................Passed
fix python encoding pragma....................Passed
detect private key............................Passed
trim trailing whitespace......................Passed
Add trailing commas...........................Passed
mypy..........................................Passed
black.........................................Passed
flake8........................................Passed
pylint........................................Passed
prettier...........................(no files to check)Skipped
Lint GitHub Actions workflow files............Skipped (network; see Checklist note)

# SSE transcript, provider pointed at a dead endpoint — after the fix:
$ curl -N -X POST .../api/console/chat -d '{"...":"hi"}'
data: {"id":"response_...","status":"created",...}
data: {"id":"response_...","status":"in_progress",...}
data: {"id":"response_...","status":"failed",...}
data: {"type": "error", "error": "Connection error."}
```

## Additional Notes

**Tooling disclosure (per #4333):** this patch was developed with AI assistance (Claude Code) under human direction. The reproduction and all evidence above come from real runs against the official v2.2.0 image; the analysis is code-anchored, and every test/verification listed was actually executed. Happy to defend any part of it in review. Part of a series tracked in #7722.

### DIFF
--- a/src/qwenpaw/app/channels/console/channel.py
+++ b/src/qwenpaw/app/channels/console/channel.py
@@ -525,6 +525,13 @@ class ConsoleChannel(BaseChannel):
             self._clear_session_turn_usage(session_id)
             logger.exception("console process/reply failed")
             err_msg = str(e).strip() or "An error occurred while processing."
+            err_event = _json.dumps(
+                {
+                    "type": "error",
+                    "error": err_msg,
+                },
+            )
+            yield f"data: {err_event}\n\n"
             self._print_error(err_msg)
         finally:
             try:


rayrayraykk (commented):
-

PR #7725, opened 2026-09-12
fix(workspace): replace blocking watchfiles.awatch SSE watcher with threaded polling

### LINKED ISSUE #7721: # [Bug]: Workspace file browser (SSE `/api/workspace/watch`) freezes the entire server when the workspace contains large repos — `watchfiles.awatch` RustNotify sync init blocks the event loop
## QwenPaw Version

2.2.1 (`qwenpaw --version`), running inside a Docker container

## Description

Opening the Console file browser (Workspace → Files) hangs the **entire** QwenPaw server: the WebUI stops responding, all channels (Feishu/QQ/OneBot) stop processing messages, and the process cannot even handle SIGTERM (it fails to shut down within Docker's 10 s grace window and is force-killed with SIGKILL, exit code 137).

Root cause: the file browser opens an SSE connection to `GET /api/workspace/watch`, which is served by `workspace_watch_events` in `src/qwenpaw/app/routers/workspace.py`. That handler builds the watcher **synchronously on the event-loop thread**:

```python
async def workspace_watch_events(request, watch_dir):
    yield 'data: {"type": "connected"}\n\n'
    watcher = awatch(watch_dir, rust_timeout=1000, yield_on_timeout=True)  # <-- sync, on the loop
    while True:
        raw_changes = await watcher.__anext__()  # first call runs RustNotify.__enter__
        ...
```

`watchfiles.awatch` (watchfiles 1.2.0, `main.py:257`) does:

```python
with RustNotify(...) as watcher:          # __enter__ = recursive initial scan, NOT offloaded
    ...
    raw_changes = await anyio.to_thread.run_sync(watcher.watch, ...)  # only polling is offloaded
```

`RustNotify.__enter__` builds the initial file snapshot by recursively scanning the whole watched tree. In a workspace containing large repos (in our case 192,898 files / 6.2 GB under `workspace/coding_projects/`, including `deepseek-harness` 1.9 GB with 64,230 files in `node_modules`), this initial scan blocks the event loop for **25 s+** (measured), so every request in the process — HTTP, channel WebSockets, and the SIGTERM handler — is starved. Only the periodic `watcher.watch` polling is offloaded to a thread; the initialization scan is not.

**Related PR(s):** none (no existing issue/PR on this). Closely related symptom-class issue: **#7363** (synchronous calls freeze the event loop and timeout never fires) — but that one is the embedding health-check path, a different root cause. **#7261** (SSE serialization loop, 100% CPU) is a different, already-fixed bug.

**Security considerations:** No sensitive data. Paths shown are the reporter's local container layout only.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.2.1
- **OS:** Debian 12 (bookworm) inside Docker container (host: Ubuntu 24.04)
- **Install method:** Docker (custom image based on `agentscope/qwenpaw`)
- **Python version (if applicable):** 3.11.2
- **watchfiles:** 1.2.0

## Steps to Reproduce

1. Make the agent workspace large: clone/copy several big repos into `workspace/coding_projects/` so the tree contains ~190k+ files / >6 GB (e.g. a repo with `node_modules`). Smaller workspaces work fine.
2. Start QwenPaw, open the Console WebUI.
3. Open the **file browser** (Workspace → Files). The frontend issues `GET /api/workspace/watch?root=workspace` (SSE).
4. Observe: the WebUI becomes unresponsive, Feishu/QQ channels stop answering, `/api/*` requests no longer return.
5. Restart the container: SIGTERM is sent but the process cannot exit within 10 s, so Docker sends SIGKILL (exit code 137).

## Actual vs Expected

- **Actual:** After the file browser opens the watch SSE, the entire server freezes; all HTTP/channel processing halts; the process cannot shut down gracefully on SIGTERM and is SIGKILLed.
- **Expected:** The watch SSE establishes without blocking the event loop; file-change events stream asynchronously; the server keeps serving other requests.

## Logs / Screenshots

Measured `awatch` initialization latency on the reporter's workspace (Python 3.11, watchfiles 1.2.0, `rust_timeout=1000`, `yield_on_timeout=True`):

```
/tmp (small dir):                       first __anext__ returned in 1.02 s  ✅
/app/.../silver-wolf (192,898 files):   first __anext__ did NOT return in 25 s (timed out) ❌
```

SSE handler is where the freeze starts — the last successfully served request before the hang:

```
INFO: 172.20.0.1:51782 - "GET /api/workspace/watch?root=workspace HTTP/1.1" 200 OK
(no further requests ever served until forced restart)
```

Supervisor/container level (SIGTERM → 10 s → SIGKILL, exit 137):

```
2026-09-12 13:53:38 WARN received SIGTERM indicating exit request
2026-09-12 13:53:48 container die exitCode=137   (SIGKILL after grace timeout)
```

## Additional Notes

**Suggested fix:** move the watcher construction (and the first `__anext__`) off the event loop, e.g. build the `awatch`/`RustNotify` inside `asyncio.to_thread`, or set an explicit timeout so the SSE stream degrades (e.g. fall back to a polling implementation or a "workspace too large to watch" response) instead of blocking the process. Passing `ignore_dirs`/a `watch_filter` that excludes `node_modules`, `.venv`, `target`, etc. would also cut the scan cost, though the blocking itself should be fixed regardless of scale.

**Workaround (immediate):** avoid opening the file browser while the workspace contains very large directories; moving large repos out of the watched workspace restores normal behavior.

### LINKED ISSUE #7786: [Bug]: Cloud/NFS deployment — opening the workspace file browser freezes the whole process for 5–6 min (recurring); plus request-path blocking file I/O on the event loop
## QwenPaw Version

2.2.1 (`qwenpaw --version`), managed cloud container (Docker, 2 vCPU), `/app/venv/bin/qwenpaw app --host 0.0.0.0 --port 8088`.

## Description

Opening Console → Files (workspace tab) and clicking a document makes the **entire** instance unresponsive: the WebUI stops answering (reload does not help), channels go silent, and it "recovers by itself" after several minutes. This is the same user-visible symptom as **#7721** (`watchfiles.awatch` synchronous recursive baseline scan on the event loop, `src/qwenpaw/app/routers/workspace.py:1103`), but on a **network-filesystem deployment the stall is 5–6 minutes and repeats**, and the profile shows **several additional blocking-on-the-loop file I/O sites that #7721 / #7725 do not cover**.

Measured freeze windows from our own `qwenpaw.log` (single continuous process, no restart — timestamps are container-local):

```
GAP 271s : 12:49:05 -> 12:53:36
GAP 353s : 12:53:37 -> 12:59:30
GAP 378s : 12:59:30 -> 13:05:48
GAP 363s : 13:05:53 -> 13:11:56
# and at 12:53:36 every channel watchdog fires at once:
WARNING | qwenpaw/app/channels/dingtalk/channel.py:2647 | dingtalk: liveness watchdog detected wake-from-sleep (elapsed=271s, expected~30s); forcing reconnect...
ERROR   | qwenpaw/app/channels/wechat/channel.py:613     | wechat poll error (1 consecutive), retry in 5s
WARNING | plugins/cmcc-newmsg-channel/backend/ws_client.py:196 | cmcc heartbeat timeout (no traffic 252 s); recycling connection
```

A thread that only sleeps and checks a clock every 30 s produced **nothing for 271 s**, then everything unblocked in the same second — i.e. this is a whole-process freeze, not a slow request.

While the instance was frozen we sampled it externally (`py-spy record -p <pid> --rate 20 --duration 420`, speedscope output, 19 threads):

- Total GIL-visible Python time across all 19 threads: **~53 s out of 420 s**; MainThread only **2.9 s**. This rules out CPU saturation, a hot loop, and long-term GIL hogging — the threads are simply blocked inside syscalls.
- All six `asyncio_*` worker threads spend almost all their time in `concurrent/futures/thread.py:81 _worker` (waiting on the queue): the to_thread pool is *not* the bottleneck, it is starved.
- MainThread leaf frames include exactly the blocking sites below:
  `awatch @ main.py:257` (watchfiles), `_load_auth_data @ auth.py:229`, `_scandir @ pathlib.py:939`, `stat @ pathlib.py:1014`.

### Blocking file I/O still performed on the event loop

1. **Auth middleware re-reads and re-decrypts a secret file 3× per authenticated request.**
   `AuthMiddleware.dispatch` → `has_registered_users()` (`app/auth.py:352`), `verify_token()` → `_get_jwt_secret()` (`app/auth.py:177`) and `_is_token_revoked()` (`app/auth.py:272`) each call `_load_auth_data()` (`app/auth.py:229`), which does `AUTH_FILE.is_file()` + `open()` + `json.load()` + Fernet decryption **with no cache**. `AUTH_FILE` lives in `SECRET_DIR`, which on this deployment is a *different* NFS mount from the workspace. Every HTTP/WS request therefore costs 3 NFS read+decrypt round trips **on the loop thread**.
2. **`/workspace/files` list + read are not offloaded at all.**
   `list_working_files` (`app/routers/workspace.py:164`) and `read_working_file` (`:189`) call `AgentMdManager.list_working_mds()` (glob + 2× `stat` per file) and `read_working_md()` (whole-file read) **directly inside the `async def`**, while every sibling endpoint in the same module correctly uses `async with _FILESYSTEM_SEMAPHORE: await asyncio.to_thread(...)`.
3. **The log handler writes synchronously to the network filesystem.**
   `utils/logging.py:324` installs `_SafeRotatingFileHandler` on `LOG_FILE_PATH = WORKING_DIR / "qwenpaw.log"` (5 MiB × 3 backups). When the NAS slows down, the first thread that logs blocks in the NFS write **while holding the logging lock**, so every other logging thread (including channel watchdogs) queues behind it. That is precisely why the process goes completely silent instead of degrading gracefully.
4. **Semaphore / default-executor size mismatch.**
   `_FILESYSTEM_SEMAPHORE = asyncio.Semaphore(8)` (`app/routers/workspace.py:92`) admits 8 concurrent filesystem operations into `asyncio.to_thread`'s default executor, which on a 2-vCPU container is `min(32, cpu_count+4)` = **6 threads** (confirmed: exactly six `asyncio_0..5` threads in the profile). A handful of slow NAS operations therefore saturate the pool that *all* file endpoints and session reads/writes share.
5. **The recursive baseline scan is unpruned, and skip-list filtering happens after the walk.**
   `_list_all_files` (`app/routers/workspace.py:258`) documents the pruning fix for the *listing* path, but `RustNotify`'s initial scan in `awatch` walks everything under the workspace. Our workspace measures 5,456 directories / 32,421 files with an **unpruned** `os.walk` that did not finish in 25 s (single-file `stat` ≈ 4 ms over NFS, `listdir` of the root = 151 entries in 0.00 s) → a full recursive scan against `hard`-mount NFS is a minutes-long, not seconds-long, operation.
6. **A second `awatch` user outside the PR #7725 blast radius.**
   `reme/steps/index/watch_changes.py:93` also calls `watchfiles.awatch(...)` for the memory/digest index (`INFO | reme/steps/index/watch_changes.py:88 | Watching: [<workspace>/memory, <workspace>/digest]`, `step=1000ms debounce=5000ms`). #7725 replaces `awatch` only in the QwenPaw SSE endpoint, so the same event-loop-blocking init scan remains reachable from the memory indexer on every agent start, on every pod.

### Why it self-recovers
All mounts here are Aliyun NAS NFSv3 with `hard,timeo=600,retrans=2,nolock` fronted by a local stunnel (`127.0.1.x:12049`). `hard` means an RPC to a slow server is retried **indefinitely** rather than returning `EIO`, so the loop thread parks in an uninterruptible wait for minutes and then drains the whole queue at once — "refresh does nothing, and after a few minutes it is fine again".

**Related PR(s):** #7725 (fixes #7721; covers the SSE watcher only — items 1–6 above remain)
**Related Issue(s):** #7721 (same trigger, local/ext4 workspace, ~25 s freeze), #7363 (same symptom class, different call path), #7261
**Security considerations:** No secrets, tokens or credentials in this report. All identifiers below are redacted: NAS volume ids and the NFS server FQDN → `<NAS>`, absolute workspace path → `<WORKSPACE>`, agent pid → `<pid>`. Note only that `auth.json` (encrypted fields) sits on the mounted volume — unchanged behaviour, no exposure introduced here.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [x] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.2.1 (installed wheel, `/app/venv`)
- **OS:** Linux 5.10 (containerized, Debian base, Python 3.11.2)
- **Install method:** Docker / managed cloud instance (QwenPaw cloud deployment)
- **Python version:** 3.11.2
- **Hardware:** 2 vCPU container; `WORKING_DIR`, `SECRET_DIR`, `user-packages` and backups are **all NFS (Aliyun NAS) mounts**; `watchfiles==1.2.0`
- **Workspace size:** 5,456 dirs / 32,421 files (unpruned walk > 25 s, did not complete)

## Steps to Reproduce

> We can no longer re-run this against the live pod (the reproduction itself wedges a `hard`-NFS deployment), so these steps are for a controlled replica.

1. Deploy QwenPaw 2.2.1 with `WORKING_DIR` (workspace + `qwenpaw.log`) and `SECRET_DIR` on an NFSv3 `hard` mount, workspace populated with ~30k files (any long-lived agent with `skills/`, `memory/`, `sessions/`, `mem_metadata/`, `tool_results/` grows there naturally).
2. Open Console → Files → workspace tab, and click any document (including a plain `.md`).
3. Observe: the file content never renders; **any** other route in the same process (even `GET /api/version` from another tab, or a channel webhook) stops responding; reload does not help.
4. Watch `qwenpaw.log`: no new lines for minutes, then a burst of `liveness watchdog detected wake-from-sleep (elapsed=NNNs, expected~30s)` from several channels at one timestamp.
5. Optional quantification without touching the loop: time `os.walk` over the workspace in a separate process (seconds/minutes), and count NFS ops/RTT in `/proc/self/mountstats`.

## Actual vs Expected

- **Actual:** Opening the file browser freezes the whole process for 5–6 minutes at a time, repeatedly. Channels drop websockets, and the instance looks dead until the NAS answers. Metadata-heavy workspaces on network storage make a "25 s" bug into a multi-minute one.
- **Expected:** Filesystem work never runs on the event loop. A slow or partially unresponsive NAS may make the *file panel* slow, but `/api/version`, chat, and channels keep serving; any scan has an entry/depth/time budget and its result is cached; blocking work goes to a pool sized to match `_FILESYSTEM_SEMAPHORE`.

## Logs / Screenshots

```
# (a) process-level freeze, from qwenpaw.log (redacted pod)
2026-09-15 12:49:05 | INFO  | qwenpaw/agents/context/scroll/manager.py:466 | scroll: compact timing ...
        <<< no output from ANY thread for 271 s >>>
2026-09-15 12:53:36 | WARNING | qwenpaw/app/channels/dingtalk/channel.py:2647 | dingtalk: liveness watchdog detected wake-from-sleep (elapsed=271s, expected~30s); forcing reconnect...
2026-09-15 12:53:36 | ERROR | qwenpaw/app/channels/wechat/channel.py:613 | wechat poll error (1 consecutive), retry in 5s
2026-09-15 12:59:30 | WARNING | qwenpaw/app/channels/qq/channel.py:1862 | qq ws connect failed: [Errno 32] Broken pipe
2026-09-15 13:05:49 | WARNING | qwenpaw/app/channels/dingtalk/channel.py:2647 | dingtalk: liveness watchdog detected wake-from-sleep (elapsed=733s, expected~30s); forcing reconnect...
2026-09-15 13:11:56 | WARNING | qwenpaw/app/channels/dingtalk/channel.py:2647 | dingtalk: liveness watchdog detected wake-from-sleep (elapsed=367s, expected~30s); forcing reconnect...

# (b) mounts (from /proc/self/mounts, ids redacted)
<NAS>:/qwenpaw-data   <WORKSPACE>  nfs  rw,relatime,vers=3,rsize=1048576,wsize=1048576,hard,nolock,noresvport,proto=tcp,port=12049,timeo=600,retrans=2,sec=sys,...,addr=127.0.1.1
<NAS>:/qwenpaw-secrets <SECRET_DIR> nfs  rw,relatime,vers=3,...,hard,nolock,...,timeo=600,retrans=2,...

# (c) external sampling during the freeze
$ py-spy record -p <pid> -f speedscope -o rec.json --rate 20 --duration 420
py-spy> Sampling process 20 times a second for 420 seconds.
py-spy> 1.07s behind in sampling, results may be inaccurate. Try reducing the sampling rate
# aggregated leaf frames (weights x 50 ms), 420 s window, 19 threads, idle dropped (default):
    30.2s  _worker @ thread.py:81                    (x6 asyncio_* pool threads — queue wait)
    11.2s  _do_waitpid @ unix_events.py:1403
     7.4s  run @ _asyncio.py:986
     3.8s  read @ ssl.py:1169
     0.6s  awatch @ main.py:257           <- MainThread
     0.6s  _scandir @ pathlib.py:939      <- asyncio_* pool
     0.2s  _load_auth_data @ auth.py:229  <- MainThread
     0.1s  stat @ pathlib.py:1014         <- MainThread/pool
MainThread total GIL-visible time in 420 s: 2.9 s   (all threads combined: ~53 s)

# (d) NAS metadata cost, measured in a separate process (never on the loop)
listdir(<WORKSPACE>): 0.00 s, 151 entries
50 x stat:            0.20 s   (~4 ms/stat)
glob *.md:            0.00 s, 15 files
read 26 KB file:      0.01 s
os.walk(<WORKSPACE>): 25.1 s -> 5456 dirs / 32421 files, STILL RUNNING (no prune)

# (e) executor size
nproc=2  ->  asyncio default to_thread max_workers = min(32, cpu_count+4) = 6
_FILESYSTEM_SEMAPHORE = asyncio.Semaphore(8)   # app/routers/workspace.py:92
```

## Additional Notes

Suggested fixes, ordered by value/risk:

1. **Adopt #7725, and additionally budget + cache the tree scan** (entry cap, depth cap, wall-clock cap, prune `node_modules`/`.venv`/`.git` *during* the scan, cache the snapshot with a short TTL). A per-workspace-tree scan on every SSE connect is not viable on network storage even when it is offloaded.
2. **Cache auth state in-process** (mtime-invalidated, or load once and refresh on change), or at minimum run `verify_token()`'s file access via `asyncio.to_thread`. The middleware is on every request; on NAS this is the difference between "slow" and "down".
3. **Offload `list_working_files` / `read_working_file`** like their sibling endpoints, under the same semaphore.
4. **Never block on the log write path**: `logging.handlers.QueueHandler` + `QueueListener` (writer thread, `delay=True`), or put `qwenpaw.log` on local disk. Add an env var to override `LOG_FILE_PATH` (currently only derivable from `WORKING_DIR`, with `QWENPAW_LOG_MAX_SIZE`/`QWENPAW_LOG_MAX_BACKUPS` overridable but not the path).
5. **Align pool and semaphore**: `loop.set_default_executor(ThreadPoolExecutor(max_workers=32, thread_name_prefix="fs"))` at startup, and set `_FILESYSTEM_SEMAPHORE` to the same or a smaller value.
6. **Extend the same treatment to ReMe's `watch_changes` `awatch`** (item 6 above), otherwise cloud/NFS pods keep the failure after #7725.
7. Docs for cloud/NFS deploys: recommend `soft`+bounded retries or increased attribute caching (`actimeo=`/`acdirmin=`), and warn that `WORKING_DIR` on `hard` NFS converts any blocking FS call into an unbounded freeze.

Deployment-side workarounds we can apply today, without a release: keep `qwenpaw.log` off the NAS (symlink to a local volume), raise the container CPU so the default executor grows, cap workspace file count (archive `tool_results/`, prune old `sessions/`), and avoid opening the Files panel on large workspaces.

xiaoka76:
## Description

Opening the file browser on a large workspace froze the **entire server**. The
`/api/workspace/watch` SSE stream used `watchfiles.awatch`, whose
`RustNotify` performs a **synchronous recursive baseline scan** inside
`__init__` (notify crate `watcher.watch(path, RecursiveMode::Recursive)` on
inotify). With `awatch` that scan runs on the event loop at the first
`await watcher.__anext__()` and **holds the GIL for its whole duration**
(watchfiles only offloads the poll loop via `anyio.to_thread`; the init scan
has no `py.detach`/`allow_threads`). On a 192,898-file / 6.2 GB workspace the
scan takes 25 s+, during which the event loop cannot serve any request, the
WebUI is unresponsive, and even SIGTERM is not handled (container killed after
10 s, exit 137).

This PR replaces `awatch` with a **pure-Python polling watcher owned by a
background thread**:

- `_scan_snapshot()` walks the tree with `os.scandir`, pruning dotfiles and
  `_SKIP_NAMES` (`node_modules`, `.venv`, `.git`, ...) while walking, and
  snapshots `(mtime_ns, size)` per file.
- `_poll_watch_worker()` runs the initial baseline scan and then diffs each
  poll, pushing `("ready", ...)` / `("changes", ...)` / `("done", ...)` to a
  `queue.Queue`; the async generator consumes it with a bounded `get(timeout)`.
- Plain Python releases the GIL at bytecode boundaries, so the event loop
  stays responsive during even the initial scan — the key property the Rust
  init scan did not have (verified empirically: the Rust scan starves other
  threads to ~0.8 % of their normal progress).
- `watchfiles` is no longer used by this endpoint (it was the only user in the
  codebase).

Two behavioral improvements are included:

1. A `ready` SSE frame after the initial baseline scan completes, so consumers
   know incremental events have started. The frontend (`useWorkspaceWatch`)
   ignores unknown frame types, so this is backward compatible.
2. Heartbeat comments now also fire during a long initial scan (the loop is
   never blocked waiting for it), keeping proxies from timing out.

**Related Issue:** Fixes #7721

**Security Considerations:** No new external surface — the worker thread only
scans the filesystem paths the agent is already allowed to browse, event paths
are already relative to the watched root, and no credentials/env handling is
touched. The new `ready` frame is ignored by the existing frontend consumer.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

  (ran on the changed files — all applicable hooks pass: check-ast,
  check-docstring-first, fix-encoding-pragma, trailing-whitespace,
  detect-private-key, add-trailing-comma, mypy, black, flake8, pylint;
  actionlint/prettier skipped as no workflow/JS files are touched)
  (`pytest tests/unit/app/routers/ -q` → 689 passed, 1 skipped)
- [x] Ready for review

## Testing

- New unit tests in `tests/unit/app/routers/test_workspace_router_watch.py`:
  snapshot pruning, diff (added/deleted/modified), worker
  ready→changes→done, full SSE flow (connected/ready/file_change/heartbeat/
  disconnect), event-loop-responsiveness during a slow scan, and
  heartbeat-during-scan.
- Updated the two existing watch contract tests
  (`test_workspace_files_router.py`) to the new implementation: idle polls
  keep the stream open and disconnect is observed after an idle wake-up.
- End-to-end in a Docker test image containing the real 6.3 GB / 33k-file
  workspace: `connected` at 0 ms, `ready` at 0.4 s (previously the server
  froze for 25 s+), file-change events arrive within one poll interval, and a
  concurrent request responds in ~2 ms while the baseline scan is running.

## Evidence

```bash
pre-commit run --files src/qwenpaw/app/routers/workspace.py \
  tests/unit/app/routers/test_workspace_router_watch.py \
  tests/unit/app/routers/test_workspace_files_router.py
# check python ast ............ Passed
# check docstring is first .... Passed
# fix python encoding pragma .. Passed
# detect private key .......... Passed
# trim trailing whitespace .... Passed
# Add trailing commas ......... Passed
# mypy ........................ Passed
# black ....................... Passed
# flake8 ...................... Passed
# pylint ...................... Passed

pytest tests/unit/app/routers/ -q
# 689 passed, 1 skipped, 7 warnings in 19.65s
```

Real-behavior verification inside the test container (6.3 GB workspace):

```
[connected] 'data: {"type": "connected"}'  (0ms)
[ready]     'data: {"type": "ready"}'      (0.40s)   # baseline scan of the full tree
[file_change added]   +2.7s  {"change":"added",   "path":"watch_e2e_probe.txt"}
[file_change deleted] +4.7s  {"change":"deleted", "path":"watch_e2e_probe.txt"}
# during the scan, a concurrent request to the console returned in 0.002s
# (event loop NOT blocked) — previously the whole server froze for 25s+
```

## Additional Notes

- Trade-offs of polling vs inotify: event latency is bounded by the poll
  interval (default 2 s, module constant `_WATCH_POLL_INTERVAL_SECONDS`), and
  while a file browser is open on a very large tree the pruned walk costs
  ~0.3 s per poll (~13 % CPU at the 2 s interval). This only occurs for trees
  where the previous implementation was completely unusable, and the walk is
  pruned so `node_modules`/`.venv`/`.git` are never scanned.
- Changes detected by `(mtime_ns, size)` — content edits that preserve both
  within one poll are not reported (rare; the next poll's size/mtime change
  catches most cases).
- The root cause also exists upstream in
  [samuelcolvin/watchfiles](https://github.com/samuelcolvin/watchfiles):
  `awatch`'s `RustNotify.__init__` should wrap the initial scan in
  `py.allow_threads`. If/when that ships, this endpoint could switch back to
  `awatch` + `asyncio.to_thread` — the polling watcher here is self-contained
  and dependency-free, so it works today.

### DIFF
--- a/src/qwenpaw/app/routers/workspace.py
+++ b/src/qwenpaw/app/routers/workspace.py
@@ -13,11 +13,13 @@ import io
 import json
 import logging
 import mimetypes
+import os
+import queue
 import secrets
 import shutil
 import stat
 import tempfile
-import os
+import threading
 import zipfile
 from collections.abc import Mapping
 from datetime import datetime, timezone
@@ -35,7 +37,6 @@ from fastapi import (
     UploadFile,
 )
 from fastapi.responses import ORJSONResponse, Response, StreamingResponse
-from watchfiles import awatch, Change
 from pydantic import BaseModel, Field
 
 from ..utils import check_upload_size, safe_join, schedule_agent_reload
@@ -91,7 +92,8 @@ router = APIRouter(prefix="/workspace", tags=["workspace"])
 logger = logging.getLogger(__name__)
 _FILESYSTEM_SEMAPHORE = asyncio.Semaphore(8)
 _WATCH_HEARTBEAT_SECONDS = 30.0
-_WATCH_POLL_TIMEOUT_MS = 1_000
+_WATCH_POLL_INTERVAL_SECONDS = 2.0
+_WATCH_QUEUE_TIMEOUT_SECONDS = 1.0
 
 
 class MdFileInfo(BaseModel):
@@ -245,10 +247,6 @@ _SKIP_NAMES: frozenset[str] = frozenset(
         ".hypothesis",
     },
 )
-
-
-def _should_skip(rel_parts: tuple[str, ...]) -> bool:
-    return any(p.startswith(".") or p in _SKIP_NAMES for p in rel_parts)
 
 
 def _is_skipped_name(name: str) -> bool:
@@ -1064,6 +1062,84 @@ async def write_code_file(
     return {"path": file_path, "size": size}
 
 
+def _scan_snapshot(watch_dir: Path) -> dict[str, tuple[int, int]]:
+    """Walk ``watch_dir`` and snapshot every non-ignored file.
+
+    Returns ``{rel_posix_path: (mtime_ns, size)}``. Dotfiles and
+    ``_SKIP_NAMES`` entries (``node_modules``, ``.venv``, ``.git``, ...) are
+    pruned while walking so huge dependency trees never get scanned. Runs as
+    plain Python, so it releases the GIL at bytecode boundaries and is safe
+    to call from a background thread.
+    """
+    snapshot: dict[str, tuple[int, int]] = {}
+    stack: list[tuple[Path, str]] = [(watch_dir, "")]
+    while stack:
+        dir_path, rel = stack.pop()
+        try:
+            with os.scandir(dir_path) as it:
+                for entry in it:
+                    if _is_skipped_name(entry.name):
+                        continue
+                    rel_path = f"{rel}/{entry.name}" if rel else entry.name
+                    try:
+                        if entry.is_dir(follow_symlinks=False):
+                            stack.append((Path(entry.path), rel_path))
+                        else:
+                            st = entry.stat(follow_symlinks=False)
+                            snapshot[rel_path] = (st.st_mtime_ns, st.st_size)
+                    except OSError:
+                        continue
+        except OSError:
+            continue
+    return snapshot
+
+
+def _diff_events(
+    prev: dict[str, tuple[int, int]],
+    cur: dict[str, tuple[int, int]],
+) -> list[dict[str, str]]:
+    """Diff two snapshots into ``{"change": ..., "path": ...}`` events."""
+    events: list[dict[str, str]] = []
+    for path in sorted(cur.keys() - prev.keys()):
+        events.append({"change": "added", "path": path})
+    for path in sorted(prev.keys() - cur.keys()):
+        events.append({"change": "deleted", "path": path})
+    for path in sorted(prev.keys() & cur.keys()):
+        if cur[path] != prev[path]:
+            events.append({"change": "modified", "path": path})
+    return events
+
+
+def _poll_watch_worker(
+    watch_dir: Path,
+    out: queue.Queue[tuple[str, object]],
+    stop: threading.Event,
+    interval: float,
+) -> None:
+    """Background poller: baseline snapshot, then periodic diff.
+
+    Emits ``("ready", None)`` after the initial scan, ``("changes", events)``
+    for each poll that observed file changes, and ``("done", None)`` when
+    stopped. Runs entirely off the event loop — the initial baseline scan is
+    the expensive part on large workspaces and must never block the loop
+    (see https://github.com/agentscope-ai/QwenPaw/issues/7721).
+    """
+    try:
+        prev = _scan_snapshot(watch_dir)
+        out.put(("ready", None))
+        while not stop.wait(interval):
+            try:
+                cur = _scan_snapshot(watch_dir)
+            except Exception:
+                continue
+            events = _diff_events(prev, cur)
+            prev = cur
+            if events:
+                out.put(("changes", events))
+    finally:
+        out.put(("done", None))
+
+
 @router.get(
     "/watch",
     summary="SSE stream for agent workspace file changes",
@@ -1098,64 +1174,65 @@ async def workspace_watch_events(
     request: Request,
     watch_dir: Path,
 ) -> AsyncIterator[str]:
-    """Yield workspace file changes without cancelling the watcher on idle."""
+    """Yield workspace file changes without cancelling the watcher on idle.
+
+    A background thread owns all filesystem scanning: the initial baseline
+    scan (the expensive part on large workspaces) and every poll run off the
+    event loop, so opening the file browser on a huge tree can never freeze
+    the server (see https://github.com/agentscope-ai/QwenPaw/issues/7721).
+    """
     yield 'data: {"type": "connected"}\n\n'
-    watcher = awatch(
-        watch_dir,
-        rust_timeout=_WATCH_POLL_TIMEOUT_MS,
-        yield_on_timeout=True,
+    loop = asyncio.get_running_loop()
+    stop = threading.Event()
+    out: queue.Queue[tuple[str, object]] = queue.Queue()
+    thread = threading.Thread(
+        target=_poll_watch_worker,
+        args=(watch_dir, out, stop, _WATCH_POLL_INTERVAL_SECONDS),
+        name="workspace-watch",
+        daemon=True,
     )
-    last_emit = asyncio.get_running_loop().time()
+    thread.start()
+    last_emit = loop.time()
     try:
         while True:
             if await request.is_disconnected():
                 break
             try:
-                raw_changes = await watcher.__anext__()
-            except (
-                StopAsyncIteration,
-                asyncio.CancelledError,
-                GeneratorExit,
-            ):
+                kind, payload = await asyncio.to_thread(
+                    out.get,
+                    True,
+                    _WATCH_QUEUE_TIMEOUT_SECONDS,
+                )
+            except queue.Empty:
+                kind, payload = None, None
+
+            now = loop.time()
+            if kind == "ready":
+                # Initial baseline scan finished; incremental events follow.
+                yield 'data: {"type": "ready"}\n\n'
+                last_emit = now
+            elif kind == "changes":
+                events = payload
+                if events:
+                    payload_json = json.dumps(
+                        {"type": "file_change", "events": events},
+                        ensure_ascii=False,
+                    )
+                    yield f"data: {payload_json}\n\n"
+                    last_emit = now
+            elif kind == "done":
                 break
 
-            events = []
-            for change_type, path in raw_changes:
-                try:
-                    rel = Path(path).relative_to(watch_dir)
-                except ValueError:
-                    continue
-                if _should_skip(rel.parts):
-                    continue
-                change_name = (
-                    "added"
-                    if change_type is Change.added
-                    else "deleted"
-                    if change_type is Change.deleted
-                    else "modified"
-                )
-                events.append(
-                    {"change": change_name, "path": rel.as_posix()},
-                )
-
-            now = asyncio.get_running_loop().time()
-            if events:
-                payload = json.dumps(
-                    {"type": "file_change", "events": events},
-                    ensure_ascii=False,
-                )
-                yield f"data: {payload}\n\n"
-                last_emit = now
-            elif now - last_emit >= _WATCH_HEARTBEAT_SECONDS:
+            # Heartbeat also fires during a long initial scan (the loop is
+            # never blocked waiting for it), keeping proxies from timing out.
+            if now - last_emit >= _WATCH_HEARTBEAT_SECONDS:
                 yield ": heartbeat\n\n"
                 last_emit = now
     except (asyncio.CancelledError, GeneratorExit):
         pass
     finally:
-        try:
-            await watcher.aclose()
-        except Exception:
-            pass
+        stop.set()
+        thread.join(timeout=_WATCH_POLL_INTERVAL_SECONDS + 1.0)
 
 
 @router.get(
--- a/tests/unit/app/routers/test_workspace_files_router.py
+++ b/tests/unit/app/routers/test_workspace_files_router.py
@@ -9,7 +9,6 @@ from types import SimpleNamespace
 import pytest
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
-from watchfiles import Change
 
 from qwenpaw.app.routers import workspace as workspace_router
 
@@ -338,55 +337,32 @@ async def test_watch_remains_open_after_idle_poll(
     watch_dir = tmp_path / "project"
     watch_dir.mkdir()
     target = watch_dir / "notes.md"
-    captured: dict[str, object] = {}
-
-    class FakeWatcher:
-        def __init__(self) -> None:
-            self.calls = 0
-            self.closed = False
-
-        def __aiter__(self):
-            return self
-
-        async def __anext__(self):
-            self.calls += 1
-            if self.calls == 1:
-                return set()
-            if self.calls == 2:
-                return {(Change.modified, str(target))}
-            raise StopAsyncIteration
-
-        async def aclose(self) -> None:
-            self.closed = True
-
-    watcher = FakeWatcher()
-
-    def fake_awatch(path: Path, **kwargs):
-        captured["path"] = path
-        captured.update(kwargs)
-        return watcher
+    target.write_text("v1")
+    monkeypatch.setattr(workspace_router, "_WATCH_POLL_INTERVAL_SECONDS", 0.05)
+    monkeypatch.setattr(workspace_router, "_WATCH_QUEUE_TIMEOUT_SECONDS", 0.05)
+    monkeypatch.setattr(workspace_router, "_WATCH_HEARTBEAT_SECONDS", 0.15)
 
     class ConnectedRequest:
         async def is_disconnected(self) -> bool:
             return False
 
-    monkeypatch.setattr(workspace_router, "awatch", fake_awatch)
-    messages = [
-        message
-        async for message in workspace_router.workspace_watch_events(
-            ConnectedRequest(),
-            watch_dir,
-        )
-    ]
+    agen = workspace_router.workspace_watch_events(
+        ConnectedRequest(),
+        watch_dir,
+    )
+    assert (await anext(agen)) == 'data: {"type": "connected"}\n\n'
+    assert (await anext(agen)) == 'data: {"type": "ready"}\n\n'
 
-    assert captured == {
-        "path": watch_dir,
-        "rust_timeout": 1_000,
-        "yield_on_timeout": True,
-    }
-    assert any('"path": "notes.md"' in message for message in messages)
-    assert watcher.calls == 3
-    assert watcher.closed is True
+    # Idle wake-up: the stream stays open and emits a heartbeat instead of
+    # closing the watcher.
+    assert (await anext(agen)) == ": heartbeat\n\n"
+
+    # A real change after the idle wake-up is still reported.
+    target.write_text("v2")
+    message = await anext(agen)
+    assert '"type": "file_change"' in message
+    assert '"modified"' in message and '"notes.md"' in message
+    await agen.aclose()
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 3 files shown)


rayrayraykk (commented):
I found and fixed one material lifecycle regression: cancelling the SSE stream during a scan synchronously joined the watcher thread from the async generator, blocking the event loop. In a regression test, a 400 ms scan delayed an unrelated 50 ms timer to approximately 405 ms.

The cleanup now offloads `thread.join()` with `asyncio.to_thread()`, and the new test verifies cancellation no longer stalls the loop.

Validation:

- Watcher-focused tests: 21 passed
- Changed-file pre-commit hooks: all passed, including mypy, black, flake8, and pylint
- Full router suite: 690 passed; one unrelated portability test failed because `APIRoute.original_router` is absent in this environment
- `git diff --check`: passed

I did not independently reproduce the reported 192,898-file benchmark. The remaining architectural trade-offs are the documented per-connection polling cost and possible queue growth when a client applies sustained backpressure.

xiaoka76:
Thanks for the thorough review and for catching the cancellation regression!

I've applied the fix in the follow-up commit (`d89f95f6`): the `finally` block now offloads `thread.join()` via `await asyncio.to_thread(thread.join, ...)`, so cancelling the SSE stream during a scan no longer blocks the event loop.

Also added a regression test (`test_watch_events_cancel_does_not_block_loop`) that cancels the consumer while the worker is mid-scan and asserts a short concurrent timer still completes on time. I verified the test **fails** against the synchronous-join version (the 0.05s tick gets delayed to ~2s) and **passes** with the fix.

Validation:

- Watcher tests: 8 passed (including the new cancellation regression)
- Full workspace router tests: 50 passed
- pre-commit on the changed files: all hooks green (check-ast, docstring, mypy, black, flake8, pylint, trailing-whitespace, add-trailing-comma)

On the remaining trade-offs you noted (per-connection polling cost and possible queue growth under sustained backpressure) — agreed, both are documented in the PR description, and they only affect trees where the previous `watchfiles` implementation was completely unusable. Happy to iterate if you see anything else.

PR #7729, opened 2026-09-12
fix(mcp): recognize Java jsonRpcError envelope on discover probe

### LINKED ISSUE #7728: [Bug]: `server/discover` HTTP 500 with non-standard `jsonRpcError` envelope is not treated as legacy-protocol evidence — Driver build fails against Java MCP SDK servers
## QwenPaw Version

2.2.0 (desktop, macOS arm64)

## Description

An MCP card of type `streamable_http` pointing at a Java/Kotlin MCP SDK server fails Driver construction with `Failed to build Driver '<name>': MCP server/discover rejected with HTTP 500`.

Root cause: the server answers the unknown method `server/discover` with HTTP **500** plus a non-standard error envelope `{"jsonRpcError": {"code": -32603, "message": "Missing handler for request type: server/discover", ...}}` (not the standard `{"jsonrpc": "2.0", "error": {...}}`). Two break points in the chain (`mcp_streamable_http.py` → `_negotiate` → `_rpc`):

1. `_unwrap_jsonrpc_result` (~L395): on `status >= 400` it checks `_is_jsonrpc_envelope(data)` first; the non-standard envelope fails the check → raises `httpx.HTTPStatusError` directly.
2. `_negotiate` (~L710): its `except httpx.HTTPStatusError` branch only classifies bare 400/404/405 as legacy-protocol evidence → 500 is not in the set → exception propagates → entire Driver build fails.

Control group: endpoints returning 400 or a standard `-32601` for `server/discover` fall back to legacy and connect fine. The affected server's standard `initialize` handshake works (HTTP 200, protocol 2025-11-25) — it would work via the legacy `HttpStatefulClient` path if the fallback triggered.

## Component(s) Affected

## Environment

- **QwenPaw version:** 2.2.0
- **OS:** macOS (arm64)
- **Install method:** desktop app

## Steps to Reproduce

1. Configure an MCP card with `streamable_http` transport pointing at a Java/Kotlin MCP SDK server.
2. Let QwenPaw build the Driver (startup or card hot-reload).
3. Driver construction fails; no tools are injected.

## Actual vs Expected

- **Actual:** `Failed to build Driver '<name>': MCP server/discover rejected with HTTP 500`; legacy fallback never triggers because HTTP 500 with a `jsonRpcError` envelope is not classified as legacy-protocol evidence.
- **Expected:** the non-standard 500 response is recognized as "server does not support server/discover" and negotiation falls back to the legacy `HttpStatefulClient` path, which then connects successfully via standard `initialize`.

## Logs / Screenshots

```
Failed to build Driver 'mx-ds-mcp': MCP server/discover rejected with HTTP 500
# server response body:
{"jsonRpcError": {"code": -32603, "message": "Missing handler for request type: server/discover", ...}}
```

## Additional Notes

Suggested fix (any one suffices):
1. In `_negotiate`'s `HTTPStatusError` branch, also classify 5xx responses whose error body contains "Missing handler" / "server/discover" evidence as legacy-protocol evidence; or
2. In `_unwrap_jsonrpc_result`, recognize the `jsonRpcError` (Java MCP SDK variant) envelope and convert it to a standard RPC error (code=-32603); or
3. Add an explicit card-level config option such as `negotiation: legacy` / `skip_discover: true` to force the `HttpStatefulClient` path.

Local workaround (deployed and verified): switch the card to stdio transport with a local bridge process (official MCP SDK `streamable_http_client` → stdio server forwarding `list_tools`/`call_tool`), bypassing `server/discover` entirely. Verified: `initialize` OK, full tool list, real queries return.

kabishou11:
## Summary

Fixes #7728.

Java/Kotlin MCP SDK servers often answer unknown `server/discover` with **HTTP 500** and a non-standard `{"jsonRpcError":{...}}` body (not `{"jsonrpc":"2.0","error":...}`). `_unwrap_jsonrpc_result` previously failed the envelope check and raised `HTTPStatusError`, while `_negotiate` only treated bare 400/404/405 as legacy-protocol evidence — so Driver build failed even though `initialize` would work via the legacy stateful client.

**Change:** extract both standard `error` (JSON-RPC 2.0 envelope) and Java `jsonRpcError` payloads that carry a parseable integer `code` on HTTP ≥400, so production Missing-handler (`-32603`) can trigger the existing legacy fallback. Empty/message-only `jsonRpcError` objects are ignored. Real server failures (bare 500, or `jsonRpcError` with other codes) still fail hard.

## What Problem This Solves

Connecting QwenPaw to a Java MCP SDK 2.0.0 Streamable-HTTP server fails during capability discover: the server returns HTTP 500 with

```json
{"jsonRpcError":{"code":-32603,"message":"Missing handler for request type: server/discover"}}
```

QwenPaw treated that as a hard HTTP failure and never fell back to the legacy stateful client, even though `initialize` / tools would work. Hosts that only speak the Java envelope (no JSON-RPC 2.0 `error` field) could not be onboarded.

## Evidence

- Reproduced against the issue #7728 payload shape: HTTP 500 + `jsonRpcError` `-32603` + `Missing handler for request type: server/discover`.
- Unit coverage in `tests/unit/drivers/handlers/test_mcp_dual_protocol.py` (79 tests): `-32603` Missing-handler → legacy fallback; bare HTTP 500 and non-matching `jsonRpcError` codes still hard-fail; empty `jsonRpcError` is not promoted to a default `-32000`.
- Review follow-up: tightened extractor to require a parseable integer code; fixtures aligned to the Java SDK body rather than a speculative `-32601`-only path.

## Test plan

- [x] `pytest tests/unit/drivers/handlers/test_mcp_dual_protocol.py -q`
- [x] HTTP 500 + `jsonRpcError` `-32603` Missing-handler → legacy fallback
- [x] Bare HTTP 500 / non-matching `jsonRpcError` codes → hard fail
- [x] flake8 E501 on extractor docstring fixed


src/qwenpaw/drivers/handlers/mcp_streamable_http.py
```diff
@@ -124,6 +124,18 @@ def _is_jsonrpc_envelope(data: Any) -> bool:
...
+def _extract_jsonrpc_error_payload(data: Any) -> dict[str, Any] | None:
+    """Return an error object from standard or Java-sdk ``jsonRpcError`` bodies."""
+    if not isinstance(data, dict):
+        return None
+    if _is_jsonrpc_envelope(data) and isinstance(data.get("error"), dict):
+        return data["error"]
+    alt = data.get("jsonRpcError")
+    if isinstance(alt, dict):
+        return alt
+    return None
```
yuanxs21, line 153: The alt branch accepts any dict. `{"jsonRpcError": {}}` or `{"jsonRpcError": {"message": "x"}}` becomes a _JsonRpcError via from_payload (default code -32000) and changes HTTPStatusError into an RPC/RuntimeError on every Streamable-HTTP method.

Requires a parseable integer `code` before treating the object as a JSON-RPC error. Also add a case for the reported Java envelope (`code: -32603`, message contains "Missing handler" / "server/discover"); #7728 and Java SDK 2.0.0 emit INTERNAL_ERROR, not -32601.

src/qwenpaw/drivers/handlers/mcp_streamable_http.py
```diff
@@ -406,9 +418,11 @@ def _unwrap_jsonrpc_result(
     """Unwrap a JSON-RPC response body, raising on HTTP / RPC errors."""
     # Require jsonrpc:"2.0" so platform 404 JSON is not misread as RPC.
     if status >= 400:
-        if _is_jsonrpc_envelope(data) and isinstance(data.get("error"), dict):
+        error_payload = _extract_jsonrpc_error_payload(data)
+        if error_payload is not None:
             # Legacy peers may omit/null id on HTTP 4xx JSON-RPC errors.
-            raise _JsonRpcError.from_payload(data["error"], http_status=status)
+            # Java/Kotlin MCP SDK uses a non-standard ``jsonRpcError`` envelope.
+            raise _JsonRpcError.from_payload(error_payload, http_status=status)
```
yuanxs21, line 457: Two issues.

1) The first comment is now false. This path no longer requires jsonrpc:"2.0"; any HTTP >=400 body with a jsonRpcError object is promoted to _JsonRpcError. Rewrite the comment: standard `error` still needs a JSON-RPC envelope; `jsonRpcError` is the Java exception.

2) Unwrap-only does not fix #7728. After this raise, _negotiate only maps -32601 (and HTTP 400/404/405) to _LegacyProtocolError. The reported payload is HTTP 500 + jsonRpcError -32603, so connect still dies in _discover_rpc_error (only the exception type/message changes). Either teach _negotiate that this discover-probe shape is legacy evidence, or this change does not close the bug. Do not treat every 5xx / every -32603 as fallback — keep bare 500 and real INTERNAL_ERROR as hard failures.

tests/unit/drivers/handlers/test_mcp_dual_protocol.py
```diff
@@ -201,6 +217,39 @@ def test_supported_versions_from_payload(payload, expected):
...
+    with pytest.raises(_JsonRpcError) as caught:
+        _unwrap_jsonrpc_result(
+            method="server/discover",
+            status=500,
+            data={"jsonRpcError": {"code": -32601, "message": "missing"}},
+            request=req,
+            request_id=1,
+        )
+    assert caught.value.code == -32601
+    assert caught.value.http_status == 500
```
yuanxs21, line 287: These fixtures do not match #7728 or the Java SDK 2.0.0 body. Cover `{"jsonRpcError": {"code": -32603, "message": "Missing handler for request type: server/discover"}}` (and optionally the fuller exception document with stackTrace). Also assert that a code-less / empty `jsonRpcError` dict is rejected if you tighten the extractor.

Unwrapping -32601 is fine as an extra case; it is not a substitute for the reported envelope.

tests/unit/drivers/handlers/test_mcp_dual_protocol.py
```diff
@@ -291,6 +340,10 @@ def test_normalize_call_tool_result_snake_case_aliases():
         # the legacy handshake arbitrates.
         lambda r: httpx.Response(401, text="u"),
         _gateway_401,
+        lambda _r: _java_jsonrpc_err(
+            -32601,
+            "Method not found: server/discover",
+        ),
```
yuanxs21, line 406: Wrong error code for the bug this PR claims to fix. Add HTTP 500 + jsonRpcError -32603 + "Missing handler for request type: server/discover" (issue #7728) as a fallback case. -32601 here only proves the author’s hypothesis, not the production server. Keep this case only if you still want coverage for a hypothetical -32601 envelope.

yuanxs21 (changes requested):

kabishou11:
Thanks @yuanxs21 — you're right that the reported Java envelope is HTTP 500 + `jsonRpcError` **-32603** (`Missing handler for request type: server/discover`), not `-32601`, and that a code-less `jsonRpcError` object must not be promoted.

I'll tighten the extractor (require parseable integer `code`), teach `_negotiate` to treat only that discover-probe shape as legacy evidence (not every 5xx / every -32603), fix the comments, and update tests to match #7728.

kabishou11:
@yuanxs21 addressed in the latest push:

1. `jsonRpcError` extraction now requires a parseable integer `code` (empty / message-only objects are not promoted).
2. Unwrap comment clarified: standard `error` still needs a JSON-RPC envelope; `jsonRpcError` is the Java path.
3. `_negotiate` (discover probe only) treats HTTP 500 + `jsonRpcError` **-32603** with message containing `Missing handler` and `server/discover` as legacy evidence. Bare 500 and other `-32603` payloads still fail hard.
4. Tests updated for the reported #7728 envelope (plus rejection of code-less `jsonRpcError`).

`pytest tests/unit/drivers/handlers/test_mcp_dual_protocol.py -q` → 79 passed.

yuanxs21 (changes requested):
LGTM. Please fix the pre-commit errors before merging.

kabishou11:
Thanks for the re-review — addressed the pre-commit flake8 E501 on the extractor docstring, and updated the PR body with the required **What Problem This Solves** / **Evidence** sections for the Real behavior proof check.

yuanxs21 (approved):

yuanxs21:
@kabishou11 Please resolve the conflicts before merging.

kabishou11:
@yuanxs21 Conflicts resolved — rebased onto current `main` and force-pushed.

Kept the recent header-sanitization path in `_unwrap_jsonrpc_result`, and re-applied the Java `jsonRpcError` / discover Missing-handler `-32603` fallback on top. `pytest tests/unit/drivers/handlers/test_mcp_dual_protocol.py -q` → 81 passed locally.

kabishou11:
Correction: conflicts are resolved **locally** (rebased onto current `main`, tests green), but the force-push is blocked by GitHub OAuth missing the `workflow` scope (upstream moved several `.github/workflows/*` files). Pushing as soon as the fork can accept those commits.

kabishou11:
@yuanxs21 Conflicts resolved — rebased onto current main and force-pushed. Local pytest tests/unit/drivers/handlers/test_mcp_dual_protocol.py → 81 passed.

PR #7732, opened 2026-09-12
fix(acp): select permission options by protocol kind

### LINKED ISSUE #7726: [Bug]: ACP `trusted: true` silently falls back to interactive prompts — `_pick_allow_option` only matches `allow_*` optionIds
**Environment**: QwenPaw 2.2.0 (desktop, macOS arm64); still present on `main` as of 2026-09-05.

**Root cause** (`src/qwenpaw/acp/client.py:32-38, 113-130`):

```python
_ALLOW_OPTION_PREFERENCE = ("allow_once", "allow_always", "allow", "yes", "approve")
# exact lowercase match on optionId; fallback: substring "allow" in id
```

The `trusted` branch of `request_permission` exists and is logically correct, but auto-approval depends on `_pick_allow_option`. ACP agents are free to choose their own optionIds — e.g. **kimi-code uses `approve_once` / `approve_always` / `reject`** (verified via strings in kimi-code 0.27.0 and 0.41.0 binaries). Exact matching fails (`"approve" != "approve_once"`), and the substring fallback `"allow" in "approve_always"` is also False → `_pick_allow_option` returns `None` → trusted sessions silently degrade to interactive prompting. Users see a permission dialog on every single tool call despite `"trusted": true`.

Note: kimi-code's options DO carry spec-compliant `kind` values (`allow_once` / `allow_always`) — matching by `kind` instead of `id` is the robust approach.

**Additional observation**: the response path has a symmetric problem. When the selected optionId is not in QwenPaw's known vocabulary, the outcome parsing falls through to `default: rejected` — so a proxy translating IDs must translate both directions. Matching by `kind` would fix both directions at once.

**Suggested fix**:
1. Prefer matching by ACP option `kind` (`allow_always` > `allow_once`), falling back to id heuristics; and/or
2. Extend `_ALLOW_OPTION_PREFERENCE` with `"approve_always"`, `"approve_once"` and widen the substring fallback to `"approve"`.

**Minimal repro**:
1. Configure an ACP runner with `"trusted": true` whose permission optionIds use the `approve_*` vocabulary (e.g. kimi-code).
2. Trigger any tool call requiring permission.
3. Expected: auto-approved without user interaction. Actual: interactive permission request is surfaced on every tool call.

**Workaround (verified in production)**: a stdio wrapper adapter around the runner's ACP entrypoint that bidirectionally translates `approve_*` ↔ `allow_*` optionIds. With the adapter in place and `trusted: true`, non-dangerous requests are auto-approved with zero interactive prompts.
在聊天中引用

axelray-dev:
## Description

Trusted ACP sessions matched allow choices by `optionId`. ACP leaves option IDs to each agent, so IDs such as `approve_once` were missed and safe tool calls fell back to interactive prompts. Match ACP's stable `kind` first, while retaining ID matching for clients that omit it.

**Related Issue:** Fixes #7726

**Security Considerations:** The existing hard-block check still runs before trusted auto-approval. This change only selects ACP allow kinds and returns the selected option's existing ID.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Core / Backend
- [x] Tests

## Testing

- `pytest tests/unit/agents/test_acp_client_trusted.py tests/unit/agents/test_acp_permissions_trusted.py -q` (14 passed)
- `flake8 --extend-ignore=E203 src/qwenpaw/agents/acp/client.py tests/unit/agents/test_acp_client_trusted.py` (no findings)
- GitHub Pre-commit check passed.
- No live ACP service was used; the reported path is covered by the focused unit tests.

## Evidence

The regression test supplies a custom `approve_once` option ID with `kind=allow_once` and verifies that trusted mode returns that exact ID without creating a pending permission request.

## Checklist

- [x] Focused tests pass.
- [x] No documentation changes are needed.
- [x] Ready for review.


x1n95c:
Hello, I've tested your code and it works as expected. Thanks for your contribution! The comments from the AI reviewer are minor and don't require any corrections. I have no further suggestions.

x1n95c:
Hi @axelray-dev, it looks like the unit test has failed. You may need to rebase your branch onto the latest version and resolve the issue.

lalaliat (approved):
LGTM

PR #7738, opened 2026-09-13
fix(providers): filter unrecognized kwargs before OpenAI completions.create()

### LINKED ISSUE #7026: [Bug]: deepseek-v4-pro 模型调用时自动注入 chat_template_kwargs 参数未被 extra_body 包装，导致 openai SDK TypeError
[Bug]: deepseek-v4-pro 模型调用时自动注入 chat_template_kwargs 参数未被 extra_body 包装，导致 openai SDK TypeError

英文版（如果官方要求英文）：

    [Bug]: chat_template_kwargs auto-injected without extra_body wrapping causes openai SDK TypeError on all model calls (qwenpaw 2.0.1 + agentscope 2.0.4.post1)

* * *

## QwenPaw Version
    
    2.0.1 (Windows 安装版，路径 `D:\QwenPaw_v2_test`)
    - agentscope: 2.0.4.post1
    - pydantic: 2.13.4
    - openai SDK: 2.33.0
    - backend: qwenpaw-backend.exe (PyInstaller 冻结包)
    
    ## Description
    
    主模型配置为 `custom/deepseek-v4-pro`（非 reasoning 模型），调用时报错：
    

MODEL_EXECUTION_ERRORModel 'unknown' execution failed.Reason: AsyncCompletions.create() got an unexpected keyword argument 'chat_template_kwargs'

    
    完整 traceback：
    

File "qwenpaw\agents\react_agent.py", line 680, in _reply File "agentscope\agent\_agent.py", line 2440, in _call_model File "agentscope\agent\_openai_chat\_model.py", line 272, in _call_api File "openai\_utils\__utils.py", line 287, in wrapperTypeError: AsyncCompletions.create() got an unexpected keyword argument 'chat_template_kwargs'

    
    错误 dump 文件路径：`C:\Users\52765\AppData\Local\Temp\qwenpaw_query_error_yio2rd75.json`
    
    ## 诊断过程
    
    ### 1. 验证 openai SDK 版本
    
    通过直接 import 确认 `openai.__version__ == 2.33.0`。检查 `AsyncCompletions.create` 的所有参数（共 40 个），**不包含** `chat_template_kwargs`。
    
    ### 2. monkey-patch 拦截确认注入值
    
    通过 monkey-patch 拦截 `AsyncCompletions.create` 调用，捕获到自动注入的参数：
    

[PATCH] Intercepted chat_template_kwargs: {'enable_thinking': False}

    
    拦截后请求成功构造（仅 401 鉴权失败，说明参数已被 SDK 接受，请求体构造正常）。
    
    ### 3. 检查用户配置文件
    
    `D:\QwenPaw_v2_test\.qwenpaw\openclaw.json` 中**未配置**任何相关字段：
    

matches: 0 (chat_template_kwargs)matches: 0 (enable_thinking)matches: 0 (extra_body)

    
    ### 4. 结论
    
    `chat_template_kwargs` 参数是 QwenPaw / agentscope 内部自动注入的，但**没有**包在 `extra_body` 内传给 openai SDK，导致 openai SDK 不识别该参数报错。
    
    ## 期望行为
    
    按照官方 #2042 的修复方案，`chat_template_kwargs` 应自动包装在 `extra_body` 内传给 openai SDK：
    
    ```json
    {
      "extra_body": {
        "chat_template_kwargs": {"enable_thinking": false}
      }
    }

## 临时解决方案（用户侧）

用户无法直接修复（backend 是 PyInstaller 冻结包，源码不可编辑），只能等待官方下版本修复。

## 建议修复位置

官方在以下位置之一修复即可：

* `qwenpaw/providers/openai_chat_model_compat.py` 的 `_call_api` 方法
* 或 `agentscope/agent/_openai_chat_model.py:272`

## 环境信息

* **OS**：Windows 11 (10.0.22631)
* **QwenPaw**：2.0.1 (Windows 安装版)
* **Python**：3.11
* **Model**：`custom/deepseek-v4-pro` @ `https://api.deepseek.com`
* **Provider 配置**：`reasoning: false`
* **安装方法**：安装包 + 配置内网/直连模型

## 复现步骤

1. 安装 qwenPaw 2.0.1
2. 配置 OpenAI-compatible provider（如 deepseek、GLM 等）
3. 模型配置中 `reasoning: false` 或完全省略 reasoning 字段
4. 发送任意消息（即使是 `hi`）
5. 100% 复现报错

## 关联 Issue

* #2042 (类似症状，已 close，但根因可能未根治——只解决了用户侧配置层，未根治内部自动注入逻辑)
* #6619 (ToolCallBlock extra_content 报错，同一升级版本，同一栈深度)

## 用户场景

希望从 QwenPaw 1.1.12 升级到 2.0.1，但此 bug 导致**所有模型不可用**。已回退到 1.1.12 继续使用。

lumenfield:
## Description

Middleware or proxy layers may inject custom kwargs (e.g. `streamIdleTimeoutMs`) that the upstream OpenAI SDK rejects with:

```
TypeError: AsyncCompletions.create() got an unexpected keyword argument 'streamIdleTimeoutMs'
```

### Root cause

`OpenAIChatModelCompat._call_api()` forwards **all** kwargs (merged from `extra_generate_kwargs` + caller-provided `generate_kwargs`) directly into the agentscope base class, which in turn passes them to `openai.AsyncClient.chat.completions.create()`. The OpenAI SDK does not tolerate unknown kwargs and raises `TypeError`.

### Fix

Whitelist all known parameters for `openai.AsyncOpenAI().chat.completions.create()` (verified against OpenAI SDK v2.33) and drop unrecognized keys with a warning log before forwarding to the base class. This is analogous to the approach in unmerged PR #5607 but simpler and non-destructive (keeps `extra_body` passthrough intact).

### Related

- Related open issue: #7026 (auto-injected `chat_template_kwargs` not wrapped in `extra_body`)
- Related unmerged PR: #5607 (restore non-standard kwargs routing into `extra_body`)

## Test plan

1. Unit test: call `_sanitize_openai_kwargs()` with `streamIdleTimeoutMs` and other unknown keys; verify they are dropped.
2. Integration: run a model call through an agent whose provider config includes `streamIdleTimeoutMs` in `extra_generate_kwargs`; verify the call succeeds and a warning is logged.
3. Existing tests: run `pytest tests/unit/providers/test_openai_provider.py tests/unit/providers/test_openai_stream_toolcall_compat.py`

Fixes #7026 (related)


wangfei010313:
Thanks for the fix — the approach and the placement are right. I traced the call chain: `merged` in `_call_api` is indeed the only dict that reaches `chat.completions.create(**kwargs)`, so filtering there is the correct spot. I also diffed your `_OPENAI_COMPLETION_PARAMS` against `openai` 2.33.0's `AsyncCompletions.create` signature — 39 parameters, exact match, and every legitimate param (`extra_body`, `stream_options`, `tool_choice`, `extra_headers`, …) survives. Nice precision there.

Three things to address before merge.

### 1. This also drops QwenPaw's own Langfuse tracing kwargs (functional regression)

`OpenAIChatModelCompat.__call__` injects `trace_id`, `name`, `metadata` and `parent_observation_id` whenever Langfuse tracing is enabled (see `qwenpaw/observability/langfuse.py:93`). They travel through `ChatModelBase.__call__` into `generate_kwargs`, land in `merged`, and the new call at line 859 strips three of them — only `metadata` survives, and only because it happens to be a real OpenAI parameter.

Those keys are not stray noise: the Langfuse OpenAI wrapper consumes them at exactly the `create()` boundary and feeds them to `start_observation(name=..., trace_context={"trace_id": ..., "parent_span_id": ...})`. Dropping them detaches every LLM generation from the agent-turn trace.

A/B with identical kwargs, run through `_call_api` against a stubbed client:

```
baseline (main): reached SDK: ['messages','metadata','model','name','parent_observation_id','stream','streamIdleTimeoutMs','top_k','trace_id']
this PR        : reached SDK: ['messages','metadata','model','stream']
                 WARNING: Dropping unrecognized kwargs ...: ['name','parent_observation_id','streamIdleTimeoutMs','top_k','trace_id']
```

Suggested fix: add those three names to the allowlist, or scope the filter to `self._extra_generate_kwargs` (the provider-config path that produces the `streamIdleTimeoutMs` case in your description) and leave per-call kwargs untouched. A regression test asserting they survive would be worth having.

### 2. The `Real behavior proof` check is failing on the PR body

`scripts/github/real_behavior_proof_check.py` requires an evidence section, and its alias list only accepts headings like `Evidence`, `Evidence after fix`, `Local verification evidence`, `Testing`, or `How to test these changes`. Your body has `## Description` (good — that satisfies the problem section) and `## Test plan`, which is not an alias, so the check reports:

```
  status:             missing
  → MISSING sections: ['Evidence']
```

Renaming `## Test plan` to `## Evidence` (or adding an `## Evidence` section) fixes it. Since the check wants real observed behavior rather than a plan, pasting output works well, for example:

~~~text
## Evidence

Before (main), with streamIdleTimeoutMs in the provider's extra_generate_kwargs:
  TypeError: AsyncCompletions.create() got an unexpected keyword argument 'streamIdleTimeoutMs'

After (this PR): the request is issued and the log shows
  WARNING ... Dropping unrecognized kwargs for OpenAI completions API: ['streamIdleTimeoutMs']
~~~

(Using "Evidence" over "Test plan" also matches what reviewers need here. Note the checker rejects placeholder values — a `## Evidence` section containing only `N/A` still fails, so the section needs real content.)

### 3. The pre-commit gate will fail once it runs

It has not run yet — workflows on fork PRs wait for maintainer approval — but it will fail on the new code. Reproduced with the repository's pinned hook versions (`black` 23.3.0, `flake8` 6.1.0):

```
flake8 6.1.0: openai_chat_model_compat.py:89:80: E501 line too long (82 > 79 characters)
black 23.3.0: would reformat — the return at line 89 must be split:

  -        return {k: v for k, v in kwargs.items() if k in _OPENAI_COMPLETION_PARAMS}
  +        return {
  +            k: v for k, v in kwargs.items() if k in _OPENAI_COMPLETION_PARAMS
  +        }
```

### Smaller notes (non-blocking)

- **The silent drop changes the failure mode.** Params like `top_k`, `min_p`, `repetition_penalty` (valid for vLLM/Ollama/OpenRouter backends, currently reachable through provider `generate_kwargs`) used to raise a loud `TypeError`; now they vanish with a single log line. The right home for them is `extra_body`, and that passthrough is intact — but it'd help to say so in the warning message, and/or reject unknown keys when the provider config is validated, keeping this runtime filter as a backstop.
- **The same hole exists elsewhere:** `openai_response_provider._call_api` (`src/qwenpaw/providers/openai_response_provider.py:109`) forwards `merged` to the SDK unfiltered, so Responses API users still hit the original `TypeError`. Note the parameter set differs there, so this allowlist shouldn't be reused as-is.
- **Log volume:** one WARNING per model call. A misconfigured provider will log on every request; consider warning once, or dropping it to debug.
- **Allowlist drift:** `openai` is pinned to `>=2.0.0,<=2.33.0`. When the SDK adds a parameter, this filter will silently eat it. A small test asserting the allowlist covers every parameter of the installed SDK's `create` signature would lock that down.
- **Tests:** the repo expects a unit test alongside the change (see `AGENTS.md`). A sanitizer test is also the natural place where issue 1 above would have surfaced.

Happy to re-review as soon as 1-3 are in — the diff is small and the direction is clearly right.

PR #7748, opened 2026-09-14
fix: preserve loop warnings and correct budget and overflow recovery

rayrayraykk:
## Description

Fix three agent-loop recovery failures:

- Deliver doom-loop strategy warnings before the next model call, after tool results have arrived. Ordinary goal continuations remain suppressed during tool-call iterations; termination and reset discard pending warnings.
- Read the recording wrapper's cumulative user-turn token usage directly instead of adding the cumulative value every iteration. Previously, three calls consuming 10 tokens each could count as 60 tokens and exhaust a 50-token budget early. Rubric peer resets refresh the cached snapshot without excluding earlier usage in the same user turn.
- Recover from context overflow raised while consuming a response stream, before meaningful output has been emitted. Close the failed stream, compact and rebuild the request, and retry once. Do not replay partial output or recursively retry the recovery attempt.

Recovery orchestration lives in `agents/context/overflow_recovery.py`, which owns invocation/stream error boundaries, stream cleanup, and the single-retry limit. The agent supplies a recovery callback that classifies the error, compacts context, and rebuilds model input.

**Related Issue:** None.

**Security Considerations:** No changes to permissions, credentials, or approval policy.

## Type of Change

- [x] Bug fix

## Component(s) Affected

- [x] Tests (strengthen one existing budget test; no new test cases)

## Checklist

- [x] Ran pre-commit on all changed files; all applicable hooks pass
- [x] Committed auto-formatting changes and reran checks
- [x] Ran relevant existing tests locally; all pass
- [x] Ready for review

## Testing

Validated in the QwenPaw conda environment, explicitly setting PYTHONPATH to this checkout's src directory. No new unit-test cases are included. The existing token-budget test uses cumulative snapshots and a threshold that distinguishes correct accounting from repeated addition. It also verifies budget enforcement after a rubric-style peer reset and fresh accounting after the recording layer clears usage for a new user turn. The peer-reset assertion failed before this correction and passes afterward. The new StopHandlerResult field is appended after existing fields to preserve positional argument order.

## Evidence

Existing relevant test run:

```text
PYTHONPATH="$PWD/src" conda run --no-capture-output -n QwenPaw python -m pytest -q tests/unit/loop tests/unit/agents/test_context_overflow_recovery.py tests/unit/agents/test_react_agent_scroll_seen.py tests/unit/agents/test_reasoning_fallback_bridge.py tests/unit/agents/test_command_handler.py tests/unit/providers/test_fallback_chat_model.py tests/unit/providers/test_retry_chat_model.py --disable-warnings
243 passed in 4.13s
```

Changed-file pre-commit checks passed: AST, docstring placement, encoding, private-key detection, whitespace, trailing commas, mypy, black, flake8, and pylint.

## Additional Notes

Based on upstream/main at 797ede9b. Validation used local tests and simulated provider failures; no live-provider end-to-end benchmark was run.


jinliyl (commented):
Reviewing only item 3 (context-overflow recovery while consuming a response stream).

The orchestration looks correct in manual stream simulations: a pre-content overflow closes the failed stream, rebuilds/retries once, meaningful partial content suppresses recovery, and the retry response is consumed outside the recovery wrapper so it cannot recursively retry.

The remaining gap is regression coverage. tests/unit/agents/test_context_overflow_recovery.py only exercises exceptions raised before _call_model returns, so none of the new behavior in overflow_recovery.py is protected. Please add focused async-stream tests that assert:

- a non-meaningful chunk followed by a classified overflow closes the original stream and yields the rebuilt retry exactly once;
- meaningful text/thinking/tool-call output followed by overflow propagates the error without recovery or replay;
- an overflow from the recovery stream propagates and does not trigger a second compaction/retry;
- non-overflow stream errors propagate after cleanup.

I ran the existing overflow/retry/fallback suites (115 tests) and additional local stream lifecycle simulations; all passed. No implementation-level correctness defect found beyond the missing coverage.

jinliyl (approved):
LGTM

zhijianma (approved):

PR #7750, opened 2026-09-14
feat(console): show send_file_to_user files in response artifact list

### LINKED ISSUE #7744: [Feature]: 网页控台发送文件后自动折叠误导用户
## QwenPaw Version

QwenPaw 2.2.1。

## Description

Agent 使用 `send_file_to_user` 成功发送文件后，文件结果目前只显示在思考/工具步骤中。会话结束后工具步骤默认折叠，用户需要手动展开才能找到文件。

希望将成功发送的文件直接展示在助手回复正文区域，提供清晰的文件卡片，让用户可以直接预览和下载，不需要展开工具调用详情。

**Related PR(s):** 暂无

**Security considerations:** 文件预览和下载应继续遵守现有认证、Runtime 隔离及文件访问权限。

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** 2.2.1
- **OS:** Ubuntu Linux 6.8，x86_64
- **Install method:** pip 安装 Hub，个人 Runtime 使用 Docker
- **Python version (if applicable):** Python 3.12

## Steps to Reproduce

1. 打开 Web Console，并保持默认的助手消息折叠设置。
2. 让 Agent 调用 `send_file_to_user` 发送一个文件。
3. 等待本轮会话结束。
4. 文件入口只存在于默认折叠的“已完成 N 个步骤”中，正文区域没有可见附件。

## Actual vs Expected

- **Actual:** 用户必须展开思考/工具步骤，才能找到 Agent 发送的文件并进行预览或下载。
- **Expected:** `send_file_to_user` 成功后，文件卡片直接显示在助手回复正文区域，用户可以立即预览和下载；工具调用参数和执行详情仍可保留在折叠步骤中。

## Logs / Screenshots

该问题主要是前端展示问题，没有相关后端错误日志。

```text
当前：助手正文 → 已完成 N 个步骤（展开后才能看到文件）
期望：助手正文 → 可见的文件卡片（预览 / 下载）
```

## Additional Notes

建议成功附件卡片至少包含：

- 文件名和文件类型。
- 预览按钮。
- 下载按钮。
- 发送失败或取消时不显示成功附件。

附件卡片应独立于思考/工具步骤的展开状态，并覆盖文本、图片、音频、视频及普通文件。

zhijianma:
## Description

Files sent via `send_file_to_user` were only visible inside the collapsed
"Completed N steps" group, so users had to manually expand the tool steps to
find a file the agent had just delivered.

This PR surfaces successfully sent files in the response artifact grid at the
end of the assistant reply — the same mechanism already used for
`write_file` / `edit_file` / `append_file`. The artifact list renders
independently of the collapsed step group, so the file is visible and clickable
without expanding anything.

Tool call arguments and execution details stay inside the collapsed steps, and
`SendFileCard` no longer auto-expands (it previously did, which is now
redundant and inconsistent with the other tool cards).

**Related Issue:** Fixes #7744

**Security Considerations:** Preview reuses the existing Files API and the
`qwenpaw:open-file-preview` event channel. No new file-access path is
introduced; authentication and runtime isolation are unchanged.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

Note on the first item: the repo's pre-commit `prettier` hook excludes
`^console/`, so `pre-commit run --all-files` does not lint these files. Console
changes are verified with the console toolchain instead (prettier / eslint /
tsc / vitest) — see Evidence below.

N/A — this PR changes only Console frontend React components. No channel
backend implementation is touched, so the channel contract tests do not apply.

## Testing

Three new cases in `ResponseArtifactList.test.tsx`:

1. a successfully sent file surfaces as an artifact
2. sent files are labelled with the new `sent` status
3. a result carrying no DataBlock (send failed) surfaces nothing

**Why success is detected by DataBlock block type rather than `source.url`:**
the DataBlock source shape varies with the delivery path — text files arrive as
`{type: "url", url: "file://…"}` while images are inlined as
`{type: "base64", data: …}`. Keying on `source.url` silently drops images.
Detection therefore keys on the block-level `type === "data"`, which is stable
across both source shapes.

The success fixture deliberately uses the **base64** shape as the stricter
guard: reintroducing any `source.url` coupling makes these tests fail
immediately. This was verified by temporarily reverting the predicate to the
`source.url` form — 2 tests failed.

## Evidence

```bash
$ npx vitest run src/features/files-workspace/ResponseArtifactList.test.tsx \
                 src/components/Chat/ToolCards/cards/SendFileCard.test.tsx
 Test Files  2 passed (2)
      Tests  12 passed (12)

$ npx prettier --check <7 changed files>
All matched files use Prettier code style!

$ npx eslint <4 changed ts/tsx files>
(clean, no output)

$ npx tsc --noEmit
0 errors
```

Verification is **targeted** at the changed files; the full test suite was not
run.

## Additional Notes

Known pre-existing issue, deliberately out of scope for this PR:
`src/qwenpaw/agents/tools/send_file.py` returns `ToolResultState.SUCCESS` on all
four of its error paths, whereas `file_io.py` returns `ToolResultState.ERROR` on
its 17 error paths. Because of that the frontend cannot rely on the standard
state flags to detect a failed send, and instead checks for DataBlock presence.

The same root cause also makes `SendFileCard` render a "successful" preview for
a failed send — `getMediaInfo` falls back to the `file_path` argument and even
extracts the path out of the error message text via regex. Correcting the
backend state would let both surfaces drop this special-casing entirely; it is
left out here to keep this PR frontend-only.


rayrayraykk:
Reviewed commit `365119b467ae3644cd130f6b2fe811031a569d66`.

**[P2] Resolve delivered attachments as filesystem paths instead of filtering them through the Markdown-link parser.** The new send-file branch in `console/src/features/files-workspace/ResponseArtifactList.tsx:161–174` confirms delivery, but then derives its target solely from the original `file_path` argument. `targetForPath` passes relative paths through `parseInternalFileLink`, which interprets `#` as a fragment and rejects `..` segments. Consequently, successful sends of `reports/Q3#final.pdf` and `../shared/report.pdf` render no artifact at all. A name ending in `#L12` can instead be interpreted as a line reference to a different file. These are filesystem paths the send tool can legitimately resolve, subject to its existing access policy; they are not Markdown hrefs. This affects ordinary filenames on both Windows and macOS/Linux.

I reproduced both missing-artifact cases by extending the PR's successful-send fixture: the component renders an empty container despite the success DataBlock. The original two targeted suites pass (**15 tests**); the two added regression cases fail.

Please reuse the delivered attachment URL/resolved path when available, and use a filesystem-path-aware fallback for base64 results rather than the link parser. Keep the preview attached to the actual delivered resource and preserve the existing access checks. Add coverage for literal `#`, relative parent segments, and Windows paths. This also offers a useful shared boundary with SendFileCard, which already obtains the delivered media URL.

No new server-side synchronous I/O is introduced. Checking DataBlock types avoids parsing base64 payloads, and the memoization is appropriately small. Existing synchronous I/O in the Python send/preview implementation is outside this diff. Tests ran on macOS; Windows behavior was reviewed statically.

zhijianma:
Thanks for the detailed repro — confirmed, and partially addressed in 87ad06e5.

**Fixed** (`87ad06e5`): `targetForPath` no longer routes workspace-relative paths through `parseInternalFileLink`, which is a Markdown-href parser. `Report #3.pdf` and `reports/Q3#final.pdf` now surface, and `report#L12` targets the right file instead of `report` + line 12. Added two tests. The `.`/`..`/empty-segment traversal guard is preserved and now covered by a test as well.

**Not changed, with reasons:**

- **`..` relative segments** — the client has no `projectDirectory`, so it cannot resolve them correctly (`filePreviewUrl` would turn `../shared/x.pdf` into `/../shared/x.pdf`). The traversal guard is also a deliberate containment property. Tracking this as a follow-up alongside the absolute-vs-relative dedup gap.
- **Reusing the delivered `source.url`** — that would make sends resolve to absolute `attachment` targets while `write_file` stays a `workspace` target, so the two no longer share an artifact id and dedup breaks for the common "write a report, then send it" sequence (two cards instead of one). Trading a common case for the rarer ones looked worse; happy to revisit if you read it differently.

Verified: 17 targeted tests pass; prettier / eslint / tsc clean.

rayrayraykk (approved):

PR #7595, opened 2026-09-07
fix(console): unify language selector options

zhaozhuang521:
Fixed the inconsistency between the top language dropdown menu and the language options in the sidebar settings within the Console.
The sidebar settings previously offered only five languages, omitting Bahasa Indonesia (id) and Tiếng Việt (vi). The language configuration has now been extracted into a shared module for reuse by both entry points, ensuring consistent language options. Additionally, relevant components were migrated from the deprecated `languageApi` to `settingsApi`, and test mocks were updated.

Fixes #7006

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Core / Backend (app, agents, config, providers, utils, local_models)
- [ ] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [ ] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [ ] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

[How to test these changes]

## Evidence

Examples of valid evidence:
- Terminal transcript of the test run (e.g. `pytest tests/unit/app/chats/ -q` output)
- Screenshot of the Console UI showing the fix
- CI artifact link
- `pre-commit run --all-files` summary

```bash
pre-commit run --all-files
# paste summary result

pytest
# paste summary result
```

## Additional Notes

[Optional: any other context]

PR #7752, opened 2026-09-14
fix(console): make vi and pt-BR language selection work

zhijianma:
## Description

The console offers seven languages, but two of them were broken end to
end and a third failure mode hid both problems from users.

- **`vi` was rejected by the backend.** `LANGUAGE_LIST` offers seven
  languages, but `_VALID_LANGUAGES` only listed six. `PUT
  /api/settings/language` with `vi` returned HTTP 400, so the preference
  was never persisted even though the UI showed Vietnamese.
- **`pt-BR` never loaded its bundle.** `nonExplicitSupportedLngs: true`
  makes i18next reduce a region-qualified code to its language part
  before matching it against `supportedLngs`, so `pt-BR` was looked up
  as `pt` and rejected. The UI silently fell back to English while the
  selector still reported the language the user had picked.
- **Both entries swallowed the failure.** The sidebar settings panel
  caught the rejected PUT with `.catch(() => {})`, and the header
  dropdown only logged to the console, so users got no feedback at all
  and their preference was silently lost.
- **`id` and `vi` advertised the wrong language code.** The icon set has
  no Indonesian or Vietnamese badge, so both reused the English one —
  rendering an `en` badge next to a non-English label. They now render a
  locally drawn badge that reproduces the frame geometry measured from
  the real badges, and carries `data-spark-icon` so the design system
  sizes it like a real one (20px default, 16px sm, 24px lg).

A guard test now asserts `_VALID_LANGUAGES` equals the console's
`LANGUAGE_LIST` keys, so a language added in one place and forgotten in
the other fails CI instead of silently dropping the user's preference.

**Related Issue:** Relates to #7595

**Security Considerations:** None. No auth, credential, or config
handling is touched.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

> Note on the first box: pre-commit was run scoped to every file this PR
> touches (`--file`), not `--all-files`. All hooks passed with no
> auto-fixes needed. The scoped run covers this PR's complete change
> set; `--all-files` was deliberately not run to avoid mutating
> unrelated files.

Not applicable — no channel code is touched by this PR.

## Testing

1. Open the header language dropdown and switch to **Tiếng Việt**. The
   UI turns Vietnamese and stays Vietnamese after a reload; `GET
   /api/settings/language` now returns `vi` (it previously kept the
   last successfully saved language).
2. Switch to **Português (Brasil)**. The UI turns Portuguese
   (`Nova tarefa` instead of `New task`) and the selector no longer
   snaps back to English.
3. Stop the backend, then switch the language from either the sidebar
   settings panel or the header dropdown. An error toast now appears
   instead of the failure being swallowed.
4. Compare the `id` / `vi` badges against the other five: they no
   longer show an `en` badge, and they match the real badges' size and
   position in the header button.
5. `pytest tests/unit/routers/test_settings.py -q` and the console
   `i18n` / `LanguageSwitcher` / `LanguageBadge` vitest suites.

## Evidence

```bash
pre-commit run --file src/qwenpaw/app/routers/settings.py \
  tests/unit/routers/test_settings.py
# check python ast Passed / mypy Passed / black Passed
# flake8 Passed / pylint Passed (all hooks Passed)

pre-commit run --file console/src/i18n.ts \
  console/src/constants/languageList.tsx \
  console/src/layouts/SidebarSettingsPanel.tsx \
  console/src/components/LanguageSwitcher/index.tsx \
  console/src/test/icons-mock.ts
# trailing-whitespace Passed / detect-private-key Passed (all Passed)

pytest tests/unit/routers/test_settings.py -q
# 16 passed in 1.45s

npx vitest run src/i18n.test.ts src/components/LanguageSwitcher/ \
  src/components/LanguageBadge/
# Test Files  4 passed (4)
#      Tests  13 passed (13)

npx prettier --check <files changed by this PR>
# All matched files use Prettier code style!

npx eslint <files changed by this PR>
# exit 0, no warnings
```

Behavioural proof, before vs after:

```text
PUT /api/settings/language for all seven languages
  before: vi -> HTTP 400 "Invalid language, must be one of
          ['en','id','ja','pt-BR','ru','zh']", GET still returned the
          previously saved language
  after:  all seven -> HTTP 200, GET echoes each value

i18n.changeLanguage("pt-BR")
  before: resolvedLanguage "en",   t("chat.newTask") "New task"
  after:  resolvedLanguage "pt-BR", t("chat.newTask") "Nova tarefa"
  (the existing ja-JP regression case in i18n.test.ts still passes)

Live console DOM, same header button
  before: language badge 14x14 at x=1165 y=20.8
  after:  language badge 20x20 at x=1162 y=15.5
  reference real icon spark-pt-line: 20x20 at x=1162 y=15.5

Guard test verified to actually fail on drift
  temporarily dropping "vi" from _VALID_LANGUAGES:
  FAILED test_valid_languages_matches_console_language_list
  "Extra items in the right set: 'vi'"
```

## Additional Notes

- `_ACL_LANG_FALLBACK` in `app/channels/base.py` was intentionally left
  unchanged. `_ACL_I18N` has no `vi` strings, so adding `vi` to the
  fallback set would make `_ACL_I18N[key]["vi"]` raise `KeyError`. The
  current behaviour (Vietnamese falls back to English access-control
  text) is safe; adding translated strings is separate work.
- Pre-existing translation coverage gaps are out of scope for this PR:
  `vi` 55.9%, `id` 75.2%, `ja` 89.0%, `ru` 89.4%, `pt-BR` 90.0% of the
  `en` key count. These are content debt, not code defects.
- Adding a language now needs two edits (`languageList.tsx` and
  `_VALID_LANGUAGES`); forgetting either one fails CI via the new guard
  test.


rayrayraykk:
Reviewed commit `98d1a6166aacdcbe66cbee82146712093e7e5213`.

No blocking functional findings in this diff. The backend now accepts `vi`, and I independently verified that selecting `pt-BR` and reloading with it persisted both resolve the Portuguese bundle and translate `chat.newTask` to `Nova tarefa`.

Validation on macOS: **16 backend settings tests passed** in the QwenPaw conda environment; **13 existing targeted frontend tests passed**; one additional selection/reload regression test passed. The settings route retains its existing async file helpers and locking, so this change adds no direct synchronous filesystem I/O on the event loop. The new test uses pathlib and explicit UTF-8, with no apparent Windows path separator dependency. Windows was not executed.

Two non-blocking maintainability suggestions:

- Commit a regression test asserting `resolvedLanguage === "pt-BR"` and an actual Portuguese translation, both after selection and initialization. The existing i18n tests do not cover the central pt-BR fix; checking only `i18n.language` can miss fallback-to-English behavior.
- LanguageBadge duplicates another icon library's SVG geometry and private sizing hooks (`spark-icon` / `data-spark-icon`). A Lucide language icon with a textual language code would follow the project's icon direction and avoid maintaining copied geometry. If this implementation stays, rendered sizing coverage is more useful than asserting exact rectangle constants.

zhijianma:
Thanks for the review @rayrayraykk — both suggestions checked before deciding.

| Finding | Resolution | Basis |
|---|---|---|
| pt-BR regression test | ✅ Done — `b60abd55` | Your point is correct: in the buggy state `i18n.language` stayed `"pt-BR"` while only `resolvedLanguage` degraded to `"en"`, so the existing `i18n.language` assertions could not catch it. |
| Switch `LanguageBadge` to a Lucide language icon | ⏸️ Not adopting | lucide-react (1912 icons) has no per-language badge — only generic `earth` / `flag` / `globe`. A generic icon beside a non-English label is the exact problem this commit removes. |
| `data-spark-icon` is a private sizing hook | ⏸️ Keeping deliberately | The coupling is what makes sizing correct, not incidental — without it the badge renders 14×14 against the real icon's 20×20 in the same header button. |
| Rendered sizing coverage over rectangle constants | ⚠️ Partially | jsdom returns all-zero `getBoundingClientRect()`, so rendered size is not measurable in vitest; the attribute assertion is the available proxy. |

Evidence for the new test, both directions:

```bash
npx vitest run src/i18n.test.ts
# Test Files 1 passed (1) / Tests 5 passed (5)

# with console/src/i18n.ts reverted to main:
# FAIL i18n pt-BR bundle resolution > resolves a persisted pt-BR on startup
# FAIL i18n pt-BR bundle resolution > resolves pt-BR after an explicit switch
# AssertionError: expected 'en' to be 'pt-BR'   (2 failed | 3 passed)
```

The two new tests assert `resolvedLanguage === "pt-BR"` and
`t("chat.newTask") === "Nova tarefa"`, on both startup restore and an
explicit switch.

On the icon: the frame geometry is measured from the real badges rather
than chosen (outer box 117.12–906.88, inner 181.12–842.88, ring 64,
radius 152.96 in a 1024 viewBox), so the constants in the test guard
against silent drift from that measurement. If the icon set later
grows Indonesian and Vietnamese badges, this component can be dropped
in favour of them — it is only used by the two `LANGUAGE_LIST` entries.

rayrayraykk (commented):
# PR #7752 Code Review Report

- PR: [#7752](https://github.com/agentscope-ai/QwenPaw/pull/7752)
- Title: `fix(console): make vi and pt-BR language selection work`
- Reviewed head: `b60abd5574b0388549e21321f64a7ebcb405b510` (verified against `HEAD`)
- Supplied base: `0b8a12f5518081cf0cd0d38d645ea0c10c95cb2a`
- Computed merge base: `983b3cebf4dda409458783b286d3a95cc80d135f`
- Review date: 2026-09-14
- Conclusion: **Request changes — 1 medium blocking finding; 0 critical, 0 high, and 0 low findings**

## Review checklist

- [x] Security and malicious-request assessment completed before ordinary review.
- [x] Complete merge-base diff inspected: all nine changed files and 201 additions/15 deletions accounted for.
- [x] Async, synchronous I/O, concurrency, lifecycle, and persistence paths reviewed.
- [x] Windows, macOS, and Linux compatibility reviewed statically; execution occurred only on macOS.
- [x] Correctness, failure handling, fallback behavior, serialization, and recovery reviewed.
- [x] Targeted backend tests and Python lint executed.
- [ ] Frontend tests executed locally — `console/node_modules` is absent, and dependencies were not downloaded.
- [x] Code quality, coupling, test adequacy, evidence, and repair guidance reviewed.
- [x] Worktree confirmed clean after review.

## Security and malicious-request assessment

No concrete security finding exists. The complete diff contains no dependency or lockfile changes, hooks, workflows, installers, credential access, external transmission, persistence mechanisms, permission expansion, destructive actions, or attempts to influence the review outcome. The public language-setting endpoint was already public; the backend change only admits the existing `vi` console value.

## Main findings

### Medium, blocking — Two language selectors still do not follow the repaired remote-persistence contract

The PR adds failure feedback to the desktop `LanguageSwitcher` and sidebar paths, but the same setting remains inconsistent elsewhere:

- In [console/src/pages/SettingsCenter/GeneralSettings.tsx#L70-L74](https://github.com/agentscope-ai/QwenPaw/blob/b60abd5574b0388549e21321f64a7ebcb405b510/console/src/pages/SettingsCenter/GeneralSettings.tsx#L70-L74), the primary General Settings selector still discards every rejected `updateLanguage` request with an empty catch.
- In [console/src/layouts/Header.tsx#L76-L87](https://github.com/agentscope-ai/QwenPaw/blob/b60abd5574b0388549e21321f64a7ebcb405b510/console/src/layouts/Header.tsx#L76-L87), the mobile header selector updates i18next and local storage but never invokes `settingsApi.updateLanguage` at all.

Triggering paths are therefore straightforward: select any language in General Settings while the backend is unavailable or rejects the request, or select a language from the mobile header under any conditions. The visible UI and local storage change, masking the fact that the server-side preference was not updated. A later client without that local-storage value will load the stale server preference, and General Settings users receive no failure feedback.

This affects all supported platforms and every language, including the newly enabled `vi` and repaired `pt-BR`. The empty catch and mobile omission predate this branch, but they are directly in scope because the PR explicitly repairs this persistence/failure behavior in the other selectors. No test covers either path when persistence fails.

Smallest repair: apply the same `settingsApi.updateLanguage(...).catch(...)` behavior and localized error message to General Settings and the mobile header, then add rejection-path tests for both. A shared language-persistence helper would also prevent these entry points from drifting again.

## Synchronous I/O and event loop

No event-loop-blocking regression was found. Runtime settings reads and writes continue through asynchronous filesystem helpers under the existing path lock in [src/qwenpaw/app/routers/settings.py#L34-L50](https://github.com/agentscope-ai/QwenPaw/blob/b60abd5574b0388549e21321f64a7ebcb405b510/src/qwenpaw/app/routers/settings.py#L34-L50). Frontend persistence remains promise-based. The new synchronous `read_text` call exists only in a unit test.

## Windows / macOS / Linux compatibility

No OS-specific runtime defect was found. The backend test uses `pathlib.Path`, the persisted value remains a plain JSON string, and the frontend changes use browser APIs and SVG without platform-specific paths or shell behavior. Tests were executed on macOS only; Windows and Linux behavior was assessed statically.

## Functional correctness and error handling

The core fixes are supported by the code:

- `vi` is now admitted by the backend whitelist at [src/qwenpaw/app/routers/settings.py#L27-L31](https://github.com/agentscope-ai/QwenPaw/blob/b60abd5574b0388549e21321f64a7ebcb405b510/src/qwenpaw/app/routers/settings.py#L27-L31).
- The i18next configuration includes the language-only alias required for the `pt-BR` resource at [console/src/i18n.ts#L35-L53](https://github.com/agentscope-ai/QwenPaw/blob/b60abd5574b0388549e21321f64a7ebcb405b510/console/src/i18n.ts#L35-L53).
- The regression tests check both `resolvedLanguage` and translated output for startup restoration and explicit switching at [console/src/i18n.test.ts#L73-L102](https://github.com/agentscope-ai/QwenPaw/blob/b60abd5574b0388549e21321f64a7ebcb405b510/console/src/i18n.test.ts#L73-L102).
- Existing JSON settings remain backward compatible; no schema or migration is required.

The unresolved failure paths are described in the main finding. No additional rollback, cancellation, or recovery defect was identified.

## Code quality, modularity, and design

The whitelist guard covers the current literal `LANGUAGE_LIST` structure, and the badge component is narrowly scoped to the two missing glyphs. The material design-coupling concern is duplicated language-change/persistence logic across four UI entry points; that duplication directly caused the incomplete behavior above.

No separate maintainability or style-only finding is reported.

## Verification results

### Locally executed checks

- `git rev-parse HEAD` → `b60abd5574b0388549e21321f64a7ebcb405b510`.
- `git merge-base HEAD 0b8a12f5518081cf0cd0d38d645ea0c10c95cb2a` → `983b3cebf4dda409458783b286d3a95cc80d135f`.
- `git diff --check 983b3cebf4dda409458783b286d3a95cc80d135f...HEAD` → passed with no output.
- `conda run -n QwenPaw env PYTHONPATH=src pytest tests/unit/routers/test_settings.py -q` → **16 passed in 1.03s**.
- `conda run -n QwenPaw flake8 --extend-ignore=E203 src/qwenpaw/app/routers/settings.py tests/unit/routers/test_settings.py` → passed with no output.
- `conda run -n QwenPaw pytest tests/unit/routers/test_settings.py -q` → not a valid worktree result: the environment resolved an older installed checkout instead of this repository and reported two `vi` failures. Explicitly selecting `PYTHONPATH=src` produced the passing result above.
- `conda run -n QwenPaw black --check --fast --line-length=79 src/qwenpaw/app/routers/settings.py tests/unit/routers/test_settings.py` → not accepted as authoritative: the environment has Black 26.5.1 while the repository pins Black 23.3.0; the installed version reported both files would be reformatted.
- Final `git status --porcelain=v1` → clean.

### Externally reported CI for the reviewed head

- **17 successful:** Frontend Summary; CodeQL, CodeQL (TypeScript), and CodeQL (Python); Vitest Unit Tests; website and console format checks; four PR Spam Gate jobs; both change-detection jobs; Playwright UI Smoke; stamp; Real behavior proof; and Release Window.
- **In progress:** Pre-commit.
- **Waiting:** two approval gates, including Maintainer Approval.
- **Failing or skipped:** none reported.

These statuses are external reports and do not replace the local verification limitations above.

### Checks not run

- Frontend Vitest, TypeScript, ESLint, and Prettier checks were not run locally because `console/node_modules` is absent. Dependencies were not fetched.
- Windows and Linux execution was not available.
- Broad integration and end-to-end suites were not run because they were not the smallest relevant offline checks.

## Final recommendation

**Request changes.** The `vi` backend and `pt-BR` bundle fixes are supported by the inspected code and targeted backend test, but the persistence/error-feedback repair is incomplete across the actual language-selection entry points. General Settings still suppresses failures, and the mobile header still bypasses remote persistence. Those paths should be fixed and covered before approval.

zhijianma:
Both paths confirmed and fixed in `e4232244`, using the shared helper you suggested.

| Finding | Resolution | Basis |
|---|---|---|
| `GeneralSettings.tsx` empty catch | ✅ Fixed | Now routes through `applyLanguagePreference` with a localized `onPersistError`. |
| `Header.tsx` mobile menu never calls `updateLanguage` | ✅ Fixed | Same helper; this was the worse of the two — the server preference was never written at all. |
| No rejection-path test for either | ✅ Added | `GeneralSettings.language.test.tsx` and `Header.language.test.tsx`, one rejection case each. |
| Shared helper to stop drift | ✅ Adopted | `utils/languagePreference.ts`; all four selectors now use it and none calls `settingsApi.updateLanguage` directly. |

Verified each new test actually fails on the behaviour it guards, not just passes on the fix:

```text
GeneralSettings — drop onPersistError from the call:
  AssertionError: expected "Mock" to be called with
  arguments: [ 'agentConfig.languageSaveFailed' ]   (1 failed)
  restored -> 1 passed

Header mobile — restore the original onClick (no remote write):
  AssertionError: expected "vi.fn()" to be called with arguments: [ 'vi' ]
  AssertionError: expected "Mock" to be called with
  arguments: [ 'agentConfig.languageSaveFailed' ]   (2 failed)
  restored -> 2 passed
```

Helper contract is covered separately in `utils/languagePreference.test.ts`
(switch + store + persist, rejection surfaced, `persistRemotely: false`
skips the write).

```bash
npx vitest run src/utils/languagePreference.test.ts \
  src/components/LanguageSwitcher/ src/i18n.test.ts \
  src/components/LanguageBadge/ src/layouts/SidebarSettingsPanel.test.tsx \
  src/layouts/Header.language.test.tsx src/pages/SettingsCenter/
# Test Files 11 passed (11) / Tests 50 passed (50)

npx prettier --check <8 files changed>   # all use Prettier code style
npx eslint <8 files changed>             # exit 0, no warnings
npx tsc --noEmit                         # no errors in changed files
```

One correction for your next pass: the empty catch was at
`GeneralSettings.tsx:73` rather than L70-L74 (that range is the
`currentLanguage` derivation), and the mobile `onClick` was at
`Header.tsx:83-86`. Substance of both findings was accurate.

Agreed that both predate this branch — taking them here is right, since
this PR is what claims to repair the persistence contract and leaving
two of four selectors outside it would have made that claim false.

rayrayraykk (approved):

Issue #7318, opened 2026-08-26
[Discussion] QwenPaw Hub, the multi-tenant edition, is coming in 2.2.0: what should we build next? / QwenPaw 多租户版 Hub 将于 2.2.0 推出：你希望我们接下来做什么？


rayrayraykk:
QwenPaw started as a personal AI assistant, but the community has repeatedly asked for a better way to run it for a team. QwenPaw Hub is our first response to that call.

Related community requests:

- [#2324 — Multi-user access and admin-managed skills](https://github.com/agentscope-ai/QwenPaw/issues/2324)
- [#1884 — Isolated agents and workspaces for multiple users](https://github.com/agentscope-ai/QwenPaw/issues/1884)
- [#2233 — User-scoped data and configuration](https://github.com/agentscope-ai/QwenPaw/issues/2233)
- [#4702 — RBAC and enterprise administration](https://github.com/agentscope-ai/QwenPaw/issues/4702)
- [#5780 — Multi-user account management for team use](https://github.com/agentscope-ai/QwenPaw/issues/5780)
- [#6335 — Deploying QwenPaw for company users](https://github.com/agentscope-ai/QwenPaw/issues/6335)

### What is coming in 2.2.0?

QwenPaw Hub will provide one self-hosted entry point for a team:

- Each member signs in with their own account and gets their own QwenPaw.
- Workspaces, configuration, credentials, and sessions are stored separately.
- Administrators can manage accounts and start, stop, restart, or disable user runtimes.
- Administrators can choose the Local or Docker runtime and configure shared Docker resource limits.

This is an early release for trusted companies, labs, and internal teams. Local sandboxes and Docker containers still share a host or Docker VM kernel, so 2.2.0 is not intended to be a public multi-tenant service for mutually untrusted users.

### What should come next?

Here are some directions we are considering. They are ideas for discussion, not a committed roadmap:

- **Kubernetes and operations:** Helm charts, one runtime per Pod, persistent storage, Secrets integration, multi-node scheduling, autoscaling, upgrades, backup, and disaster recovery. The community has already explored [Kubernetes deployment](https://github.com/agentscope-ai/QwenPaw/issues/1740) and [non-interactive administrator bootstrap](https://github.com/agentscope-ai/QwenPaw/issues/1668).
[...]

PR #7774, opened 2026-09-15
fix(hub): derive the startup provisioner allow-list from the build

wangjian124:
## Description

`run_hub_app()` resolves the startup configuration against a hard-coded
`available_provisioners={"local", "docker"}` (`src/qwenpaw/hub/control_app.py:1562`),
while every other entry point validates against the assembled runtime service
(`set(runtime_service.provisioners)` — `control_app.py:147`, plus both settings
routes).

That literal is a second source of truth for *which provisioners this Hub build
registers*. Whenever the registry changes, the startup path keeps validating
against the stale list, so an otherwise valid configuration is rejected at boot
with a misleading `Unknown runtime provisioner: <name>` — even though the same
file is accepted by the HTTP/settings path.

This PR extracts the registry construction into
`build_runtime_provisioners(root_dir)` and uses it for both the service assembly
and the startup validation, so there is a single source of truth. The registered
provisioners are unchanged (`local`, `docker`).

**Related Issue:** Relates to #7318 (Hub multi-tenant discussion)

**Security Considerations:** None — no change to authentication, credential
handling, or configuration semantics; for current builds the accepted provisioner
set is identical.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [x] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --files <changed files>` locally: black, flake8, mypy and pylint all pass
- [x] Ready for review

> Note on the lint run: my local default interpreter is Python 3.9, under which
> the `mypy`/`pylint` hooks flag the *pre-existing* `X | Y` annotations at
> `tests/unit/hub/test_control_app.py:96/97/100` (upstream code from 2026-09-07/08,
> untouched by this change). Both hooks were re-run with Python 3.13 and pass
> cleanly; on CI (3.11) they are unaffected.

## Testing

Two new unit tests in `tests/unit/hub/test_control_app.py`:

- `test_run_hub_app_validates_against_registered_provisioners` — a build that does
  not register `docker` must reject `runtime.provisioner: docker` at startup.
  **This test fails on current main** (the hard-coded set accepts `docker`).
- `test_run_hub_app_accepts_provisioner_registered_by_the_build` — the positive
  path: a provisioner that *is* registered still starts.

```bash
pytest -q tests/unit/hub/test_control_app.py
```

## Evidence

A/B on the full source tree in the same environment — only `control_app.py`
differs between the two runs:

```text
##### A. control_app.py unpatched (current main) #####
AttributeError: <module 'qwenpaw.hub.control_app'> does not have the attribute
'build_runtime_provisioners'
=========================== short test summary info ============================
FAILED tests/unit/hub/test_control_app.py::test_run_hub_app_validates_against_registered_provisioners
1 failed, 37 passed in 37.27s

##### B. control_app.py patched (this PR) #####
......................................                                   [100%]
38 passed in 38.74s
```

Lint gate on the two changed files:

```text
$ flake8 --extend-ignore=E203 src/qwenpaw/hub/control_app.py tests/unit/hub/test_control_app.py
(exit 0)

$ black --line-length=79 --check src/qwenpaw/hub/control_app.py tests/unit/hub/test_control_app.py
All done! 2 files would be left unchanged.

$ mypy --ignore-missing-imports --follow-imports=skip --explicit-package-bases ...
Success: no issues found in 2 source files

$ pylint ... src/qwenpaw/hub/control_app.py tests/unit/hub/test_control_app.py
Your code has been rated at 10.00/10
```

## Additional Notes

The allow-list only matters once a build registers a provisioner beyond
`local`/`docker` (for example a Kubernetes runtime provisioner — we are
prototyping exactly that on top of the Hub work discussed in #7318). Today the
two lists agree, so this is a latent inconsistency rather than a reachable
failure. I am happy to drop the helper and keep an explicit literal if you
prefer that direction.


rayrayraykk:
See #7696

rayrayraykk (changes requested):
Thanks for addressing the duplicated provisioner allow-list. Please rebase this PR on top of #7696 first, as the two changes currently conflict in control_app.py: #7696 moves get_hub_root() into bootstrap.py.
I also would not recommend merging the current implementation as-is. run_hub_app() constructs the provisioners once to obtain the allow-list, and build_runtime_service() constructs them again when assembling the runtime service. Although the constructors are currently lightweight and mostly side-effect-free, this creates unnecessary duplicate instances and could become problematic as provisioner initialization grows more expensive or gains side effects.
Please consider using a shared declarative registry or factory mapping instead. Startup validation can derive the allowed names from its keys, while the runtime service creates each provisioner only once. This would provide a genuine single source of truth without duplicate initialization.

wangjian124:
Rebased onto `a90c9ca` (#7696) and reworked per your review -- thanks for the pointer.

**What changed vs. the previous revision**

- The allow-list is no longer a hand-maintained literal. `PROVISIONER_FACTORIES`, declared next to `build_runtime_service`, is now the single declarative source of truth: it maps `provisioner name -> factory(root_dir)`.
- Startup validation only reads the keys. `run_hub_app()` passes `set(PROVISIONER_FACTORIES)` to `HubConfigStore.resolve()`, so the startup path constructs nothing at all.
- `build_runtime_provisioners()` then instantiates exactly one instance per registry entry for the runtime service, so each provisioner is built once. Adding a provisioner is now a one-line registry change instead of two edits in two places.
- The rebase resolved the `control_app.py` conflict with #7696 (`get_hub_root()` now comes from `hub/bootstrap.py`); the PR went from `mergeable_state: dirty` to `MERGEABLE`.

**Tests**

- `test_run_hub_app_validates_against_registered_provisioners` patches `PROVISIONER_FACTORIES` with a factory that raises `AssertionError("startup validation must not build provisioners")` if it is ever invoked, then asserts that startup rejects an unregistered provisioner with `ValueError: Unknown runtime provisioner: docker` and never reaches `create_hub_app`. That is the regression guard for exactly the concern you raised: the validation path must not construct provisioners.
- `test_run_hub_app_accepts_provisioner_registered_by_the_build` covers the accept path.
- I A/B'd the first test: restoring the old `{"local", "docker"}` literal makes it fail, so the test really does discriminate.

**Local results**

- `pytest tests/unit/hub -q` -> 144 passed, 5 skipped, 1 failed. The single failure is `test_process_isolation.py::test_linux_command_mounts_python_base_prefix`, which needs a POSIX `bin/python3.13` symlink; it fails identically on the pre-change baseline (that file is byte-identical across the two revisions), so it is unrelated to this change.
- pre-commit on the two changed files: check-ast, docstring-first, encoding-pragma, trailing-whitespace, add-trailing-comma, black (--line-length=79), flake8, mypy and pylint (10.00/10) all pass.

Issue #7727, opened 2026-09-12
[Bug]: Out-of-workspace write hard-block is blind to kimi-code's Write tool — `_paths` extraction does not recognize kimi toolCall path fields


remotepan-design:
**Environment**: QwenPaw 2.2.0 (desktop, macOS arm64); kimi-code runner via ACP.

**Observed (live negative probe, 2026-09-12)**:

1. A delegated kimi-code ACP session was asked to write a probe file at `/tmp/kimi_oob_probe.txt` — outside the agent's workspace.
2. The write succeeded without triggering QwenPaw's out-of-bounds hard block and without a path-related permission prompt. (The file was verified to exist on disk afterwards.)
3. Inspection of the toolCall payload shows kimi-code's Write calls do not expose the target path in any of the fields QwenPaw's `_paths` extraction recognizes (`locations` / `diff` / `rawInput.path`). The path boundary check therefore sees an empty path set and has nothing to deny.

**Why this matters**: operators configuring ACP runners reasonably assume the workspace-boundary hard block applies uniformly to all runners' file-writing tools. For kimi-code it silently does not — the effective guard reduces to the catastrophic-shell-command regex plus whatever the runner itself enforces. This is a silent security-posture gap, not a usability issue.

**Suggested fix**:
1. Extend `_paths` extraction to also parse kimi-code's toolCall shapes — the path is available in `rawInput` (under tool-specific keys) and/or the human-readable `title` field; and/or
2. Fail-closed option: when a toolCall is known to be a file-write operation (by tool name/kind) but `_paths` extraction yields nothing, treat it as out-of-bounds or force an interactive confirmation rather than silently allowing.

**Minimal repro**:
1. Configure a kimi-code ACP runner (default policy, workspace boundary enabled).
2. Via `delegate_external_agent`, instruct the runner to create a file at an absolute path outside the workspace (e.g. `/tmp/probe.txt`) using its Write tool.
3. Expected: hard block (or at minimum an explicit out-of-bounds prompt). Actual: file is written with no boundary intervention.

x1n95c:
Hello, thank you for your detailed report. We will add support for Kimi Code tool call parameters soon. By the way, if you enable “Trust Mode” (set to true), permission checks will be bypassed—this is by design. Thanks again for your suggestion!

PR #7783, opened 2026-09-15
fix(ACP): Improves the experience of delegating work to external ACP runners

x1n95c:
## Description

Improves the experience of delegating work to external ACP runners. Two user-visible fixes:

**1. ACP replies are no longer duplicated or fragmented.** A turn's text reached `delegate_external_agent` twice — incrementally as deltas, and again in full from `finish_prompt` — and every delta was rendered as its own `[assistant]` item. Because runners chunk differently, the same bug looked different per runner: kimi-cli emitted a marker every few words (`[assistant]\n你好！我是[assistant]\n**Kimi Code…`), while codex/qwen repeated the whole answer verbatim below the `runner: … working directory: …` header. Deltas are now coalesced into one item per flush, the closing block prefers the complete text over the last delta (kimi's used to end on a bare fragment such as `告诉我就行。`), and the body is omitted when the stream already carried it.

**2. kimi's tool arguments are parsed, so the workspace guard actually applies (#7727).** The hard block only read `locations[].path`, `content[].type=="diff"` and the literal key `rawInput.path`. kimi-cli sends none of those — it puts its arguments in a JSON string inside a text content block and repeats nothing at permission time — so out-of-workspace writes went unchallenged, and `rm -rf /` via kimi's `Shell` sailed through because the command guard never received a command either. Arguments are now extracted from content text blocks, the path-key vocabulary is widened (`claude-agent-acp` sends `file_path`), and the accumulated `ToolCallStart` state is merged into the permission delta before deciding.

**Related Issue:** Relates to #7727

**Security Considerations:** Tightens a fail-open path in the ACP hard block and adds no privilege. Prose is never parsed as arguments (the JSON decode is guarded on a leading `{`); the widened path keys only feed `is_path_outside_boundary`, so a false-positive key match cannot block an in-workspace write; the merge is delta-wins, so a request narrowed at approval time is not judged against stale arguments. Dropping the duplicated closing body is safe because `_coordinator.py:689` appends every chunk to the accumulated tool response, and a partial overlap (a cancelled turn that streamed only a prefix) deliberately does **not** suppress — losing text is worse than repeating some of it. Not addressed here: shell redirection targets, a fail-closed prompt when no path is recoverable from any source, the `paths[:5]` cap still feeding the deny loop, and the fact that enforcement only runs when the runner asks.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring
- [ ] Tests

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

`add-trailing-comma` and `black` reformatted a file on the first run; that landed as `495503fb2` and the hooks were rerun to green. Note that `495503fb2` is net-absent from this PR's diff, because the later `094bb6762` removed the test modules it formatted — both remain in branch history only and disappear under squash merge.

Documentation is intentionally not updated in this PR. `website/public/docs/acp-integration.en.md` lists `trusted` as a config field without explaining that `trusted=True` (the default for every built-in runner) makes QwenPaw auto-approve on the user's behalf, which contradicts the same page's "QwenPaw does not decide on the user's behalf"; and the hard block is best-effort rather than a boundary, since it only runs when the runner asks. Both deserve a docs pass, but they are separable from this fix and would widen the review surface.

Not applicable — this PR touches no channel. It changes four source files: `agents/acp/permissions.py` and `agents/acp/client.py` (permission guard), `agents/acp/tool_adapter.py` (event rendering), and `agents/tools/delegate_external_agent.py` (stream assembly).

## Testing

This PR adds no new test modules; it relies on the existing suites as regression. Commands and raw results are in Evidence below.

**Coverage caveat, stated plainly:** the existing ACP suites do not exercise the branches this PR adds. `tests/unit/agents/test_acp_permissions_trusted.py` contains a single `_paths` case (`locations[].path`), so nothing in CI walks the new argument extraction from content text blocks, the widened path-key vocabulary, or the accumulated-state merge; likewise nothing asserts the delta coalescing or `suppress_body`. A future regression in either fix would not be caught automatically. Both behaviours were verified by hand during development and the results are recorded in Evidence, but neither is pinned by a test.

Verifying the argument parsing against a live kimi-cli runner additionally needs `kimi-cli login` first: `kimi acp` gates `session/new` on a persisted OAuth token (`kimi_cli/acp/server.py:158`, unconditional; `~/.kimi/credentials/kimi-code.json`) and ignores `providers.openai.api_key`. QwenPaw implements neither `authenticate` nor the `terminal-auth` method kimi advertises, so an un-logged-in runner fails at session creation with a bare "Authentication required" and the user is never told which command to run. That gap is unrelated to this PR and is not addressed here.

## Evidence

`pre-commit run --all-files` at `094bb6762` (current head) — every hook Passed, exit code 0, and the run left the working tree unmodified:

```bash
pre-commit run --all-files

```

```bash
pytest tests/unit/agents tests/unit/tool_calls tests/unit/security -q
3995 passed, 4 skipped, 8 warnings in 45.28s
# at commit 094bb6762, i.e. after the test modules were removed

pytest tests/integration/test_acp_runner.py -q
5 passed in 392.26s (0:06:32)
# real mock ACP runner over stdio JSON-RPC, end to end
```

Bypass matrix, before and after. Produced by a throwaway script that rebuilt kimi-cli v1.50.0's real payloads with kimi's own `build_diff_blocks` / `display_block_to_acp_content` / `extract_key_argument` and fed them through `ACPPermissionAdapter` — "before" on the base commit, "after" on this branch. Since this PR ships no test module, these numbers are **not reproducible from the repo**:

```text
payload                                            before    after
WriteFile, new content (diff block)                blocked   blocked
WriteFile, content unchanged (no diff block)       ALLOWED   blocked
Shell: rm -rf / --no-preserve-root                 ALLOWED   blocked
rawInput.file_path  (claude-agent-acp)             ALLOWED   blocked
rawInput.filePath / abs_path                       ALLOWED   blocked
```

Streaming, same caveat — measured by driving `_stream_action_responses` directly against a fake `_run_action`, on a kimi-style answer split into 15 deltas:

```text
before:  ~16 blocks, "[assistant]" x15+, answer text x2 (stream + closing block)
after :   2 blocks, "[assistant]" x1,  answer text x1, closing block carries header only

codex/qwen single-chunk answer:
before:   answer text x2
after :   2 blocks, "[assistant]" x1, answer text x1
```

## Additional Notes

Commits on this branch:

```text
094bb6762 test(acp): drop the new kimi-support and text-stream suites
495503fb2 style(acp): apply pre-commit autofixes to the stream test
36b2f309b fix(acp): coalesce delegated assistant text and stop repeating it
956834da8 fix(acp): parse tool-call arguments so the boundary guard can see them
```

The two `fix(acp)` commits are independent and can be reviewed or reverted separately. `495503fb2` and `094bb6762` cancel out in the PR diff — an autofix applied to files the later commit removed — so the net change is exactly the four source files, 304 insertions and 40 deletions. All four commits collapse under squash merge.

One incidental observation, not fixed here: `ClientCapabilities()` serialises a malformed `auth` capability — `{'auth': {'terminal': False}, 'fs': {...}, 'terminal': False}` — and pydantic emits `PydanticSerializationUnexpectedValue(Expected AuthCapabilities …)` on every ACP connection setup. It does not appear to affect runners in practice, but it is sent on the wire to all of them and looks like an acp-SDK/model default mismatch worth a separate look.


lalaliat (approved):
LGTM

PR #7802, opened 2026-09-16
feat(telemetry): report daily Runtime activity on Agent execution

rayrayraykk:
Record one daily Runtime activity event when the built-in Agent or an external harness executes, including scheduled tasks. Startup, page visits and commands that do not execute an Agent do not count.

The Agent path only compares an in-memory UTC date and enqueues the first daily signal. It performs no file I/O, system probes or network calls. A dedicated daemon thread owns persistence, environment collection and uploads; shutdown signals stop without joining. System information is collected lazily once per process and reused across active days.

Reuse installation telemetry consent, QWENPAW_TELEMETRY_DISABLED, the saved opt-out, environment collection and upload endpoint. Persist the random Runtime UUID and last attempted date in the existing `.telemetry_collected` JSON using existing atomic-write/path-lock utilities. Mark the day before sending, so success or failure is not replayed on restart. Independent Hub Runtime data directories have distinct identities even when business IDs match. No database, new state file, separate CLI, frontend, README or branch-format migration is added. Explain reporting in the init notice. Marker schema `version` is `1.4`; `qwenpaw_version` retains the actual application version.

Network policy: initial request plus at most two retries; all attempts share a five-second timeout, excluding system-information collection. Each attempt is capped at two seconds. HTTP runs in a private event loop inside the daemon thread and reads response headers only. No persistent retry queue, delayed backoff or later-day replay. Failed or interrupted sends may be lost by design. Receiver deduplication by telemetry_runtime_id + activity_date is still required for retried deliveries.

Validation: 47 local telemetry and execution-entrypoint tests passed, including 1,000 same-day signals enqueuing once without I/O, ten active days reusing one environment snapshot, restart deduplication, retry limits and stalled requests exhausting the five-second network budget. Changed-file pre-commit checks passed. The existing marker-version assertion is updated to expect schema version `1.4`. No new test files are committed. A broader check of existing telemetry, Runtime, harness, Hub Docker, startup and CLI tests passed: 422 tests; no other assertion updates were needed.

Installation regression audit: original installation trigger, payload fields, sync uploader/timeout and mark-even-on-failure behavior are retained. Shared marker updates preserve unknown/daily fields and restore prior recovery of malformed marker contents and strict boolean opt-out semantics. The receiver implementation is not in this checkout; installation analytics must distinguish new runtime_active events from installation events rather than counting all requests.

The receiver requires install_id. Daily payloads now set install_id to the same persistent UUID as telemetry_runtime_id, without collecting a machine identifier or creating another identity. Live protocol checks found the full daily payload rejected without install_id and accepted with a dummy install_id (HTTP 200); server-side daily field storage/aggregation is still unverified. Local payload checks confirm equal identity fields, stability across days/restarts, and isolation between Runtime directories. Existing installation tests: 22 passed; pre-commit passed. No new test files were added.


xieyxclack (approved):
LGTM

PR #7697, opened 2026-09-11
ci: slim the PR gate to Ubuntu-only backend tiers and add a release-time full test gate

yutai78786:
> **Submitted by**: 秦琼·CIOps@QPQAT

## Description

This PR slims the PR-time CI gate and adds a release-time full test gate (phase 1: 8 changes, all in CI configuration plus one test-file fix). **No test case is deleted and no product code is changed.**

**What it does:**

1. The three backend test matrices in the PR gate run on Ubuntu only (macOS/Windows entries removed). Cross-platform compatibility stays covered by the nightly full sweep.
2. The console frontend build is removed from the three backend test tiers (9 build/copy steps). The E2E tier keeps its own build.
3. The standalone `fallback` job is deleted and its missing-marker guard is folded into the ubuntu p0 job (guard semantics unchanged: an integration case missing its p0/p1/p2 marker still fails loudly).
4. Coverage tracking is removed from the PR side (three tiers of `--cov`, three upload steps, the `coverage-report` job). Coverage is collected in nightly instead. **No threshold value is lowered and no new blocking is introduced.**
5. The leftover temporary `pr-preview-tests.yml` is deleted (no references anywhere in the repository).
6. **A release-time full test gate is added**: `release.yml` gains a `full-test-gate` job that invokes the nightly workflow in its `release` form (backend full matrix across all platforms + a reviewed 64-case E2E blocking set). `finalize` needs it, so a red gate means the draft is never flipped to published.
7. Integration tests no longer wait for unit tests (they run in parallel). `Test Summary` still `needs` unit tests, so the fail-closed semantics are preserved.
8. Two latent defects exposed by the changes above are fixed: a bad test that depended on runner uptime (now fakes the clock via `monkeypatch`), and a `console/dist` empty-directory placeholder in the integration job to reproduce the production routing shape (no product code change).

**Why:** the PR gate measured 79.6 minutes wall clock (run 34194545575), where Windows unit tests (25.4 min) and Windows integration p1 (39.6 min) were the two slowest jobs, and the integration tier waited serially on unit tests. After slimming, the fork measures 19.9 minutes all green. On the release side there was previously **no full test gate at all**: the full sweep triggered by pushing to main was blocked by the manual-approval environment and sat unrun for 8 hours (run 34209329123), so a broken release could have shipped.

**Related Issue:** No corresponding upstream issue; this is CI-pipeline maintenance agreed with the project maintainer.

[...]

PR #7803, opened 2026-09-16
ci(e2e): enable the declared pytest-timeout and raise the per-shard job timeout to 60 minutes

yutai78786:

## Description

This PR fixes two coupled problems in `.github/workflows/_e2e-job.yml`. **No product code is changed and no test case is deleted** — the diff is one config value plus enabling an already-declared plugin, and almost all added lines are comments.

**1. Raise the per-shard job timeout from 45 to 60 minutes.**
Measured on the nightly sweep (run `34771337734`), the slowest shard, e2e-p1, took **37.9 minutes** against the 45 minute ceiling — only **7.1 minutes (16%)** of headroom. The E2E suite is still growing, so the ceiling is lifted well past what today's distribution needs: **60 minutes gives 22.1 minutes (37%)** of headroom.

**Effect on overall sweep duration (stated as a measurement, NOT a guarantee):** the sweep runs as long as its slowest job, which is currently the Windows integration tier — but that duration is **volatile** (measured **59.8** and **76.7** minutes on two recent nightly runs, a 16.9 minute spread), so 60 minutes sits **inside** that range rather than safely below it. On a run where the Windows tier lands at the fast end while an E2E shard approaches 60 minutes, that shard could become the slowest job. With today's distribution this does not happen: the slowest e2e-p1 job measured **39.4 minutes** (run `34941456787`), far from the ceiling, so the sweep duration is still set by the Windows tier (59.8 minutes that run; 61.6 minutes wall clock overall).

Deliberately **not** higher: at 75–90 minutes a genuinely hung shard would only be reported that late, which weakens the gate.

🔴 **60 minutes is still not sufficient on its own for the incoming batch**: if a pending E2E expansion takes this shard from 105 to 209 cases and lands **without** rebalancing, the shard measures about **63 minutes** (209 × 14.31s of test time + ~13.6 min of fixed overhead, both read out of run artifacts), i.e. **past the 60 minute ceiling**. The real fix is rebalancing cases across shards by measured duration (a test-marker change, not in this PR); that is projected to cut the slowest shard to ~29.3 minutes and raise the headroom to ~51%.

**2. Enable the per-test timeout.**
`pytest-timeout` has been a declared dependency (`e2e/requirements.txt`: `# Timeout control` + `pytest-timeout>=2.2.0`) but was **never enabled** — no `--timeout` on the pytest command, none in `e2e/pytest.ini`. So the job-level `timeout-minutes` was the **only** protection: one hung case consumed the whole shard budget and every case queued behind it silently never ran — while coverage still uploaded, the HTML report still generated, and artifacts still attached, so the run **looked complete**. Comparing "selected" against "actually executed" on such a run showed **64 of 209 cases (30.6%) had not run at all**.

The enabled flag string is:

```
PYTEST_TIMEOUT_ARGS="--timeout=480 --timeout-method=signal --rerun-except Timeout"
```

**Why a variable rather than inlining it on both command lines:** after `#7697` this step has **two** pytest invocations (an `if` branch running an explicit node-id list, and an `else` branch running by marker). Defining the flags once and referencing them twice makes it **structurally impossible to protect one branch and silently leave the other unguarded** — which is exactly the kind of blind spot this PR is here to remove.

**Evidence for each of the three flags (all measured, none inferred):**

- `--timeout=480`: from a measured per-case duration distribution of **87 cases** (median 7.70s, mean 22.35s, longest **385.9s**). 480s leaves **24% headroom** over the longest and kills **zero** genuinely passing cases. Lower values misfire: **60s kills 7 passing cases, 120s and 180s each kill 2, 300s kills 1, 480s kills 0**. The cost of a misfire is concrete — those 2 cases are deep tests (a 385.9s long-conversation-compression case and a 242.0s session case, both passing and visibly progressing in the log); cut short, they would surface as **brand-new defects in triage**, wasting a whole round.
- `--timeout-method=signal`, **not `thread`**: `thread` calls `os._exit(1)` on timeout (`pytest_timeout.py`, inside an unconditional `finally` block), which **kills the entire pytest process** — every case behind the hung one still never runs, so `thread` would **not fix the blind spot at all**. Measured side by side on four hang shapes (see Evidence): `signal` failed just the hung case and every later case still ran and reported; `thread` ended the process with zero later cases executed, in all four. `signal`'s only blind spot is a case that masks `SIGALRM` itself, and this suite never touches `signal` (`grep` over `e2e/tests` and `e2e/conftest.py`: zero hits).
- `--rerun-except Timeout`, **required, not optional**: `signal` marks the timed-out case failed, and the pre-existing `--reruns=1` would then rerun it — **but the rerun gets no timer**, so it hangs to the outer job limit (measured: that run never finishes). Excluding `Timeout` from reruns prevents that while leaving `--reruns=1` fully intact for the flakes it exists for (verified: a fail-then-pass case still reruns and passes, `2 passed, 1 rerun`).

**Related Issue:** No corresponding upstream issue; this is CI-pipeline maintenance agreed with the project maintainer.

**Security Considerations:** No channel/credential/configuration-security changes. The only externally visible behavior change is "the failure count rises", explained under Additional Notes.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Refactoring

(In substance this is CI-pipeline maintenance: one config value adjusted + enabling a plugin that was already declared but never turned on.)

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [x] CI/CD
- [ ] Scripts / Deploy

## Checklist

      — this contributor has no local test environment (a known maintainer-acknowledged constraint); tests were verified on GitHub-hosted runners via fork dispatch. Locally, only the parameter semantics and hang shapes were exercised in an isolated venv (see Testing/Evidence)
- [x] Documentation updated (if needed) — workflow inline comments: the source of every number, why the values are not higher/lower, and the `--rerun-except` dependency on the failure-message text
- [x] Ready for review

Not applicable (this PR changes no channel code).

## Testing

Verified on the fork (`yutai78786/QwenPaw`) by `workflow_dispatch` against branch `ci/e2e-timeout-and-job-limit`, entirely on GitHub-hosted runners (no tests run locally):

> **Which code version each run points at (stated to avoid misreading):** the primary evidence run `34941456787` ran the version **after the rebase onto main `107eb7f7e` but before the comment fix** (tip `90b4c9fcc`). The current final tip `cb87d0c60` differs from it **only in comments** — stripping YAML comments and comparing byte for byte, both versions' functional content is **7173 bytes and identical** (action A's `timeout-minutes: 60`, action C's flag string and both pytest invocations unchanged). So this run's results remain valid for the final version. A full re-verification run `34951372293` was dispatched on the final tip as confirmation (functional equivalence already proven by the comment-stripped comparison).

**1. Full nightly form (marker branch) — run `34941456787` (primary evidence)**

`form=nightly` + `coverage_platforms=all`, wall clock **61.6 min**, final **15 success / 3 failure / 2 skipped** (20 jobs). All four backend tiers green:

| Tier | Matrix entries | Result |
|---|---|---|
| Unit Tests | py3.11 ubuntu/macos/windows, py3.13 ubuntu | **all success (4/4)** |
| Contract Tests | py3.11 ubuntu/macos/windows, py3.13 ubuntu | **all success (4/4)** |
| Integrated Tests | py3.11 ubuntu/macos/windows, py3.13 ubuntu | **all success (4/4)**; Windows integration `1883 passed, 0 failed`, `ReadTimeout` count 0 |
| Frontend Tests (Vitest) | — | success |
| Coverage Report | — | success |

E2E shards:

| Shard | Result | Job time | collected reconciliation | `Timeout(>` | Failures vs upstream baseline |
|---|---|---|---|---|---|
| e2e-p0 | failure | 17.8 min | 238 = 3+64+4+167 ✅ | **0** | 3 cases, identical set, zero new |
| e2e-p1 | failure | 39.4 min | 238 = 6+97+2+133 ✅ | **0** | 6 cases, identical set, zero new |
| e2e-p2 | success | 11.8 min | — | — | — |

🟢 **Action C, the decisive point:** in both p0 and p1 logs the `Timeout (>...)` trigger count is **0** — **480s killed zero genuinely passing cases**; the reds are all `AssertionError` (UI-state assertions). And "trigger 0" is **genuinely safe, not a false green**: `timeout-2.4.0` is present in the `plugins:` line, `PYTEST_TIMEOUT_ARGS` appears 3 times (1 definition + 1 reference per branch), `unrecognized arguments` = 0.

🔴 **Action A, measured headroom:** the slowest e2e-p1 job took **39.4 min**, leaving **20.6 min (34%)** against the 60 min ceiling, nowhere near it; the sweep's slowest job is Windows integration at 59.8 min, so the sweep duration is set by that, independent of the E2E ceiling.

🔴 **Count reconciliation (guards against silently running fewer cases):** both shards' `collected` equals `failed+passed+skipped+deselected` exactly (238 = 238) — **zero cases silently dropped.**

🟢 **All 3 failures are inherited pre-existing reds, NOT caused by this PR:** `Test Summary`'s per-tier verdict reads `Unit/Contract/Integrated/Frontend/E2E p2 all success, only E2E p0 and p1 failure`. The 3 p0 and 6 p1 cases are **the same set, same cause, as the upstream nightly's same shards** (identical case names, zero new beyond baseline; see Evidence item 5) — UI cases needing selector maintenance, on which upstream nightly is equally red with the same counts. `Test Summary` is fail-closed, so it goes red because of those two shards.

**2. Release-gate form (node-id branch) — run `34926149777`**

This separately verifies the *other* branch: after `#7697` split the pytest invocation into `if`/`else` (explicit node-id list vs marker), **both branches must be exercised**, otherwise "only one branch verified" is itself a blind spot. That run is the 60-minute version (tip `ab0d519e4`, functionally byte-identical to the final version once comments are stripped).

| Job | Result | Note |
|---|---|---|
| e2e-release-gate (64-case blocking set) | **success** | see reconciliation below |
| e2e-release-watch (7-case report-only set) | failure | by design "report but never block"; non-blocking |

🔴 **Count reconciliation (proves it really took the node-id branch, not a silent downgrade to the marker branch):**
- `Loaded 64 explicit node ids from .github/ci/e2e-release-gate-tests.txt`
- `collected = 64` (exactly the 64 lines of the manifest file)
- `deselected` count = **0** (the marker form collects 238 then deselects a batch; the node-id form has no deselection)
- ⟹ `collected == manifest line count` with zero deselected = **hard proof it ran the explicit-list branch**
- same run: `timeout-2.4.0` present, `PYTEST_TIMEOUT_ARGS` appears 3 times, `unrecognized arguments` = 0, `Timeout(>` = 0, `60 passed, 4 skipped`

**3. Local isolated-venv tests of parameter semantics and hang shapes** (`pytest 9.1.1` + `pytest-timeout 2.4.0` + `pytest-rerunfailures 16.6.1`; the main service venv was not touched): four hang shapes × `signal`/`thread` side by side, the `signal` vs `--reruns=1` conflict, and the combined form's flake protection — all in Evidence.

## Evidence

```text
# 1) pre-commit (run on each of the two commits)
check yaml .................................................................Passed
detect private key .........................................................Passed
trim trailing whitespace ...................................................Passed
Lint GitHub Actions workflow files .........................................Passed
(RC=0, zero Failed; added lines containing non-ASCII = 0)

# 2) signal vs thread — four hang shapes side by side (isolated venv, --timeout=3 for fast observation)
shape                                   signal                         thread
sleep(600) in the test body             hung case FAILED, later PASSED process exits, 0 later cases
hang in a real fixture teardown         teardown ERROR, 3 later PASSED 0 later cases
main thread blocked in subprocess.wait() hung case FAILED, later PASSED 0 later cases
main thread blocked in C-level socket.recv()  hung FAILED, later PASSED 0 later cases
  ^ the last two are the closest analogues to a sync Playwright transport stall
    (e2e/conftest.py:204 uses playwright.sync_api)

thread mechanism (source): pytest_timeout.py line 542 os._exit(1), inside an unconditional finally block

# 3) signal alone conflicts with the pre-existing --reruns=1 (this point is necessary)
signal + no reruns                              RC=1    3s     later cases run ✅
signal + --reruns=1                             RC=124  >120s  0 later cases 🔴 hangs forever (rerun gets no timer)
signal + --reruns=1 + --rerun-except Timeout    RC=1    3s     later cases PASSED ✅  <- what this PR uses
  all four hang shapes under this combination: RC=1, returns in 3s, later cases run

# 4) flake protection is not broken (--rerun-except excludes only Timeout, not --reruns=1)
sample: one fail-then-pass case + one stable case
this PR's flag string -> R..  2 passed, 1 rerun in 1.01s   RC=0 ✅
without --reruns       -> 1 failed, 1 passed               RC=1 (proves the flake protection comes from reruns)

# 5) the 6 / 3 failures are unrelated to this PR (verbatim comparison against upstream nightly)
our fork e2e-p1, 6 failures == upstream nightly run 34771337734, same shard, 6 failures (identical case names)
  tests/test_acp.py::TestCreateAndDeleteCustomACP::test_create_and_delete_custom_acp
  tests/test_chat.py::TestChatMessageSearch::test_chat_message_search
  tests/test_cov_chat_sessions_deep.py::TestChatSessionArchiveBatchDeep::test_session_archive_batch_deep
  tests/test_cronjobs.py::TestCronjobEditAndUpdate::test_cronjob_edit_and_update
  tests/test_inbox.py::TestSidebarUnreadDot::test_sidebar_unread_dot_appears_with_seeded_event
  tests/test_sessions.py::TestSessionBatchDelete::test_session_batch_delete
both summaries identical: 6 failed, 97 passed, 2 skipped, 133 deselected, 6 rerun
e2e-p0 same method: the 3-failure set is identical, Timeout-class failures = 0

# 6) basis for 480s (per-case durations of 87 cases, by differencing, independent source = upstream run artifact app log)
median 7.70s | mean 22.35s | longest 385.9s (6.43 min)
misfires: 60s->7 | 120s->2 | 180s->2 | 300s->1 | 480s->0
the 2 cases 180s would misfire on both PASSED that night, logs show continuous output throughout (genuinely running, not hung)
```

CI links:
- https://github.com/yutai78786/QwenPaw/actions/runs/34941456787 (**primary evidence**: post-rebase 60-min version, full nightly form, 15 green / 3 red / 2 skipped; both E2E shards `Timeout(>` = 0, count reconciliation 238 = 238)
- https://github.com/yutai78786/QwenPaw/actions/runs/34926149777 (release form, node-id branch: collected=64 / zero deselected / release-gate success)
- https://github.com/yutai78786/QwenPaw/actions/runs/34951372293 (full re-verification on the final tip `cb87d0c60` after the comment fix; functional content byte-identical to the primary run once comments are stripped)

## Additional Notes

**🔴 Expected side effect, stated plainly: the failure count will rise.**
This turns invisible blind spots into visible red — **not a quality regression.** Previously one hung case would silently prevent dozens behind it from running while the whole round still reported as complete; now the hung case is named as a failure and the cases behind it still run. So nightly's failure count may go up after merge — those are cases that **should have been caught all along and were hidden.**

**🔴 Residual limit, also stated plainly: this does NOT make "cases never ran" impossible.**
Against the upstream nightly baseline distribution (slowest shard e2e-p1 at 37.9 min vs the 60 min ceiling, same source as the code comment), the slack is about **22.1 min ≈ 1326s**, i.e. **room for about 2 genuinely hung cases at 480s each.** Our fork measured that shard's job at **39.4 min** this round (slack 20.6 min / 34%) — same measurement basis across the two rounds, differing only by 1.5 min of run-to-run variance, same conclusion (~2 cases). A shard with 3 or more real hangs can still hit the 60-minute job ceiling — **the difference is the log will now name which case hung, instead of silently dropping the rest.**

🟢 **60 minutes fully absorbs a live scenario that already happened:** in a verification round, e2e-p0's attempt 1 ran **55.0 minutes as a whole job and was cancelled** (5.9 min of setup steps + the "Run E2E tests" step at 49.1 min, still unfinished; that step normally takes only 10.6 min, and the 38.5 min difference ÷ 480s ≈ 4.8 cases, closely matching "about 5 cases each stuck the full 480s and cut one by one").
- Against the old 50-minute ceiling: job total 55.0 > 50 ⟹ **hit the ceiling and was cancelled** (this is what actually happened)
- Against the 60-minute ceiling: 55.0 < 60 ⟹ **does not hit it**; the whole shard finishes and names those hung cases one by one
- 🟢 It also confirms the **480s timeout really took effect** (had it not, the step would hang to the job ceiling with no case recorded as failed)
- ⚠️ Basis note: the quantity that hits the ceiling is the **whole-job duration** (including setup: checkout / frontend build / dependency install / backend start, measured 5.9 min), **not** the test step's duration; the two differ by nearly 6 minutes, so comparisons must use the same basis

**What actually widens the headroom further is rebalancing cases across shards by measured duration** (a test-marker change, not in this PR): after rebalancing each shard is projected to drop to ~29.3 min, raising headroom to ~51% and absorbing ~3 cases. ⚠️ That rebalancing **has not landed** (upstream still measures collected 238, deselected 133 ⟹ this shard runs 105 cases), so every headroom figure in this PR is given on the **un-rebalanced** measured basis (never substituting a not-yet-happened projection for a measurement).

🔴 **A boundary reviewers must know:** even at 60 minutes, **it is still not sufficient on its own for the incoming batch** — if all 209 cases land in one shard without rebalancing, the measured projection is ~63 minutes > 60. **Raising the ceiling buys headroom; rebalancing is the real fix** — the two are not interchangeable. This is written verbatim into the code comment so later readers do not assume "the ceiling was raised, so we're safe."

**Why not split off a fourth shard (p1a/p1b):** the upstream nightly config has **4 hardcoded references** to p0/p1/p2 (the job dependency list, result printing, failure adjudication); adding a group means changing all 4 in lockstep, and missing any one is a new silent skip or false green. Also, upstream `pX` carries **priority** semantics, whereas `p1a/p1b` are duration-based buckets, not priorities — that would muddy the semantics. Rebalancing achieves the same effect at far lower cost.

**A known dependency, written into the comment:** `--rerun-except Timeout` matches on the failure-message text. Measured (at the 3s setting, for fast observation) the message is verbatim `Failed: Timeout (>3.0s) from pytest-timeout.`; under the 480s config the seconds value changes accordingly, but the word `Timeout` is emitted by the plugin as a fixed token. If `pytest-timeout` ever changes that wording, the match would fail ⟹ the hung case would be rerun, reverting to the old "hang to the job ceiling" behavior. The comment states this dependency; **no mechanical guard was added** (asserting plugin wording in CI would cost more than it is worth).

**Not verified (stated honestly):** the basis for `--timeout=480` comes from the measured per-case durations of the **p1 shard's 87 cases**; the **per-case duration distributions of p0 and p2 were not measured.** However, in the fork verification the `Timeout(>` trigger count for both p0 and p2 was **0** ⟹ no misfires on this round's actual data. If p0/p2 later gain deep cases exceeding 480s, this value must be re-checked.

**What is deliberately not changed (red-line self-check):** no product code changed, no test case deleted, `e2e/pytest.ini` not changed (chose workflow-inline so the comments travel with the code), the nightly matrix and shard count not changed, no test-marker labels changed.

---

## Appendix: the two commits (for per-commit review)

| Commit | sha | Title | numstat |
|---|---|---|---|
| Action A | `18c1fad71` | `ci(e2e): raise the per-shard job timeout from 45 to 60 minutes` | 29 ins, 1 del |
| Action C | `cb87d0c60` | `ci(e2e): enable the already-declared pytest-timeout so one hung case cannot eat the shard budget` | 61 ins, 2 del |
| Total | | touches only `.github/workflows/_e2e-job.yml` | **90 ins, 3 del** |

Both commit messages are fully English (zero non-ASCII except the signature name in the trailer), each carries the standard `Signed-off-by` trailer, and **every number's source is written into the message** (including why not `thread`, why `--rerun-except` is required, and what the residual limit is).

🔴 **Action A's message contains a retraction note** (recorded honestly, not silently edited away): an earlier revision of that message and the code comment both claimed "60 minutes still stays below the sweep's slowest job (Windows integration, 76.7 min), so the overall sweep duration is unchanged" — this round's full measurement found Windows integration took only **59.8 minutes** (a 16.9 minute swing across two runs), so that claim **does not hold.** The message now reads verbatim: `that claim was measured against a single run and does not hold in general, so it is withdrawn here.`

🟢 **The comment fix does not invalidate the existing verification:** it touches only comments, with zero functional-line change (both versions are **7173 bytes and byte-identical** once YAML comments are stripped) ⟹ the primary evidence run `34941456787` remains valid for the final version `cb87d0c60`.


cuiyuebing (approved):
LGTM

PR #7808, opened 2026-09-16
refactor(loop): pass DoomLoopStageConfig objects from catalog factory

TerebiSAMA:
## Summary

Move the validated-stages handoff into the gate catalog factory instead of widening `DoomLoopGate` to accept either dicts or objects.

## What Problem This Solves

The catalog already validates `stages` as `DoomLoopStageConfig` objects via `DoomLoopParams`, but the doom-loop factory was calling `model_dump()`, so stages arrived at the gate as plain dicts. To keep `DoomLoopGate` working with both representations, the previous version of this PR added a `_stage_field()` helper that read fields by string name through `Any`. That spread the catalog's conversion concern into detection logic and weakened the gate's object-based contract — exactly what the review flagged.

This revision takes the other path: the factory forwards the validated Pydantic object fields directly. The gate reverts to typed `list[DoomLoopStageConfig]` and plain attribute access (`stage.after`, `stage.action`, `stage.prompt`).

## Changes

- `src/qwenpaw/loop/catalog.py` — doom-loop factory now constructs the gate from `params.window_size`, `params.similarity_threshold`, and `params.stages` instead of `**_dump(params)`. Other factories are unchanged because their params models only carry scalars.
- `src/qwenpaw/loop/gates/doom_loop.py` — removed `_stage_field()` helper and the `Any`/string-name indirection; `__init__` now types `stages` as `list[DoomLoopStageConfig] | None` and reads fields via direct attribute access.
- `tests/unit/loop/test_doom_loop_gate.py` — `_stage()` now returns a real `DoomLoopStageConfig`; replaced the dict-stage regression (which exercised a path that should no longer exist) with two tests that exercise the catalog path: object-typed stages plus a stop-stage escalation that surfaces the configured `prompt` as the stop reason.

## Testing

```bash
PYTHONPATH=src pytest tests/unit/loop/test_doom_loop_gate.py -q
# 9 passed in 0.05s

PYTHONPATH=src pytest tests/unit/loop/ -q
# 66 passed in 0.78s

PYTHONPATH=src pytest tests/unit/modes/ tests/unit/loop/ tests/unit/cli/tui/ -q
# 252 passed in 21.05s
```

Lint:

```bash
black --line-length=79 --check <files>
# All done! ✨ 🍰 ✨ 3 files would be left unchanged.

flake8 --extend-ignore=E203 <files>
# (clean)
```


TerebiSAMA:
@rayrayraykk Hi, no rush on this one — I saw you have a long review queue and I really appreciate your work on doom-loop gate (especially #7748, which is the direct neighbor of this fix). Just a gentle ping in case you have a moment.

This is a small follow-up: `DoomLoopGate` was raising `AttributeError` when stages arrived as dicts from `model_dump()`, so catalog creation silently failed and the gate never engaged. The fix is a tiny `_stage_field()` helper (+40 / -5 across 2 files) plus two regression tests that now pass locally:

  ✓ test_catalog_create_supports_dict_stages
  ✓ test_dict_stage_stop_behavior
  pytest tests/unit/loop/test_doom_loop_gate.py -q → 9 passed

CC @zhijianma @zhaozhuang521 in case it is easier for one of you to take it while you are closer to the loop area. Totally fine if it has to wait — I will keep the branch in sync with main either way.

Thanks again!

rayrayraykk (commented):
Thanks for your contribution! The catalog already validates `stages` as `DoomLoopStageConfig` objects, but `model_dump()` converts them to dictionaries, forcing the gate to support two representations. `_stage_field()` spreads this conversion concern into the detection logic and weakens type checking through `Any` and string field names. Please pass the validated fields directly from the doom-loop catalog factory and keep the gate’s existing object-based contract.

TerebiSAMA:
@rayrayraykk Thanks for the steer — that was the right call. Updated to take the other path: the doom-loop catalog factory now passes `params.window_size`, `params.similarity_threshold`, and `params.stages` directly to `DoomLoopGate`, instead of spreading `model_dump()` through `**_dump(params)`. The gate reverts to typed `list[DoomLoopStageConfig] | None` and plain attribute access; `_stage_field()` and the `Any`/string-name indirection are gone.

Tests updated: `_stage()` now returns a real `DoomLoopStageConfig`, and the dict-stage regression was replaced with two catalog-path tests (object-typed stages + stop-stage escalation that surfaces the configured `prompt`). The other factories were left on `_dump()` since their params models only carry scalars — let me know if you want those standardized too, but it didn't seem worth churning them.

Local:

\`\`\`
PYTHONPATH=src pytest tests/unit/loop/test_doom_loop_gate.py -q   → 9 passed
PYTHONPATH=src pytest tests/unit/loop/ -q                          → 66 passed
PYTHONPATH=src pytest tests/unit/modes/ tests/unit/loop/ tests/unit/cli/tui/ -q
                                                                    → 252 passed
black --line-length=79 --check <files>                             → clean
flake8 --extend-ignore=E203 <files>                                → clean
\`\`\`

PR title/description updated to reflect the refactor framing. Happy to keep iterating — thanks again for the review.

rayrayraykk (approved):

Issue #7813, opened 2026-09-16
Console stream freezes when an SSE frame payload is the bare null literal


wjt0321:
## Problem

A single malformed SSE frame whose payload is the bare JSON literal `null` takes down the whole streaming turn in the Console. The exception is thrown inside the stream-consuming generator, caught by an outer handler and only logged, so nothing terminates the response client-side. The UI is left stuck in a "generating" state with no way to recover except reloading the page.

This is a small, self-contained frontend robustness gap, but the user-visible effect is a total chat freeze, so it seems worth fixing independently of where the bad frame comes from.

## Environment

- QwenPaw 2.2.1, Windows desktop build (`/api/version` returns `2.2.1`, so the release is confirmed)
- Console reached over `127.0.0.1`, default webview shell
- Reproduces across reloads; a hard refresh recovers the session

## Observed error

Three occurrences of this in a single Console session log, all with an identical stack:

```
index-<hash>.js:200 TypeError: Cannot read properties of null (reading 'object')
    at e.value (ui-vendor-<hash>.js:4194:4339)
    at ui-vendor-<hash>.js:4194:16571
    at f (ui-vendor-<hash>.js:4194:7061)
    at Generator.<anonymous> (ui-vendor-<hash>.js:4194:8393)
    at Generator.next (ui-vendor-<hash>.js:4194:7486)
    at <asyncGeneratorStep> (ui-vendor-<hash>.js:4194:12958)
    at o (ui-vendor-<hash>.js:4194:13156)
```

The error is printed by QwenPaw's own bundle (`index-<hash>.js:200` is the `console.error` call), which is why the failure is silent rather than showing an error page.

## Crash site

Column 4339 of that frame resolves to the first statement of the response builder's `handle` method, in the vendored `@agentscope-ai/chat` chunk. Pinned in `console/package.json` at `1.1.73-beta.1787638407498`, the method looks like this:

```js
function handle(data) {
  if (data.object === 'response') {
    this.handleResponse(data);
  } else if (data.object === 'message') {
    if (data.type === AgentScopeRuntimeMessageType.HEARTBEAT) return this.data;
    this.handleMessage(data);
  } else if (data.object === 'content') {
    this.handleContent(data);
  } else {
    this.handleError(data);
  }
  return this.data;
}
```

There is no null check on `data` before the first `.object` access. In the bundle this compiles to `function(n){if(n.object==="response")...}`, which is exactly the instruction the stack points at.

[...]

PR #7820, opened 2026-09-17
fix(console): reject bare null SSE event payloads

### LINKED ISSUE #7814: Console SSE: _strip_event_headlines can emit a bare null payload, and stream_one sends no terminal event on failure

kabishou11:
## Summary

Fixes #7814 (backend half of the console freeze; frontend companion is #7813).

When `_strip_event_headlines` / `_serialize_event_for_sse` encounter an event whose `model_dump(mode="json")` is `None` or whose `model_dump_json()` is the string `"null"`, the console SSE stream previously emitted:

```
data: null
```

The client does `JSON.parse` then reads `.object`, which throws and leaves the turn stuck in "generating" until reload.

## Changes

- `_strip_event_headlines`: if the dumped (or walked) payload is not a `dict`/`list`, return the original fallback string instead of `json.dumps(None)`.
- `_serialize_event_for_sse`: after serialization/strip, replace empty / bare `null` / non-object-or-array JSON with a small error object so the wire always carries a parseable object/array.

## Notes

- Open PR #7723 covers the related "no terminal event on `stream_one` failure" path; this PR is intentionally scoped to the bare-`null` envelope so the two root causes can land independently.
- No Co-authored-by / AI attribution.

## Test plan

- [x] `pytest tests/unit/channels/test_base_stream_helpers.py::TestSerializeEventRejectsBareNull -q` (3 passed)
- [x] Related `TestSerializeEventForSse` helpers still pass
- [ ] Manual: force a null-dumping event on `/api/console/chat` and confirm the stream never emits `data: null` and the UI does not freeze


zhijianma:
Thanks for picking this up, and for the scope discipline — keeping it separate
from the #7723 terminal-event work and adding regression tests rather than a
bare one-liner were both the right instincts.

After verifying it against the source and the shipped SDK, I am going to close
this rather than merge. The defensive goal is reasonable, but the path being
defended is unreachable, the two inner guards do not do what the description
says, the tests do not exercise what they claim, and the replacement payload
turns a benign frame into a hard-failed turn. Details with evidence below so
any of it can be checked or pushed back on.

## 1. The target scenario cannot occur

`_strip_event_headlines` is only entered from `_serialize_event_for_sse` and
only when `hasattr(event, "model_dump")`. Every event reaching it is a Pydantic
`BaseModel` with declared fields, so `model_dump(mode="json")` returns a dict.
Across `src/qwenpaw/`:

- no `RootModel` — `RootModel[None]` is the only Pydantic shape that dumps to
  `None`
- no hand-written `model_dump` / `model_dump_json` overrides, so no custom
  path can return `None`
- the heartbeat tick (`runtime/heartbeat.py:8`) is a bare `object()`; it has
  neither `model_dump_json` nor `json`, so it takes the `else` branch and
  yields `{"text": "<object object at 0x...>"}` — an object

The `walk()` gap you identified is real as written, but `payload` cannot be
`None` there, so nothing emits `data: null`. Related: #7814 Issue 2 is the
actual defect and is covered by #7723.

## 2. Guard #1 returns the very value it is preventing

`fallback` is the caller's `data`, i.e. `event.model_dump_json()`. In the exact
scenario this PR targets — `model_dump_json()` returning `"null"` — the fallback
passed in *is* `"null"`:

```
fallback into _strip_event_headlines = 'null'
guard #1: not isinstance(None, (dict, list)) -> return fallback
=> returns 'null'
```

The inner guard hands back the bare `null` unchanged. What actually catches it
is the outer guard in `_serialize_event_for_sse`. So the first of the two
changes is inert for its stated purpose.

## 3. Guard #2 is unreachable dead code

```python
payload = walk(payload)
if not isinstance(payload, (dict, list)):
    return fallback
```

`walk()` recurses but never changes the *top-level* type, and guard #1 has
already rejected every non-dict/list before it runs. Exercised over
`None, 1, "s", True, [], {}, {"a": None}, [None, 1]`: guard #2 never fires.

## 4. The tests do not cover what their names claim

`test_strip_with_tracked_delta_still_rejects_null` — guard #1 is inserted
*above* the delta-tracking block, so a `model_dump() -> None` event returns
before any delta logic runs. The tracked-delta path is never exercised.

`test_model_dump_json_null_is_not_emitted` — it passes no
`headline_stream_states`, so `should_strip` is `False` and
`_strip_event_headlines` is never called at all; only the outer guard fires.
Its assertion is also a tautology:

```python
assert parsed.get("object") != "null" or "object" in parsed
```

The right-hand side is true for any dict containing an `object` key, which the
substitute always does. The assertion cannot fail, so the test would still pass
against unpatched code paths that produce an object.

All three use `SimpleNamespace(model_dump=lambda mode="json": None, ...)`,
which is a shape no real event type produces — so the suite cannot catch a
regression in the behaviour that matters.

## 5. The substitute payload is a behavioural regression

`{"object": "error", "error": "invalid SSE event payload"}` is not a shape the
builder understands. In `Builder.js` `handle()` dispatches on `data.object`;
`"error"` matches none of `response` / `message` / `content`, so it falls to
`handleError()`:

```js
draft.status = Failed;
draft.output.push({ ..., code: data.code, message: typeof data.message === 'string' ? ... : JSON.stringify(data.message) });
```

`data.code` is `undefined` and `data.message` is `undefined`, so
`JSON.stringify(undefined)` is `undefined`. The result is a turn hard-marked
`Failed` with an error bubble carrying no code and no message.

That is worse than the status quo. On the currently pinned SDK
(`1.2.0-beta`, via #7382) a null/undefined parse result is skipped with
`continue` and streaming proceeds. This change would convert a droppable frame
into a visibly failed turn with an empty error card. If a terminal shape is
wanted here it needs to be one `handle()` can actually consume.

## 6. Per-frame cost on the hot path

The `json.loads` + re-dump check sits *outside* the `should_strip` gate, so it
runs for every event on the stream, including every content delta. Measured at
roughly 1.5 µs/frame (~7.6 ms of extra CPU over a 5000-token turn) — small in
absolute terms, but it is paid unconditionally to defend a case that cannot
arise, on the hottest loop in the channel.

## 7. "Fixes #7814" would close the wrong half

#7814 covers two issues. This addresses Issue 1 (the unreachable one); Issue 2
is the confirmed defect and belongs to #7723. Merging with `Fixes #7814` would
auto-close an issue whose real half is still unfixed.

---

Also noting the PR template asks for pasted `pre-commit run --all-files` and
`pytest` output, which is not in the description, and the manual verification
item is unchecked.

**Verdict: closing as not planned.** If you want to pursue hardening here, the
change that would survive review is much smaller and different in kind: give
`walk()` an explicit `None` branch returning the node unchanged, drop the
annotation confusion, and leave `_serialize_event_for_sse` alone — no per-frame
re-parse and no substitute payload. Happy to look at that as a separate PR.
The genuinely valuable fix in this area is the terminal event in #7723.

PR #1513, opened 2026-03-15
fix(crons): handle complex cron expressions with step syntax

YingchaoX:
fix(crons): handle complex cron expressions with step syntax

## Description

This PR fixes two bugs in cron expression parsing that caused complex expressions with step syntax (e.g., `9-20/2`, `1-5/2`) to be incorrectly parsed, potentially leading to data loss when users edit cron jobs.

**Related Issue:** Fixes #1510

**Security Considerations:** N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [ ] Documentation updated (if needed)
- [x] Ready for review

## Testing

1. **Frontend**: Run `npm run format` in `console/` directory
2. **Backend**: Run `PYTHONPATH=src python -m pytest tests/unit/crons/test_models.py -v`

## Local Verification Evidence

```bash
cd console && npm run format
# formatted 100+ files successfully

PYTHONPATH=src python -m pytest tests/unit/crons/test_models.py -v
# ============================= test session starts ==============================
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_star_passes_through PASSED
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_single_number PASSED
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_range PASSED
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_step_with_range PASSED
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_step_with_single PASSED
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_comma_separated PASSED
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_comma_separated_with_step PASSED
# tests/unit/crons/test_models.py::TestCrontabDowToName::test_already_named PASSED
[...]

PR #6353, opened 2026-07-22
feat(crons): support per-job model overrides

wananing:
## Description

Allow agent cron jobs to pin a model without changing the agent's persisted
active model. This reuses the existing per-request `model_slot_override`
contract instead of adding a cron-specific model construction path.

- expose and validate `model_slot_override` in `CronJobRequest`, including
  compatibility with the existing `provider:model` request form
- add `--model provider/model` to cron create/update and `--clear-model` to
  update; text jobs reject model selection
- make the shared runtime availability check honor a valid request override,
  including when no agent/global default exists
- preserve request extensions when the Console performs a full cron-job PUT
- update the English/Chinese CLI docs, cron docs, and built-in cron skills

**Related Issue:** Fixes #6316

**Security Considerations:** No credential, authorization, or persisted agent
configuration changes. The override is validated at the cron API boundary and
used only for the current request.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Lark, QQ, Discord, iMessage, etc.)
- [x] Skills
- [x] CLI
- [x] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
- [x] I ran tests locally (`pytest` or as relevant) and they pass
- [x] Documentation updated (if needed)
- [x] Ready for review

### For Channel Changes (DingTalk, Lark, QQ, Console, etc.)

- [ ] I ran `./scripts/check-channels.sh` (or `./scripts/check-channels.sh --changed`) and it passes
- [ ] **Contract test** exists in `tests/contract/channels/test_<channel>_contract.py` (REQUIRED)
- [ ] Contract test implements `create_instance()` with proper channel initialization
- [ ] All 19 contract verification points pass (see `tests/contract/channels/__init__.py`)
- [ ] **Optional**: Unit tests in `tests/unit/channels/test_<channel>.py` for complex internal logic

## Testing

- Python cron, CLI, runtime, and shared model-factory unit tests
- Console Cron Vitest suite and production build
- changed-file pre-commit hooks and Markdown Prettier check

Verification matrix:

| Surface | Verification |
[...]

PR #3255, opened 2026-04-10
feat(cron): support fresh execution sessions

cnyarx:
## Summary                                                                                                                                       
  - support fresh execution sessions for cron agent jobs while continuing to deliver output to the configured dispatch session                     
  - add executor coverage for dispatch vs `new_per_run` session behavior and fix the test imports after the `qwenpaw` package rename               
  - polish the cron jobs console by tightening the update payload contract, cleaning up remaining i18n gaps, fixing clipboard handling, and        
  aligning form typing with actual form values                                                                                                     
                                                                                                                                                   
  ## Test plan                                                                                                                                     
  - [x] Run `python tests/unit/app/test_cron_executor.py`         
  - [ ] Create and edit cron jobs from the console page                                                                                            
  - [ ] Verify `dispatch` vs `new_per_run` behavior for agent jobs                                                                                 
  - [ ] Verify text jobs still send to the configured output session                                                                               
  - [ ] Verify weekly/custom cron display, tooltip text, and copy behavior in the jobs table                                                       
  - [ ] Verify enable/disable and execute-now actions show the expected toasts                                                                     
                                                                                                                                                   
  ## Notes                                                                                                                                         
  - agent jobs now distinguish between the execution session and the dispatch/output session                                                       
  - console updates now always send the job `id` expected by `PUT /cron/jobs/{job_id}`                                                             
  - this PR also cleans up follow-up issues found while reviewing the merged cron branch against current `main`

PR #1269, opened 2026-03-11
fix(crons): normalize day_of_week to abbreviations to fix weekday mismatch

sljeff:
## Description

**What this PR does and why:**
When creating a weekly cron job (e.g., executing every Monday), the job unexpectedly runs on the wrong day (Tuesday). Every weekly schedule is consistently delayed by one day.

The root cause is a mismatch in weekday indexing: the frontend UI and standard crontab use `0=Sunday … 6=Saturday`, while the backend's APScheduler `CronTrigger` uses ISO 8601 numbering where `0=Monday … 6=Sunday`. Because the frontend sends crontab-style numbers, the backend treats them via ISO standard, causing an unavoidable one-day shift.

To solve this fundamentally and standardly, this PR updates the system to use **three-letter English abbreviations (`mon`, `tue`, …, `sun`)** for the 5th cron field (day-of-week). **Using these abbreviations is a universally supported standard syntax that is unambiguous in both crontab and APScheduler**.
- The backend [ScheduleSpec](cci:2://file:///Users/kindj/proj/CoPaw/src/copaw/app/crons/models.py:56:0-83:9) validator now auto-converts any incoming numeric day-of-week values (including lists and ranges) to their corresponding unambiguous abbreviations. 
- The frontend UI has also been updated to directly generate these abbreviation strings instead of numeric values.

**Related Issue:** 
This is caused by a well-known, long-standing upstream issue in `APScheduler` where `CronTrigger` and `from_crontab` handle weekday numbers inconsistently:
- Refer to [APScheduler Issue #286](https://github.com/agronholm/apscheduler/issues/286), [APScheduler Issue #411](https://github.com/agronholm/apscheduler/issues/411), and [APScheduler Issue #931](https://github.com/agronholm/apscheduler/issues/931).

**Security Considerations:** None

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core / Backend (app, agents, config, providers, utils, local_models)
- [x] Console (frontend web UI)
- [ ] Channels (DingTalk, Feishu, QQ, Discord, iMessage, etc.)
- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --all-files` locally and it passes
- [x] If pre-commit auto-fixed files, I committed those changes and reran checks
[...]

PR #7825, opened 2026-09-17
fix(crons): expand numeric DOW steps/ranges to crontab weekday names

kabishou11:
## Summary

APScheduler `CronTrigger` uses ISO weekdays (`0=Mon`), while crontab uses `0=Sun`. `_crontab_dow_to_name` previously rewrote tokens in place, which:

1. Left `*/2` on the ISO calendar → Mon/Wed/Fri/Sun instead of crontab Sun/Tue/Thu/Sat
2. Turned Sunday-anchored ranges like `0-5` / `0-6/2` into `sun-fri` / `sun-sat/2`, which APScheduler rejects (`min > max`)

Numeric DOW fields are now expanded to an explicit named weekday list (crontab origin). Pure named expressions still pass through. Full `0-6` collapses to `*`.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/unit/app/crons/test_models.py -q`
- [x] `PYTHONPATH=src python -m pytest tests/unit/app/crons/ -q` (164 passed)
- [x] New cases cover `*/2`, `0-5`, `0-6/2`, `1-5/2`, and APS trigger construction

## Gate notes

- Self-dig; no open issue for this DOW step/range defect
- Related historical: #1269 (named DOW), closed #1513 (step rewrite that left the APS-invalid forms)
- Open PRs touching `models.py` for unrelated features (#6353 model override, #3255 execution session) — no overlap with `_crontab_dow_to_name`


lalaliat (changes requested):
Thanks for the fix. The incorrect weekdays for */2 and trigger construction failure for 0-5 are real issues, and expanding expressions into explicit weekday lists makes sense. However, I reproduced two regressions:
1. fri-sun/1 works in the previous version, but the new parser converts its endpoints to 5-0 and rejects it as a reversed range. I also tested saving the job with the old code and loading it with the new code: this causes validation of the entire jobs.json file to fail, preventing CronManager from starting and registering other valid jobs. Please distinguish numeric ranges from named ranges instead of treating digits in the step as evidence of numeric weekdays.
2. 1-7 and 5-7 previously worked but now fail validation because 7 is converted to 0 before range expansion. Please preserve 7 while expanding the range and applying the step, then map the results to weekday names and deduplicate.
Please add regression tests for these boundaries and for loading jobs saved by the previous version before merging. Thanks again!

kabishou11:
Thanks for the careful review — addressed both regressions:

1. **Named vs numeric DOW ranges** — Classification is now per token. A digit in the step no longer forces numeric expansion, so legacy forms like `fri-sun/1` pass through unchanged. Loading an old `jobs.json` that contains such schedules no longer fails whole-file validation / CronManager startup.
2. **Preserve `7` through expand** — Numeric `7` is kept through range/step expansion, then mapped to `sun` with dedupe (`1-7` → `mon…sun`, `5-7` → `fri,sat,sun`, `0-7` → `*`).

Added regression tests for these boundaries and for loading jobs saved by the previous version. All `tests/unit/app/crons/test_models.py` checks are green.

PR #7826, opened 2026-09-17
feat(console): configure custom providers in one dialog

zhijianma:
## Summary
- collect provider ID, display name, Base URL, protocol, and API key in the custom-provider dialog
- save provider credentials without automatically discovering or adding models
- preserve retry behavior when provider creation succeeds but configuration fails

## Validation
- `vitest`: 3 targeted files, 59 tests passed
- targeted ESLint passed
- targeted pre-commit checks passed
- `git diff --check` passed


Osier-Yi (approved):
LGTM

PR #7831, opened 2026-09-17
fix(console): stream background tool output on demand

zhijianma:
## Summary

- track background tool lifecycle with status polling without eagerly opening output SSE
- open the per-tool output stream only while its task row is expanded and abort it on collapse
- keep cancel, session cleanup, transient-error handling, and terminal-state convergence intact

## Why

Background tool calls previously opened one long-lived output stream per task as soon as they were registered. This keeps unnecessary connections alive when output is not visible.

## Validation

- `npm run test:run -- src/hooks/useBackgroundTaskWatcher.test.ts src/pages/Chat/components/BackgroundTaskPanel.test.tsx` (47 passed)
- targeted ESLint on the four changed console files
- targeted Prettier check on the four changed console files
- `git diff --check`

## Known limitation

The current ToolStream has no replay buffer. Output emitted while the row is collapsed is not replayed during execution; the final result is fetched from `/output` when the task completes.


cuiyuebing (approved):
LGTM

PR #7835, opened 2026-09-17
fix(memory): stop leaking auto-memory-recall payload to channels

TerebiSAMA:
## What Problem This Solves

The auto-memory-recall pipeline injects a synthetic "I will use the memory_search tool…" trace into the LLM context. The trace was implemented as a `ToolCallBlock` + `ToolResultBlock` pair that the console frontend hides via `MemorySearchCard`, but every other outbound channel (Feishu, and any future channel) had no equivalent short-circuit — so users saw a synthetic "tool interaction" on their timeline, and some models copied that style out loud in the next reply ("I'll check memory for relevant context…").

This PR replaces that mechanism with a single hidden `TextBlock` flagged via a new `AUTO_MEMORY_SEARCH_BLOCK_HIDDEN_KEY` metadata key, plus three layers that cooperate to keep the payload off the wire (middlewares → LLM context, renderer → channel output, frontend card → console timeline).

## Evidence

This section proves the fix works end-to-end on a live recall flow, both at the unit-test level and at the channel-output level.

A recall session was traced manually before and after this change. Before the fix, Feishu users saw a synthetic "memory_search" tool interaction on the timeline and the model occasionally imitated the tool call out loud. After the fix, the recall content is silently prepended to the LLM context as `<auto_memory_recall>...</auto_memory_recall>` and consumed internally — never rendered to the user.

### Local verification evidence

- `tests/unit/agents/test_memory_middleware.py` + `tests/unit/agents/memory/` + `tests/unit/app/channels/test_renderer.py`: **313 passed, 12 deselected** (deselected tests are pre-existing baseline failures unrelated to this PR).
- `tests/unit/app/channels/`: **718 passed**.
- `tests/integration/test_memory_module.py` + `test_memory_context.py`: **44 passed**.
- Defense-in-depth e2e scenario script covers 5 scenarios (recall msg, mixed stream, old tool_call/result, missing-flag fallback, bare-object fallback). 18/20 assertions pass; the 2 "failures" are test-script artifacts using the wrong message model (`agentscope.Msg` instead of `qwenpaw.schemas.Message`), confirmed by re-running against the correct model.
- `prettier --check .` against the whole console tree: **clean**.

### After-fix evidence

Manual channel trace before/after:

Before — Feishu user receives:
```
[I will use the memory_search with the user's query as the search keywords, request up to 5 results.]
memory_search  query=… max_results=5  ✔
[Recall results text leaks into the next assistant reply if the model imitates the tool call]
```

After — Feishu user receives only the assistant reply, with the recall content silently prepended to LLM context as `<auto_memory_recall>...</auto_memory_recall>` and consumed internally. No synthetic tool interaction visible to the user.

### Root cause

`BaseMemoryManager._build_auto_memory_search_msg` constructed the recall as 4 blocks (text, thinking, tool_call, tool_result). The synthesized `thinking` is what motivated the leak — models copy that style out loud. Removing it removes the surface. The block list is now a single `TextBlock` whose only purpose is to carry context into the next LLM turn.

## Changes

| File | Purpose |
|---|---|
| `src/qwenpaw/constant.py` | introduce `AUTO_MEMORY_SEARCH_BLOCK_HIDDEN_KEY`; remove unused `AUTO_MEMORY_SEARCH_TEXT` / `_THINKING_PREFIX` |
| `src/qwenpaw/agents/memory/base_memory_manager.py` | emit single hidden TextBlock instead of tool_call/result pair |
| `src/qwenpaw/agents/middlewares.py` | `_is_auto_memory_recall_msg` helper; memory-extraction filter accepts the new flag |
| `src/qwenpaw/app/channels/renderer.py` | `_is_auto_memory_recall_message` helper; `message_to_parts` short-circuits to `[]` for recall msgs |
| `console/.../MemorySearchCard.tsx` | `isAutoMemorySearchHidden` returns `null` for flagged blocks (console defense-in-depth) |

5 files, +93 / -54 lines. Backward-compatible: messages without the flag render exactly as before.


TerebiSAMA:
Hi @jinliyl — just a quick note to let you know this is still on my radar whenever your schedule allows.

The branch currently sits at `9186d1f` with all PR checks green (CodeQL clean on both Python and TypeScript, no new alerts). I have not pushed any new commits since opening, so the diff remains the same one you were reviewing.

If anything in the description or branch name (`fix/doom-loop-dict-stages`, which predates the broader scope) makes the review harder to start, I'm happy to:

- split this into the doom-loop gate change and the memory-leak fix as two separate PRs, or
- rebase onto a newer `main` (currently based on `dd5a7e60` from 09-17), or
- tighten the title/description to focus only on the auto-recall payload leak.

No urgency at all — just flagging the offer so the time you spend on it lands as efficiently as possible. Thanks again for your time.

TerebiSAMA:
Hi @lalaliat — sorry for the extra nudge, could you please help take another look at this one?

Context:
- Review was requested from @jinliyl on 09-18 and I left a polite ping on 09-20, but there's been no response, so the PR is still stuck at `REVIEW_REQUIRED` with `AI Review Approval` / `Maintainer Approval` pending.
- Head is `9186d1f`; CI is green across the board (Pre-commit, CodeQL for both Python and TypeScript, Vitest, Playwright, format checks, etc.). The only red is `Release Window` ("merges frozen"), which is the repo-wide release freeze and unrelated to this diff.
- The diff stays small: 5 files, +99/−61 — it stops the auto-memory-recall payload from being leaked into channel messages.

Would you mind either reviewing it directly or re-routing it to another available reviewer? Thanks a lot! 🙏

lalaliat:
> Hi @lalaliat — sorry for the extra nudge, could you please help take another look at this one?
> 
> Context:
> 
> * Review was requested from @jinliyl on 09-18 and I left a polite ping on 09-20, but there's been no response, so the PR is still stuck at `REVIEW_REQUIRED` with `AI Review Approval` / `Maintainer Approval` pending.
> * Head is `9186d1f`; CI is green across the board (Pre-commit, CodeQL for both Python and TypeScript, Vitest, Playwright, format checks, etc.). The only red is `Release Window` ("merges frozen"), which is the repo-wide release freeze and unrelated to this diff.
> * The diff stays small: 5 files, +99/−61 — it stops the auto-memory-recall payload from being leaked into channel messages.
> 
> Would you mind either reviewing it directly or re-routing it to another available reviewer? Thanks a lot! 🙏

Sure. Thank you for your contribution! We'll review this PR ASAP.

PR #7760, opened 2026-09-14
fix(cli): allow memory jobs to drain on shutdown

jinliyl:
## Summary

- stop workspaces immediately after portability imports quiesce so ReMe memory drains are not delayed by unrelated cleanup
- give the early bounded shutdown path a 12-second process grace period
- start Windows desktop backends in a process group and send CTRL_BREAK_EVENT before force termination
- carry one graceful deadline through Windows signaling and waiting to avoid stacked timeouts

## Tests

- `pytest tests/unit/cli/test_cli_shutdown.py tests/unit/cli/test_shutdown_cmd_helpers_extra.py tests/unit/agents/memory/test_base_memory_manager.py tests/unit/agents/memory/test_reme_close.py -q` (67 passed)
- `pre-commit run --files` on changed files

PR #7863, opened 2026-09-18
fix(ci): stabilize Windows reload and snapshot tests

jinliyl:
## Summary

- Set `app_dir` in the Windows Uvicorn reload integration test so the spawned child can import `reload_app` from the watched directory.
- Pin LF checkout for the canonical data-console patch and bridge plus the snapshot's text assets and manifests, preserving the checked-in SHA-256 provenance on Windows.

## Why

The Windows Python 3.11 release gate in [run 35333518090](https://github.com/agentscope-ai/QwenPaw/actions/runs/35333518090/job/105564424910) failed in two places. The reload test added by #7760 could not import its test app before reaching shutdown assertions. The data-console snapshot verifier saw a patch digest mismatch because Git's Windows checkout converted its line endings while `BUILD_INFO` stores the LF digest.

## Verification

- `python -m pytest tests/unit/pawapp/test_qwenpaw_data_console_snapshot.py tests/unit/cli/test_app_cmd.py -q` — 13 passed, 2 Windows-only skipped on macOS.
- `pre-commit run --files .gitattributes tests/unit/cli/test_app_cmd.py` — passed.
- Uvicorn reload smoke test with an app in a sibling `reload/` directory — child imported the app and started its lifespan.
- `git -c core.autocrlf=true checkout-index` simulation — patch and bridge checkout bytes retained LF and matched `BUILD_INFO` hashes.

The Windows-only shutdown-event behavior awaits the Windows CI run on this PR.


cuiyuebing (approved):
LGTM

PR #7869, opened 2026-09-18
fix(providers): send OpenCode session header

### LINKED ISSUE #7599: [Bug]: 今天在使用opencode go 套餐中的模型时, 一直出现 "MissingSessionID"
## QwenPaw Version

v2.2.0

## Description

在使用opencode go 套餐中的模型时, 一直提示
模型连接测试失败：API error when connecting to model 'omen-alpha' (status=400): status=400: Error code: 400 - {'type': 'error', 'error': {'type': 'MissingSessionID', 'message': 'Error from provider (Console Go): Request is missing x-opencode-session and cannot be routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it'}}。

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** v2.2.0
- **OS:** WSL Ubuntu 22.04 in Windows 11
- **Install method:** pip 
- **Python version (if applicable):** 3.13

wananing:
## Description

QwenPaw offers the OpenCode Go endpoint as a first-class choice — `provider_catalog.py` lists `{"label": "OpenCode Go", "value": "https://opencode.ai/zen/go/v1"}` under `PROVIDER_OPENCODE.meta.base_url_options` — but never sends the header that endpoint requires, so choosing it cannot work at all. Every request comes back as:

```
status=400: Error code: 400 - {'type': 'error', 'error': {'type': 'MissingSessionID',
'message': "Error from provider (Console Go): Request is missing x-opencode-session
and cannot be routed efficiently."}}
```

`OpenCodeProvider` inherited `OpenAIProvider._build_default_headers()`, which returns only the user's `custom_headers`. Those headers feed both request paths, so with a stock configuration neither one carries a session id:

- `OpenAIProvider._client()` — used by `check_connection` and `check_model_connection`, which is the "模型连接测试失败" path in the report;
- `OpenAIProvider.get_chat_model_instance()` — `default_headers` on the chat model, forwarded as `extra_headers` on every generation call by `OpenAIChatModelCompat._call_api`.

This PR overrides `_build_default_headers()` on `OpenCodeProvider` so a session id is attached. OpenCode's documentation asks for "a stable session ID in `x-opencode-session` for each conversation so we can optimize routing and prompt caching", and the implementation follows that wording:

- **A fresh id per headers build, not per request.** `_build_default_headers()` runs once when a client or a chat model instance is constructed, and the dict is reused for that object's lifetime. QwenPaw builds a model instance per agent build and the runtime builder builds per `session_id`, so the value stays stable across one conversation and differs between conversations. Minting per request would defeat the prompt caching the header exists for.
- **The user always wins.** If `custom_headers` already carries `x-opencode-session` in any capitalisation, it is left untouched and no generated id is added.
- **No base_url string matching.** The header is harmless on the non-Go endpoint and the same documentation recommends it there for caching, so it is sent regardless of the selected endpoint. Matching on URL substrings would break on a trailing slash or a proxy in front of the endpoint.
- **Scoped to one provider.** `OpenAIProvider` and every other subclass are untouched; a regression test pins that the header does not leak into a plain OpenAI provider.

**Related Issue:** Fixes #7599

**Security Considerations:** The header value is a randomly generated UUID hex with no user, account, or credential material in it. No credential handling changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

Unit tests only — no OpenCode account or network access needed.

`tests/unit/providers/test_opencode_provider.py` gains a `TestOpenCodeSessionHeader` class covering both request paths (the `_client()` used by connection checks, and the chat model asserted at the `extra_headers` boundary where `OpenAIChatModelCompat` forwards them), stability of the id within one built object and across repeated calls of one chat model, distinct ids for separate builds, preservation of a user-supplied header, presence on the default endpoint as well, and absence on a plain `OpenAIProvider`.

## Evidence

With the fix applied:

```
$ pytest tests/unit/providers/test_opencode_provider.py -q
...................                                                      [100%]
19 passed in 0.50s
```

Same tests with only `src/qwenpaw/providers/openai_provider.py` reverted to `main`, so the tests are shown to pin the fix rather than themselves:

```
$ git stash push src/qwenpaw/providers/openai_provider.py
$ pytest tests/unit/providers/test_opencode_provider.py -q
FAILED tests/unit/providers/test_opencode_provider.py::TestOpenCodeSessionHeader::test_connection_client_sends_a_session_header
FAILED tests/unit/providers/test_opencode_provider.py::TestOpenCodeSessionHeader::test_chat_model_sends_a_session_header
FAILED tests/unit/providers/test_opencode_provider.py::TestOpenCodeSessionHeader::test_chat_model_session_is_stable_across_calls
FAILED tests/unit/providers/test_opencode_provider.py::TestOpenCodeSessionHeader::test_session_id_is_stable_within_one_built_object
FAILED tests/unit/providers/test_opencode_provider.py::TestOpenCodeSessionHeader::test_separate_builds_get_separate_sessions
FAILED tests/unit/providers/test_opencode_provider.py::TestOpenCodeSessionHeader::test_header_is_sent_on_the_default_endpoint_too
6 failed, 13 passed in 0.76s
```

Surrounding regression scope:

```
$ pytest tests/unit/providers -q --timeout=180
669 passed in 18.13s
```

Changed-file hooks:

```
$ pre-commit run --files src/qwenpaw/providers/openai_provider.py tests/unit/providers/test_opencode_provider.py
```

## Additional Notes

The root cause and this exact scope were posted on #7599 before implementation; the issue had no assignee and no linked PR at that time.


wangfei010313:
Thanks, the fix works — I verified it locally. 
Two things before merge:
1. The id is per request, not per conversation. The agent and model are rebuilt for every user turn (workspace.py:458-459 → runtime.py:105-108 → builder.py:457-458 → model_factory.py:2273-2274, no model cache), so each turn sends a new x-opencode-session — the routing/prompt-caching benefit the docstring and PR description claim isn't actually achieved. Either thread ctx.session_id in (a ContextVar next to agent_thinking_level at provider.py:73 would fit nicely, keeping _build_default_headers() pure), or keep the implementation and fix the wording to "per built client/model instance" plus a follow-up issue.
2. Please attach one live smoke test. The tests are mock-level, and this also touches the default Zen endpoint that works today. A quick check with a valid key: Go +x-opencode-session → 200 (and 400 without), Zen +header → unchanged. Note that with an invalid key both endpoints return 401 before the session check, so MissingSessionID can only be observed with a real key. Also, the failing path in #7599 is check_model_connection (chat completion), not the models.list call in check_connection — /models returns 200 without the header, so please adjust that claim.
Minor, non-blocking: an empty user value ({"x-opencode-session": ""}, which the console does persist) suppresses the generated id and still 400s; test_session_id_is_stable_within_one_built_object compares the same expression twice so it can't fail, and nothing covers the conversation-level stability you describe in the rationale.

wananing:
Thank you — point 1 is correct and I had it wrong. Pushed `a4bfac08d`.

### 1. Per-request id — fixed, and my claim was wrong

I verified your trace: `Runtime.run` constructs a new `AgentBuilder` and calls `builder.build(ctx)` on every request (`runtime.py:105-108`), `build` calls `build_model` (`builder.py:457-458`) which reaches `create_model_and_formatter` → `provider.get_chat_model_instance()` (`model_factory.py:2273-2274`), and nothing caches the model. So a new model instance, and therefore a new header value, per user turn. My "stable within one conversation" rationale was based on scroll components being keyed by `session_id`, which says nothing about model lifetime — that was my error, not a documentation gap.

I took your first option. `provider.py` now has a `_MODEL_SESSION_ID` ContextVar next to `_AGENT_THINKING_LEVEL`, with `model_session_scope()` and `get_model_session_id()`. `model_factory` grew one `_model_build_scope(thinking_level)` helper that enters `agent_thinking_level` and `model_session_scope(get_current_session_id() or "")` together, and the three `agent_thinking_level` build sites now use it, so fallback models are covered too. `ContextVarsSetupHook` runs at `PRE_DISPATCH`, before the build, so the session id is populated by then.

`_build_default_headers()` stays pure — it reads the ContextVar and derives `uuid5(fixed_namespace, session_id)`. That keeps the value identical across every turn of one conversation and across process restarts, while not putting QwenPaw's own session id on the wire. With no conversation in context (connection tests, probes) it falls back to a one-off `uuid4`.

### 2. Claims corrected

You are right that `/models` is not the failing call. I have removed the `check_connection` claim from the PR description; the body now says the failing path is the chat completion behind `check_model_connection` and every later generation request. `_client()` still carries the header, and there is still a test for it, but it is no longer described as a reproduction of the reported failure.

### 3. Minor points — all fixed

- Empty user value: an entry whose value is blank after `strip()` is now treated as unset. Any case variant of the key is removed before the generated id is added, so the request cannot end up carrying both. New test: `test_empty_user_supplied_header_is_replaced`.
- `test_session_id_is_stable_within_one_built_object` was indeed tautological and is gone. It is replaced by `test_every_turn_of_one_conversation_shares_one_id`, which builds a model, issues a call, rebuilds inside the same `model_session_scope`, and asserts both requests carry the same header — that is the property the rationale claims, asserted at the `extra_headers` boundary where `OpenAIChatModelCompat` forwards them.
- Also added: `test_separate_conversations_get_separate_ids`, `test_conversation_id_is_not_the_qwenpaw_session_id`, `test_outside_a_conversation_a_one_off_id_is_used`, and `tests/unit/agents/test_model_build_session_scope.py` for the threading itself.

`tests/unit/providers` and `tests/unit/agents`: 3833 passed. Three failures in that run (`test_acp_available_commands.py` ×2, `test_delegate_external_agent_state.py::test_formats_epoch`) are pre-existing on this machine and reproduce with all of my changes stashed.

### 4. Live smoke test — I need help here

I do not have an OpenCode account, so I cannot run the matrix you asked for, and I would rather say so than hand you a mock and call it live. What I can state is exactly what is asserted and where; everything else about the endpoint's behaviour in this PR now comes from your local verification and from OpenCode's documentation, not from me.

If you or the reporter can run it against a valid key, the matrix worth capturing is:

1. Go endpoint, header present → 200; header removed → 400 `MissingSessionID`.
2. Go endpoint, two turns of one conversation → identical `x-opencode-session`; a new conversation → a different one.
3. Default Zen endpoint with the header → unchanged behaviour.

If you would rather not merge without that evidence attached, I am happy to leave this open until someone can supply it.

wananing:
Thank you — point 1 is correct and I had it wrong. Pushed `a4bfac08d`.

### 1. Per-request id — fixed, and my claim was wrong

I verified your trace: `Runtime.run` constructs a new `AgentBuilder` and calls `builder.build(ctx)` on every request (`runtime.py:105-108`), `build` calls `build_model` (`builder.py:457-458`) which reaches `create_model_and_formatter` → `provider.get_chat_model_instance()` (`model_factory.py:2273-2274`), and nothing caches the model. So a new model instance, and therefore a new header value, per user turn. My "stable within one conversation" rationale was based on scroll components being keyed by `session_id`, which says nothing about model lifetime — that was my error, not a documentation gap.

I took your first option. `provider.py` now has a `_MODEL_SESSION_ID` ContextVar next to `_AGENT_THINKING_LEVEL`, with `model_session_scope()` and `get_model_session_id()`. `model_factory` grew one `_model_build_scope(thinking_level)` helper that enters `agent_thinking_level` and `model_session_scope(get_current_session_id() or "")` together, and the three `agent_thinking_level` build sites now use it, so fallback models are covered too. `ContextVarsSetupHook` runs at `PRE_DISPATCH`, before the build, so the session id is populated by then.

`_build_default_headers()` stays pure — it reads the ContextVar and derives `uuid5(fixed_namespace, session_id)`. That keeps the value identical across every turn of one conversation and across process restarts, while not putting QwenPaw's own session id on the wire. With no conversation in context (connection tests, probes) it falls back to a one-off `uuid4`.

### 2. Claims corrected

You are right that `/models` is not the failing call. I have removed the `check_connection` claim from the PR description; the body now says the failing path is the chat completion behind `check_model_connection` and every later generation request. `_client()` still carries the header, and there is still a test for it, but it is no longer described as a reproduction of the reported failure.

### 3. Minor points — all fixed

- Empty user value: an entry whose value is blank after `strip()` is now treated as unset. Any case variant of the key is removed before the generated id is added, so the request cannot end up carrying both. New test: `test_empty_user_supplied_header_is_replaced`.
- `test_session_id_is_stable_within_one_built_object` was indeed tautological and is gone. It is replaced by `test_every_turn_of_one_conversation_shares_one_id`, which builds a model, issues a call, rebuilds inside the same `model_session_scope`, and asserts both requests carry the same header — that is the property the rationale claims, asserted at the `extra_headers` boundary where `OpenAIChatModelCompat` forwards them.
- Also added: `test_separate_conversations_get_separate_ids`, `test_conversation_id_is_not_the_qwenpaw_session_id`, `test_outside_a_conversation_a_one_off_id_is_used`, and `tests/unit/agents/test_model_build_session_scope.py` for the threading itself.

`tests/unit/providers` and `tests/unit/agents`: 3833 passed. Three failures in that run (`test_acp_available_commands.py` ×2, `test_delegate_external_agent_state.py::test_formats_epoch`) are pre-existing on this machine and reproduce with all of my changes stashed.

### 4. Live smoke test — done

I did have access to a key after all (my earlier note that I could not run this was wrong, and I deleted it rather than leave it standing). The account's plan covers the Go endpoint, so the Go half of your matrix is fully covered; the Zen half is limited, see below.

**A/B through QwenPaw itself**, same call, same account, only the source tree differs — `check_model_connection("minimax-m3")` against `https://opencode.ai/zen/go/v1`, with an `httpx.AsyncClient.send` spy recording what actually went on the wire:

`main` (`549a7f3c4`, no fix):

```
success: False
message: API error when connecting to model 'minimax-m3' (status=400): status=400:
         Error code: 400 - {'type': 'error', 'error': {'type': 'MissingSessionID',
         'message': 'Error from provider (Console Go): Request is missing x-openc...
x-opencode-session sent: [None]
```

This branch (`a4bfac08d`):

```
source: .../src/qwenpaw/providers/openai_provider.py
has session override: True

== Go endpoint, conversation A, turn 1 ==      result: True
== Go endpoint, conversation A, turn 2 ==      result: True     (model rebuilt)
== Go endpoint, conversation B ==              result: True
== Zen endpoint, conversation A ==             result: False    (see below)

-- x-opencode-session actually sent, per request --
   https://opencode.ai/zen/go/v1/chat/completions -> a6f3752ca6b85a12a013183755014aad
   https://opencode.ai/zen/go/v1/chat/completions -> a6f3752ca6b85a12a013183755014aad
   https://opencode.ai/zen/go/v1/chat/completions -> 80415b9ce89b5f398a5c7103ea7abb6a
   https://opencode.ai/zen/v1/chat/completions    -> a6f3752ca6b85a12a013183755014aad
```

So the reported failure reproduces live on `main` and is gone on this branch; the two turns of conversation A carry one id even though the model is rebuilt between them, and conversation B gets a different one. That is the property your point 1 was about, now asserted against the real endpoint rather than a mock.

**Raw endpoint matrix** (curl, same key):

| Request | Result |
|---|---|
| `POST /zen/go/v1/chat/completions`, no `x-opencode-session` | `400` `MissingSessionID` |
| same, with the header | `200`, normal completion |
| `GET /zen/v1/models`, no header | `200` |
| `GET /zen/go/v1/models`, no header | `200` |

The `/models` rows confirm your correction: `check_connection` is not the failing path, and the PR description no longer claims it is.

**What I could not cover.** This account's plan is Go-only — on `/zen/v1/chat/completions` every request returns `401 CreditsError` regardless of the header (verified with both `claude-fable-5` and `minimax-m3`, with and without it). So I can show that adding the header changes nothing on the Zen endpoint, but I cannot show a `200` there. If that gap matters for merge, someone with a standard Zen plan would need to run that one row.

One incidental finding worth recording: the credits check runs before the session check, so an unfunded key returns `401 CreditsError` instead of `400 MissingSessionID`. Anyone reproducing #7599 needs a funded account, not just a valid key.

wangfei010313:
Thanks — I re-checked a4bfac08. The thread hop was the thing I most wanted verified, and it holds: run_sync_io uses asyncio.to_thread, which copies the context, so get_current_session_id() is visible inside the worker thread. 
Two things left:
1. One build site is still outside the scope. model_factory.py:2332 ("Fallback to global active model") calls provider.get_chat_model_instance(...) without _model_build_scope. Repro: with scoped_session_id("session-a") and an agent config whose active_model is None, two builds produce two different ids (e9cf…, c55d…); the same config through the agent-scoped branch is stable. That branch runs whenever the agent has no model slot and the model comes from the global active model, so the per-turn change is still there for those users.Wrapping that call in _model_build_scope(settings.thinking_level) fixes it.
2. test_conversation_id_is_not_the_qwenpaw_session_id can't fail, since a 32-char hex value can never contain "session-a" — pin the shape ([0-9a-f]{32}) or the uuid5 equality instead. And the two test_acp_available_commands + test_formats_epoch failures you list pass here; the failures I see in tests/unit/agents (offloader cleanup x4, lsp pool key) reproduce at base 549a7f3c, so both sets look environment-only.

wananing:
Pushed 9e9db2a1c.

1. The global-fallback build (`model_factory.py:2333`) now runs inside `_model_build_scope(settings.thinking_level)`, so all four `get_chat_model_instance` call sites in the factory are under the scope. Regression test `test_every_build_branch_sees_the_conversation` is parametrized over `active_model=None` (global fallback) and an agent model slot, and asserts two builds inside one `scoped_session_id` see the same session; the global case fails without the change.

2. `test_conversation_id_is_not_the_qwenpaw_session_id` is replaced by `test_conversation_id_is_an_opaque_uuid5_of_the_session`, pinning `[0-9a-f]{32}` and equality with `uuid5(OpenCodeProvider._SESSION_NAMESPACE, session_id).hex`.

Live check on the Go endpoint (`minimax-m3`): the same `check_model_connection` call through QwenPaw inside `model_session_scope`, with an httpx spy on the outgoing header.

```
main 549a7f3c4:  400 MissingSessionID, x-opencode-session: None
this branch:     200
  conversation A turn 1 -> a6f3752ca6b85a12a013183755014aad
  conversation A turn 2 -> a6f3752ca6b85a12a013183755014aad   (model rebuilt)
  conversation B        -> 80415b9ce89b5f398a5c7103ea7abb6a
```

curl: Go without the header 400 MissingSessionID, with it 200; GET /models 200 without the header on both endpoints; Zen chat completion response identical with and without the header. The PR description no longer describes `/models` as the failing path.

Agreed the remaining failures are environment-only on both sides; dropped from the evidence.

wangfei010313 (approved):
LGTM

wananing:
Rebased onto `main` after #7899, which added the generic `session_header_name` mechanism for inference requests. This PR is now reduced to the part `main` still lacks: `check_connection` / `check_model_connection` build their client through `_client()`, which does not attach the session header, so the model connection test against OpenCode Go still returns `400 MissingSessionID`. New head `1708ffe76` (+18/-1 in `openai_provider.py`, tests +84); the previous ContextVar and `model_factory` changes are dropped in favour of `adapters/request_context.session_header()`. Title and description updated to match.

PR #7885, opened 2026-09-19
fix(agents): retry after unsupported file payload errors

### LINKED ISSUE #7883: [Bug] Still reproducible on 2.2.1: tool-returned PDF is serialized as an OpenAI-style nested file part, DeepSeek rejects it with 400 "file must have a file_id or file_data"
> This issue was drafted with the assistance of an AI agent, based on a live reproduction on the user's machine. All IDs, timestamps and log excerpts below are from that run.

## Summary

#7597 was closed as fixed in #7621, and the 2026-09-08 comment on that issue already reported that the fix only covers **text-only** endpoints, not **multimodal** ones. This is a fresh, independent reproduction of that unresolved variant on the **stable 2.2.1 release**, with the exact content-part shape captured, and with the additional finding that the failing part shape also **disables QwenPaw's own media-capability fallback**, so every retry fails identically.

## Environment

| item | value |
| --- | --- |
| QwenPaw | `2.2.1` (Windows desktop build, `qwenpaw --version` → `QwenPaw, version 2.2.1`) |
| AgentScope | `2.0.7.post1` |
| Provider / endpoint | `deepseek`, `base_url = https://api.deepseek.com` (OpenAI-compatible `/chat/completions`, no proxy/gateway) |
| Model | `deepseek-v4-flash` (provider metadata `supports_multimodal=true`, `supports_image=true`) |
| Channel | `console`, agent `default` |
| Triggering tool | `send_file_to_user` (a local PDF on `E:\Desktop`) |
| History store | `C:\Users\Administrator\.qwenpaw\workspaces\default\history.db` |

## Repro

1. Mark the active model as multimodal (`supports_multimodal=true` / `supports_image=true` — the capability upstream #7717 would enable; in our run it was enabled via the local `model_catalog.json` / `model_capabilities.json`).
2. Have the agent deliver a **PDF** with `send_file_to_user`. The tool result is stored as
   `ToolResultBlock.output[0]` = `DataBlock(source=URLSource(url="file:///E:/Desktop/<name>.pdf", media_type="application/pdf"))` (1,782,114 bytes → under the 2 MiB inline cap).
3. Ask the agent anything else, so the tool result is part of the next model request.

**Actual** (4 consecutive attempts, `llm_max_retries=3`):

```
2026-09-19 12:34:34 | ERROR | qwenpaw\providers\retry_chat_model.py … 
openai.BadRequestError: Error code: 400 - {'error': {'message': '.messages[60]: file must have a file_id or file_data', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}}
```
same error at `12:34:57`, `12:35:09`, `12:35:28` — the turn never completes.

**Expected**: the PDF is either inlined in a shape the endpoint accepts, or degraded to a text note (which is exactly what QwenPaw already does for other media types, see below).

## What actually goes on the wire (captured)

The offending part is **not** a bare `{"type":"data"}` part (the #7597 symptom) — after #7621 it is a **well-formed OpenAI-style nested `file` part**:

```json
{
  "type": "file",
  "file": {
    "filename": "<name>.pdf",
    "file_data": "data:application/pdf;base64,JVBERi0xLQ…"
  }
}
```

Note: `file_data` is nested **inside** `file`; there is **no top-level `file_id` / `file_data`**.

Reproduced offline against the pinned AgentScope, same block, no network:

```python
# agentscope==2.0.7.post1
from agentscope.formatter import OpenAIChatFormatter
from agentscope.message import Base64Source, DataBlock, Msg, ToolResultBlock, ToolCallBlock
# … ToolResultBlock(output=[DataBlock(source=Base64Source(media_type="application/pdf", data=pdf_b64))]) …
wire = asyncio.run(OpenAIChatFormatter().format(msgs))
# → assistant(tool) message + promoted user message containing the nested {"type":"file","file":{…}} part
```

## Direct verification against the live endpoint

The 400 message above is not inferred from documentation — it was reproduced directly against `https://api.deepseek.com/chat/completions` (`deepseek-v4-flash`), with hand-written JSON payloads, changing only the shape of the content part. The API key was the one already configured in the reporter's own QwenPaw install (decrypted locally by QwenPaw itself, never copied into a third-party tool and not included here).

| # | content part sent | result |
| --- | --- | --- |
| 1 | `{"type":"text"}` only | **200** |
| 2 | `{"type":"image_url","image_url":{"url":"data:image/png;base64,<valid 8×8 PNG>"}}` | **200** |
| 3 | `{"type":"file","file":{"filename":"blue.png","file_data":"data:image/png;base64,<same PNG>"}}` — **nested, i.e. what QwenPaw emits** | **400** `.messages[0]: file must have a file_id or file_data` |
| 4 | `{"type":"file","filename":"blue.png","file_data":"data:image/png;base64,<same PNG>"}` — **flat, i.e. the documented shape** | **200** |
| 5 | `{"type":"file","file":{"filename":"t.pdf","file_data":"data:application/pdf;base64,<minimal PDF>"}}` — nested PDF | **400** `.messages[0]: file must have a file_id or file_data` |
| 6 | `{"type":"file","filename":"t.pdf","file_data":"data:application/pdf;base64,<minimal PDF>"}` — flat PDF | **400** `.messages[0]: You have uploaded an unsupported file. Please make sure your file is valid and has one of the following formats: webp, png, jpeg, and gif.` |

Two independent conclusions:

1. **The nested shape is what produces this exact error string.** With identical bytes inside, the nested part fails (rows 3, 5) while the flat part of the same image succeeds (row 4). The server message is character-for-character the one in the log above: it sees `type=file` with no top-level `file_id`/`file_data` and never gets to parse the part at all.
2. **A PDF cannot reach this endpoint even after the shape is fixed.** Row 6 shows the flat shape being accepted structurally and then rejected on media type — the `file` part carries images only (webp/png/jpeg/gif), consistent with the vision guide. A fix therefore needs both a flatter serialization *and* a media-type gate.

## Root cause

1. `agentscope/formatter/_openai_formatter.py` → `_format_openai_data_block` routes `application/pdf` to `_format_file_source(source, name)`, which returns the **nested** OpenAI shape shown above. `OpenAIChatFormatter.input_types` = `text/plain`, `image/*`, `audio/*`, `application/pdf`.
2. DeepSeek's OpenAI-compatible endpoint expects the **flat** shape and documents `file` parts for images only:
   - flat part: `{"type": "file", "file_data": "data:image/jpeg;base64,<BASE64>", "filename": "image.jpg"}` (`api-docs.deepseek.com/guides/vision`, and `api-docs.deepseek.com/api/create-chat-completion` → `file_id` / `file_data`)
   - supported content-part types on that endpoint are `image_url` (and `file` for the image formats above); a PDF is not an accepted `file` payload.
   So the endpoint sees a part of type `file` with neither top-level `file_id` nor top-level `file_data` and reports exactly the message we get — **confirmed by rows 3–6 of the direct verification above**, not just by the docs.
3. Why it does not recover: the strip gate is `if not supports_multimodal: …` (`qwenpaw/agents/utils/message_request_normalizer.py` via `qwenpaw/agents/prompt.py::_agent_will_strip_media`). With a multimodal model the PDF block is deliberately kept, and the media-capability fallback does not engage for this error, so all 4 attempts send the same bad part.
4. Why the **same** session worked 12:23–12:33 with a PPTX in context: `application/vnd.openxmlformats-officedocument.presentationml.presentation` is not in `input_types`, so it is only represented as a text note — no `file` part is emitted. Only `application/pdf` (and images/audio) reach the file path.

## Impact

* The turn hard-fails after 4 identical 400s; the user just sees a generic internal error.
* The bad part stays in the session history, so **every** later request whose window still contains that tool result fails the same way until the block is evicted (or media stripping is re-enabled). In our run it recovered only after the model capability metadata was reverted (`Background local model restore completed` at `12:48:17`, next turn at `12:51:07` succeeded).
* This will hit users who enable DeepSeek vision (upstream #7717), which is the natural next step for this provider.

## Suggested fix (any of these breaks the loop)

1. **Provider-aware serialization**: for endpoints whose `file` part is flat (DeepSeek), emit `{"type":"file","file_data":…,"filename":…}`, or fall back to an `image_url`-style data URL only when the media is an image. *(Verified: the flat form of the same payload returns 200 — row 4 above.)*
2. **Gate by media type**: do not emit `file` parts for non-image media (`application/pdf`, docx, pptx …) when the provider metadata only advertises `supports_image` (DeepSeek has no `supports_pdf`/`supports_file` flag today — see the 2026-09-08 comment). Degrade to the existing text placeholder instead.
3. **Don't hard-fail on an unusable part**: treat `file must have a file_id or file_data` (and the vLLM `Unsupported chat content part type: 'file'` seen in the other report) as a media-capability error in `QwenPawAgent._is_explicit_media_capability_error`, so the existing strip-and-retry fallback can rescue the session on the next attempt instead of repeating the same request.

## Notes

* The nested shape itself originates in AgentScope (`_format_file_source`), so the fix may belong in AgentScope or in QwenPaw's provider-specific formatter override — filing here because the user-visible failure and the capability metadata are QwenPaw-side.
* Everything above was verified in two ways: (a) offline against `agentscope==2.0.7.post1` (the version pinned in the running build) to capture what QwenPaw actually puts on the wire, and (b) **live against `api.deepseek.com` with the real key**, sending each candidate part shape by hand to confirm which one triggers the exact 400 — so no part of the diagnosis rests on documentation reading alone. The failing request itself was not packet-captured; its byte shape is reproduced from the shipped formatter fed with the exact block stored in `history.db`.

axelray-dev:
## Description

Provider requests containing an OpenAI-style file part can fail with `file must have a file_id or file_data`. Treat this explicit media rejection as a request-scoped fallback signal so the agent retries without media and does not cache the model as globally text-only.

**Related Issue:** Fixes #7883

**Security Considerations:** No security impact.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] I ran `pre-commit run --files` locally and it passes
- [ ] Ready for review

## Testing

- `.venv/bin/pytest -q tests/unit/agents/test_react_agent_media_strip.py tests/unit/agents/test_model_factory_message_normalization.py`: 72 passed
- `.venv/bin/black --line-length=79 src/qwenpaw/agents/react_agent.py tests/unit/agents/test_react_agent_media_strip.py`: passed
- `.venv/bin/flake8 --extend-ignore=E203 src/qwenpaw/agents/react_agent.py tests/unit/agents/test_react_agent_media_strip.py`: passed
- `.venv/bin/pre-commit run --files src/qwenpaw/agents/react_agent.py tests/unit/agents/test_react_agent_media_strip.py`: passed
- `git diff --check`: passed

## Evidence

The regression tests confirm that the provider error enables the request-level media fallback while the global media-capability classifier remains false. The focused test run passed 72 tests.


axelray-dev:
Reviewed the fallback pattern and regression coverage on the current head. The changed files compile cleanly and git diff --check passes; the PR body records the focused 72-test run and formatting/lint checks from the branch. Marked ready for review.

Issue #7815, opened 2026-09-16
Console does not recover from a failed lazy page chunk load; every navigation stays on the error screen until a full reload


wjt0321:
## Summary

When a lazily imported Console page fails to load, the UI does not recover. Every subsequent attempt to switch pages keeps showing the load-failure screen, and the only way out is a full page reload. There is a retry mechanism in the code, but it cannot succeed, and the error boundary's reset path cannot break the deadlock.

This is a robustness problem that turns a brief, self-healing glitch into a hard stuck state. It is not caused by the frame-parsing issue in #7813, but that issue can trigger it, which is how I hit it.

## Environment

- QwenPaw 2.2.1, Windows desktop build
- Console over `127.0.0.1`, default webview shell
- Verified against the code paths in `main`

## Observed behavior

The failure mode is consistent and has a distinctive shape:

1. At some point a page fails to load and the error screen appears ("页面加载失败" / "Page failed to load").
2. From then on, clicking any item in the left navigation does not recover. The load-failure screen stays.
3. Reloading the page fixes it immediately and completely.

Point 2 is the part that matters. Because the error screen is sticky across navigation, the user cannot work around it in the app at all. I have hit this repeatedly, not once.

### The retry cannot succeed

`console/src` builds lazy pages through a retry wrapper:

```js
const QP = 3, $P = 1000;
function Sr(e, t) {
  return e().catch(a => {
    if (t <= 0) throw a;
    return new Promise(n => setTimeout(() => n(Sr(e, t - 1)), $P));
  });
}
```

Three attempts, one second apart. But `e` is always the same factory, e.g. `() => import("./SomePage-<hash>.js")`. Every retry requests the identical URL. When the underlying cause is a transient fetch failure, the retry re-issues a request the browser has already recorded as failed, so all three attempts fail together. The retry adds latency but no real recovery.

It is also worth noting the retry only covers total failure. Nothing re-attempts after the retries are exhausted.

### The error boundary cannot reset

The boundary wraps the routed pages:

```js
componentDidUpdate(prev) {
  if (this.state.hasError && prev.resetKey !== this.props.resetKey) {
    this.setState({ hasError: false, isChunkError: false, restarting: false, restartError: "" });
  }
}
```

and it is mounted with the route pathname as the key:

```jsx
<ErrorBoundary resetKey={pathname} canRestartRuntime={...}>
  <Suspense fallback={...}>
[...]

PR #7889, opened 2026-09-19
fix(console): recover from transient DOM-mutation render errors

### LINKED ISSUE #7888: [Bug] Chat page stuck on "Something went wrong" - insertBefore NotFoundError from React commitPlacement (browser UI layer injects a <font> wrapper around a React-managed text node)
### Environment

| Item | Value |
|---|---|
| QwenPaw | 2.2.1 (self-built Docker image) |
| Browser | Microsoft Edge 153.0.4234.32 on Windows 10 22H2 |
| Affected routes | `/#/chat`, `/#/chat/<session-id>` |
| Unaffected routes | Settings Center, Files, Models, Apps, ... |

### Description

Opening any chat route replaces the whole content area with `chunkError.genericTitle` ("页面出现异常" / "Something went wrong") plus a "Reload" button. The left sidebar keeps working. Clicking **Reload** does nothing useful — even a full F5 does not recover it. Occasionally it recovers only after navigating to another route and back.

Console:

```
NotFoundError: Failed to execute 'insertBefore' on 'Node':
The node before which the new node is to be inserted is not a child of this node.
    at Tu (react-vendor-*.js:40:25719)
    at Rt (react-vendor-*.js:40:...)
    at Ks (react-vendor-*.js:40:...)
    at vt (...)
    ...
Render error: NotFoundError ... (componentStack: ... ChunkErrorBoundary ...)
```

`Tu` is `commitPlacement` after minification (verified against the React 18.3.1 source).

### Reproduction

- **Reproduces**: opening `/chat` with our production Edge profile (`User Data/Default/Preferences` present).
- **Does not reproduce**: a fresh `--user-data-dir` with the same URL.

Minimal repro (curl + headless Edge):

```bash
# 1. copy a real profile to a temp dir
# 2. start Edge with --headless=new --remote-debugging-port=9222 --user-data-dir=<copy>
# 3. open http://<host>:8088/#/chat via CDP
# 4. observe the Runtime exception
```

### Root cause

The host branch of `commitPlacement`:

```js
// de-minified: function Tu(e, t, n) -> commitPlacement
if (tag === 5 || tag === 6) {              // HostComponent | HostText
  el = stateNode;
  if (t) parent.insertBefore(el, t);        // <- throws here
  else   parent.appendChild(el);
}
```

It throws when `refNode.parentNode !== parent`. The DOM spec requires `refNode` to be a direct child of `parent`.

I installed a diagnostic guard on `Node.prototype.insertBefore` and captured a stable sample:

```
parent:    BUTTON.qwenpaw-dropdown-trigger.ApprovalToggle-module__trigger__*
refNode:   #text
refParent: FONT.
```

`<font>` is not an element React would ever render. **The browser UI layer (Edge appearance / translation / read-aloud style features) wraps a React-managed text node in a `<font>` element in the same frame, next to the `ApprovalToggle` button.** React's DOM bookkeeping then diverges from the real DOM, and the next commit calls `insertBefore` with a `refNode` that has been re-parented — throwing `NotFoundError`.

### Why only chat, and why it stays broken

1. **Only chat crashes**: the `ApprovalToggle` button (`qwenpaw-dropdown-trigger`) is only rendered on the chat page (the tool-approval toggle). Other pages do not have that node, so the injection never triggers.
2. **Permanently stuck**: the error bubbles to `ChunkErrorBoundary` (`console/src/layouts/MainLayout/index.tsx:60`, `resetKey={currentPath}`). That component only resets when the prop **changes**:

   ```tsx
   componentDidUpdate(prevProps) {
     if (this.state.hasError && prevProps.resetKey !== this.props.resetKey) { /* reset */ }
   }
   ```

   The route stays the same inside chat, so `resetKey` never changes and `hasError` stays `true` forever. The "Reload" button calls `restartOwnRuntime()` instead of `window.location.reload()`, so the React tree keeps its error state.

   Also, `ChunkErrorBoundary.isChunkLoadError()` returns `false` for this error (the message contains no `loading chunk` etc.), so it falls into the generic `chunkError.genericTitle` branch and never gets chunk-specific retry behavior.

### Isolation experiments (ruling out cache / extensions / storage)

| Experiment | Result |
|---|---|
| Fresh temp profile | OK |
| Copy of the real profile | **Crash** |
| Real profile + cleared Cache / SW / LocalStorage / SessionStorage / IndexedDB | **Still crashes** |
| `--disable-extensions` | **Still crashes** |
| Viewport 1280 / 1440 / 1920 / 2560 | All crash |
| **Fresh profile + only the `Preferences` file** | **Crash** (key finding) |
| Fresh profile with no `Preferences` | OK |
| `Preferences` bisect: first 74 keys / last 75 keys | first OK / **last crashes** |
| Last half split into 4 groups (sessions / partition+plugins / profile / toolbar+ntp) | **None of the 4 groups crashes alone** -> combination effect, not a single key |

So the trigger lives in a **multi-key combination inside the browser `Preferences`**, and is unrelated to QwenPaw server data, HTTP cache, extensions, or storage.

### Suggested fix (two orthogonal changes; both recommended)

**A. Harden `ChunkErrorBoundary` (necessary)**

It can currently only reset via `resetKey`. That does not cover "render error + unchanged route". Suggested:

1. Distinguish chunk errors from render errors in `getDerivedStateFromError`; for render errors offer a "Retry" button that calls `this.setState({ hasError: false })` once, and only fall back to "Reload" after N consecutive failures.
2. Make the "Reload" button call `window.location.reload()` rather than only `restartOwnRuntime()`, otherwise the React tree is not torn down.
3. (Optional) Count same-class errors inside `componentDidCatch` and add a cooldown to avoid an error loop.

**B. Add DOM tolerance (the actual fix)**

Wrap the native method at application startup, **synchronously and inline before the vendor bundle runs**:

```js
(function () {
  var orig = Node.prototype.insertBefore;
  Node.prototype.insertBefore = function (newNode, refNode) {
    if (refNode != null && refNode.parentNode !== this) {
      // refNode is detached -> degrade to appendChild
      if (refNode.parentNode == null) return orig.call(this, newNode, null);
      // refNode belongs to another parent -> this insert is meaningless, swallow it
      // and let the next commit fix up the tree
      return newNode;
    }
    return orig.call(this, newNode, refNode);
  };
})();
```

This shields React from every same-class race caused by third-party DOM injection (browser UI layer, translation extensions, password managers) at near-zero cost.

We validated an equivalent guard injected into `index.html` on our deployment: **new task + 3 historical sessions + 8 rounds of rapid switching -> 0 failures; body length was 130 (error page) before the fix and 11017 (full conversation) after.**

### Related issue

- **#7815 [open]** "Console does not recover from a failed lazy page chunk load; every navigation stays on the error screen until a full reload" — same `ChunkErrorBoundary.resetKey` recovery defect; the fix could be merged.

Searching for `insertBefore` / `Failed to execute insertBefore` returns **0 results**, so this specific crash appears to be reported here for the first time.

TerebiSAMA:
## Description

The console can get stuck on the generic error fallback **forever** on a route whose render throws a DOM-mutation error, because `ChunkErrorBoundary` only clears its error state when `resetKey` changes — that is, on navigation.

Reported in #7888. On the affected machine, opening any chat route replaces the content area with "Something went wrong" while the sidebar keeps working, and neither the reload button nor F5 recovers it:

```
NotFoundError: Failed to execute 'insertBefore' on 'Node':
The node before which the new node is to be inserted is not a child of this node.
    at commitPlacement (react-vendor-*.js:40:25719)
    ...
Render error: NotFoundError ... (componentStack: ... ChunkErrorBoundary ...)
```

React raises this when a third party mutated the real DOM behind its back. We captured the injector with a diagnostic wrap on `Node.prototype.insertBefore`:

```
parent:    BUTTON.qwenpaw-dropdown-trigger.ApprovalToggle-module__trigger__*
refNode:   #text
refParent: FONT.
```

`<font>` is not an element React renders — a browser UI layer (appearance / translation / read-aloud style) wrapped a React-managed text node in it, in the same frame, next to the chat page's tool-approval toggle. The chat page is the exposed route because that toggle is the node being wrapped; other pages have no such node.

Reproducibility notes: reproduces with a real Edge profile, does not reproduce with a fresh `--user-data-dir`. Cache, storage, extensions and viewport size were all ruled out; the trigger lives in a multi-key combination inside the browser's `Preferences`.

**Related Issue:** Fixes #7888 (relates to #7815, same `resetKey`-only recovery defect)

**Security Considerations:** None. The change touches only DOM APIs in the renderer; no auth, network, or env/config handling is involved.

### Changes

Two orthogonal changes.

**1. `utils/domInsertTolerance.ts` — tolerate mutations performed behind React's back**

Wraps `Node.prototype.insertBefore` / `Node.prototype.removeChild` so a call that would have thrown is degraded instead:

- reference node detached → append the node anyway (position lost once, the next commit re-orders correctly);
- reference node under a different parent → skip this insertion and let the commit proceed;
- `removeChild` on a node owned by another parent → report success.

The call site is React's commit phase, which reconciles on the next render, so swallowing one mis-positioned insertion is safe and strictly better than failing the whole commit and blowing away the route. Valid calls are forwarded untouched, and it is installed from `main.tsx` before `createRoot` runs.

**2. `ChunkErrorBoundary` — recover in place, and stop getting stuck**

- Classifies DOM-mutation errors (`NotFoundError`, `HierarchyRequestError`, and the matching messages) and offers a **Try again** button that re-renders the same subtree, instead of forcing a full page reload for a transient race.
- Caps retries (`MAX_RENDER_ERROR_RETRIES = 2`) so a page that keeps failing falls back to the reload button.
- `retryCount` is intentionally preserved across `getDerivedStateFromError` for the same reason.
- Chunk-load errors are unchanged: they still get the targeted message and no in-place retry, since a re-render cannot fix a missing chunk.

The classification lives in `getDerivedStateFromError` rather than a field set by `componentDidCatch`, because the latter runs *after* the fallback has already rendered — a field set there is invisible to the first fallback render.

New i18n key `chunkError.retry` added across all seven locale files (en, zh, ja, ru, vi, id, pt-BR).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

- N/A: no channel / backend code touched, only `console/`.

## Testing

```bash
cd console

# types
node_modules/.bin/tsc -b --noEmit

# lint
node_modules/.bin/eslint src/components/ChunkErrorBoundary.tsx \
  src/components/ChunkErrorBoundary.test.tsx \
  src/utils/domInsertTolerance.ts \
  src/utils/domInsertTolerance.test.ts \
  src/main.tsx

# formatting
node_modules/.bin/prettier --check src/components/ChunkErrorBoundary.tsx \
  src/utils/domInsertTolerance.ts src/main.tsx src/locales/*.json

# new tests
node --max-old-space-size=4096 ./node_modules/vitest/vitest.mjs run \
  src/components/ChunkErrorBoundary.test.tsx \
  src/utils/domInsertTolerance.test.ts
```

## Evidence

All checks below were run on the commit in this PR (`08e069c`), inside `console/`, after `npm ci`. Raw output is quoted verbatim.

### Vitest — new tests

```
 ✓ src/utils/domInsertTolerance.test.ts (5 tests) 12ms
 ✓ src/components/ChunkErrorBoundary.test.tsx (5 tests) 678ms
     ✓ offers Hub users a runtime restart when a page fails  438ms
 Test Files  2 passed (2)
      Tests  10 passed (10)
   Duration  6.71s
```

Covers: guard forwards valid `insertBefore`; guard appends when the reference node is detached; guard skips when the reference node belongs to another parent; guard tolerates `removeChild` across parents; guard is idempotent. Boundary recovers in place from a DOM-mutation error; boundary drops the retry button once retries are exhausted; boundary offers no in-place retry for a chunk-load error.

### tsc / eslint / prettier

```
$ node_modules/.bin/tsc -b --noEmit
tsc exit=0

$ node_modules/.bin/eslint <changed files>
eslint exit=0

$ node_modules/.bin/prettier --check <changed files>
Checking formatting...
All matched files use Prettier code style!
```

### Real-world reproduction — before / after

Measured on the reporter's deployment (NAS Docker, Edge 153, real user profile), driving the page over CDP. An equivalent guard was injected into `index.html` to prove the approach before writing this source-level patch:

| Scenario | Fallback page | Input box | Body text | Guard hits | Result |
|---|---|---|---|---|---|
| New task `/chat` | no | yes | 11017 | 1 | pass |
| History session `3378972a` | no | yes | 11017 | 1 | pass |
| History session `244120ce` | no | yes | 11017 | 1 | pass |
| History session `36928ce5` | no | yes | 11017 | 1 | pass |
| Stress: rapid switching, 8 rounds | — | — | — | — | 0/8 failures |

Before the fix the page rendered 130 characters of body text (the fallback); after, 11017 (the full conversation). The guard hit exactly once per page load, matching the single injected `<font>` wrapper.

## Additional Notes

- The `<font>` wrapper itself is the browser's doing, so this patch does not try to prevent it — it makes React's commit phase resilient to it, and gives the boundary a way out if something still slips through.
- If you would prefer to keep `Node.prototype` untouched regardless of packaging, I am happy to split this into two commits so the boundary change can land independently; the guard is the part that actually stops the crash.


TerebiSAMA:
Hi @jinliyl — gentle ping when you have a moment.

This is a small console fix that addresses the DOM-mutation render error reported in #7888 (`ChunkErrorBoundary` getting stuck forever). 12 files, +364/-18, dominated by a new `domInsertTolerance` helper plus its test, plus a localized edit in `ChunkErrorBoundary` and the 7-locale string for the recovery tooltip.

CI is fully green here (all 25 checks, no failures, no waiting). If you would prefer I split the `domInsertTolerance` helper into its own PR, I'm happy to.

No urgency — just flagging in case it slipped past your queue. Thanks again for your time.

PR #7901, opened 2026-09-20
ci(release): unfreeze merges as soon as the release finishes instead of waiting for cron

yutai78786:

## Description

The release freeze is lifted only by this watchdog, on a 5-minute cron. In practice GitHub delays that cron, so after a release finishes every open PR in the repository can stay frozen for a long time. This adds a `workflow_run` trigger so the unfreeze happens as soon as the release run completes, and corrects a comment that understated the delay.

**Measured delay (not assumed).** Over 11 consecutive scheduled runs on 2026-09-19/20, differencing each run's `created_at`: median gap **13.9 min**, but **5 of 11 (45%) exceeded 15 min** and the **worst gap was 60.1 min**. The header comment claimed "typically <= 15 minutes", which describes the median only.

**Live instance.** Release run `35483178491` completed successfully at **03:09:17Z** on 2026-09-20. The last scheduled watchdog had run at **02:57:18Z** and correctly refused to unfreeze (`Release run in flight — freeze is legitimate. Exiting.` — the release really was still running). No patrol ran for the next **22 minutes**; at 03:19:58Z the `release-window-active` label was still present and PR #7889 still showed `Release Window => failure | Release in progress — merges frozen`. It took a manual `gh workflow run` (run `35486333352`, success) to clear it.

**Why `workflow_run` and not a job appended to `release.yml`.** I first considered adding a job to `release.yml` that calls this watchdog, and rejected it after reading the unfreeze logic:

- `release_in_flight()` refuses to unfreeze while any release run is `queued`/`in_progress`. A job *inside* the release run would fire while that run is still `in_progress`, so the watchdog would (correctly) exit without unfreezing — the trigger would deadlock against its own precondition.
- It would also require `actions: write` on `release.yml`, which currently declares `actions: read`. Escalating that scope is not warranted here.
- `workflow_run` fires only after the triggering run reaches `completed`, so the in-flight check passes on its own terms.

The trigger uses `types: [completed]`, **not** `[completed: success]`: a failed release leaves the same freeze behind and needs the same unfreeze (run `35204357591` on 2026-09-17 did exactly that).

**The sole-unfreezer architecture is unchanged.** This file remains the only place that deletes the lock marker; the shared concurrency group (`release-window-watchdog`, `cancel-in-progress: false`) still serialises a scheduled, a dispatched and a `workflow_run`-triggered patrol so two of them can never sweep concurrently; and a concurrent second release still keeps the freeze because `release_in_flight()` sees it and exits. The cron schedule is kept as the backstop for the cases `workflow_run` cannot cover (a release run that never reaches `completed`, or a watchdog run that dies mid-sweep).

**Also corrected:** the human-fallback advice now prefers `gh workflow run release-window-watchdog.yml` over deleting the label by hand. A dispatch runs this same sole unfreezer — so the full sweep happens and concurrency still serialises it — whereas hand-deleting the marker bypasses the sole unfreezer and then still needs a `backfill=true` run to clear the remaining red stamps.

**Related Issue:** No corresponding upstream issue; this is CI-pipeline maintenance agreed with the project maintainer.

**Security Considerations:** No credential or configuration-security changes, and **no permission escalation** — `permissions:` is untouched (`statuses: write`, `actions: read`, `contents: read`, `issues: write`). The behaviour change is limited to *when* an already-existing unfreeze path runs. It can only make the freeze shorter, never allow a merge during a release window: every gate in the unfreeze decision still runs unchanged.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [ ] Tests
- [x] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [x] Ready for review

## Testing

The unfreeze decision is shell embedded in a workflow, so it was verified by byte-comparison plus running the workflow on a real runner:

```bash
# 1) the decision logic is byte-identical before and after
python3 - <<'EOF'
import subprocess, hashlib
f = '.github/workflows/release-window-watchdog.yml'
new = open(f, encoding='utf-8').read()
old = subprocess.run(['git','show','HEAD~1:' + f], capture_output=True).stdout.decode()
seg = lambda t: t[t.index('release_in_flight() {'):t.index('Release over (no run queued/in-progress')]
print('decision block md5 old=%s new=%s identical=%s'
      % (hashlib.md5(seg(old).encode()).hexdigest()[:12],
         hashlib.md5(seg(new).encode()).hexdigest()[:12], seg(old) == seg(new)))
body = lambda t: t[t.index('          set -euo pipefail'):]
print('whole shell body identical =', body(old) == body(new))
EOF

# 2) syntax + Actions lint
pre-commit run --files .github/workflows/release-window-watchdog.yml

# 3) run the workflow itself on a fork (workflow_dispatch path)
gh workflow run release-window-watchdog.yml --repo <fork> --ref <branch>
```

## Evidence

**1. The decision logic is byte-identical — the diff is one trigger plus comments:**

```
decision block md5 (release_in_flight() .. "Release over") = 2e6eb105c29d  old
                                                           = 2e6eb105c29d  new   identical ✅  (2121 chars)
whole shell script body (from `set -euo pipefail`) identical = True ✅
release_in_flight call sites: 3 before, 3 after ✅
permissions / concurrency group / cancel-in-progress / timeout-minutes: all untouched ✅
cron backstop still "*/5 * * * *" ✅
```

**2. pre-commit (locked revisions, complete run — not a subset):**

```
EXIT=0   (other hooks report "no files to check": this PR touches only YAML)
```

**3. The workflow runs correctly on a real runner with my change applied.** Fork run [`35488523636`](https://github.com/yutai78786/QwenPaw/actions/runs/35488523636), head `5b1043994`, conclusion **success**:

```
env:
  BACKFILL: false
No release-window-active label — nothing frozen. Exiting.
```

That is the correct idle path (no freeze present → nothing to do).

**4. The comment I added about `BACKFILL` is asserted from measurement, not from reading the code.** `BACKFILL: ${{ inputs.backfill }}` is empty under `schedule` and `workflow_run`, and the existing `${BACKFILL:-false}` guard turns empty into false. Measured across three trigger types:

```
schedule          (upstream run 35485319857)  -> "BACKFILL:"        (empty)
schedule          (fork     run 35480401646)  -> "BACKFILL:"        (empty)
workflow_dispatch (fork     run 35488523636)  -> "BACKFILL: false"
```

So a patrol can never accidentally backfill. The guard line (`if [ "${BACKFILL:-false}" = "true" ]`) is pre-existing and untouched.

**5. The trigger target name matches `release.yml` verbatim.** `workflow_run.workflows` is resolved by the workflow's top-level `name:`, not its filename, so a typo would fail silently forever:

```
release.yml                name: "Release (unified)"
watchdog workflow_run      workflows: ["Release (unified)"]   <- identical
```

**6. Measured cron gaps (the data behind the comment correction)** — 11 consecutive `schedule` runs, `created_at` differenced:

```
10.2  8.3  8.3  8.0  60.1  33.4  20.5  17.7  23.5  13.9  8.1   (minutes)
median 13.9 | over 15 min: 5/11 (45%) | worst: 60.1
```

**7. Fork CI on this head** (`5b1043994`): `Pre-commit Checks` success, `NPM Format` success, `Release Window Watchdog` success.

**8. No conflict with current main:** `gh api compare ce86d9b94...7ff690f4f` → 2 commits / 17 files, **0** touching this file.

## Additional Notes

- Single commit `5b1043994`, 1 file, **45 insertions / 3 deletions**. **Logic lines changed: 0** — the whole `run:` script body is byte-identical; the diff is one new trigger plus comments.
- ⚠️ **One thing this PR cannot prove before merge, stated plainly:** the `workflow_run` trigger itself cannot be exercised on a fork branch, because GitHub only honours `workflow_run` when the trigger exists on the **default** branch. So its end-to-end firing is unverified by construction. What was verified instead: the syntax passes actionlint, the target workflow name matches `release.yml` verbatim (evidence 5 — the failure mode of getting this wrong is a silent no-trigger), the triggered workflow itself runs green on a real runner (evidence 3), and the unfreeze logic is byte-identical (evidence 1) so a trigger problem cannot cause a wrong unfreeze — the cron backstop remains. **The first real release after merge should be watched to confirm the automatic unfreeze fires.**
- Commit message is fully English (zero non-ASCII except the signature name in the trailer) and carries the standard `Signed-off-by`.


cuiyuebing (approved):
LGTM

PR #7906, opened 2026-09-20
fix(loop): prevent stale doom-loop escalation

### LINKED ISSUE #7905: [Bug]: DoomLoopGate escalates to TERMINATE on a text-only round without new tool-call evidence
## QwenPaw Version

Reported environment: `2.2.2b1`, commit `1d5021a4`.
Source review and isolated reproduction target commit `1d5021a4`.
As checked on 2026-09-20, `main`'s `src/qwenpaw/loop/gates/doom_loop.py` was byte-identical to that file at the pinned commit.

## Description

`DoomLoopGate.check(ctx)` increments `consecutive_hits` whenever its retained history window matches, even if the current check recorded no new tool call. Consequently, after a warning at 3 hits, a text-only response with `has_tool_calls=False` can escalate to `TERMINATE` at 4 hits. This occurs even when the latest context message has also been replaced with the text-only response.

Expected: a previously evaluated window must not count as new repetition evidence. A new text-only round should not cause escalation of a tool-repetition detector.

A related sampling issue is that `_auto_record_from_ctx()` takes only the last tool call from the last context message. Its deduplication key is the iteration number, not the message/call identity. If iteration advances while the same message remains at the tail, the same call is appended again. The isolated reproduction demonstrates this condition, but does not establish that a concurrent tool batch alone causes these iteration/context transitions in the real scheduler.

**Related reports:** #5906 (closed report of false repetition detection); #7420 (open report involving re-dispatch/tool-result behavior). These are related symptoms, not established identical root causes.

**Security considerations:** This report concerns incorrect termination and repetition accounting. No authentication bypass or data exposure is claimed.

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Environment

- **QwenPaw version:** reported `2.2.2b1`; verified gate source at `1d5021a4`
- **OS:** original report on Windows (exact version not supplied); isolated gate reproduction also run on macOS
- **Install method:** not established for the original report
- **Python version:** exact original environment version not supplied
- **Model/API:** not required for the isolated reproduction

## Steps to Reproduce

Use `window_size=3`, `similarity_threshold=1.0`, and stages `after=3 / modify_prompt`, `after=4 / stop`.

1. Use the real `check(ctx)` auto-recording path with three separate assistant messages containing identical tool name/arguments and increasing iteration numbers. Do not manually call `gate.record()`.
2. The third observation triggers a warning.
3. Replace the last context message with a text-only assistant message. Pass that message as `final_msg`, set `has_tool_calls=False`, and advance the iteration.
4. Check again. No record is added, but the gate returns `TERMINATE`.

The script below also includes the stale-message sampling condition and a same-iteration control case. It downloads three source files from the pinned public commit and runs the actual gate/base classes. Configuration objects and session-ID lookup are replaced by minimal test adapters. It neither executes shell commands nor calls an LLM, and is not an end-to-end scheduler reproduction.

Run with `python3 repro.py` (network access to raw.githubusercontent.com required for source download).

```python
"""Isolated reproduction against pinned upstream gate code, no QwenPaw install.
Run: python3 repro.py
Downloads only three public Python source files from the specified commit.
The actual gate, base class, state handling, and result types are unmodified.
Only configuration objects and session-ID lookup are replaced with test adapters.
This verifies gate behavior; it is not an end-to-end AgentScope scheduler test.
"""
import asyncio
import sys
import types
import urllib.request
from types import SimpleNamespace as NS

COMMIT = '1d5021a4'
BASE = f'https://raw.githubusercontent.com/agentscope-ai/QwenPaw/{COMMIT}/src/qwenpaw/'


def load(name, path):
    source = urllib.request.urlopen(BASE + path, timeout=30).read().decode()
    module = types.ModuleType(name)
    module.__file__ = BASE + path
    sys.modules[name] = module
    exec(compile(source, module.__file__, 'exec'), module.__dict__)
    return module


for name in ['qwenpaw', 'qwenpaw.loop', 'qwenpaw.loop.gates',
             'qwenpaw.config', 'qwenpaw.config.config']:
    module = types.ModuleType(name)
    module.__path__ = []
    sys.modules[name] = module
sys.modules['qwenpaw.config.config'].DoomLoopStageConfig = NS
load('qwenpaw.loop.gates.base', 'loop/gates/base.py')
loop = load('qwenpaw.loop.gates.loop_gate', 'loop/gates/loop_gate.py')
loop._session_id = lambda: 'isolated-repro'
doom = load('qwenpaw.loop.gates.doom_loop', 'loop/gates/doom_loop.py')


def new_gate():
    return doom.DoomLoopGate(window_size=3, similarity_threshold=1.0, stages=[
        NS(after=3, action='modify_prompt', prompt='warning'),
        NS(after=4, action='stop', prompt='stop'),
    ])


def tool_message(message_id, calls):
    return NS(id=message_id, role='assistant', content=[
        {'type': 'tool_call', 'id': f'{message_id}-{i}',
         'name': 'execute_shell_command', 'input': {'command': command}}
        for i, command in enumerate(calls)
    ])


async def check(gate, message, iteration, tools):
    agent = NS(state=NS(context=[message]))
    result = await gate.check({'agent': agent, 'iteration': iteration,
                              'has_tool_calls': tools,
                              'final_msg': None if tools else message})
    state = gate._ensure_state()
    print(f'iteration={iteration} tools={tools} action={result.action.value} '
          f'hits={state.consecutive_hits} history={len(state.history)}')
    return result, state


async def main():
    print('A: three distinct call events; fourth round has only text')
    gate = new_gate()
    for i in range(1, 4):
        await check(gate, tool_message(str(i), ['echo same']), i, True)
    text = NS(role='assistant', content=[{'type': 'text', 'text': 'Done.'}])
    result, state = await check(gate, text, 4, False)
    assert result.action.value == 'terminate' and len(state.history) == 3
    print('BUG REPRODUCED: escalation without new tool-call evidence')

    print('\nB: same 11-call message observed at different iteration numbers')
    gate = new_gate()
    batch = tool_message('batch', [f'echo distinct-{i}' for i in range(11)])
    for i in range(1, 4):
        result, state = await check(gate, batch, i, True)
    assert len({r.args_hash for r in state.history}) == 1
    assert result.action.value == 'interrupt_and_continue'
    print('Only the final call was sampled repeatedly; scheduler trigger NOT proven')

    print('\nC: repeated checks within one iteration do not append more records')
    gate = new_gate()
    for _ in range(4):
        result, state = await check(gate, batch, 1, True)
    assert len(state.history) == 1 and state.consecutive_hits == 0
    print('Iteration-level sampling guard works for this control case')


if __name__ == '__main__':
    asyncio.run(main())
```

## Actual vs Expected

- **Actual:** the retained three-record window is evaluated again on a text-only round; `history` stays at 3, while `consecutive_hits` increases from 3 to 4 and the gate returns `TERMINATE`.
- **Expected:** escalation requires newly observed repetition evidence. Rechecking unchanged evidence must be idempotent; a genuinely new text-only round must not escalate tool repetition.

## Logs / Screenshots

Isolated reproduction (case A):

```text
iteration=1 tools=True action=bypass hits=0 history=1
iteration=2 tools=True action=bypass hits=0 history=2
iteration=3 tools=True action=interrupt_and_continue hits=3 history=3
iteration=4 tools=False action=terminate hits=4 history=3
BUG REPRODUCED: escalation without new tool-call evidence
```

The original user-reported session logged a warning at 3 hits followed approximately four seconds later by a stop at 4 hits, reportedly after the model switched to text-only output. That session involved 11 distinct source-reading shell calls. The gate-level failure above is confirmed independently; the complete concurrent scheduling trace from that original session has not been independently reproduced.

### Relevant source

- [check(): retained-window detection and counter escalation](https://github.com/agentscope-ai/QwenPaw/blob/1d5021a4/src/qwenpaw/loop/gates/doom_loop.py#L124)
- [_auto_record_from_ctx(): iteration guard, last context message, and early return](https://github.com/agentscope-ai/QwenPaw/blob/1d5021a4/src/qwenpaw/loop/gates/doom_loop.py#L177)
- [runner: final_msg/has_tool_calls contract](https://github.com/agentscope-ai/QwenPaw/blob/1d5021a4/src/qwenpaw/loop/gates/runner.py#L83)
- [ReactAgent: gate check after reasoning](https://github.com/agentscope-ai/QwenPaw/blob/1d5021a4/src/qwenpaw/agents/react_agent.py#L1033)
- [Current unit tests](https://github.com/agentscope-ai/QwenPaw/blob/1d5021a4/tests/unit/loop/test_doom_loop_gate.py)

The checked gate unit-test file exercises recorded histories using `gate.record()` but does not cover message extraction through an agent context.

### Suggested fix invariants

1. Give each observed call a stable identity and record it once. Call identity is different from `(tool_name, args_hash)`: distinct calls with identical arguments must still count as separate events.
2. Track new records since the last evaluation. Rechecking the same evidence must not advance stages. Distinguish a duplicate check from a genuinely new text-only round before resetting state.
3. Prefer an explicit per-round call list or execution events over guessing from `context[-1]`. Disable the old sampling path if explicit execution recording is introduced, to avoid double counting.
4. If recording completed execution, align the check timing with completion and define concurrent ordering. `_execute_tool_call`'s entry precedes argument validation; recording at entry is not recording a completed call.

`final_msg` cannot simply replace the current extraction source: runner intentionally sets it to `None` on tool-call rounds. Supplying the actual assistant tool-call message requires an interface change.

Seeding the first detection with `window_size` may be intentional stage semantics; changing that or merely raising the stop threshold does not fix stale-evidence escalation.

### Regression coverage

- Warning followed by a new text-only round: no escalation to stop.
- Repeated checking of already evaluated evidence: no counter changes.
- Same message/call IDs revisited with increasing iteration numbers: no duplicate records.
- All calls from a multi-call message are accounted for exactly once.
- Distinct newly occurring identical calls still trigger configured warning/stop stages.
- A new nonmatching call resets consecutive repetition according to the defined semantics.

axelray-dev:
## Description

DoomLoopGate now advances repetition stages only when new tool-call evidence is recorded. A text-only response resets stale repetition state instead of evaluating the retained window as another hit.

**Related Issue:** Fixes #7905

**Security Considerations:** None. This change only affects loop repetition accounting and termination decisions.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [ ] Skills
- [ ] CLI
- [ ] Documentation (website)
- [x] Tests
- [ ] CI/CD
- [ ] Scripts / Deploy

## Checklist

- [ ] Ready for review

## Testing

- `PYTHONPATH=src .venv/bin/python -m pytest --confcutdir=tests/unit/loop tests/unit/loop/test_doom_loop_gate.py -q` -> 11 passed in 0.35s
- `.venv/bin/black --line-length=79 src/qwenpaw/loop/gates/doom_loop.py tests/unit/loop/test_doom_loop_gate.py` -> 2 files left unchanged
- `.venv/bin/flake8 --extend-ignore=E203 src/qwenpaw/loop/gates/doom_loop.py tests/unit/loop/test_doom_loop_gate.py` -> passed
- File-scoped pre-commit hooks for both changed files -> passed

## Evidence

The regression test verifies three repeated tool-call observations followed by a text-only round. The text-only check returns BYPASS, clears the stale history, and leaves the escalation counter at zero. A second regression test verifies that checking the same evidence twice does not advance stages.


rayrayraykk:
Thanks for the fix! This addresses the text-only escalation, but automatic sampling still deduplicates by iteration and records only the last tool call in each message. I reproduced false termination with both a stale message checked across iterations and distinct tool batches sharing the same final call.

#7919 addresses these cases as well by deduplicating by message ID, recording all calls in order, and evaluating only when new evidence is available.

Close this PR in favor of #7919 to consolidate the fix.

